PR #2, opened 2018-05-17
Blankslates

aaronshekey:
I've created a draft of what blankslates could be in a [Google Doc](https://docs.google.com/document/d/1OkK5nU4BCPP--SaQ5EhLxeKn1PL-cp8ZFRAR0o93wyk/edit#).


aaronshekey:
This was blocked by the fact that we didn't have illustrations at the time. If you've got a good set of illustrations we can drop into this, I'm happy to wrap it up and document!

TedGoas:
Hmm, what if we got a few generic illustrations (sized and colored correctly)? I'm happy to lead this using your Google Doc above.

PR #27, opened 2018-06-04
Negative 1px classes

hellohynes:
Adds negative 1px margins.


package.json
```diff
@@ -5,7 +5,7 @@
     "type": "git",
     "url": "https://github.com/StackExchange/Stacks.git"
   },
-  "version": "0.16.0",
+  "version": "0.16.1",
```
aaronshekey: This needs to be updated in the `package-lock.json` file as well. Make sure you're running `npm version 0.16.1`

hellohynes: I am. Not sure why `package-lock.json` isn't getting updated.

Issue #30, opened 2018-06-08
Descenders in .s-input are being cut off


aaronshekey:
The descenders within our `.s-inputs` are being cut off at various sizes. This is observed in Safari on High Sierra.

![screen shot 2018-06-08 at 12 57 39 am](https://user-images.githubusercontent.com/1369864/41141309-2c9e8828-6ab7-11e8-8260-0eeb9d163c9e.png)

PR #31, opened 2018-06-08
s-input bugs

### LINKED ISSUE #29: .s-inputs are rendering with the default OS font
As observed on the Teams join workflow, it appears that inputs with `s-input` applied are inheriting the OS-level fonts, and not `ff-sans` as intended. I think we need to add our standard fonts to `s-inputs`, text areas, etc.

screen shot 2018-06-08 at 12 54 23 am

Sadly, San Francisco looks soooo good here, but it's not intended in this case.

hellohynes:
This PR _decreases_ input padding (from `0.8em` to `0.7em`) and _increases_ line-height (from `@lh-sm` to a custom value `1.352307692`). This keeps inputs the same visual height as before, but  descenders aren't clipped anymore in Safari or Firefox.

Also it adds `font-family: inherit` to `.s-input`, `.s-textarea`, `.s-select > select` instead of not declaring a font-family.

Closes #29 #30


hellohynes:
@aaronshekey Minor update here that adjusts both the top and bottom padding for elements along with the line-height. A little more resilient IMO than modifying the top or bottom padding.

hellohynes:
Oh and I tested this in Chrome (Mac), Firefox (Mac), and Safari.

PR #33, opened 2018-06-08
s-tag

### LINKED ISSUE #32: Component: Tags
Add in tags into the Stacks system.

hellohynes:
This PR adds in `.s-tag` to Stacks. It updates class names accordingly:

| Old Class | » | New Class |
| --------- | -- | --------- |
| `.post-tag` | » | `.s-tag` |
| `.moderator-tag` | » | `.s-tag .s-tag__moderator` |
| `.required-tag` | » | `.s-tag .s-tag__required` |
| `.disliked-tag` | » | `.s-tag .s-tag__muted` |
| `.favorite-tag` | » | **Removed** (wasn't used anywhere) |
| `.company-tag` | » | **Removed** (unneeded style)  |
| `.geo-tag` | » | **Removed** (unneeded style)  |

Closes #32


lib/src/components/_stacks-tags.less
```diff
@@ -0,0 +1,119 @@
...
+.s-tag {
+    display: inline-flex;
+    position: relative;
+    align-content: center;
+    justify-content: center;
+    min-width: 0;
+    padding: 0.545em;
+    border-style: solid;
+    border-width: 1px;
+    border-radius: @tags-border-radius;
```
aaronshekey: This likely won't need to be themed after the new Unified Themes ship. I think we can define `@br-sm` here. PS. These corners look _great_.

lib/src/components/_stacks-tags.less
```diff
@@ -0,0 +1,119 @@
...
+
+    &:hover {
+        text-decoration: none;
+    }
+}
+
+//  $$  DISMISS ICON
+//      Style adjument to @Svg.ClearSm
+//  ---------------------------------------------------------------------------
+.s-tag--dismiss {
```
aaronshekey: Could we deviate from the current style here? The red circle could be problematic if themes choose a different tag color. Would something like this work?

screen shot 2018-06-08 at 11 59 23 am

lib/src/components/_stacks-tags.less
```diff
@@ -0,0 +1,120 @@
...
+//  ===========================================================================
+//  $   BASE STYLE
+//      The same style is applied to all tags. Do not modify the core style.
+//  ---------------------------------------------------------------------------
+.s-tag {
+    display: inline-flex;
+    align-content: center;
+    justify-content: center;
+    min-width: 0;
+    padding: 0 (@su4 + @su2);
```
aaronshekey: Should we just add an `@su6` already?

lib/src/components/_stacks-tags.less
```diff
@@ -0,0 +1,120 @@
...
+    border-style: solid;
+    border-width: 1px;
+    border-radius: @tags-border-radius;
```
aaronshekey: I don't think we'll actually need to customize this. I think it can just be `@br-sm`.

lib/src/stacks-dynamic.less
```diff
@@ -19,10 +19,11 @@
 
 //  --  COMPONENTS
 @import "components/_stacks-buttons";
+@import "components/_stacks-banners";
```
aaronshekey: This should be above banners, if we're alphabetizing.

lib/src/components/_stacks-tags.less
```diff
@@ -0,0 +1,120 @@
...
+        color: fade(@color, 90%);
+    }
+
+    &:hover {
+        border-color: darken(@border, 5%);
+        background-color: darken(@background, 5%);
+        color: darken(@color, 5%);
+
+        .s-tag--dismiss {
+            background-color: fade(@color, 10%);
```
aaronshekey: I've been sitting with this one, and I don't think we need the intermediate hover state on `.s-tag--dismiss`

lib/src/components/_stacks-tags.less
aaronshekey: Working on a new tiny icon for this one.

lib/src/components/_stacks-tags.less
hellohynes: Haha… Maybe?

lib/src/stacks-dynamic.less
hellohynes: 🙄 😀

hellohynes:
@aaronshekey Should we do the same thing with tags as we did with badges for hover states? Tags which are interactive have hover states; otherwise they don't. Thoughts?

aaronshekey:
Ah, I was just reminded. There are often times when tags will appear within a header or larger type. How will these render there? For example: https://stackoverflow.com/tags/ios/info

aaronshekey:
> Should we do the same thing with tags as we did with badges for hover states?

Yeah, I think this makes sense, though I think 99% of the time our tags will be links.

hellohynes:
> There are often times when tags will appear within a header or larger type. How will these render there? For example: https://stackoverflow.com/tags/ios/info

I've considered adding in sizes. At least a medium and a large.

PR #36, opened 2018-06-12
Badges & Number Counts

### LINKED ISSUE #34: Component: Badges

### LINKED ISSUE #35: Component: Number Counts

hellohynes:
This PR adds badges and number counts into Stacks. Fairly straightforward. This addition does represent a break from our current way of implementing tags in 2 ways:

 1. All classes have been updated (see table below)
 2. Badge images are no longer pulled in via a CSS sprite, but are saved out as individual files.

| Old Class | » | New Class |
| --------- | -- | --------- |
| `.badge` | » | `.s-badge` |
| `.badge1-alternate` | » | `.s-badge .s-badge__gold` |
| `.badge2-alternate` | » | `.s-badge .s-badge__silver` |
| `.badge3-alternate` | » | `.s-badge .s-badge__bronze` |
| `.badge-tag` | » | `.s-badge .s-badge__tag` |
| `.bounty-indicator` | » | `.s-badge .s-badge__bounty`  |
| `.user-show-new .user-panel .mini-counts` | » | `.s-badge .s-badge__votes`  |
| `.user-show-new .user-panel .mini-counts .answer-accepted` | » | `.s-badge .s-badge__answered`  |
| `.user-show-new .user-panel .mini-counts` | » | `.s-badge .s-badge__votes`  |
| `.user-show-new .user-rep .rep-amount .rep-up` | » | `.s-badge .s-badge__rep`  |
| `.user-show-new .user-rep .rep-amount .rep-down` | » | `.s-badge .s-badge__rep-down`  |
| `.top-bar .indicator-badge` | » | `.s-badge .s-badge__mini`  |
| `.top-bar .indicator-badge._important` | » | `.s-badge .s-badge__mini .s-badge__important`  |

## Open Questions

 1. Should we save out SVGs out individually or should we include them via CSS using base64? The downside to this is that we will need to generate multiple class names to deliver different images versus the current method which has one class name (`.s-badge--image`) which is concerned only with placing the icon—not about the icon itself.

Closes #34 
Closes #35


aaronshekey:
Most design systems refer to these elements either as tags or badges. They're all effectively the same element, a rounded rectangle with some text and possibly an image or a delete button. What do we think of calling these something arbitrary like `s-bubble` since tags and badges mean such strong things on Stack Overflow? _Then_ we use a modifier to specify whether it's a badge or a tag. Consider the following:

`<div class="s-bubble s-bubble__tag">c#</div>` A tag
`<div class="s-bubble s-bubble__badge">editor</div>` A badge
`<div class="s-bubble">208</div>` Just a number bubble

hellohynes:
My initial thought here is to not create a unifying class (such as `s-bubble`), for a few reasons:

1. We tried doing this recently with `.s-notify`, which on the surface seemed to support _all_ notification banners. Yet in retrospect we found that we needed to split out system banners (`.s-banner`) and hero banners (`.s-hero`) into their own areas. Granted banners are more complicated than tags or badges, but the there's a recent history here of overdoing unification.
2. One of the frequent confusions people had when we tried to do everything through `s-notify` is understanding when you had to modify a modifier. We don't allow double modifier classes such as `s-bubble__tag__gold`. As I've talked with people on the team, when components become too abstracted, it increases the learning curve and makes it harder for people to properly utilize them.
2. Using your proposed method, we would now have to add _3_ classes instead of 2 (current). A gold tag number count wouldn't be `s-tag s-tag__gold`, but now would be `s-bubble s-bubble__tag s-bubble__tag__gold`?
4. Another thought: I don't see a need to unify these right now. If the visual styles for tags and badges needed to be aligned more, sure. But they don't. I don't think it's a problem that tags and badges look the same. That might not always be the case. Keeping them distinct allows for changes to be made within each one without affecting the other.

lib/src/components/_stacks-badges.less
```diff
@@ -0,0 +1,124 @@
...
+    border-color: @border;
+    background-color: @background;
+    color: @color;
+
+    .s-tag--dismiss {
+        background-color: transparent;
+        color: fade(@color, 90%);
+    }
+
+    &:hover {
```
aaronshekey: Should only have a hover state when badges are a link? There are use cases for non-link badges. For example `s-badge__mini` shows up as a bounty count in our filters, but isn't itself a link, so it doesn't need a hover state. 🤔

lib/src/components/_stacks-badges.less
```diff
@@ -0,0 +1,124 @@
...
+    .badge-styles(transparent, @green-400, @white);
+}
+.s-badge__rep {
+    .badge-styles(@green-400, @white, @green-500);
+}
+.s-badge__rep-down {
+    .badge-styles(@red-400, @white, @red-500);
+}
+.s-badge__mini {
+    align-self: flex-start;
```
aaronshekey: Should we be so prescriptive with the alignment here? This seems like it'll be overridden a lot.

lib/src/components/_stacks-badges.less
```diff
@@ -0,0 +1,124 @@
...
+    padding-right: @su4;
+    padding-left: @su4 - 1;
+    font-size: @fs-fine;
+    line-height: 1.8;
+    .badge-styles(transparent, @blue-600, @white);
+
+    &.s-badge__important {
```
aaronshekey: In the documentation's list of classes, it's not clear that this should be paired with `.s-badge__mini`. Does it need to be?

docs/components/badges.html
```diff
@@ -0,0 +1,164 @@
...
+<span class="s-badge s-badge__tag">
+    <span class="s-badge--image"><img src="../{{ site.img }}/BadgeBronze.svg" aria-hidden="true"></span>
+    <span>swift</span>
+</span>
+{% endhighlight %}
+        <div class="stacks-preview--example">
+            <div class="grid gs4">
+                <span class="grid--cell s-badge s-badge__tag">
+                    <span class="s-badge--image"><img src="../{{ site.img }}/BadgeGold.svg" aria-hidden="true"></span>
+                    <span>css</span>
```
aaronshekey: I don't think we need the `<span>`s surrounding the text. We should included as little markup as possible in our examples.

docs/assets/img/BadgeSilver.svg
```diff
@@ -0,0 +1 @@
+<svg aria-hidden="true" class="svg-icon iconBadgeSilver native" width="16" height="16" viewbox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path fill="#8C9298" fill-rule="evenodd" d="M8 5c1.657 0 3 1.343 3 3s-1.343 3-3 3-3-1.343-3-3 1.343-3 3-3"/></svg>
```
aaronshekey: These won't have the `svg-icon icon[X] native` in production, so we should kill those classes and make sure they're still showing up how we want them to via `.s-badge--image`

docs/components/badges.html
```diff
@@ -0,0 +1,164 @@
...
+</span>
+<span class="s-badge s-badge__bronze">
+    <span class="s-badge--image"><img src="../{{ site.img }}/BadgeBronze.svg" aria-hidden="true"></span>
+    <span>8234</span>
+</span>
+{% endhighlight %}
+        <div class="stacks-preview--example">
+            <div class="grid gs4">
+                <span class="grid--cell s-badge s-badge__gold">
+                    <span class="s-badge--image"><img src="../{{ site.img }}/BadgeGold.svg" aria-hidden="true"></span>
```
aaronshekey: Can the class be applied directly to the `img` to reduce the markup?

lib/src/components/_stacks-badges.less
hellohynes: Oh, that's a tough one.

lib/src/components/_stacks-badges.less
hellohynes: Would we use `s-badge__important` on it's own?

docs/assets/img/BadgeSilver.svg
hellohynes: I was unsure how to do this because I need `native` so `svg-icon` CSS doesn't re-color them. These would be delivered by `Design/Icons` I would imagine, correct?

lib/src/components/_stacks-badges.less
aaronshekey: The docs kinda make it look like you _could_. Not saying we need to. The `s-badge__important` seems to be a modifier of `s-badge__mini`?

docs/components/badges.html
hellohynes: Sure

docs/assets/img/BadgeSilver.svg
aaronshekey: Just remove all the classes and they should display in their native colors. These won't be delivered by `Design/Icons`. They'll be static assets in our CDN within the sites folder.

lib/src/components/_stacks-badges.less
hellohynes: What if we had the `:hover` state only apply when you use an `<a>` tag for a badge? If you intend the badge to be a link, use a `<a>` tag—otherwise use a `<span>`?

aaronshekey: Yep, that's what I'm saying 👍

lib/src/components/_stacks-badges.less
```diff
@@ -0,0 +1,125 @@
...
+    padding: 0 (@su4 + @su2);
+    border: 1px solid transparent;
+    border-radius: @br-sm;
+    font-size: @fs-caption;
+    font-weight: normal;
+    line-height: 2;
+    text-decoration: none;
+    vertical-align: middle;
+    white-space: nowrap;
+    cursor: pointer;
```
aaronshekey: We don't need this `cursor: pointer`, since that'll come from the anchor tag :v:

lib/src/components/_stacks-badges.less
hellohynes: How would you change this?

PR #40, opened 2018-06-20
Remove font rem mixin

hellohynes:
This PR removes the font `.mx-fs(…)` mixin. As all supported browsers support `rem` units, there's no need for us to deliver a `px` fallback anymore.


lib/src/components/_stacks-modals.less
```diff
@@ -27,16 +27,11 @@
...
     bottom: 0;
     left: 0;
-    -webkit-box-align: center;
-        -ms-flex-align: center;
-            align-items: center;
-    -webkit-box-pack: center;
-        -ms-flex-pack: center;
-            justify-content: center;
+    align-items: center;
+    justify-content: center;
```
aaronshekey: We should double check the removal of these prefixes if this wasn't intentional.

balpha: That seems fine to me. I haven't checked, but none of these sound like something that's necessary in any browser we care about.

hellohynes: Yeah, I think these were leftover from the previous verbose postcss-less browser support we were using.

PR #43, opened 2018-06-21
Overall site IA and layout changes

aaronshekey:
We plan to add more sections to Stackoverflow.design. This PR begins to address _how_. As an added benefit, this work in progress is also responsive.

A few goals:
1. We need to add other sections beyond our product (eg. Email, Brand, Copy guidelines)
2. The site should be viewable in any viewport.
3. We should have a landing page that makes sense with multiple sections.
4. We should dogfood as much as possible.


hellohynes:
## Some IA thoughts

 1. Looking through other Design Systems, the website IA varies widely because the audiences differ widely. Some are made for enterprise environments—others for smaller internal product teams.
 2. Let's use language we're familiar with—not language others use.

## A Proposed IA

Before we moved repos, I had drafted a sample IA structure. This is still up for discussion, but here is the general idea we had: *Stacks is the Design System for Stack Overflow.* It is composed of the following areas:

 - Brand
   - Principles
   - Logos
   - Colors
   - Typography
   - Patterns
   - Illustration
   - Photography
   - Videography
 - Voice & Tone
   - *TBD*
 - Product
   - Overview
   - Changelog
   - FAQs
   - Variables
   - Utilities
   - Objects (Things like grids and layouts)
   - Components
   - Patterns
 - Email
   - Overview
   - Changelog
   - FAQs
   - Components
   - Patterns
 - Design Team Info

I'm having a hard time remembering more than this, but it's a solid start. Before we jump too far into implementation it would be good if we explored some visual designs. [Our current homepage](https://stackoverflow.design) could be sharper. We should explore some ideas there and on the documentation pages.

TedGoas:
I dig it Aaron, excited to see where this goes. I also remember and like @hellohynes's IA above, very flexible. Only thing we might watch out for a burying the most popular pages or hiding them behind another menu, though I'm not sure if that's an issue here.

aaronshekey:
I've changed the scope of this PR to focus on shipping a responsive layout. Punting on adding additional sections in this PR. Just a few outstanding items before this is shippable and our documentation is responsive:

- [ ] Fix search dropdown styling on smaller viewports
- [ ] Fix banner and notice presentation on smaller viewports

PR #44, opened 2018-06-22
make the .XXX-unset classes work in IE

balpha:
Stacks is providing a bunch of `.FOOBAR-unset` atomic classed. All of these classes literally set the `unset` value on the corresponding properties.

Since `unset` doesn't exist in IE, which Stacks has to support, this doesn't work. Fortunately, `unset` is just minor syntactic sugar for either `inherit` or `initial`, depending on whether the property is inherited or not.

Therefore, we can continue providing the Stacks user with `...-unset` classes that do the right thing, and that *also* work in IE.

This PR does that by replacing `unset` with `inherit` for `font-style`, `text-align`, `text-transform`, `white-space`, and `list-style-type`, and with `initial` for `display`, `box-sizing`, `vertical-align`, and `position`.

It does *not* change the class names; they still contain the word "unset", for the following reasons:

- The semantics are identical; these class *do* have the same effect as `unset`, except that they work in IE.
- If/when we drop IE support, we can change the values back, without having to modify a ton of existing useage.
- These classes *already exist*, so chances are they're being used. Removing (including renaming) them would be a breaking change.

I'm not sure about the documentation. I could imaging some sort of footnote on the impacted classes, like "this doesn't actually generate the `unset` value, which is not available in all supported browsers, but it behaves identically"). Open to opinions.


aaronshekey:
I think it'd be worth renaming all of these to `x-inherit` since our atomic classes are usually very literal.  We'd need to update any references to unset in the docs as well.

balpha:
I think it's okay in this case, because the semantics are still like `unset` -- the classes do *exactly* the same things that `: unset` does (namely, `inherit` if it's an inherited property, `initial` if not). Except that it also works in IE.

hellohynes:
image

balpha:
@aaronshekey @hellohynes description added with additional discussion items

hellohynes:
@balpha I added in a notice on the various pages that had unset classes list. Also I merged in `develop`. There were two new unset properties merged in last week.

PR #45, opened 2018-06-22
responsify `.fd-row` as well as the various `.fw-` `.ai-` and `.as-` classes

balpha:
Rationale:
- For `.fd-row`: all other `.fd-` classes were already responsive -- seems strange for this one not to be. And I've actually already had a need for it in practice.
- For the other ones: When building responsive layouts, changing flex wrapping and direction at a specific breakpoint is a common approach. *Some* of the flex classes were already responsive, but some important ones weren't. For example, if you switch to column direction via `sm:fd-column`, chances are that you *also* need to change the item alignment on the cross axis (which is now the vertical one). It was already possible on the main axis (the `.jc-...` classes were already responsive), but oftentimes you need to change both.


balpha:
ah thanks @aaronshekey, forgot to do that

PR #46, opened 2018-06-22
Docs update: Add in keyboard shortcuts for search

hellohynes:
Small upgrade for the docs site. This PR adds some simple keyboard shortcuts for Stacks search.

- Pressing `/` focuses on the search bar
- Pressing `ESC` while focused on the search bar blurs focus


aaronshekey:
Cool! This currently doesn't work on the homepage, probably because the JS just isn't included there?

hellohynes:
Huh. The JS files were never included in the homepage layout.

aaronshekey:
So... about that. Clicking a link in the sidebar won't work when the JS is included on the homepage, since it's trying to replace a content `div` that doesn't actually exist. We'll need to split the searching JS from the `main.js` and then include _that_ new file in both home and subpages. 😜

hellohynes:
(╯°□°）╯┻━┻

PR #47, opened 2018-06-25
Create page layout without table of contents in right aside

TedGoas:
Per our meeting last week, this creates a page template without the in-page table of content menu to the right. The main column is extended to take up this space.

I created [a dummy page](https://deploy-preview-47--stacks.netlify.com/resources/no-toc-test) to view these changes in a browser. This file won't be included in the final commit.

---

**Note:** I removed `.stacks-content` because it didn't seem to do anything except assign a `width` and `max-width`, both of which were changed.


aaronshekey:
Digging into this, the more I want to delete all the JS that generates the menu and just do this on the jekyll side. Easier said than done. Not a lot of prior art out there for that style of jekyll plugin.

Locally, I've fixed the error that occurs when the menu element doesn't exist, but our AJAX page loader doesn't support flipping between pages that _have_ the menu and pages that do _not_. It'll take a bit more work to go through that.

Any chance you could take a look, @balpha?

hellohynes:
@aaronshekey Can't we include the build menu js only on the pages that we want?

hellohynes:
> Digging into this, the more I want to delete all the JS that generates the menu and just do this on the jekyll side. Easier said than done. Not a lot of prior art out there for that style of jekyll plugin.

So funny story—that's how it used to be.

hellohynes:
@aaronshekey This is older, but this is the basic idea: ["Jekyll Liquid sub-navigation not properly selecting"](https://stackoverflow.com/questions/37279381/jekyll-liquid-sub-navigation-not-properly-selecting/37288977#37288977). We would still need some JS IMO to properly highlight the menu IMO. But this would generate the full menu at least.

hellohynes:
@aaronshekey Take a look at this PR when you can. This splits up the JS files so we aren't throwing errors about menus that don't exist.

aaronshekey:
@hellohynes You've got a double slash in there so the search script isn't loading `<script src="https://stackoverflow.design//assets/js/search.js"></script>` `{{site.url}}` has the trailing slash built in, so kill the `/` between `{{site.url}}` and `{{site.js}}` :v:

hellohynes:
@aaronshekey Weird thing is if I remove the slash locally, it breaks

Issue #62, opened 2018-07-11
Applying s-link to a button element should strip all default button styling


aaronshekey:
As reported by Dean, there are contexts where buttons need to appear as links. We can make links look like buttons, and we should support the corollary.

hellohynes:
@aaronshekey Can you describe further what you mean by, "there are contexts where buttons need to appear as links"? By default all `s-btn` appear as text links. Is these buttons sans padding?

PR #63, opened 2018-07-11
Allow s-link to be applied to button elements

aaronshekey:
This addresses #62.


rossipedia:
umm... shouldn't that be `s-link` not `s-btn`?

aaronshekey:
> umm... shouldn't that be s-link not s-btn?

👮

hellohynes:
If `s-link` == `s-btn`, then what's the point of `s-link`?

aaronshekey:
Lots of conversation in `#team-design` about this today.

We have an action that's in a banner, but the action is mutating and potentially dangerous, requiring it to be a `POST` within a form. 

image

In this case, this button needs to look and function like a link. We can make links look like buttons, so why not allow for buttons to look like links? [Bootstrap](https://getbootstrap.com/docs/4.1/components/list-group/#links-and-buttons) does this with their navigation component.

Ultimately, this is an edge case, but something Stacks didn't support.

PR #66, opened 2018-07-12
.s-anchors for styling all descendant <a> elements

balpha:
This idea was slightly controversial when we talked about it internally, so here's an example implementation as a basis for discussion.

This implementation allows one level of nesting, e.g.

```
<div class="s-allanchors s-allanchors__danger">
    <a href="#">dangerous</a>
    <div class="s-allanchors s-allanchors__default">
        <a href="#">plain ol' blue</a>
    </div>
</div>
```

You'll probably never need that, but it's possible.

More than one level is not supported (it's a question of luck, or rather, CSS ordering, whether it'll work). But if you find yourself needing that, you're *probably* doing something wrong?

Unlimited amounts of nesting *would* be possible by using custom properties (a.k.a. CSS variables), but we need to support IE11 for the time being.


balpha:
[Direct link to the new documentation](https://deploy-preview-66--stacks.netlify.com/components/links#Styling_all_anchor_elements_inside_a_container)

balpha:
Suggestion from @hellohynes: rename the class to just `.s-anchors` (works for me)

aaronshekey:
Just merged all the work that came before this, so I had to restructure the links page a little bit. What if we just called this helper `s-links`? There's a nice symmetry to `s-link` and `s-links`. Thoughts?

balpha:
Re: your changes to the docs page, are you intentionally talking about "child links"? This class doesn't just work with direct children, but with any decendants.

I also think the documentation should be very clear about applying to `<a>` elements specifically. "Links" in Stacks is something different.

PR #70, opened 2018-07-16
Add progress bars into Stacks

hellohynes:
This takes the `.s-progress-bar` work done within Teams and Profiles and adds it into Stacks using the `.s-progress` name. Some of the clever widths and border radii have been cleaned up.

This adds in the basic success (green) progress bar, brand (orange), info (blue), as well as gold, silver, and bronze badge progress bars.


lib/src/components/_stacks-progress-bars.less
```diff
@@ -0,0 +1,182 @@
...
+    border-radius: @br-sm;
+    background-color: @progressbar-fill-bgc;
+}
+
+
+//  ===========================================================================
+//  $   VARIATION PROGRESS BACKGROUNDS
+//  ---------------------------------------------------------------------------
+.s-progress__brand .s-progress--fill {
+    background-image: linear-gradient(to right, #F58907 0%, #F16720 100%);
```
aaronshekey: Should these be based on color variables?

docs/components/progress-bars.html
```diff
@@ -0,0 +1,121 @@
...
+title: Progress Bars
+description: A component that visually communicates the completion of a task or operation. It shows the task's progress and remainder.
+---
+<section class="stacks-section">
+    {% header {element: 'h2', text: 'Base Style'} %}
+    <p class="stacks-copy">The progress bars use background gradients for the fill color. In order to properly reveal the changing color of the bar, the progress "fill" is the actual background gradient. The actual bar is a gray mask that slowly shrinks to the right to reveal the background gradient underneath.</p>
+    <div class="stacks-preview">
+{% highlight html linenos %}
+<div class="s-progress">
+    <span class="s-progress--bar" role="progressbar" aria-valuemin="0" aria-valuemax="100" aria-valuenow="33" style="width: calc(100% - 33%)"></span>
```
aaronshekey: Is there a way to ship this without modifying the style directly?

docs/components/progress-bars.html
```diff
@@ -0,0 +1,121 @@
...
+        </div>
+    </div>
+</section>
+
+<section class="stacks-section">
+    {% header {element: 'h2', text: 'Fill Alternates'} %}
+    <p class="stacks-copy">In addition to the standard progress bar, there is also a brand (orange) and info (blue) progress bar color variation.</p>
+    <div class="stacks-preview">
+{% highlight html linenos %}
+TBD
```
aaronshekey: We should flesh this out before shipping.

docs/components/progress-bars.html
```diff
@@ -0,0 +1,121 @@
...
+{% endhighlight %}
+        <div class="grid gs16 gsy fd-column stacks-preview--example">
+            <div class="grid--cell">
+                <label class="s-label d-inline-flex my8 fw-normal" for="brandprogress">Brand</label>
```
aaronshekey: Do we need proper labels for this? Not sure what we gain by having them clickable / selectable since they can't be adjusted like form elements 🤔

lib/src/components/_stacks-progress-bars.less
hellohynes: These are specific HEX values that either pre-existed (and don't exist as color variables currently), or were identified within the Visual Guide.

docs/components/progress-bars.html
hellohynes: Not that I know of. This is based on the current implementation of progress bars where the bar is updated on page load and the width value is calculated in Razor.

docs/components/progress-bars.html
hellohynes: ¯\_(ツ)_/¯ I mainly used labels for the example. We should probably consider adding in a hidden label for all progress bars though to help with accessibility.

docs/components/progress-bars.html
balpha: I would say this is a valid use for inline styles. Moving this to classes would be pretty awkward, especially if the value changes dynamically via JS.

However, can we make it so that we really only set the percentage as a width, not 100%-x via calc? E.g. by putting the gradient as the background on the `s-progress`, having `s-progress-bar` with a transparent background, and `s-progress--fill` taking up the remaining space on the right (via flexbox)?

Actually, the fill could just be an `::after`, so you would only need one child element.

hellohynes: Oh that's an interesting approach. You might not need any child elements at all then. 🤔 

Let me try that out.

balpha: I don't think you can get rid of the one child element, since you can't set an inline style on a pseudo-element and thus couldn't set the width.

docs/components/progress-bars.html
```diff
@@ -0,0 +1,117 @@
+---
+layout: page
+title: Progress Bars
+description: A component that visually communicates the completion of a task or operation. It shows the task's progress and remainder.
+---
+<section class="stacks-section">
+    {% header {element: 'h2', text: 'Base Style'} %}
+    <p class="stacks-copy">The progress bars use background gradients for the fill color. In order to properly reveal the changing color of the bar, the progress "fill" is the actual background gradient. The actual bar is a gray mask that slowly shrinks to the right to reveal the background gradient underneath.</p>
```
aaronshekey: Liar! 😛

hellohynes: 🤭

PR #73, opened 2018-07-17
make s-input also gain focus styling via a .has-focus status class and (if supported) via :focus-within

balpha:
This makes it possible for custom-built components that *look* like a textbox but actually aren't (think: tag editor) to use `s-input` to style themselves *and* be able to get focus highlighting.

`:focus-within` makes that work automatically if the browser supports it; otherwise, the component can set the `has-focus` class via JS.

Note that `:focus-within` is not going to have any negative impact on the *usual* case where an `.s-input` is an actual `<input>`, so it's not necessary to exclude that case in the selector.


lib/src/components/_stacks-inputs.less
```diff
@@ -320,17 +320,32 @@ fieldset {
...
+
+    .s-input:focus,
+    .s-input.has-focus,
+    .s-textarea:focus,
+    .s-select > select:focus {
+        @focus-style();
+    }
+
+    // This needs to be separate from the one above, because in browsers that don't support :focus-within
+    // it's a syntax error, rendering the whole selector null and void.
```
aaronshekey: Of course, hahah.

Issue #94, opened 2018-08-09
Component: Tooltips


hellohynes:
We need to add in a basic tooltip style like so:

![image](https://user-images.githubusercontent.com/1544281/43919341-a120a444-9be3-11e8-9dd5-8aa375564499.png)

This would be similar to popups, but would use the `title=""` attribute for content instead of Javascript.

aaronshekey:
I argue that the popovers component should handle this use case.

PR #110, opened 2018-08-15
Tooltips and Popovers

hellohynes:
This PR addresses issues #94 and #109.

It componentizes the arrows so that they could be applied to a popover or tooltip. These are largely working _except_ **I don't have a great way (YET) of using classes to position items via CSS.**

The reason? Popovers and tooltips are not child elements of the item controlling them. Their siblings. Instead we will _have to use Javascript_. And while a library such as [Popper.JS](https://popper.js.org/) gives us a lot of what we need, it also adds too much bloat.

@balpha or Ben Kelly: Thoughts on how we could do this?


hellohynes:
@aaronshekey re: having tooltips have a different look — I do and here's why: it creates visual distinction and reinforces their purpose.

I reviewed a number of systems while working on this. My research found that systems were all over the place when naming these components. Some called them popovers. Others tooltips. Other menus. They tended to create one component that could be modified, which is a fine approach.

Yet as the more I work on and talk with people using Stacks, there's a increasing desire for components to have _some_ opinion about when, where, and how they should be used. Based on previous discussions the team has had around tooltips and popovers, I've come down with the following distinctions:

 1. **Tooltips** — Has black background / white copy. For text-only items which further describes an item. This would be used occasionally.
 2. **Popovers** — Has white background / dark text copy. Should be used for menus, dropdown menus, additional UI controls, or larger copy + image help panels. This would be used most of the time IMO.

I could see us splitting popovers further by adding dropdowns.

Another thought too is that tooltips, popovers, and dropdowns all have different accessibility concerns as well. Breaking them into different documentation areas allows us to deal with those concerns individually without needing someone to read all of the popover documentation.

hellohynes:
Also thanks for kudos. I picked up what you started in the `s-popover` branch and ran with it. 😎

bnickel:
One feature that tag watching needs (at least with its current design) is the ability to arbitrarily shift arrows to keep the popover reasonably within page bounds.

In the ideal scenario, the tag we hover over is situated far enough from the right or left side of the screen that we can center the popover immediately above or below it, but if the popover would spill into the left nav or beyond the right sidebar, we constrain its x-coordinates to fit within the page.  When we do that, we shift the arrow left or right so it remains centered on the tag:

```javascript
const centerX = ($el) => $el.offset().left + $el.outerWidth() / 2;
const arrowOffset = centerX($source) - centerX($popover);
$arrow.css('margin-left', arrowOffset + 'px');
```

tag watching arrow shifting

With the available options in this PR, the options would be to either let the content overhang, possibly outside of the window's bounds or opt for `.s-popover__top-left` which significantly shifts the popover's position.

Because the arrow is a `:before` pseudo-element, we don't have direct access to change its properties.  There are some solutions, like [using CSS vars](https://stackoverflow.com/a/46572990/860000) that works everywhere but IE11 or we could [inject styles](https://stackoverflow.com/a/7330454/860000).  But probably the cleanest option would be to have an additional configuration  `.s-popover__top_configurable` with a `.s-popover--arrow` child element that gets styled like `.s-popover__top:before`.

bnickel:
The existing CQL implementation has a similar issue on mobile where there's no optimal style without bounds checking logic.  A fully customizable approach would be nice, because then we could have a script like `show(popover, sourceElement, boundingElement)` that figures our the right placement for the popover and its arrow.

aaronshekey:
Excellent points, @bnickel. I believe [Popper](https://popper.js.org) has support for this out of the box if you look at the [Custom flip behavior](https://popper.js.org/#example3).

bnickel:
@aaronshekey Yep, looks like Popper is doing what I do (but more universally), using `<div class="popper__arrow" style="left: ..."></div>` to render the arrow.

bnickel:
Since I already have a couple JS implementations of popover placement, I'd be happy to write a reusable placement function when this is merged.

hellohynes:
@bnickel I hadn't thought about the need for arbitrarily shifting the arrows. In that case, it would be better to make `.s-arrow` it's own element rather than a pseudo element of the tooltip or popover itself. I'll change that back.

Also, could you point me to the production JS files that you created for tag watching? That's a fine place to start for this IMO. If you'd like to help, that would be great!

aaronshekey:
What I would like to see is Stacks including a JS bundle. This bundle would include:

1. Popper.JS as a dependency
2. Custom interactions that define behavior for things like `.s-tooltip`, `.s-popover`, and `.s-modal`.

Ideally, by adding the proper data attribute (or `js-*` class), you'd get this behavior for free. Much like Stacks classnames, I think these identifiers will likely have to be namespaced as well.

This JS bundle could be delivered via NPM the same way as our CSS. (eg. Version `0.21.0` brings `.s-popover` and introduces `stacks.min.js`.)

As for our prior conversations about [Stimulus](https://stimulusjs.org)... I think we could hold off on including another way of writing JS for now. I find the documentation to be a little opaque and perhaps regular ol' jQuery will work here (or vanilla!) `¯\_(ツ)_/¯ `

aaronshekey:
JavaScript bundling is being explored in #112 👍

aaronshekey:
So far the documentation on these is slightly confusing. The description of these seems a little backwards, right? We either need to mention that it's the _arrow_ that's being described by the class name, or the position of the popover itself.

screen shot 2018-08-22 at 11 55 12 pm

hellohynes:
@aaronshekey Popover documentation hasn't been updated yet. I was waiting on JS bundling to wrap up first.

PR #112, opened 2018-08-16
Start bundling JavaScript

aaronshekey:
This PR does a few things:

1. Re-organizes our `/lib` directory into `/lib/css/` while adding `/lib/js/`
2. Adds `stacks.js` where we'll write our behavior for components.
3. Adds Stimulus as a dependency
4. Adds grunt tasks to concatenate our dependencies and behavior JS into a single `stacks.min.js` that's shipped in `/dist/js/`
5. Includes our JavaScript bundle into the docs site's `<head>`.
6. Builds a first component—sortable tables.

This prepares us to be able to ship JavaScript for Code / Careers / prototypes. You'll be able to include `<link rel="stylesheet" href="https://unpkg.com/@stackoverflow/stacks/dist/css/stacks.min.css">` and `<script src="https://unpkg.com/@stackoverflow/stacks/dist/js/stacks.min.js"></script>` to receive all the _stuff_ automagically.


aaronshekey:
Here's where I'm struggling with Stimulus.

When working with modals, I'd like to embed a few example modals into the page arbitrarily. These modals will then need to be shown by a specific button, or multiple buttons.

Consider our use of modals on Stack Overflow. We have a million flag buttons that each need to show the same flag dialog.

Our embedded modals will look something like:

``` html
<aside class="s-modal" aria-hidden="true">
    <div class="s-modal--dialog">
        <h1 class="s-modal--header">…</h1>
        <p class="s-modal--body">…</p>
        <div class="grid gs8 gsx s-modal--footer">
            <button class="grid--cell s-btn s-btn__primary" type="button">…</button>
            <button class="grid--cell s-btn" type="button">…</button>
        </div>
        <button class="s-modal--close s-btn s-btn__muted">
            @Svg.ClearSm
        </button>
    </div>
</aside>
```

As far as I understand Stimulus, we would add a controller to that modal like so:

``` html
<aside class="s-modal" aria-hidden="true" data-controller"s-modal">
```

But now the targets and actions are scoped to the modal itself. How do I wire up the buttons to show _this_ particular modal. Do we create another controller and add that to `<html data-controller="modal-show-thing">`. And how do we connect `this` button with `that` modal given an arbitrary number of buttons and modals on a single page?

aaronshekey:
It looks like there has been some similar confusion in this [pair](https://github.com/stimulusjs/stimulus/issues/73) of [issues](https://github.com/stimulusjs/stimulus/issues/35). It seems we can potentially nest these controllers. 🤔

balpha:
@aaronshekey I think my answer would be, "Does it really matter?" It's not Stack's thing to worry about what causes the modal to be shown; Stacks is there to provide the modal style (and, if the user so desires, some basic functionality like the close button, lightbox background, etc.).

> We have a million flag buttons that each need to show the same flag dialog.

But all of these flag dialogs are going to have different content. It's not like it's *always* the same, identical dialog. The logic that fills these dialogs with their content is the one responsible for showing the dialog, not Stacks. For example, if we need to *wait* until the dialog is shown (as in the case of the flag dialog, which gets its data via an AJAX request), you *can't* just say, "Whenever X is clicked, show Y."

The flag button may very well be a *Stimulus* component (provided by the Q&A code), but not a *Stacks* component. The markup may look like this (all of this is just spitballing):

```html
<span data-controller="se-post-actions" data-post-id="12345">
    <button data-action="se-post-actions#flag"></button>
    ...
</span> 
```

Then `SEPostActionsController.flag()` may perform an AJAX request to get necessary post flag data, fill the modal with the appropriate content, and launch it. But all of this is custom code written by the Stacks user.

Now, if we *think* it makes sense to provide some sort of default "modal-showing button" mechanism, where the single purpose of a button is to show a dialog, then we can do that. I would imaginge the connection between the two to happen via an ID, because that's the standard way of doing this kind of thing in HTML (think `<label for="...">` or `aria-controls="..."`).

```html
<button class="s-btn" data-controller="s-modal-launcher" data-modal-id="mymodal">...</button>
...
<div id="mymodal" class="s-modal" data-controller="s-modal">...</div>
```

I feel the main question is, is this really a common enough pattern to provide such functionality in Stacks? A modal will almost always have custom functionality that the Stacks *user* (i.e., say, the Q&A JavaScript) provides.

hellohynes:
@balpha To make sure I'm understanding you correctly, you're suggesting that Stacks provides basic hooks for projects to write JS on top of, correct? For example, we could write the basic JS for opening and closing a modal, and then a project (like Q&A) could fit that function call into their JS—or they could use the basic open/close functionality we provide if they don't need to do anything custom. Am I understanding you correctly?

balpha:
@hellohynes Yes, exactly. Well put.

hellohynes:
@balpha Would we write our JS in Typescript file that could be imported then?

balpha:
@hellohynes I could go either way, TS or plain JS. But you're right, we should probably make a decision on that.

Not sure I understand what the "imported" part of your question refers to.

hellohynes:
I mean a TS( (or JS) file that could be compiled by the project using Stacks.

balpha:
Yeah, we should probably provide two versions:

1. A complete, stand-alone JS file ready to be included in an HTML page via `<script>` tag that works out of the box (initializes Stimulus, registers all components)
2. A set of source files that a project user could compile themselves in some way, possibly providing some ways of customization (not sure how that would work yet).

This would somewhat mirror what we do with CSS right now (1. a complete stand-alone ready-to-use CSS file, and 2. the complete LESS sources that can be compiled & customized).

aaronshekey:
Agree with _how_ we'll deliver this. It should mirror our delivery of CSS, as @balpha says.

> Now, if we think it makes sense to provide some sort of default "modal-showing button" mechanism, where the single purpose of a button is to show a modal, then we can do that.

I'm arguing we need this. There ought to be a default way to properly initialize our components with their assumed default behavior. I'd quickly be frustrated by needing to wire up custom functions to show a modal, or popover, or toast, especially if we're delivering how to close them, dismiss them, etc.

To me, leaving out the default initialization would be a little like not including a background color on a component we will know will always require a background color. We should choose the default background color, and allow for overrides in the same way our component / atomic classes interact.

Is that possible? I'm not sure 🤣 

Stimulus makes perfect sense to me _within_ the component, but I may need explore a proof of concept for the higher-level listeners. In my latest commit, I've explored what that _might_ look like, though I'm still a little fuzzy about how we wire specific buttons to specific modals within the component. Something to noodle on.

balpha:
@aaronshekey I think we're talking about different things here. I wasn't talking about initializing a component with its default behavior -- that's exactly what a Stimulus controller would take care of.

Hynes and I were talking about specifying *in the markup* on, say, a button, that clicking this button will open a particular modal (see the last code block in [this comment](https://github.com/StackExchange/Stacks/pull/112#issuecomment-415948542)).

My argument was that *that* kind of thing would be rare, because almost always a button will do more than *only* open a modal that already exists (hidden) in the page.

aaronshekey:
> Hynes and I were talking about specifying in the markup on, say, a button, that clicking this button will open a particular modal

Sorry for being unclear. I think we need to do exactly this. Every modal will need to be shown, so I think we could build that default case that users can amend.

aaronshekey:
@balpha This now has the latest `develop` in it and is ready for your table component javascript 🎉

aaronshekey:
Todo:

- [ ] IE polyfill delivery
- [ ] Coordinate where the NPM bundling occurs
- [ ] Re-write the Stacks updater in Core / Careers to move JS files around
- [ ] Update Documentation to reference JS being included

docs/product/components/tables.html
```diff
@@ -631,6 +633,76 @@
...
+                <div class="grid--cell lh-xl">
+                    <b>Note:</b> Using <code class="stacks-code bg-powder-200">js-…</code> classes is not really the optimal way of doing this, and will probably replaced with something better eventually. When that happens, the <code class="stacks-code bg-powder-200">js-…</code> mechanism will be deprecated but continue to be supported for a while, so you have ample time to update things.
+                </div>
+            </div>
+        </li>
+    </ol>
+
+    <p class="stacks-copy">By default, the data is sorted by the content of the cell. If you need to use a different value, for example because your cell contains a human-readable date, add a <code class="stacks-code">data-s-table-sort-val</code> attribute to the cell.</p>
+
+    <p class="stacks-copy">If a column contains any data that is not an integer, the data will be sorted lexicographically. Otherwise it will be sorted numerically, with empty cells being considered the lowest number.</p>
```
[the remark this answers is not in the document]
balpha: There's a subtle difference (at least in how developers tend to use the words). Alphabetically means "a" comes before "Z", but lexicographically "a" comes after "Z" because lowercase letters have higher codepoints.

docs/product/guidelines/javascript.html
```diff
@@ -5,40 +5,37 @@
...
 
-    <p class="stacks-copy">While Stacks is first and foremost a CSS library, it also provides commonly used functionality for some components via JavaScript code. Using this functionality is optional. If you only need the styling parts of Stack, you’re free to ignore the provided JavaScript. The converse is not true: The JavaScript components work under the assumption that the Stacks CSS is available.</p>
+    <p class="stacks-copy">While Stacks is first and foremost a CSS library, it also provides commonly used functionality for some components via JavaScript. This functionality is optional. If you only need the styling parts of Stack, you’re free to ignore the provided JavaScript. The converse is not true: The JavaScript components work under the assumption that the Stacks CSS is available.</p>
 
-    <p class="stacks-copy">In order to use the JavaScript components, do the following.</p>
+    <p class="stacks-copy">Stacks JavaScript is currently included within various Stack Overflow projects automatically. If you’re working on a Stack Overflow project, chances are it’s already available for you! If not, reach out to us and we’ll work on getting it setup.</p>
```
balpha: This isn't true. Shouldn't we wait before claiming this until it's actually the case?

docs/product/guidelines/javascript.html
```diff
@@ -5,40 +5,37 @@
...
+        <li>Include the file <code class="stacks-code">dist/js/stacks.min.js</code> in your page. For example, if you use the unpkg CDN, add the tag <code class="stacks-code">&lt;script src="https://unpkg.com/@stackoverflow/stacks/dist/js/stacks.min.js"&gt;&lt;/script&gt;</code> to your HTML. See <a href="{{ "/product/guidelines/using-stacks" | relative_url }}">Using Stacks</a> for more information on Unpkg and installing Stacks via NPM.</li>
 
-        <li>If you need to support Internet Explorer 11, you also have to include the file <code class="stacks-code">dist/js/stacks.polyfills.min.js</code>. This file has to be included before the main JavaScript file.</li>
+        <li>If you need to support Internet Explorer 11, you also have to include a polyfill <code class="stacks-code">dist/js/stacks.polyfills.min.js</code>. This file has to be included before the main JavaScript file.</li>
```
balpha: For the record, the official spelling in copy text is "unpkg" with a lowercase U. I have no strong feelings about this though 😊

lib/js/stacks.js
```diff
@@ -0,0 +1,43 @@
...
+            } else {
+                Controller.prototype[prop] = data[prop];
+            }
+        }
+        return Controller;
+    }
+
+    Stacks.addController = function addController(name, data) {
+        if (!/^s-/.test(name)) {
+            throw "Stacks-created Stimulus controller names must start with \"s-\".";
```
bnickel: Is the thought here that Stacks would have its own Stimulus application just for Stacks components and consuming sites would have their own Stimulus application that they'd add their own controllers to?  For example, if a site wanted a component called `so-comment`, they'd create their own application and use something like `StackOverflow.addController('so-comment', { ... })`

bnickel:
With regards to how we show (or hide!) a modal, that's gonna be tricky no matter how you slice it.  Imagining we have an application `stackApplication` and a method `show`, our flag function could look like this:

```javascript
StackOverflow.addController('so-comment', {
    flag: async function() {
        let dialogElement = await loadDialog();
        stackApplication.getControllerForElementAndIdentifier(dialogElement, 's-modal').show();
    },
    ...
});
```

That dialog may have two controllers, the Stacks `s-modal` controller and the StackOverflow `s-comment-flagging-modal` controller (that's possible).

Then the submit action would be something like:

```javascript
StackOverflow.addController('so-comment-flagging-modal', {
    submit: async function() {
        let result = await submitFlag();
        if (result.success) {
           stackApplication.getControllerForElementAndIdentifier(this, 's-modal').hideAndRemove();
        }
    },
    ...
});
```

It would be good to see this fleshed out more.

lib/js/stacks.js
bnickel: If we do have multiple applications, is Stimulus smart enough to ignore actions like `data-action="so-comment-flagging-dialog#submit"` in applications that don't have a `s-comment-flagging-dialog` controller or will it spit out a warning/error?  Do we need to namespace `data-so-action` and `data-s-action` in the schemas?

balpha: I don't have a a settled opinion on how this works eventually.

The reason that this exception is thrown here is that I don't want anyone to rely on a JavaScript API that isn't supported. We have a tendency to move from "don't use this function, it's an internal thing" to "this function cannot be removed or changed ever" pretty quickly.

In step 1, Stacks offers ready-made Stimulus controllers (or, in the *very* first step, one such controller), nothing else.

docs/product/components/tables.html
```diff
@@ -631,6 +633,76 @@
             </table>
         </div>
     </div>
+
+    {% header h3 | JavaScript sorting %}
```
JonHMChan: At some point, I'd revisit the documentation for these components and make the API clearer. The step-by-step guidance is nice, but it'd be really nice to see something like a table with the available `data-` attributes, appropriate values, and what elements to apply them to in a component. I could glance at it quickly and go.

Not a blocker for getting JS out, but something to consider for the future.

Gruntfile.js
```diff
@@ -27,59 +27,149 @@ module.exports = function(grunt) {
...
+                }
+            },
+            stacks_partials: {
                 // these files are immediately deleted because we don't distribute them directly, but we must ensure
                 // that they compile (e.g. if a static file accidentally accesses the dynamic options, the full bundle will compile
                 // fine, but stacks-static alone will blow up)
                 files: {
-                    'lib/css/stacks-static.css': 'lib/src/stacks-static.less',
-                    'lib/css/stacks-dynamic.css': 'lib/src/stacks-dynamic.less',
+                    'tmp/css/stacks-static.css': 'lib/css/stacks-static.less',
```
aalear: Not 100% sure how this works here, but should this be `'dist/css/stacks-static.css'` to match the .gitignore above?

balpha: No, these are compiled into `tmp/ ` and then immediately deleted again ([here](https://github.com/StackExchange/Stacks/blob/feature/javascript-bundle/Gruntfile.js#L160)) -- we don't ship compiled versions of these, they just exist so that Core can compile the two parts separately.

Compiling them right here (into the temp folder) only serves the purpose of making sure that they *do* compile separately.

docs/product/components/tables.html
aaronshekey: Huge 👍 on this one. This Tables page could use both an available classes table, as well as a new JS data attributes table. On it!

bnickel:
@balpha I was responding to something @aaronshekey said earlier.  I'm looking at the PR from the perspective of:

1. **How can we consume Stimulus in Stack Overflow?**  What will our controllers look like?  How will our controllers interact with Stacks controllers?  When we go about implementing Stack Overflow controllers, will there be any conflicts that result in a breaking API change in Stacks?
2. **How will controllers and logic I've already implemented in the SO codebase for Stacks CSS port to Stacks JS?** Here, I'm thinking about `s-modal` and `s-toast`.  How will I interact with with these other modules both from Stack Overflow Stimulus code as well as Stack Overflow jQuery code?

I ended up writing a demo to test my inline comment at https://chrome-pine.glitch.me.  It looks like everything is fine, Stimulus applications just ignore names they don't recognize.  We can have separate application spaces for Stack Overflow and Stacks controllers, and a Stacks controller can live inside a Stack Overflow controller.

For the record, I am **ecstatic** about the idea of using Stimulus in our codebase.  This is 🔥🔥🔥!

lib/js/stacks.js
```diff
@@ -0,0 +1,60 @@
...
+    Stacks.controllers = Object.create(null);
+
+    function StacksController () {
+        Stimulus.Controller.apply(this, arguments);
+    }
+    StacksController.prototype = Object.create(Stimulus.Controller.prototype, {
+        constructor: { value: StacksController, enumerable: false, writable: true }
+    });
+    StacksController.prototype.getElementData = function(element, key) {
+        return element.getAttribute("data-" + this.identifier + "-" + key);
```
bnickel: FWIW, there's a "private" function, `this.data.getFormattedKey` that does the same conversion, only it'll also desnake key (eg. `fooId -> data-s-table-foo-id`).  Really, it might be worth a PR to Stimulus to add an optional element parameter to `this.data.get` so you could do `this.data.get('some-attribute', this.someTarget)` instead of ` this.getElementData(this.someTarget, 'some-attribute')`

docs/product/components/tables.html
```diff
@@ -631,6 +633,76 @@
...
+                        … <!-- CODE ONLY -->
+                    </th>
+                </tr>
+            </thead>
+            <tbody>
+                <tr><td>Winter</td><td data-s-table-sort-val="12">December</td><td>2</td></tr>
+                <tr><td>Spring</td><td data-s-table-sort-val="3">March</td><td>13</td></tr>
+                <tr><td>Summer</td><td data-s-table-sort-val="6">June</td><td>25</td></tr>
+                <tr><td>Fall</td><td data-s-table-sort-val="9">September</td><td>13</td></tr>
+                <tr data-s-table-sort-to="bottom" class="fw-bold">
```
bnickel: Why isn't this row in a `<tfoot>`?  Is this just to demonstrate the functionality?

lib/js/stacks.js
balpha: I want to play with it a bit whether this way of doing things even makes sense. The Basecamp people are very careful with their choice of what APIs they add (which I appreciate), so maybe this won't even turn out to be a good idea in practice. We'll see :)

docs/product/components/tables.html
balpha: Yep, that's all.

balpha:
> and a Stacks controller can live inside a Stack Overflow controller.

@bnickel I assume you mean the controllers can live inside the same application?

Yeah, maybe. But for the time being, any JavaScript API that Stacks creates (like `Stacks.addController`) are purposefully undocumented and shouldn't be used outside of Stacks.

This is too new for us to commit to a specific set of APIs, and once it's given the "official blessing" by documenting it, we can no longer break it.

balpha:
@bnickel BTW, my current (totally not finalized) thinking about communication between Stacks components and other JS, where data attributes aren't enough, is to use the default mechanism for communication with DOM elements: Events.

bnickel:
@bnickel I mean you can have:

```
<div data-controller="so-reloader">
    <div data-controller="s-table">...</div>
</div>
```

Two different applications with overlapping or nested trees are are working correctly.  I was worried there might be conflicts when the `so-reloader` contained a `data-target="s-table.whatever"` but it handled it fine.  So, if someone at Stack Overflow wanted to start using Stimulus for their own stuff, of if Stacks wanted to replace their use of jQuery with Stimulus, they shouldn't have to worry about conflicting with whatever the Stacks library is doing.

Issue #69, opened 2018-07-13
Notices should respond to different viewports


aaronshekey:
Our toast notifications don't appear centered horizontally at smaller viewports.

bnickel:
There are a couple problems:

1. The toast isn't actually centered.  It's left is at `50% - 16rem` and its right is at `50% + 26rem`.  I'm assuming this is a correction made for the left sidebar, but the left sidebar isn't always there.  Not knowing enough about the Stacks design philosophy, I'd say it should be truly centered with `margin-left: -21rem` and then consuming sites should override its `left` rule in cases when left/right nav affects the available space.  For SO, that would look something like:

    ```less
    IF_LEFT_NAV_VISIBLE {
        .s-notice__toast {
            left: calc(50% +82px);
        }
    }
    ```

2. We aren't responsive.  I would just say if viewport width < 44rem, unset width and left-margin, and instead pin right and left to an appropriate distance from the screen edge.

hellohynes:
Good suggestions @bnickel

PR #115, opened 2018-08-21
Center toasts using flex

aaronshekey:
This PR is a work in progress meant to address #69. Instead of relying on a left position of 50%, which can often lead to blurry half-pixel alignments, this makes `s-notice` a child of `s-notice-toast`.

I realize this is a little weird with our naming conventions, but you can see what I'm trying to attempt in the updated documentation on this branch.


bnickel:
When the toast is active, I'm not able to click on content to the right or the left of it.  This can be problematic on SO, since the default duration of toasts has historically been 20 seconds.

aaronshekey:
@bnickel I've solved this using pointer events. :v:

PR #116, opened 2018-08-21
A few minor email refinements

### LINKED ISSUE #100: "Grab the Code" button should display the code
Each email template has a button to grab the code, but it launches a window displaying the rendered email instead of the actual code os you have to `view source` to get the code. Both Hynes and Pawel mentioned this.

This button should be updated to show the source code.

---

Possible solution: I've used [this thing](https://pypi.org/project/githubusercontent/) to do this in other projects. I don't think we have to install anything, only follow the project's `url` conventions.

aaronshekey:
This PR closes #100, removes transitions from various elements, and rounds the corners properly on email containers. What do we think, @TedGoas?


TedGoas:
Why'd we remove transitions on buttons & tags? I think emails should match the site, just curious.

---

Thoughts about rounding the email frame:
* [Mobile clients that don't support `<style>`](https://www.campaignmonitor.com/css/style-element/style-in-head/) won't apply the `bar0` class and will have round corners. I think we can take a different approach to work around this, though.
* The bottom section of an email can appear in different locations in the code, so it's up to the developer to 1. realize this, and 2. know where the rounded corners should be. This is my main concern, as it's a small thing that can be easily forgotten or overlooked.
* I'm not sure how this will play with foreground and background images (eg. will they get rounded/clipped if their parent is too? On the web, yes. In email, 🤷‍♀️). Needs testing.
* Why do this? Honest question, not trying to be a jerk. Are we rounding corners on major containers now?

aaronshekey:
> Why'd we remove transitions on buttons & tags? I think emails should match the site, just curious.

On Stack Overflow, we have transitions but those are inherited from legacy CSS. On Stacks-only implementations of links and buttons, there are no transitions. In my opinion, they were too slow and didn't add to the experience. There were also moments when transitions were applied to `all` instead of individual CSS properties causing things to be unintentionally animated.

> Why do this?

Every component in Stacks is rounded. As far as I understand, avoiding corner rounding in the past was due to technical limitations, not an aesthetic choice. In all new components and UI refactors, we've been rounding corners. You can see this on sites like http://apple.stackexchange.com. To my eye, the sharp corner, especially when paired when a medium button looks like an oversight and not an aesthetic choice.

In this case, I've chosen `@br-md` since the box is nesting `@br-sm`s inside it.

Of course I defer to your judgment to find a reliable way to actually build it _if_ one exists, but I think it's worth it to match all the other components in Stacks. :v:

TedGoas:
👍 I'm happy to build and test `border-radius` and see if we can get this working reasonably.

PR #117, opened 2018-08-22
Add Copy Code JS

hellohynes:
This PR adds in a copy code button to all examples on Stacks.

Example:

screen recording 2018-08-22 at 10 30 am-1


aaronshekey:
@hellohynes I hate to be a stickler on this one but I've found a few issues that may be worth reverting for _now_. 

- [ ] Copy button doesn't show up via AJAX page load, only on hard refreshes.
- [ ] When scrollbars are visible, it doesn't look great.

docs/assets/js/code-copy.js
```diff
@@ -0,0 +1,52 @@
+$(document).ready(function() {
+    var addCodeButton = $(document).find("figure.highlight");
+    var addTarget = $(document).find("td.code > pre");
+    var button = "<button class='ps-absolute baw0 bar-sm t0 r0 p8 bg-black-900 fc-white c-pointer js-copy-code' role='button' title='Copy code'><svg class='svg-icon iconCopy' xmlns='http://www.w3.org/2000/svg' width='24' height='24'><path d='M16 1H4c-1.1 0-2 .9-2 2v14h2V3h12V1z'/><path d='M15 5H8c-1.1 0-1.99.9-1.99 2L6 21c0 1.1.89 2 1.99 2H19c1.1 0 2-.9 2-2V11l-6-6zM8 21V7h6v5h5v9H8z'/></svg></button>";
```
aaronshekey: It might be worth adding this as an include since it'll get repeated everywhere.

docs/assets/js/code-copy.js
```diff
@@ -0,0 +1,52 @@
...
+    var toast = "<aside class='s-toast js-toast-popup'><div class='s-notice js-toast-notice' role='status' aria-hidden='true'><div class='grid jc-space-between ai-center'><div class='grid--cell js-toast-copy'></div></div></div></aside>";
```
aaronshekey: We may consider adding this toast into every page via `header.html`, since it's possible it'll get used on every page.

PR #118, opened 2018-08-22
Additional width & height classes

### LINKED ISSUE #74: Precentage-based min/max-width classes?
Right now we have [atomic min/max-width classes](https://stackoverflow.design/base/width-height#max-width) only for some (seemingly arbitrary, although I assume there's reasoning behind them) fixed widths, with the exception of 100%.

Especially in a responsive world, percentage-based is something I need much more often than fixed values.

What do we think about adding those?

There would be a bit of naming discussions required. Right now `.wmx1` through `wmx12` refer to ~105px steps, while `.wmx100` refers to a percentage. So there's precedent for mixing "units" here.

Since a 10% min/max width is probably rarely helpful, we probably wouldn't have overlap if we go with, for example, `.wmx25`, `.wmx33`, `.wmx50`, `.wmx67`, `.wmx75`, or something like that.

Or we could be explicit like under my original suggestion for the naming scheme, e.g `.wmx-50pc` -- but we haven't done that anywhere else.

Any thoughts?

aaronshekey:
This PR:

1. Adds a few more responsive classes to min / max widths.
2. Adds static height classes.
3. Adds min heights / max heights and makes some responsive.
4. Adds some percentage-based min / max widths and closes #74 

Have a look!


docs/_data/product/atomic-definitions.yml
```diff
@@ -222,31 +222,46 @@ width-static:
...
     output: "max-width: 211px;"
     responsive: false
+  - class: "wmx25"
+    output: "max-width: 25%;"
+    responsive: false
   - class: "wmx3"
     output: "max-width: 316px;"
     responsive: false
+  - class: "wmx33"
+    output: "max-width: 33%;"
```
ludwiczakpawel: I was thinking about values like `33%` and `66%` – is having them that common? Having three elements with width `33%` means that there will be `1%` remaining. It's usually not a lot, but it feels to me like an icon with blurred pixels :D 

A bit crazy, but how about using `calc(100% / 3)` instead (and `calc(100% / 3 * 2)` for 66%)?

aaronshekey: Perhaps we don't need thirds at all, but I think relying on calc would totally work in this case. I threw these in there for the conversation, but I'm leaning away from them, especially since our grid has things like `grid__allcells4` and the like. 🤔 

@hellohynes @balpha Y'all have a strong opinion on keeping 1/3 and 2/3 percentage widths? I'm leaning toward just dropping them off, especially since we don't have `w33`. Oops!

balpha: No strong opinions here, happy with dropping them.

PR #121, opened 2018-08-28
Email: Better Team Name/Avatar

### LINKED ISSUE #97: Email: Team avatar should be height / width instead of padding
Pawel pointed out the current [Team icon](https://stackoverflow.design/email/patterns#team-identification) is padding-based. Since letters are shaped differently, this can make the icon seem "off."

Making it height/width based would be better.

TedGoas:
This makes the Team avatar/icon square (instead of being shaped by the letter shape and padding). It also makes the `Team Name` text appear more consistently with the site:

screen shot 2018-08-28 at 11 35 15 am

---

Resolves #97


TedGoas:
@aaronshekey Shoot! I forgot about those. I'll change, thanks!

aaronshekey:
> I forgot about those

Think it's worth making those look more like a proxy ala this [Dribbble shot](https://dribbble.com/shots/2962967-Code-review-illustrations)? Sorry to reference something _I_ did. 🤦‍♂️

TedGoas:
@aaronshekey Yep, there's another to-do about upgrading these graphics somehow. What's currently there doesn't look right, but boxing out everything doesn't really convey what's _actually in_ the template. Gotta think about this some more, but don't want it to block us.

PR #124, opened 2018-08-30
Validation states links

hellohynes:
This PR alters all `<a>` tags within `.s-input-message` to match the current validation state color. It also adds `text-decoration: underline` to all `s-input-message a` tags.

## Examples

**Warning**

image

**Danger**

image

**Success**

image


lib/src/components/_stacks-inputs.less
```diff
@@ -409,7 +413,15 @@ fieldset {
     }
 
     .s-input-message {
-        color: shade(@yellow-600, 25%);
+        color: #7C6400;
```
ludwiczakpawel: Do you think this could use `@yellow-800` instead of hex value? It's almost the same color.

And for `a`, one line below, maybe use `@yellow-900`?

Otherwise, LGTM.

Issue #81, opened 2018-07-26
Component: Data Tables


TedGoas:
# Problem
[Our current tables](https://stackoverflow.design/components/tables) are don't support complex data sets out of the box. Alone, `s-table` is pretty barebones. Modifiers can be used to push text around, but there aren't codified patterns or recommendations for tables with more complex data. Eg. more than words, data that can be sorted or acted upon. We're currently going at it piecemeal and @UXPiper and I would like to not do that anymore.

# Proposal
Codify a few more patterns to support some use cases that we've come up against with more complex data tables.

## What to include:
- [ ] Improved column headers
- [ ] Sortable column headers
- [ ] Actions in table cells, and how to display them (eg. in a popover menu or not)
- [ ] Graphics in table cells (icons, badges, mini-graphs, etc.)
- [ ] Blank state ([related to this](https://5ad67b65e39e7c12b475dde8--stacks.netlify.com/components/blankslates))

## What _not_ to include:
* Title and text around the table.
* Search, sorting, and filtering
* Pagination, show more/less rows

---

Examples in the wild:

<img width="856" alt="screen shot 2018-07-26 at 12 22 21 pm" src="https://user-images.githubusercontent.com/1172461/43275778-c41631c6-90d0-11e8-8108-9dae148ce06b.png">

<img width="846" alt="screen shot 2018-07-26 at 12 21 57 pm" src="https://user-images.githubusercontent.com/1172461/43275792-cbfc118a-90d0-11e8-8ea9-58d3d035931e.png">

<img width="1045" alt="screen shot 2018-07-26 at 12 23 10 pm" src="https://user-images.githubusercontent.com/1172461/43275799-d3e93ab2-90d0-11e8-90c0-d054b3f03a30.png">

<img width="857" alt="screen shot 2018-07-26 at 12 31 40 pm" src="https://user-images.githubusercontent.com/1172461/43275803-d8cd33da-90d0-11e8-9b5e-2d0205ae7fa0.png">

TedGoas:
@hellohynes @aaronshekey @UXPiper I created a branch for this... couple questions before I start.

1. Should we add this on to [the existing tables page](https://stackoverflow.design/components/tables)? I think _yes_, but am open if you think it should be it's own component.

2. What should we call these more complex tables?
[ ] Dashboard table
[ ] Data table (well _all_ tables are data tables)
[ ] Complex table
[ ] something else

hellohynes:
[...]

PR #129, opened 2018-09-07
Extending tables

TedGoas:
This addresses everything in #81 and a whole lot more. We upgraded `s-table` and added a bunch of options.

---

**Note:** the JavaScript bits haven't been hooked up yet, so the table sorting and popover menu examples here don't work.


lib/src/components/_stacks-tables.less
```diff
@@ -46,21 +46,39 @@
         padding: @su8;
         border-top: 1px solid @brc-lighter;
```
hellohynes: All `@brc-lighter` variables should be phased out. We rarely use them.

lib/src/components/_stacks-tables.less
```diff
@@ -46,21 +46,39 @@
...
-        vertical-align: top;
+        vertical-align: middle;
+        border-right: 1px solid @brc-lighter;
+        &:last-child {
+            border-right: none;
+        }
+    }
+    th {
+        font-weight: normal;
+        color: @fc-dark;
```
hellohynes: If you set the font color to dark here, there's no need to restate the color below in `thead th`.

lib/src/components/_stacks-tables.less
```diff
@@ -115,22 +133,61 @@
...
+        tr:last-child td {
+            border-bottom: none;
         }
-
         tbody + tbody {
-            border-top: 2px solid @brc-darker;
+            border-top: none;
         }
     }
+    &.s-table__borderless-x {
```
hellohynes: I would change this name to `.s-table__bx0`. It brings it more line in with how we do directional things like this with padding and margins.

lib/src/components/_stacks-tables.less
```diff
@@ -197,6 +254,34 @@
...
+            background-color: transparent;
+            border-top: none;
+            border-right: none;
+        }
+    }
+}
+
+//  $$  Sortable Table Columns
+//  ------------------------------------------------------------------------
+.s-table__sortable {
```
hellohynes: I would make this a table-level class. The state `.is-sorted` would apply to the `th` still.

hellohynes:
@TedGoas All those items I asked you to change? I went ahead and did them. Let me know if I incorrectly changed something. Thanks.

docs/product/components/tables.html
```diff
@@ -29,9 +28,59 @@
...
+                    <tr>
+                        <th scope="col">Name</th>
+                        <th scope="col">Reputation</th>
+                        <th scope="col">Joined</th>
+                        <th scope="col">Last Seen</th>
+                    </tr>
+                </thead>
+                <tbody>
+                    <tr>
+                        <th><a>Aaron Shekey</a></th>
```
ludwiczakpawel: `tbody > tr > th` – is this correct use of `th`? I thought it can be used only in `tbody`..? I'm not sure to be honest, just thought I'll point it out. Feel free to ignore if I'm wrong.

docs/product/components/tables.html
```diff
@@ -446,40 +507,332 @@
...
+    <div class="stacks-preview">
+{% highlight html linenos %}
+<table class="s-table s-table__sortable">
     <thead>
         <tr>
-            <th scope="col" class="ta-left">Listing</th>
-            <th scope="col" class="s-table--cell2">Views</th>
-            <th scope="col" class="s-table--cell2">Apps</th>
-            <th scope="col" class="s-table--cell2">App CTR</th>
+            <th scope="col" class="is-sorted">Listing @Svg.ArrowDownSm</th>
```
ludwiczakpawel: Should we use any `@Svg.xxx` in Stacks code? It's dependent on our backend right now so feels like kinda _dependency_, and since Stacks supposed to be public, maybe we should skip it as well? Or include similar banner as on [Icons page](https://deploy-preview-129--stacks.netlify.com/product/resources/icons).

docs/product/components/tables.html
hellohynes: `th`s are used for any header type cell regardless of `thead`, `tbody`, or `tfoot`. ([Reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/th))

ludwiczakpawel: Oh, I didn't know that, thanks!!

docs/product/components/tables.html
hellohynes: The examples use the liquid Svg helper noted on the Icons page. The primary audience for Stacks is SO employees, so I see no problem listing the SVG helper as we use it in the product within our code snippets. Also this is a pattern we use throughout Stacks.

docs/_data/product/tables.yml
```diff
@@ -0,0 +1,25 @@
...
+    output: 58.3333333%
+  - class: .s-table--cell8
+    output: 66.6666667%
+  - class: .s-table--cell9
+    output: 75%
+  - class: .s-table--cell10
+    output: 83.3333333%
+  - class: .s-table--cell11
+    output: 91.6666667%
+  - class: .s-table--cell3
```
aaronshekey: This should be `s-table--cell100` or something, right?

lib/src/components/_stacks-tables.less
```diff
@@ -197,17 +151,172 @@
...
+
+        //  This makes the border transparent, so we need to use whitespace
+        //  to achieve the same effect a 2px gray border achieves.
+        tbody + tbody {
+            border-color: transparent;
+            border-width: @su12;
+        }
+    }
+    //  --  X-AXIS
+    //      Clear out the y-axis colors
```
aaronshekey: Is this comment correct?

lib/src/components/_stacks-tables.less
```diff
@@ -197,17 +151,172 @@
...
+            }
+        }
+        tfoot th,
+        tfoot td {
+            border-bottom-color: transparent;
+        }
+    }
+
+    //  --  Y-AXIS
+    &.s-table__by0 {
```
aaronshekey: How necessary are these classes? These tables, especially this one, are really awkward to me, aesthetically.

lib/src/components/_stacks-tables.less
aaronshekey: These class names seem backwards to me. If I'm applying `bx0`, I expect there to be no horizontal borders, but the documentation has the opposite?

docs/product/components/tables.html
```diff
@@ -204,34 +212,21 @@
             </table>
         </div>
     </div>
+</section>
 
-    {% header {element: 'h3', text: 'Condensed'} %}
-    <p class="stacks-copy">The condensed table does pretty much what it sounds like it would do: condenses the padding and line-height so that the table has a tighter feeling to it. This is suggested for data tables such as found within reports that have multiple records.</p>
+<section class="stacks-section">
+    {% header {element: 'h2', text: 'Spacing'} %}
+    <p class="stacks-copy">A table's padding can be changed to be more or less condensed.</p>
```
aaronshekey: Apostrophe

docs/product/components/tables.html
```diff
@@ -446,40 +507,332 @@
         </div>
     </div>
 </section>
-
 <section class="stacks-section">
-    {% header {element: 'h2', text: 'Unique Layouts'} %}
-    <p class="stacks-copy">These are different layouts which can be added to particular areas within a table.</p>
+    {% header {element: 'h2', text: 'Alignment'} %}
+    {% header {element: 'h3', text: 'Vertical Alignment'} %}
+    <p class="stacks-copy">The default vertical alignment is <code class="stacks-code">middle</code>. You change a table's or a specific cell's vertical alignment by using the <a href="{{ "/product/base/vertical-alignment#classes" | relative_url }}">Vertical Alignment atomic classes</a>.</p>
```
aaronshekey: Apostrophes

docs/product/resources/svg-icons/ArrowUpDownSm.svg
```diff
@@ -1 +1 @@
-<svg aria-hidden="true" class="svg-icon iconArrowUpDownSm" width="14" height="14" viewBox="0 0 14 14"><path d="M3 8h8l-4 4zm0-2h8L7 2z"/></svg>
\ No newline at end of file
+<svg aria-hidden="true" class="svg-icon iconArrowUpDownSm" width="14" height="14" viewBox="0 0 14 14"><path d="M3 8h8l-4 4zm0-2h8L7 2z"/></svg>
```
aaronshekey: What's going on here? Did we add a new line at the end of this file somehow?

docs/_data/product/tables.yml
hellohynes: This follows the same pattern as `.grid--cell#`.

lib/src/components/_stacks-tables.less
hellohynes: This was modified from @TedGoas's original class name of `s-table__borderlessx`, which I felt was a bit verbose. I read `.s-table__bx0` as x-axis borders have a 0 width. But if that's not clear, then we should change it. Would `s_table__bx-transparent` be clearer? It's still verbose, but it at least matches some of our border naming conventions.

docs/product/resources/svg-icons/ArrowUpDownSm.svg
hellohynes: It was modified earlier in the PR and this reverted it back to what was in `develop`. No changes to the icon itself.

hellohynes:
@UXPiper Good point about the spacing. I was trying to keep the spacing here consistent with other components like buttons and inputs. That said, we could probably go with `s-table__sm` and `s-table__lg` for now. I've removed `s-table__md` and updated the documentation.

Re: Bar graph — To achieve having items right-aligned means either we have to fix the width of the progress bars or we can't use `s-progress`. Neither option is great IMO. The first option let's us use `s-progress` but every time someone uses them in a table cell they'll have to pick a max-width. It won't automatically work. But recreating something that looks like something we already have isn't great either. 🤔

lib/src/components/_stacks-tables.less
UXPiper: Do you think they are all unnecessary or just borderless and no y borders?

hellohynes: We need borderless and borderless-x IMO. Borderless-y I'm struggling to come up with a use case within the product.

lib/src/components/_stacks-tables.less
```diff
@@ -197,17 +142,152 @@
...
+        }
+        tfoot th,
+        tfoot td {
+            border-bottom-color: transparent;
+        }
+    }
+
+    //  $$  NO BACKGROUNDS
+    //  ------------------------------------------------------------------------
+    &.s-table__clear th {
```
aaronshekey: I think if `.s-table__clear` is to mean no backgrounds, we can keep the header typography consistent with the default styles. I'd only keep `background-color` here.

PR #131, opened 2018-09-07
Add s-tag__danger class for when a tag is incorrect.

hellohynes:
This PR adds the following tag modifier: `.s-tag__danger`

Examples

This will be used _only_ for when a tag is incorrectly added. This is something we are adding a Teams project. Here's an example of how it would be used:

Example Usage


aaronshekey:
I like this approach, _technically_, but I do question the UX a _little_. I wonder if there's a way to avoid an error'd state, if possible. I feel like in other solutions like this, the errant auto-complete input wouldn't graduate to a tag at all. It would just remain as text. It feels funny, but barring a better solution, and considering this is a real need, I'd say I'm 👍 on it. 🤔

hellohynes:
@aaronshekey The pluses for converting it to an error tag-state:

 1. Let's the user know that what they submitted isn't correct
 2. Allows the user to clear it out with a single mouse click instead of selecting or hitting backspace a number of times.

We currently have an error state for `.post-tag`, but it outlines the `.post-tag` with a red border. It's not a huge visible change though.

PR #132, opened 2018-09-10
Fill colors for SVGs

Courtny:
**Proposed**
Let's add in the ability to reference all the Stacks colors with a `fill` helper.

**Why Do I Need It**
I find myself wanting to customize the color of embedded SVG files frequently ~ this would allow me to quickly and easily tap into all the colors we have available in Stacks.


aaronshekey:
I think we could solve this with a single new class of `fill-current` with:

``` css
.fill-current {
    fill: currentColor;
}
```

You could then add `fill-current` and `fc-xxx` to any element to achieve the same thing without adding ~70ish classes.

Courtny:
@aaronshekey That sounds like it would work!

aaronshekey:
> That sounds like it would work!

This is for non-icon SVGs, right? Otherwise @hellohynes solution absolutely works in combination with the `@Svg` helper :v:

Courtny:
@aaronshekey Yes, correct. Any SVG that is not an icon or that we have access to the svg helper for!

hellohynes:
Ah — yeah I'm not understanding the PR correctly then.

aaronshekey:
Just gotta stuff this class somewhere in the docs. Perhaps the Colors page?

Courtny:
@aaronshekey Do you think it should be a little *note* on the page below the introductory text? Or at the bottom of the page?

aaronshekey:
Good question. I'm not sure. Maybe its own section called Fill at the bottom of that page? This one's an oddball, heh!

If we were being totally consistent, it would be its own page called Files under base, but for a single class? 🤔

PR #133, opened 2018-09-11
Updated horizontal table cell borders

TedGoas:
Phew! That tables PR had a metric ton of stuff in it, huh? Glad we got it pushed across the finish line, but I think a detail around table borders fell through the cracks somewhere. I noticed it as I went to update the Manage Users / Teams styles everywhere and found that what I need need exist.

I intend Manage Users tables to look like the default, except not _all_ the vertical borders:
screen shot 2018-09-11 at 4 56 07 pm

**Note:** most of the tables I work on would look like this ☝️. I haven't built a table intense enough to need [all the borders](https://stackoverflow.design/product/components/tables#default-style), but love all the other things we've built into the 'default' style.

## Current
However when I go to remove those vertical borders with `s-table__bx`, the class removes the vertical borders and [nukes the outline around the whole table too](https://stackoverflow.design/product/components/tables#horizontal-borders):

screen shot 2018-09-11 at 5 03 16 pm
This looks just like the old tables we were trying to replace 😫. I want to keep the table's outline.

# This Pull Request changes `s-table__bx` to make the second screenshot appear like the first.
It's _not_ done, but shows what I was going for with my initial commit on that tables PR.

I'm happy to jam with someone on this. Do we need to add/modify a minimal table version that removes the table outline and the `<th>` background color? 🤷‍♀️


TedGoas:
@hellohynes I think there's a better way to do it (`s-table__bx` is doing too much as-is), but I understand the spirit of what you're asking for. I'll jam on this later today.

hellohynes:
An alternative is to have three states for borders:

1. Full (default).
2. Remove internal y-axis borders (your proposed version).
3. Remove all y-axis borders and outer x-axis outlined borders (current `s-table__bx`). 

All three have a time and place IMO. Thoughts?

TedGoas:
Yea, my mind was going in that direction, but I think `s-table__bx` should be the middle version. The class name indicates it would remove the vertical axis, but it currently _also_ removes the table outline and darkens the `<th>` border. I don't think it should do that.

A third class (`s-table__simple` or whatever, you can name it) can do all the those things to make "[simple tables](https://github.com/StackExchange/Stacks/pull/133#issuecomment-420695748)."

hellohynes:
It'd be nice if the names were somehow logically connected. Like `s-table__bx` and `s-table__bx-simple`, but yes — we're on the same page I think.

hellohynes:
Thanks for doing this @TedGoas 👍

docs/product/components/tables.html
```diff
@@ -120,16 +120,16 @@
         </div>
     </div>
 
-    {% header {element: 'h3', text: 'Horizontal Borders'} %}
-    <p class="stacks-copy">Shows only horizontal table cell borders.</p>
+    {% header {element: 'h3', text: 'Simple Borders'} %}
+    <p class="stacks-copy">Removes most of the default borders and backgrounds. Good for tables without much data that don't need to be sorted or filtered.</p>
     <div class="stacks-preview mb24">
 {% highlight html linenos %}
 <table class="s-table s-table__bx">
```
hellohynes: This should be `s-table__bx-simple`

TedGoas:
Is this good? I'd love to get this into production cause Enterprise is waiting on it.

PR #143, opened 2018-09-17
Email team avatar

TedGoas:
Just re-applying the previous edits. This branch was created before our big IA re-org, so that might have been the culprit.


aaronshekey:
Sorry this got eaten by the IA change 👎 

While we're here, think we can combine these notes into a single one at the top?

screen shot 2018-09-17 at 10 49 31 am

TedGoas:
They're unrelated. If anything, that first note could be moved out of the "Example" since it's really a guideline.

aaronshekey:
I think they can safely be combined :v:

> An email should only contain content from **one** Team. A Team's content should never appear mixed alongside content from another team in a single email. The color should match the Team's Theme Color as defined in their Display Settings..

TedGoas:
@aaronshekey I'm ok with this if you feel strongly. It's a little weird that we're talking about tactics like theme colors _before_ we've even shown an example of the thing, but the page is small enough. I'm like [a 3 out of 10](http://blog.capwatkins.com/the-sliding-scale-of-giving-a-fuck) on this one.

PR #147, opened 2018-09-21
Email footers

TedGoas:
This page documents all configurations of our email footers. I would like feedback on everything from everyone... and also specifically:

* @donnachoi Is this thorough and representative of the work we did last year and want to see going forward?
* @UXPiper Does this cover Talent?
* @aaronshekey Any thoughts on how to fix the in-page anchors (screenshot below)? And curly quotes.

---

screen shot 2018-09-21 at 1 58 29 pm


aaronshekey:
Since the plugin I wrote for headers is pretty naive, I'd just stick to how I handled it in the grid section:

screen shot 2018-09-21 at 2 18 30 pm

PR #153, opened 2018-09-26
Core table extras

TedGoas:
Move some more table stuff into Stacks and out of the core codebase. One merged in Core, this will [replace this CSS](https://gh.stackoverflow.com/Core/StackOverflow/pull/1351/commits/ed960053400661358f45d88b516d0e35a4ca9880).


lib/src/components/_stacks-tables.less
```diff
@@ -71,6 +71,11 @@
     tbody + tbody {
         border-top: @su2 solid @black-100;
     }
+
+    //  The .user-info include has some padding we don't want when used in tables.
+    .user-info {
```
aaronshekey: I don't think this belongs here, especially since Stacks doesn't provide `user-info`. I think this would be better suited for a `p0` elsewhere on Core. 🤔

lib/src/components/_stacks-tables.less
```diff
@@ -316,4 +328,8 @@
     max-width: @su24;
     padding-right: @su2;
     padding-left: @su2;
+
+    input {
```
aaronshekey: This feels dirty and temporary. Think we could comment that this is an override to bad defaults in Core?

lib/src/components/_stacks-tables.less
```diff
@@ -262,6 +267,13 @@
         color: @fc-light;
         cursor: pointer;
 
+        //  If an <a> link is used, it should appear like a <th> and not an <a>
+        th {
+            a {
```
aaronshekey: Since there's only one descendent here, I don't think we need the nesting. This would probably do it? 2/10 on this one, heh

``` css
th a {
  color: inherit;
}
```

lib/src/components/_stacks-tables.less
TedGoas: `<div class="user-info"></div>` is a shared component that we use in a lot of places. In most cases, it makes sense to have the padding:
screen shot 2018-09-26 at 1 00 54 pm

I don't think atomic CSS is possible here. I've run into this a few times now.

lib/src/components/_stacks-tables.less
TedGoas: Sure, and hopefully this will get standardized when balpha finishes his table sorting stuff and we know what markup should be in the `<th>` to make sorting happen.

lib/src/components/_stacks-tables.less
aaronshekey: I think we should add a parameter to the shared component that accepts additional classes. In the context of tables, it'd make sense to pass `.p0`. I don't think this CSS belongs in Stacks.

lib/src/components/_stacks-tables.less
TedGoas: 👍 I'm a 1/10 so I will change.

lib/src/components/_stacks-tables.less
TedGoas: Actually... should we just advise using an `.m0` on the `input` here?

aaronshekey: Yeah, now that you mention it, I think that's where my head's at too.

TedGoas: Ok, I'll update.

aaronshekey:
Ok, @TedGoas I'm going to merge this and release a new version. I can make the `user-info` changes on Core. I've also renamed some things so I can handle that as well.

PR #159, opened 2018-10-01
Fix Typo

dan-drews:


aaronshekey:
Good catch! Thanks @dan-drews 🏆

PR #163, opened 2018-10-09
Add a print class for d-none

### LINKED ISSUE #78: Support for print styles?
When debugging a print layout, it may be beneficial to include some utilities for printing. Oftentimes, with an atomic layout, we aren't actually giving an overall class to anything:

```
<div class="p12 bg-black-050 bar-md">
    <a class="s-btn s-btn__primary">Sign up</a>
</div>
```

In production, this markup is valid. If we want to hide this markup for printing layouts, it doesn't make much sense to target `s-btn` or its parent and apply `display: none;`, and it's also tedious to write a new class to target print-only styles.

Perhaps we allow for `p:d-none` meaning "Don't allow this element to appear in print layouts." Further, we could apply `p:d-block` if we _only_ want something to show when printed. 🤔

aaronshekey:
This PR closes #78 by adding `.p:d-none` and documents it. It also applies `.p:d-none` to appropriate places in our documentation.

What do we think of this approach?

screen shot 2018-10-09 at 11 04 20 am


TedGoas:
Is `.p:` too short of a prefix? What about `.pr:` to follow the responsive classes? or `.print:` to make it really obvious what the thing does? 🤷‍♀️

aaronshekey:
We've traditionally favored shorter class names, since this will likely be combined with lots of other classes. Is `.pr:d-none` more obvious than `.p:d-none`? What do @ludwiczakpawel and @balpha think?

Does obviousness matter vs. documentation?

TedGoas:
Just surfacing some other options, I'm like a [2 out of 10](http://blog.capwatkins.com/the-sliding-scale-of-giving-a-fuck) on this.

aaronshekey:
I think we can be verbose here. Switching to `.print:d-none` and figuring out where to explicitly document this class.

ludwiczakpawel:
@aaronshekey what do you think about this page? https://stackoverflow.design/product/base/display we may rethink if we ever add more print classes...

aaronshekey:
Yep, I've got a checkmark there that says a class is available for print. I think at some point, if we add more print classes, we can group Responsiveness and Print into a section like "Patterns" or something where we talk generally about classname patterns.

Issue #79, opened 2018-07-25
Support for responsive text sizing


Courtny:
**Problem**
Font sizes become unruly when responding to smaller screen size. This especially true at `sm` sizes where desktop sized headlines are visually too large for the layouts. Different scales work better for different screen sizes, and it's my hope Stacks can intelligently respond to this as a system.

**Solution I'd like**
It would be very useful to have some baked in responsive font scaling when hitting certain breakpoints. This could manifest itself as a Stacks baked in modular type scaling _(I explore this below)_ and/or the inclusion of some optional utility classes like `sm:text-smaller`, `sm:text-larger`. The utility classes would take the base font size for that type style and scale accordingly.

**Thoughts on calculating modular scaling**
On a small screen, heading sizes should probably be more uniform; a large screen has room for greater variation. We can use a modular typographic technique to fluidly transition between different scales. Maybe we could pick a ratio for small screens and a different ratio for large screens, and then work out the minimum and maximum font sizes for each heading level.

To calculate the modular scale, take the base font size and multiply it by the desired ratio to get larger numbers. For smaller numbers, divide. Repeat this process with subsequent results to get each required step in the scale:

**LARGER**

- 1em × 1.125em = 1.125em
- 1.125em × 1.125em = 1.266em
- 1.266em × 1.125em = 1.424em
- 1.424em × 1.125em = 1.602em

**SMALLER**

- 1em ÷ 1.125em = 0.889em
- 0.889em ÷ 1.125em = 0.790em
- 0.790em ÷ 1.125em = 0.702em
- 0.702em ÷ 1.125em = 0.624em

For this example, say we use 1.125 as our minimum scale and do the same calculations for 1.250 as our maximum scale. We’d then apply different steps in the scale to our heading levels.

- Minimum scale: 1.125
- Maximum scale: 1.250

Heading level | Minimum font size | Maximum font size
-- | -- | --
Heading 1 | 1.602em | 2.441em
Heading 2 | 1.424em | 1.953em
Heading 3 | 1.266em | 1.563em
Heading 4 | 1.125em | 1.250em

Using these numbers, perhaps there could be some Stacks definitions that deal with the initial scaling.

`fs-body1-min`
`fs-body1-max`

I'll try to put together a demonstration soon.

**Resources**
[Type Scale](http://type-scale.com)
[Modular Scale](http://www.modularscale.com/)

aaronshekey:
[...]

PR #170, opened 2018-10-20
Make the base font size responsive

aaronshekey:
This PR explores the smallest idea for what's required to make #79 happen. In this proof of concept, we simply swap the base font size at the smallest breakpoint. It's 13px everywhere until the smallest breakpoint, when we drop it down to 11px.

| Smallest Breakpoint | Larger Breakpoints           |
|-------------------------|--------------------------|
|  |  |
|  |  |


lib/src/base/_stacks-body.less
```diff
@@ -25,6 +25,13 @@ body {
     line-height: @lh-base;
 }
 
+@media (max-width: 640px) {
```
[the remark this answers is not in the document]
balpha: Okay, there'll be a way to do this after #171 has been merged. See [this example](https://github.com/StackExchange/Stacks/blob/bfb0bc437395e439e31878cceefd386c060ad3ef/lib/src/base/_stacks-internals.less#L76), which is almost precisely your use case.

lib/src/base/_stacks-body.less
```diff
@@ -25,12 +25,12 @@ body {
     line-height: @lh-base;
 }
 
-@media (max-width: 640px) {
+#stacks-internals #screen-sm({
```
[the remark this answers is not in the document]
balpha: It doesn't work if, like on the Q&A sites, responsiveness is controlled via CSS classes on the root element, because you get a selector like this:

```css
html.html__responsive html {
   ...
}
```

aaronshekey:
This pulls in your PR and tests, @balpha Lemme know what I screwed up, and if you'd like to commit directly to this branch, by all means 🤝

lib/css/stacks.css
```diff
@@ -8154,6 +8154,12 @@ body {
   font-size: 13px;
   line-height: 1.30769231;
 }
+@media (max-width: 640px) {
+  html ,
+  html  body {
```
aaronshekey: This whitespace is a little weird, no?

balpha: Yeah, admittedly it bothered me a little bit as well 😁 But it's not like it's an actual issue, especially in generated code that's going to be minified anyway.

aaronshekey:
After some feedback from the crew, this is definitely the approach we should take. Some issues:

- [ ] Minimize the multiplier / font size curve to improve headline presentation
- [ ] Explore ignoring font changes on the `fs-body[x]` group of classes
- [ ] Make sure the literal _base_ font size never gets shown in the UI.

ludwiczakpawel:
@aaronshekey would this affect everything we have set in `rem` units? Because for example static widths & heights (min, max too) are based on `rem` units. Breaking those might be painful :P

aaronshekey:
I've pushed a bunch of commits that considers that checklist. We now hardcode pixel values at the smallest 3 type sizes in our type scale, meaning that only the things above `fs-body1` get scaled down at the smallest viewport.

Since we never assumed base font size would never get smaller than 13px, I made sure to add `font-size: @fs-body1` to some components that didn't have them prior to this (and probably should have).

aaronshekey:
@ludwiczakpawel These widths seem to stay consistent given that the body font size is fixed at the smallest sizes, but we should revisit widths and heights. We ought to brainstorm what those look like in a separate issue. For example, I reallllly wish we had something like `.ws32` for `width: 32px;`.

Y'all, how do we feel about this approach to responsive font sizes?

lib/src/base/_stacks-body.less
```diff
@@ -25,6 +25,13 @@ body {
     line-height: @lh-base;
 }
 
+#stacks-internals #screen-sm({
```
aaronshekey: @balpha This approach isn't working in the production codebase. I was able to drop in the prior approach. Do you mind poking at `feature/responsive-typography` in production?

balpha: @aaronshekey Will do. Can you define "not working"?

balpha: So it, well "works" here -- it's just that the body font size is overridden by another CSS rule later. Is that what you're referring to?

balpha: If you comment out line 88 in `_base-primary.less`, it behaves as expected.

aaronshekey: Yep, that must be what's happening here. Thanks for tracking that down. I assumed it was a mixin issue and wasn't looking beyond that. :v:

PR #173, opened 2018-10-25
Formatted <code> in email

TedGoas:
Adds support for `code` in email.

### DIFF
--- a/docs/email/base/typography.html
+++ b/docs/email/base/typography.html
@@ -125,7 +125,7 @@ description: This page provides some common typography patterns for email, but a
     {% header h2 | Code %}
     <div class="stacks-preview mb16">
 {% highlight html linenos %}
-// Preformatted code block
+<!-- Preformatted code block -->
 <pre style="line-height: 17px; background-color: #EFF0F1; padding: 4px 8px; border-radius: 3px; overflow-x: auto;"><code style="margin: 0 0 15px; background-color: #EFF0F1; color: #242729; font-size: 13px; line-height: 17px; font-family: Consolas, Menlo, Monaco, Lucida Console, Liberation Mono, DejaVu Sans Mono, Bitstream Vera Sans Mono, Courier New, monospace, sans-serif;">
     …
 </code></pre>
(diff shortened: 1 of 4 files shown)


docs/email/base/typography.html
```diff
@@ -122,7 +122,35 @@ <h2 style="font-weight: bold; font-size: 21px; line-height: 24px; color: #242629
     </div>
 </section>
 <section class="stacks-section">
+    {% header h2 | Code %}
+    <div class="stacks-preview mb16">
+{% highlight html linenos %}
+// Preformatted code block
+<pre style="line-height: 17px; background-color: #EFF0F1; padding: 4px 8px; border-radius: 3px; overflow-x: auto;"><code style="margin: 0 0 15px; background-color: #EFF0F1; color: #242729; font-size: 13px; line-height: 17px; font-family: Consolas, Menlo, Monaco, Lucida Console, Liberation Mono, DejaVu Sans Mono, Bitstream Vera Sans Mono, Courier New, monospace, sans-serif;">
```
ludwiczakpawel: So `background-color` is specified on both elements: `pre` and `code` but `color` is specified only on `code`. Is that correct? Just making sure this is intended. 

------

Out of curiosity: how does `overflow-x: auto` behave in various email clients?

TedGoas:
@UXPiper _ALL_ of me wishes we stylized `code` with a dark theme 😂 but I'm just matching what we do on the web.

Regarding `target: blank;` I'd prefer not to force folks into that UX (they can launch in a new tab if they want). Sound reasonable?

docs/email/base/typography.html
```diff
@@ -122,7 +122,35 @@ <h2 style="font-weight: bold; font-size: 21px; line-height: 24px; color: #242629
...
+    …
+</code></pre>
 
+// Inline code
```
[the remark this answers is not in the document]
TedGoas: @ludwiczakpawel Wow I had no idea this features existed, so cool!

docs/email/base/typography.html
TedGoas: @ludwiczakpawel [support for `overflow` is _ok_](https://www.campaignmonitor.com/css/positioning-display/overflow/)... but it's supported in the places we need it most: mobile clients and Gmail. [`word-wrap` is an alternative](https://www.campaignmonitor.com/css/text-fonts/word-wrap/).

PR #175, opened 2018-10-27
Box Shadow and Opacity hovers

Courtny:
We've had a need for hover animations on occasion. These are some common ones I find myself generally lifting from Tachyons regularly.

Still need to do additional cleanup to this page design and add additional documentation but wanted to get the proof of concept out for review.


aaronshekey:
Excited to see this!

I have a couple thoughts with how we might approach hover states. I really love [Tailwind](https://tailwindcss.com/docs/state-variants/#hover)'s approach, and that's how we've modeled our responsive modifiers.

I've wanted to add `h:` class modifiers to a subset of our colors. I could see us sprinkling in hover overrides by adding `h:` to a bunch of our existing atomic classes. So, for your examples, like shadows, I think we could do something like `h:bs-sm`.

Instead of `hvr-fade` with this pattern you'd do `h:o20`, etc.

For `hvr-grow-lg`, I wonder if we ought to add scaling classes and then add a `h:` modifier.

How silly is this approach?

Courtny:
@aaronshekey That is not silly at all!

I think it all sounds reasonable. I would still recommend we have some pre-built "go-to" components developers could grab so they don't have to wonder about what opacity to apply, etc. Y'know?

So in that way I'd still like to hand folks a couple of _pre-designed_ options like I have that are off the shelf.

**ALSO** I can work on breaking down the components I have here into an approach more similar to Tailwind / what you suggested.

lib/src/atomic/_stacks-misc.less
```diff
@@ -99,18 +99,30 @@
...
-.o20                      { opacity: 0.2 !important; }
-.o30                      { opacity: 0.3 !important; }
-.o40                      { opacity: 0.4 !important; }
-.o50                      { opacity: 0.5 !important; }
-.o60                      { opacity: 0.6 !important; }
-.o70                      { opacity: 0.7 !important; }
-.o80                      { opacity: 0.8 !important; }
-.o90                      { opacity: 0.9 !important; }
-.o100                     { opacity: 1 !important; }
+.o-0                       { opacity: 0 !important; }
```
aaronshekey: I don't think these need to be renamed, do they? What's with adding the dash?

Courtny: For some reason, these will not work properly with the `\:` operator when creating the hover states without it ~ any ideas why that is? I'm not familiar with using that.

Courtny: Maybe it will work and i just didn't have the correct syntax in place? Let me know when you can chat and we can discuss before I keep moving forward.

Courtny:
OK!
Opacity hovers have been added in and documentation updated. For review @aaronshekey ~ let me know if you think we can roll this out.

PR #176, opened 2018-10-29
Add support to <blockquote>'s in email

### LINKED ISSUE #102: New Feature: block quotes
Styles for a `blockquote` tag.

TedGoas:
Like [what we did for `code`](https://github.com/StackExchange/Stacks/pull/173) last week. This would resolve #102

### DIFF
--- a/docs/email/base/typography.html
+++ b/docs/email/base/typography.html
@@ -151,6 +151,29 @@ console.log("request body: " + pm.globals.get("switch_json"));</code></pre>
     </div>
 </section>
 <section class="stacks-section">
+    {% header h2 | Blockquotes %}
+    <div class="stacks-preview mb16">
+{% highlight html linenos %}
+<blockquote style="margin: 0 0 15px; padding: 4px 10px; background-color: #FFF8DC; border-left: 2px solid #ffeb8e;">
+    <p style="padding: 4px 0; margin: 0; font-size: 15px; overflow-wrap: break-word;">Write a function that takes a string consisting of alphabetic characters as input argument and returns True if the string is a palindrome.
+        …
+    </p>
+</blockquote>
+{% endhighlight %}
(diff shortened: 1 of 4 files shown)


aaronshekey:
@TedGoas Is this good to go? Looks good to me!

docs/email/base/typography.html
```diff
@@ -150,6 +150,29 @@ <h2 style="font-weight: bold; font-size: 21px; line-height: 24px; color: #242629
         </div>
     </div>
 </section>
+<section class="stacks-section">
+    {% header h2 | Blockquotes %}
+    <div class="stacks-preview mb16">
+{% highlight html linenos %}
+<blockquote style="margin: 0 0 15px; padding: 4px 10px; background-color: #FFF8DC; border-left: 2px solid #ffeb8e;">
```
ludwiczakpawel: Out of curiosity: for `code` blocks in the previous PR you've defined font `color`. Here, for `blockquote` you're not doing so. Any reason for (not?) doing that?

TedGoas: Yes this is intentional. `<code>` can be used with or without a preceding `<pre>` tag, so it may not inherit the color it needs. The `<p>` tag is likely already set up to inherit the color from the rest of the email body, so I've left it out.

But good point! If, for some reason the text around the `<blockquote>` is styled differently, the `<blockquote><p>` may need an inline style. Though I don't see that being an issue in our case. Something to keep an eye on, though.

ludwiczakpawel: It's good to know, thanks!

TedGoas:
@UXPiper Hmm, either could work I suppose. I put them in typography because they style type, but as you say they're a little more involved than the rest of the stuff in that section. I'd like to merge this for now but am open to moving things around whenever!

Issue #161, opened 2018-10-02
Button Groups lose their border radius when each option is wrapped in its own <form> tag


TedGoas:
I noticed that when each button is wrapped in its own `<form>` tag, the corners aren't rounded because technically it's not the first or last child anymore.

<img width="1213" alt="screen shot 2018-10-02 at 9 54 39 am" src="https://user-images.githubusercontent.com/1172461/46360959-85cce280-c63a-11e8-9982-0063c72e6112.png">

TedGoas:
Also noticed that the vertical borders don't collapse.

<img width="107" alt="screen shot 2018-10-03 at 1 59 16 pm" src="https://user-images.githubusercontent.com/1172461/46429534-a8332e80-c714-11e8-991d-670c355e3fc5.png">

aaronshekey:
Yeah, that makes sense since we're applying these overrides to the button and not the parent in this case. I'm on it!

I'd also like to clean up how these things wrap when they're forced to.

PR #177, opened 2018-10-31
Bugfixes for button groups

aaronshekey:
As outlined in #161, button groups were losing their styling when required to wrap one or more of the buttons in a form element. I've added some guards for this common use case. I've also clarified this in the documentation.


docs/_data/product/button-groups.yaml
```diff
@@ -0,0 +1,7 @@
+classes:
+  - class: .s-button-group
+    applies: N/A
+    description: Base button group style.
+  - class: .s-button-group--container
```
aaronshekey: Instead of relying on `form` here, I've designed this to allow for any arbitrary wrapping element. But, is this class name stupid?

lib/src/components/_stacks-button-groups.less
```diff
@@ -17,27 +17,27 @@
 
     .s-btn {
         margin-right: 0; // With Safari in production, we need to collapse white space
-
-        // We don't want border-radii on interior buttons
-        &:not(:first-of-type):not(:last-of-type) {
-            border-radius: 0;
-        }
+        border-radius: 0; // Kill the default border radius
```
aaronshekey: I switched to removing the border radius by default here, and then re-adding it as needed to the first and last children, but I may switch back to the original approach of removing border so `s-btn-group` can actually support different button sizes and their different border radiuses. 🤔

docs/_data/product/button-groups.yaml
TedGoas: Nah I think that's good, broad name.

PR #181, opened 2018-11-09
Additional buttons modifiers

aaronshekey:
This PR adds a couple extra options to our buttons based on feedback from engineering.

It allows for stripping all button styles in `s-btn__unset`, styling a button like a link with `s-btn__link`, and moves icon-specific styling to `s-btn__icon`.


aaronshekey:
Ok, great point. The focus indicator's returned to `.s-btn__unset` since we don't have an atomic way of adding one back.

PR #184, opened 2018-11-14
Gmails conversation thread

TedGoas:
This explains how Gmail's conversation threading works and includes a CSS reset to defend against it.

---

This would resolve #184

### DIFF
--- a/docs/email/guidelines/faq.html
+++ b/docs/email/guidelines/faq.html
@@ -156,6 +156,16 @@ title: Frequently Asked Questions
 </section>
 
 <section class="stacks-section">
+    {% header h2 | What is conversation threading? %}
+    <p class="stacks-copy">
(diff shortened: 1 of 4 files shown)


Courtny:
I say, ship it!

This is purely for you to consider Ted: Do we need the line; "...though this will only be applied in email clients that support CSS styles in the <head>..." given that this occurs specifically for Gmail clients?

TedGoas:
@Courtny I think it'd good to include that line. Gmail has several products and [only some](https://www.campaignmonitor.com/css/style-element/style-in-head/) support CSS in the head. Given this _could_ be happening elsewhere (we haven't bulletproofed this yet to my knowledge) and things _could_ change (email clients occasionally make unannounced changes), I think it's practical to link to Campaign Monitor's breakdown if someone has a question about a specific email client. Agree?

Courtny:
@TedGoas Yeap, agree!

PR #186, opened 2018-11-19
Add hover classes to a select number of color stops

aaronshekey:
This PR adds a small selection of hover states for our colors. This subset adds hover classes to the stop darker than the most common color in a set. For example, `orange-500` is a common color, so `h:orange-600` has been added. Font color and background color hover classes have been added for the following stops:

- `orange-600`
- `black-900`
- `black-600`
- `black-100`
- `blue-600`
- `green-600`
- `red-600`


Courtny:
@aaronshekey What do you think about pulling these in, so I can use them as reference for #191 ?

aaronshekey:
> What do you think about pulling these in

I feel good. Lemme give this thing a sanity check since it's been over 2 weeks since this was opened :v:

Issue #174, opened 2018-10-25
Animation utilities


aaronshekey:
We should add some animation atomic classes that provide transitions.

PR #191, opened 2018-11-21
CSS Transition Utilities

Courtny:
This PR is focused on introducing a handful of simple transition utilities as #174 suggested.

**What is included here?**
The simplest, most straightforward way to animate components is through CSS transitions. I am adding these in so developers don't have to guess or invent new transitions for components.

I've provided the following:

- **A default, atomic transition style that can be used on an element.** It is not necessarily atomic, but provides a base `transition-property`, `transition-timing-function`, `transition-timing-duration`, and `transition-delay`.
-  Four (4) **Durations** classes
-  Five (5) **Properties** classes
-  Five (5) **Timing Functions** classes
-  Four (4) **Delays** classes

**Summary of generated utilities**

```css
// -- Provide a sensible default
.t { transition-duration: .25s; transition-property: all; transition-timing-function: ease-in-out; transition-delay: .1s;}

// -- Durations
.t-slower      { transition-duration: .75s; }
.t-slow        { transition-duration: .5s; }
.t-fast        { transition-duration: .15s; }
.t-faster      { transition-duration: .075s; }

// -- Properties
.t-all         { transition-property: all; }
.t-none        { transition-property: none; }
.t-bg          { transition-property: background-color; }
.t-opacity     { transition-property: opacity; }
.t-shadow      { transition-property: box-shadow; }

// -- Timing Function
.t-linear      { transition-timing-function: linear; }
.t-ease        { transition-timing-function: ease; }
.t-ease-in     { transition-timing-function: ease-in; }
.t-ease-out    { transition-timing-function: ease-out; }
.t-ease-in-out { transition-timing-function: ease-in-out; }

// -- Delays
.t-delay-long { transition-delay: .2s; }
.t-delay-longer { transition-delay: .3s; }
.t-delay-longest { transition-delay: .4s; }
.t-delay-none { transition-delay: 0s; }
```

**Class Naming**
I'm going back and forth on naming. To maintain brevity, I simply used `t-`. Something like `transition-` is quite long to write, but is clearer. I am opposed to using `trans-` at the moment as it could easily be confused with `transform` or `translate` down the line. I dunno. I could be convinced into any of them.

**What's left to do?**
I'd like to do some fancier animation examples in the docs section to showcase timing functions off.

Need to update the Classes list to reflect all the classes. Probably good to add a description column, too.

I also want to do something cleaner for the BG color class _(I just hacked it in for now for testing)_. On that note, would be nice to support the `:hover` classes for background colors... #186 

And finally, I'd need to rip out any baked in transitions in Stacks and replace with the new atomic versions.


docs/product/base/transitions.html
```diff
@@ -32,8 +32,7 @@
     <p class="stacks-copy">The base transition applies a moderate duration, timing function, and delay to an element.</p>
     <div class="stacks-preview">
 {% highlight html linenos %}
-<div class="t">…</div>
-{% endhighlight %}
+<div class="t">…</div>{% endhighlight %}
```
aaronshekey: We actually make the highlighted code flush left with the left side of the document, since it's all a `<pre>` tag. It messes with the format of the doc itself, but the output is correct :v:

Courtny: Oh wow, I did not know this! Thanks for looking out 👍

aaronshekey:
Here are some of my thoughts inline with the output CSS:

``` less
// -- Provide a sensible default
.t { transition-duration: .25s; transition-property: all; transition-timing-function: ease-in-out; transition-delay: .1s;} // 2 things here. I think our default transition time should be .1s. Try it! Part of the reason I've killed so many transitions throughout SO is they're just so slow. I also don't think the transition should be delayed by default. I want things to feel snappy.
.t { transition-duration: .1s; transition-property: all; transition-timing-function: ease-in-out; } // Perhaps this is the default. Feel it out!

// -- Durations
.t-slower      { transition-duration: .75s; }
.t-slow        { transition-duration: .5s; }
.t-fast        { transition-duration: .15s; } // What if... we killed the option to make them slow?
.t-faster      { transition-duration: .075s; }

// -- Properties
.t-all         { transition-property: all; } // Do we need this since it's included in .t?
.t-none        { transition-property: none; } // Would this be better served as a .t-unset? Perhaps it actually unsets a whole transition (Good for overrides).
.t-bg          { transition-property: background-color; }
.t-opacity     { transition-property: opacity; }
.t-shadow      { transition-property: box-shadow; }

// -- Timing Function
// Something tells me we ought to be opinionated here and just rely on the default, killing all these entirely. Or, even better, what if there's a custom easing that we all agree feels great and we rely on that?
.t-linear      { transition-timing-function: linear; }
.t-ease        { transition-timing-function: ease; }
.t-ease-in     { transition-timing-function: ease-in; }
.t-ease-out    { transition-timing-function: ease-out; }
.t-ease-in-out { transition-timing-function: ease-in-out; }

// -- Delays
.t-delay-long { transition-delay: .2s; }
.t-delay-longer { transition-delay: .3s; }
.t-delay-longest { transition-delay: .4s; }
.t-delay-none { transition-delay: 0s; }
```

I dig the naming of these, and your argument against `trans` is a good one. As I see it, most transitions will be applied like follows `<div class="t t-bg h:bg-black-025">` right? Seems sensible!

Courtny:
✅This PR is ready for review!

I made most all of the recommendations provided by @aaronshekey with the exception of completely removing `t-slow`.

I decided to leave in the ability to slow down a transition, as I have some areas in the product on Talent that could make use of this _(recommendation cards, in particular)_. I could also see it being useful for toasts.

```css
// -- Provide a sensible default
.t             { transition-duration: .1s; transition-property: all; transition-timing-function: ease-in; transition-delay: 0s;}

// -- Durations
.t-slow        { transition-duration: .25s; }
.t-fast        { transition-duration: .05s; }

// -- Properties
.t-unset       { transition-property: none; }
.t-bg          { transition-property: background-color; }
.t-opacity     { transition-property: opacity; }
.t-shadow      { transition-property: box-shadow; }

// -- Delays
.t-delay       { transition-delay: .25s; }
.t-delay-unset { transition-delay: 0s; }
```

Issue #90, opened 2018-08-02
Better `s-card` hover states


hellohynes:
Initially posted by Piper in the Stacks Slack room:
---

I have some concerns regarding `.s-card`. The hover state on the cards isn’t very recognizable. I’m not really sure why a card wouldn’t need an *obvious* hover state, to be honest. It either needs a hover state (and should be obvious) or it doesn’t. 

Anywho, I ran into the cards on this page a couple weeks ago and was excited that there seemed to be some sort of alternate option for cards that needed a more obvious hover state: 
https://stackoverflow.com/jobs/salary/results?l=Columbus%2c+OH%2c+United+States&ed=0&ex=2&ff=1&dr%5B0%5D=DataScientist&tl%5B0%5D=

I went looking at stacks to find how I could implement the shadow on hover and the way the documentation is written, it seems like I can do such a thing

> When applying a `.bs-*` class with a `.s-card` component, a :hover style is added to the base `.bs-*` box shadow.

If the above statement is true, then why would there be a  `.bs-hover` class? Well, I was hoping that it would do what the class says it would do… apply a border-shadow on hover.
But apparently this class is attached to the `.bs-*` classes, which is silly because the documentation says there already will be a shadow increase on hover.

Basically, I’m asking that you move the `.bs-hover` be attached to only the `.s-card` class. I can think of at least 3 situations where this would be useful. Including the example I sent you earlier which is currently using custom CSS to implement the shadow.

TedGoas:
✅ [Addressed here](https://github.com/StackExchange/Stacks/pull/192)

PR #192, opened 2018-11-21
Card Upgrades

### LINKED ISSUE #82: Beef Up Cards
**Problem**
Our [current cards](https://stackoverflow.design/components/cards) should be updated with an example for clickable cards, where the entire `.s-card` is linked.

**Solution**
Codified suggestions on how to design a clickable card.

**Describe alternatives you've considered**

1. Adding an extra link inside the card:
screen shot 2018-07-26 at 1 47 14 pm

2. Hyperlinking the card title:
screen shot 2018-07-26 at 1 49 10 pm

3. Do something more like a sidebar widget:
screen shot 2018-07-26 at 1 50 47 pm

---

Also may be worth creating some guidelines for when and how to use cards, [like this](https://yellowstroke.com/2018/10/03/creating-design-principles/).

TedGoas:
### Goals
- [x]  Remove Hover are regular cards
- [x]  Add support for linked cards
- [x]  Adjust the "muted" modifier
- [x]  Add guidelines and examples for different sizes

`.s-card__muted` is only used in one place and this can easily be swapped out for the regular `.s-card`: 
`/Volumes/C/code/stackoverflow/StackOverflow/Views/Channels/Onboarding/ChecklistSidebar.cshtml`

---

This will resolve #82 and #90


TedGoas:
I think there should be a visual cue that linked cards (`a.s-card`) are interactive/clickable. How do we think we should do that? Here are four options (not married to any of them):
screen shot 2018-11-21 at 4 32 50 pm

Courtny:
Shadows are very subtle to me ~ may not offer the most visual affordance. But they look nice. I like the idea of adding a blue border to cards which are clickable.

I also have some ideas based on my transition/animation work if we'd wanna get fancier.

card-click

docs/product/components/cards.html
```diff
@@ -35,39 +46,78 @@ <h1 class="fs-body3 fc-dark">Sample card title</h1>
...
+</a>
+{% endhighlight %}
+        <div class="stacks-preview--example">
+            <div class="grid grid__allcells3 md:fd-column gs16">
+                <a href="#url" class="grid--cell s-card">
+                    <h2 class="fs-body3 lh-sm fc-dark">Sample card title</h2>
+                    <p class="fs-body1 fc-medium">This is a description or explanation.</p>
+                    <p class="fs-body1 s-link">Learn more</p>
+                </a>
+                <a href="#url" class="grid--cell s-card bs-hover bs-sm">
```
ludwiczakpawel: This may not be a part of this PR (I'm not sure) but I wouldn't make `bs-hover` (or any `box-shadow` related classes) using solid colors without transparency. They look kinda off whenever you have different than white background color below element with such a shadow. In these examples, the regular shadow looks like have a transparency, but the one on hover does not.

ludwiczakpawel:
Regarding 4 proposals:
I think I could find a scenario for each of those. I mean I wouldn’t like to pick ONE OF THOSE. I’d prefer to have some options to choose from while using Stacks.

For example, for this small amount of text (as in the screenshot above) 1,2 and 3 look good, but 4 doesn’t look like clickable thing. Which may not be the best. 

If there was more text or more other content inside `.s-card` then I'd choose 1 or 3.. but I can also see a scenario for number 4 – imagine a dashboard with lots of stats, graphs, and numbers. Some of those numbers would be represented in `.s-card` – but it won’t be super important for user to know it’s clickable and (s)he can drill down. In this case I’d use number 4 – regular text, without extra (blue) decoration – this would help keep the page simple and clean (especially when you have for example 12 `.s-card` elements with those numbers next to each other.  But I'm only talking about representing clickable thing inside `.s-card`, not about shadow on hover.

So, that being said - depending on the situation I could find a use for each of those.

And the same for @Courtny proposal – I think it would work in some cases but not all of them. But that's definitely something that's worth including in Stacks.

Except that, I left some comment directly in code.

docs/product/components/cards.html
TedGoas: 👍 I don't disagree but agree this isn't part of this PR.

TedGoas:
Alright, I think this is ready for y'all to start poking at. Two notes:

1. @Courtny I'm interested in [using your transition/hover work somehow](https://github.com/StackExchange/Stacks/pull/192#issuecomment-440823253), let me know when this merges and let's take a second swing at this.

2. Josh had all `.s-cards` marked up using `<article>` tags; how do we feel about that? (vs. `<div>`s).
🤷‍♀️`<article class="s-card">...</article>`

docs/product/components/cards.html
```diff
@@ -1,73 +1,142 @@
...
+                <p class="fs-body1 fc-medium">This is a description of the card's content.</p>
+                <p class="grid gsx gs12">
+                    <button class="grid--cell s-btn s-btn__primary s-btn__sm">Call to action</button>
+                    <button class="grid--cell s-btn s-btn__sm">Cancel</button>
+                </p>
+            </article>
+            <article class="s-card">
+                <h2 class="fs-body3 lh-sm fc-dark">Base card title</h2>
+                <p class="fs-body2 fc-medium">This is a description of the card's content. Sometimes the description is two or three sentences.</p>
+                <p class="grid gsx gs12">
```
aaronshekey: Throughout Stacks, we generally use a grid spacing of `gs4` with buttons.

docs/product/components/cards.html
```diff
@@ -1,73 +1,142 @@
...
+                <h2 class="fs-body3 lh-sm fc-dark">Base card title</h2>
+                <p class="fs-body1 fc-medium">This is a description of the card's content.</p>
+                <p class="grid gsx gs12">
+                    <button class="grid--cell s-btn s-btn__primary s-btn__sm">Call to action</button>
+                    <button class="grid--cell s-btn s-btn__sm">Cancel</button>
+                </p>
+            </article>
+            <article class="s-card">
+                <h2 class="fs-body3 lh-sm fc-dark">Base card title</h2>
+                <p class="fs-body2 fc-medium">This is a description of the card's content. Sometimes the description is two or three sentences.</p>
```
aaronshekey: Remember to make a punctuation pass here.

docs/product/components/cards.html
```diff
@@ -1,73 +1,142 @@
...
     </div>
 </section>
 
+<section class="stacks-section">
+    {% header h2 | Linked Cards %}
+    <p class="stacks-copy">The <code class="stacks-code">.s-card</code> class can be applied to an <code class="stacks-code">&lt;a&gt;</code> tag for instances where a whole card should link somewhere. If possible, linked cards should visually indication that they're interactive (ex. including an <code class="stacks-code">.s-btn</code> or <code class="stacks-code">.s-link</code> somewhere).</p>
+    <p class="stacks-copy">A <code class="stacks-code">:hover</code> style for border color is automatically added to all linked cards. For linked cards with a box shadow (<code class="stacks-code">.bs-*</code>), adding a <code class="stacks-code">.bs-hover</code> class will apply a hover style to the box shadow as well.</p>
+    <div class="stacks-preview">
+{% highlight html linenos %}
+<a href="#url" class="s-card">
```
aaronshekey: We generally use the `…` character with dummy URLs.

docs/product/components/cards.html
aaronshekey: I think we can remove the `bs-hover` class / guidance here and replace with our newly-shipped `h:bs-sm` classes since those are based on our semi-transparent box-shadows.

docs/product/components/cards.html
```diff
@@ -1,73 +1,142 @@
...
-            <div class="grid">
-                <article class="s-card s-card__muted wmx3">
-                    <h1 class="fs-body3 fc-dark">Sample card title</h1>
-                    <p class="fs-caption fc-light">This is some explanation copy to explain what to do.</p>
+            <div class="grid grid__allcells4 md:fd-column gs12">
+                <article class="grid--cell s-card s-card__muted">
+                    <h2 class="fs-body3 lh-sm fc-dark">Base card title</h2>
+                    <p class="fs-body1 fc-medium">This is a description of the card's content.</p>
                 </article>
+                <a href="#url" class="grid--cell s-card s-card__muted">
```
aaronshekey: Just give these dummy links in examples a `#` for the href.

lib/src/components/_stacks-cards.less
```diff
@@ -14,42 +14,35 @@
...
 
     p:last-of-type {
         margin-bottom: 0;
     }
 }
 
+//  ============================================================================
+//  $   LINKED CARDS
+//  ----------------------------------------------------------------------------
+a.s-card:not(.s-card__muted) {
```
aaronshekey: Fancy!

aaronshekey:
Also, let's remember to take a pass at how this affects our documentation. We use linked cards on the documentation homepage.

docs/product/components/cards.html
TedGoas: done

docs/product/components/cards.html
TedGoas: done

docs/product/components/cards.html
TedGoas: done

docs/product/components/cards.html
TedGoas: done

TedGoas:
**`:hover` update:** How about the following hover behaviors?

1. No shadow → Small shadow
2. Small shadow → Medium shadow
3. Medium shadow → Large shadow

If a card already has a large shadow, it's maxed out and there's no larger shadow to transition to.

aaronshekey:
Big thumbs up from me. I can really dig it!

TedGoas:
@Courtny worth chatting in the future for sure! Regarding `:hover :focus :active`, I feel this is an appropriate place to componentize things and not use atomic classes since it doesn't add to CSS bloat.

PR #201, opened 2018-11-28
Visual regression testing

aaronshekey:
This is something I've wanted to do for a long time. Hopefully we can make this part of a more robust test suite that runs on CI for each PR.

**What does this do?**
This PR installs BackstopJS and configures a test suite that grabs screenshots for each of our pages at each breakpoint. This PR checks in the reference images that will get tested against.

**How do I use this?**
While Stacks is running and being continuously built, in another Terminal window, you can run `backstop test` and it will launch a new browser window with a comparison of your changes with the reference images. This is especially important to discover unintended visual regressions while working on a specific feature. If these changes are expected, you can run `backstop approve` and those new images will become the new reference.


aaronshekey:
We probably don't need to be tracking the reference images here, but let's try it out for a while and see what happens 🏆

PR #204, opened 2018-11-29
Stop tracking compiled CSS

aaronshekey:
This PR removes our compiled CSS from the repo. We can instead compile locally and deliver those files via NPM without tracking them. This way, we can avoid the crazy merge conflicts that we have constantly.

Any objections, @balpha?


aaronshekey:
Hmm. Looks like this doesn't build correctly in Netlify on the first run.

aaronshekey:
Ok, we can definitely have Netlify first run grunt to build the necessary CSS, but I haven't cracked the configuration just yet :v:

aaronshekey:
Ok, this PR is ready for actual consideration. It removes the compiled CSS from git, while modifying our Grunt and Netlify workflows to first build the CSS as needed and then deploy or publish to npm.

This should allow for the exact same workflows, while removing the routine merge conflicts that occur with compiled CSS. ✌️

PR #205, opened 2018-11-29
Audit and refactor email code

### LINKED ISSUE #202: Audit and refactor patterns in email templates
It's a lot to maintain. Even as someone who helped write these, I've had moments where I got lost in the code.

I'd like to remove any duplicate or superfluous inline styles and remove a design pattern or two from the promotional template.

TedGoas:
This will resolve #202 

---
**Working To Do's:**
- [x] Group <style> into CSS Resets and Progressive Enhancements
- [x] Remove the CSS reset for lists in Win Outlook because we’re not using it.
- [x] Add a reset for DPI scaling in Win Outlook
- [x] Clean up a few inconsistencies with Team ID and images.
- [x] Remove unused CSS classes
- [x] Remove superfluous table nesting
- [x] Remove unused patterns


TedGoas:
✅ Ready for review. This is mostly a refactor; there's only one visible change (mentioned below).

Ways to help:
1. If you have a Litmus account, test each template.
2. Look at the code and ask questions/leave comments on anything that looks questionable.

---

**Visual change:** I pared back the promotional email template to be more inline with emails we've done. It looks like this now:

screen shot 2018-12-03 at 8 11 56 am

docs/email/components/footers.html
```diff
@@ -96,11 +96,11 @@
             <!-- Sender Info : BEGIN -->
             <tr>
                 <td style="padding-bottom: 5px; text-align: left;">
-                    <img src="LogoSm@2x.png" align="left" alt="Stack Overflow" border="0" width="111" height="22" style="display: block;">
+                    <img src="LogoSm@2x.png" align="left" alt="Small Stack Overflow logo" border="0" width="111" height="22" style="display: block; font-family: arial, sans-serif; font-size: 15px; line-height: 15px; color: #3C3F44;">
```
ludwiczakpawel: Why would anyone using screenreader be interested in this logo being **small**?

ludwiczakpawel: Also, this is footer section where everything seems to be 12px. Why is this using `15px` font size instead? This is logo, so maybe having some bigger font size would make sense, but on the other hand it doesn't say just "Stack Overflow" – it's "Small Stack Overflow Logo"

docs/email/components/footers.html
```diff
@@ -769,7 +769,7 @@
                                     </td>
                                 </tr>
                                 <tr>
-                                    <td style="padding-bottom: 5px; font-size: 12px; line-height: 16px; font-family: arial, sans-serif; color: #9199A1; text-align: left;"><strong>Stack&nbsp;Overflow</strong>, <span class="unstyle-auto-detected-links">110 William Street, 28th Floor, New York, NY 10038</span></td>
+                                    <td style="padding-bottom: 5px; font-size: 12px; line-height: 17px; font-family: arial, sans-serif; color: #9199A1; text-align: left;"><strong>Stack&nbsp;Overflow</strong>, <span class="unstyle-auto-detected-links">110 William Street, 28th Floor, New York, NY 10038</span></td>
```
ludwiczakpawel: Can we save some bytes by shortening this class `unstyle-auto-detected-links` into something more like `link-unstyle` or whatever else?

docs/email/templates/code/promotional.html
```diff
@@ -37,15 +37,11 @@
             mso-table-rspace: 0pt !important;
         }
 
-        /* What it does: Fixes webkit padding issue. Fix for Yahoo mail table alignment bug. Applies table-layout to the first 2 tables then removes for anything nested deeper. */
+        /* What it does: Fixes webkit padding issue. */
```
ludwiczakpawel: Probably out of scope for this PR but since you're there... Can we ditch "What it does" string – it's used very often here and I'm not sure if it's required to understand the comment.

docs/email/templates/code/promotional.html
```diff
@@ -326,62 +288,46 @@
...
                                             <tr>
-                                                <td style="padding: 30px 10px; font-family: arial, sans-serif; font-size: 15px; line-height: 21px; color: #ffffff; text-align: center;" class="sm-py">
-                                                    <img src="illo-scene-1.png" width="340" height="" alt="alt_text" border="0" style="display: block; margin: auto; width: 100%; max-width: 340px; font-family: arial, sans-serif; font-size: 15px; line-height: 16px; color: #ffffff;" class="g-img">
+                                                <td class="s-btn s-btn__white" style="border-radius: 4px; background: #ffffff;">
+                                                    <a class="s-btn s-btn__white" href="#" style="background: #ffffff; border: 1px solid #ffffff; box-shadow: 0 0 0 0 rgba(0,149,255,0); font-family: arial, sans-serif; font-size: 17px; line-height: 17px; color: #0077cc; text-align: center; text-decoration: none; padding: 13px 17px; display: block; border-radius: 4px; white-space: nowrap;">Medium White Button</a>
```
ludwiczakpawel: Defininig color on a `box-shadow` that doesn't have any size seems unnecessary, right? Also, wouldn't `box-shadow: none;` work instead?

docs/email/components/footers.html
TedGoas: We can name it whatever we want but saved bytes aren't an issue and `unstyle-auto-detected-links` is more descriptive, as this unstyles _auto-detected_ links.

docs/email/components/footers.html
TedGoas: RE: Logo alt text: good point. In fact we probably don't need alt text at all since the company name is in HTML text nearby, so an `alt=""` should do. https://www.w3.org/WAI/tutorials/images/decision-tree/

docs/email/templates/code/promotional.html
TedGoas: We can remove it if you think it's superfluous.

docs/email/templates/code/promotional.html
TedGoas: You're right, we could probably remove this line entirely since there's no cascade we have to defend against in email. Wondering if this is a relic of me pasting something from our web design system.

docs/email/components/footers.html
TedGoas: Good point about font size of alt text... I'll make that 12px to match rest of the footer.

docs/email/templates/code/promotional.html
UXPiper: I actually enjoy the what it does

PR #210, opened 2018-12-11
Absolute positioning responsive classes

aaronshekey:
This PR adds a few responsive classes for absolute positioning. Sometimes you just need to pin the absolute dimensions at smaller breakpoints.


aaronshekey:
You know what, I don't think we actually need these.

PR #213, opened 2018-12-13
Codify $Num$ Counts in Button Groups

### LINKED ISSUE #208: Codify $Num$ Counts in Button Groups
**Problem**
There are several instances where numerical counts are included in [button group filters](https://stackoverflow.design/product/components/button-groups), but we don't have a codified way to design these counts. Many instances are text-only (matching the filter label), but that seems like lazy and neglected design. We've experimented with a modified badge design in the past, but it those designs never felt right.

**Let's come up with a good recommendation!**

---

screen shot 2018-12-06 at 4 49 49 pm
screen shot 2018-12-06 at 4 50 24 pm

TedGoas:
This PR fixes #208.

---

Let's start with a few options and discuss:
screen shot 2018-12-13 at 5 05 42 pm


TedGoas:
Alright y'all, ready for review! A few notes:

* I scoped this to `s-btn-group`'s using `s-btn-group__badge`.
* I mimicked the `mini-badge` styles, but bumped up the font size from `11px` to `12px`.
* There a "magic number" [here](https://github.com/StackExchange/Stacks/pull/213/files#diff-fbb86310fb6a4cb75a57bb6b05041f3fR84) that I don't love but it does the job.

aaronshekey:
image

docs/product/components/button-groups.html
```diff
@@ -64,30 +64,30 @@
 <section class="stacks-section">
     {% header h2 | Counts %}
     <p class="stacks-copy">
-        Used to display a numerical count inside a button group. Good for large data sets where it's helpful to quickly know how many items are in each filter.
+        Counts are used to display a numerical count inside a button group. This is good for large data sets where it’s helpful to quickly know how many items are behind each filter.
```
TedGoas: We should be consistent with our language here. I've been using the same style in various places. We should audit this sometime.

TedGoas:
Noticed the vertical spacing is off visually:

screen shot 2019-01-16 at 9 36 24 am

TedGoas:
@aaronshekey Hey [these are great](https://github.com/StackExchange/Stacks/pull/213#issuecomment-454594258)! I have a few legibility and accessibility concerns on a few of these towards to bottom, but overall these are looking good and I'm glad you addressed this in this PR.

Don't want to step on your toes, lemme know if you want to poke at this.

screen shot 2019-01-16 at 9 45 10 am

aaronshekey:
> Don't want to step on your toes, lemme know if you want to poke at this.

I'm considering shipping this and punting on making those styles perfect, mostly because they're so rarely used. I'd rather get a release out the door and come back to these later. Any hesitations there?

TedGoas:
@aaronshekey no hesitations, I'm with you.

Issue #216, opened 2018-12-21
Email: Improve code to negate auto-detected links


TedGoas:
Jason has [a more bulletproof way](https://www.24a11y.com/2018/accessibility-and-inclusion-with-email/) of handling this. We should add this to our templates:

```
/* iOS BLUE LINKS */
a[x-apple-data-detectors] {
   color: inherit !important;
   text-decoration: none !important;
   font-size: inherit !important;
   font-family: inherit !important;
   font-weight: inherit !important;
   line-height: inherit !important;
}

/* GMAIL BLUE LINKS */
u + #body a {
   color: inherit;
   text-decoration: none;
   font-size: inherit;
   font-family: inherit;
   font-weight: inherit;
   line-height: inherit;
}

/* SAMSUNG MAIL BLUE LINKS */
#MessageViewBody a {
   color: inherit;
   text-decoration: none;
   font-size: inherit;
   font-family: inherit;
   font-weight: inherit;
   line-height: inherit;
}
```

bnickel:
Do you have context for where this issue is happening?  Is it in user generated content or templates?  Is it just that links are getting added or that they're getting added and styled in a way that doesn't match the stylesheet rules.

TedGoas:
@bnickel it happens anywhere a phone number or postal address appears. Some email clients detect this and make ugly, default-blue links out of regular HTML text. Almost every template has Stack's postal address in the footer... and occasionally content like this appears in the email body.

We already have [some code](https://github.com/StackExchange/Stacks/blob/develop/docs/email/templates/code/transactional-short.html#L57) in our templates for this, but this code covers a few more clients.

---

<img width="438" alt="screen shot 2018-12-21 at 1 34 21 pm" src="https://user-images.githubusercontent.com/1172461/50357836-4a4d0080-0525-11e9-9a0f-a7a21437851e.png">

bnickel:
Based on the rules, I'm guessing it isn't just a matter of them sticking an `<a>` in there but also UA specific styles?  That's rough.

I just wanted to point out that if you want to completely remove auto-links, some email clients honor the `format-detection` meta-tag:

    <meta name="format-detection" content="telephone=no, address=no, email=no, date=no">

There are also hacks to prevent detection in areas by either inserting hidden characters to disrupt the recognizer or wrapping content you don't want linked in an un-styled `<a>` tag.

TedGoas:
[...]

PR #217, opened 2018-12-31
Email: Improve code to negate auto-detected links

TedGoas:
1. Added a `<meta name="format-detection">` tag as per [Brian's suggestion](https://github.com/StackExchange/Stacks/issues/216#issuecomment-449468156).

2. Left iOS untouched since they're already tightly scoped.

3. Added rules for Gmail and Samsung Mail but didn't include an `!important` because they're more globally defined and would otherwise impact non-auto-detected links.


TedGoas:
@aaronshekey does this happen often with _our_ emails? Do you have an example?

aaronshekey:
We have no actual need for auto highlighting phone numbers, so this is super safe. Since the only phone numbers we would pass around would be false positives, take your green checkmark and merge this 👍

PR #219, opened 2019-01-03
Account is suspended message

Herst:
BTW, the numbers below or equal to 12 having to be written out rule is outdated, right?


TedGoas:
@Herst 👋 Hey thanks for the PR. Good catch on making the account text more appropriate. Looks like a copy/paste error on my part. We're working from an internal doc that has our intended text, so I'd like to update your suggestion with what I should have pasted. Happy to discuss further.

---

> BTW, the numbers below or equal to 12 having to be written out rule is outdated, right?

I'm... not sure? We're currently working on a *Grammar and Mechanics* section to define some of the finer details like this. It would be great to include this in that section.

PR #221, opened 2019-01-04
Popovers

aaronshekey:
Following @balpha's advice to ship the markup and styling well ahead of our JavaScript implementation, here are popovers!

This introduces and documents the `s-popover` class. A full technical spec will follow with what I'm thinking for a JavaScript implementation.

This implementation makes the arrow completely independent of the `.s-popover` parent, leaving positioning of the popover up to whomever's implementing. It also adds an `.s-popover--close` button, for the case of a dismissible popover.


lib/css/components/_stacks-popovers.less
```diff
@@ -0,0 +1,250 @@
...
+        transition: opacity 100ms @te-smooth 0s, z-index 0s 0s, visibility 0s 0s; // Transition in
+    }
+}
+
+//  ============================================================================
+//  $   CLOSE BUTTON
+//  ----------------------------------------------------------------------------
+
+.s-popover--close {
+    position: absolute !important;
```
balpha: I strongly prefer floating this instead of positioning it absolutely. That way there is no

> need to adjust manually to make sure the close button doesn’t overlap its contents

which nobody ever does.

lib/css/components/_stacks-popovers.less
```diff
@@ -0,0 +1,250 @@
...
+//      • TOP & BOTTOM DIRECTIONS (Left, Center, Right)
+//      • LEFT
+//      • RIGHT
+//      • LEFT & RIGHT DIRECTIONS (Top, Center, Bottom)
+//
+//
+//  ============================================================================
+//  $   VARIABLES
+//  ----------------------------------------------------------------------------
+@arrow-width:                @su8;
```
balpha: Optimally these variables should not be in the global scope. If that's unavoidable, they should be at least namespaced with something like `@stacks-internals-...`.

lib/css/components/_stacks-popovers.less
```diff
@@ -0,0 +1,250 @@
...
+//  TABLE OF CONTENTS
+//  • VARIABLES
+//  • BASE STYLE
+//  • ARROWS
+//  • LOCATIONS
+//
+//  ============================================================================
+//  $   VARIABLES
+//  ----------------------------------------------------------------------------
+@popover-background: @white;
```
balpha: see below about goobal variables

lib/css/components/_stacks-popovers.less
aaronshekey: Good call!

lib/css/components/_stacks-popovers.less
```diff
@@ -0,0 +1,236 @@
...
+    border: 1px solid @black-100;
+    background-color: @white;
+    box-shadow: @bs-sm;
+    color: @fc-dark;
+    font-size: @fs-body1;
+
+    will-change: visibility, z-index, opacity; // Not supported in IE11 / Edge
+    transition: opacity 100ms @te-smooth 0s, z-index 0s 100ms, visibility 0s 100ms; // Transition out
+
+    &[aria-hidden="false"] {
```
balpha: The purpose of `aria-hidden` is to be able to hide elements from assistive technologies even though they are visible. If we tie visibility to this attribute, we lose the possibility of doing that.

We should use a state class instead (probably `is-visible`?). Because `visibility: hidden` also [hides the element from screen readers](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility#Accessibility_concerns), there is no need for an additional `aria-hidden`.

Also, we have to add the fact that `s-popover` is invisible by default and instructions on how to show it to the documentation.

lib/css/components/_stacks-popovers.less
```diff
@@ -0,0 +1,236 @@
...
+    z-index: @zi-hide;
+    opacity: 0;
+    border-radius: @br-md;
+    border: 1px solid @black-100;
+    background-color: @white;
+    box-shadow: @bs-sm;
+    color: @fc-dark;
+    font-size: @fs-body1;
+
+    will-change: visibility, z-index, opacity; // Not supported in IE11 / Edge
```
balpha: Is `will-change` solving an actual problem here or is it just there preemptively? If the latter, I would prefer removing it. To quote [MDN](https://developer.mozilla.org/de/docs/Web/CSS/will-change#Summary),

> *Don't apply will-change to elements to perform premature optimization.* If your page is performing well, don't add the `will-change` property to elements just to wring out a little more speed.

PR #224, opened 2019-01-15
Loading component

aaronshekey:
This PR explores adding a component called `.s-loading` that has a couple fixed sizes that can be added as a single div wherever you need a loading spinner. It also unifies the visuals for the `is-loading` modifier on `s-btn`.

image

### DIFF
--- a/docs/_data/nav.yml
+++ b/docs/_data/nav.yml
@@ -84,8 +84,6 @@
           url: product/components/labels
         - title: Links
           url: product/components/links
-        - title: Loading
-          url: product/components/loading
         - title: Modals
           url: product/components/modals
         - title: Notices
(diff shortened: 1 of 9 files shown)


balpha:
Two thoughts:

One, what if we make it use `currentColor` for the spinner color, like this:

spinners

This would make the spinners very easily themeable, something I know that Enterprise has a use for.

Two, the constant motion evokes a feeling of "this is never going to end" in me. A more deliberate speed curve instead of a constant rotating velocity may help to convey "work is actively being done right now". What do you think?

TedGoas:
+1 for it being themeable.

aaronshekey:
Ok, I'm going to move forward with `.s-spinner`. Unfortunately, the animations on IE11 and Edge wobble with the top-level spin, so punting on that for now.

PR #228, opened 2019-01-18
Fix form element descenders getting cropped

aaronshekey:
This PR is meant to address descenders in our form elements, specifically the select menu, getting cut off. I've increased the line-height a bit across our form elements while compensating with the padding to make sure everything stays pretty much the same size.

A weird test image
screen shot 2019-01-18 at 1 02 55 am

Using the Backstop tests, you can see things get normalized a little bit better across `s-select` and `s-input`, and multiple lines actually look better in `s-textarea`. An added benefit to the reduced padding is the `s-input__sm` is even more compact :v:


aaronshekey:
**Production**
image

**With normalization**
screen shot 2019-01-22 at 11 32 30 am

With normalization, the text input loses a couple pixels of height, but now by default when putting a select next to an input, they'll be the same height regardless of being wrapped in a grid.

I think this one's good to go. Any last thoughts, @balpha?

balpha:
> Any last thoughts, @balpha?

Only one: `@input-padding` shouldn't exist in the global scope. But you didn't newly introduce it in this PR, so I'm willing to let this one go 😛

PR #231, opened 2019-01-21
make Stacks.addController an officially supported helper

balpha:


b-kelly:
Looks good to me. As a test I've re-implemented my SO js paginator (see [gist](https://gist.github.com/b-kelly/e75ff0b30cd20777771bd0b5c66eeb7a)). Everything works great and Stimulus was a HUGE time-saver.

PR #235, opened 2019-01-23
s-expandable

balpha:
- rename `s-collapsible` to `s-expandable` and default to collapsed
- make it a stand-alone thing instead of an appendix on `s-widget`
- provide an `s-expandable-control` JavaScript component for the default behavior


balpha:
> I'm torn on the naming of this one right now.

Based on the rest of your message, I assume you're referring to the name of the JS component (`s-collapsing-control`) only, correct?

> It almost seems to me like this is just as usable as a way to *expand* and *collapse*.

Well, that's exactly what it does, so I'm not quite sure what you're saying here?

> Does something like `s-toggle` have more utility?

Can you clarify what *exactly* `s-toggle` would be do (and do differently) in your thinking there? (Also, "toggle" is such an overloaded term that I'd be wary of using it as a component name)

balpha:
@aaronshekey ping

aaronshekey:
So, as it's named, calling something that both expands and contracts an `s-collapsible` seems a little limiting, especially since we've got some examples where it's hidden by default, and is expanded when interacting with.

Currently the page calls it `Collapsible`. `Toggle` could be confusing too. Maybe we already had it with calling it `Accordion`? How about `Show / Hide Thingy` 😛 ? That's what this actually does in my mind.

aaronshekey:
We talked about this face-to-face and we think that switching this to `s-expandable` will work since most uses will default to something being collapsed. JS will be renamed appropriately.

PR #241, opened 2019-01-24
Various bugfixes

### LINKED ISSUE #87: Please make .s-description text more legible (accessibility concern)
I was referred here from [Meta.SE](https://meta.stackexchange.com/a/313437/162102).  The new [radio menu](https://stackoverflow.design/components/radio) for flag comments has two types of text, a larger bold label and a smaller (gray?) description of what that option is for.  I find the description hard to read.  A description sounds like something that should be a "generally normal" size, not like a footnote or a subscript where we expect things to be smaller.  The developer there told me that the description is using the .s-description (element? class? style?) and I should make my request here.

There seem to be two issues here, and it's possible that addressing either one of them would solve the problem (or maybe it needs both):

1. The text is very small.
2. The text is light.  I think it's gray not black, and the face is not a "heavy" one.

I don't know enough about Stacks to be able to make better suggestions or explore alternatives, sorry.

Here's the screenshot I posted over on Meta: 

image

### LINKED ISSUE #122: Both input label and description should highlight the checkbox / radio button on click.
In our comment flagging UI, clicking the description of a radio button doesn't also highlight the radio button. As reported on [meta](https://meta.stackexchange.com/questions/313454/please-wrap-the-entire-comment-flag-description-in-a-label-making-the-entir), this would allow for a much more fuzzy click. We should update our example markup and any css to enable this.

### LINKED ISSUE #214: Make links in `s-notice__important` underlined
image

The second half of this ☝️ contains a link, but it's visually identical to non-linked text. Can we make the white link underlined?

---

This uses `s-anchors s-anchors__inherit` [defined here](https://stackoverflow.design/product/components/links#descendent-anchor-examples). Perhaps we can build on that?

### LINKED ISSUE #226: Feature request: cursor:help
Should we add an atomic class for `cursor:help;`?
---
**Use Case**
Some UI elements have a `title` attribute containing helpful information and it's not immediately obvious that if the user pauses her cursor for a moment, something will appear. A `cursor:help` style could suggest something has a `title` attribute.

Sometimes we don't have time to do a proper Tooltip.

hovering over the checkmark icon
☝️ hovering over the checkmark icon

aaronshekey:
This PR closes #226, closes #214, closes #87, and closes #122 

:v:

### DIFF
--- a/lib/css/components/_stacks-notices.less
+++ b/lib/css/components/_stacks-notices.less
@@ -29,11 +29,6 @@
         box-shadow: @bs-sm;
         pointer-events: all;
     }
-
-    a:not(.s-btn) { // We don't want to overwrite our button styling
-        color: inherit;
-        text-decoration: underline;
-    }
 }
 
 & {
--- a/lib/css/components/_stacks-notices.less
+++ b/lib/css/components/_stacks-notices.less
@@ -29,6 +29,11 @@
         box-shadow: @bs-sm;
         pointer-events: all;
     }
+
+    a:not(.s-btn) { // We don't want to overwrite our button styling
+        color: inherit;
+        text-decoration: underline;
+    }
 }
 
 & {
--- a/docs/_data/product/atomic-definitions.yml
+++ b/docs/_data/product/atomic-definitions.yml
@@ -57,6 +57,9 @@ cursors:
   - class: c-not-allowed
     output: "cursor: not-allowed;"
     responsive: false
+  - class: c-help
+    output: "cursor: help;"
+    responsive: false
 
 pointer-events:
   - class: pe-auto
--- a/docs/_data/product/links.yaml
+++ b/docs/_data/product/links.yaml
@@ -14,6 +14,9 @@ single-links:
   - class: .s-link__inherit
     applies: .s-link
     description: Applies the parent element’s text color.
+  - class: .s-link__underlined
+    applies: .s-link
+    description: Adds an underline to the link’s text.
   - class: .s-link__visited
     applies: all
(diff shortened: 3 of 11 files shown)


lib/css/components/_stacks-notices.less
```diff
@@ -29,6 +29,11 @@
         box-shadow: @bs-sm;
         pointer-events: all;
     }
+
+    a:not(.s-btn) { // We don't want to overwrite our button styling
+        color: inherit;
```
balpha: This is going to apply the color with 021 specificity, which in particular means it's more specific than `.s-link` -- is that intentional?

aaronshekey: Yeah, currently the `s-link`s look a little goofy in there, so I opted for styling the `a` instead of the `s-link`. We can re-style the `s-link`, but since this is a component I thought it might be nice to not have to add `s-link` to each link, though I'm happy to.

balpha: That was not my question. You're making it impossible to use `s-link` here.

aaronshekey: This CSS didn't ship. This is an outdated conversation. Instead of very specific CSS, I added `s-link__underlined` and changed the examples to just use an `s-link s-link__inherit s-link__underlined` :v:

aaronshekey: I think I just missed pressing the `resolve conversation` button 😛

balpha: Nah, a conversation attached to a code line is usually marked as outdated when that code line is edited afterwards. Anyway, moot point. All good.

Issue #238, opened 2019-01-24
wb-break-word is not a legal atomic class


balpha:
One of the atomic classes we [offer](https://stackoverflow.design/product/base/typography#layout-classes) is `wb-break-word` for `word-break: break-word`.

The problem is that this is not a thing &ndash; `break-word` is not a legal value for `word-break`. The WebKit(ish) browsers just pretend it is, so it works in some browsers. And our users are [relying on this to work everywhere](https://gh.stackoverflow.com/Core/StackOverflow/pull/1849).

`break-word` *is* a legal value for `word-wrap` (actually `overflow-wrap`, but let's not go there).

We actually do have a `ww-break-word` class for that.

So we have two options:

1. Remove the "illegal" `wb-break-word` class and replace all its uses with `ww-break-word`, or
2. keep it, but change it to mean `word-wrap: break-word;` (basically emulating what the WebKit browsers do, but for all browsers).

aaronshekey:
I'd prefer the first option. Thanks for articulating this. It's been bugging me since the earliest days of Stacks.

PR #242, opened 2019-01-24
Kill .wb-break-word

aaronshekey:
This PR is a fix for #238. We'll need a corresponding PR in Core and Careers to double replace `wb-break-word` with `.ww-break-word`.

It's used in 16 spots on Core, but in super straightforward ways.


balpha:
Core PR created and assigned to you

balpha:
@aaronshekey I made a few additional changes here, see my commits.

We should use `ow-` classes (because that refers to the official `overflow-wrap` property). But while we still support IE, we also have to use `word-wrap` in the created CSS; once we drop support, we can remove it.

aaronshekey:
Got it. Thank you for these changes!

PR #243, opened 2019-01-25
Add atomic visibility classes

### LINKED ISSUE #222: Support for screenreader-only content
As highlighted on [meta](https://meta.stackexchange.com/questions/310596/screen-reader-users-cannot-discover-badges-and-reputation) and mentioned by @JasonPunyon in #stacks, it may be worth Stacks providing support for [wrapping text in a div visible only to screen readers](https://make.wordpress.org/accessibility/2015/02/09/hiding-text-for-screen-readers-with-wordpress-core/).

This is definitely one approach, but there may other more elegant solutions, like stronger title attributes or aria-attributes.

### LINKED ISSUE #227: Add s-spinner accessibility and usage guidance
**Is your feature request related to a problem? Please describe.**

When we show a loading indicator on the page, we should also hint to screen readers that we're loading content.

**Describe the solution you'd like**

I did this for Stack Overflow:

~~~html
<div class="s-spinner" role="alert" aria-busy="true">
    <div style="font-size: 0px; font-color: transparent">loading...</div>
</div>
~~~

When added to the DOM, VoiceOver accounts the contents of the div.  The styles could be incorporated into a `s-spinner__message` class or a `sr-only` class from #222.

Additionally, the usage guidance should suggest temporarily adding `aria-busy="true"` to the component that triggered the spinner, indicating that work is in progress.

aaronshekey:
This PR adds atomic classes for `visibility` and a utility class for showing text only to screen readers.

Closes #222 
Closes #227

### DIFF
--- a/docs/_data/nav.yml
+++ b/docs/_data/nav.yml
@@ -56,6 +56,8 @@
           url: product/base/typography
         - title: Vertical Alignment
           url: product/base/vertical-alignment
+        - title: Visibility
+          url: product/base/visibility
         - title: Width & Height
           url: product/base/width-height
         - title: Z-Index
--- a/docs/_data/product/atomic-definitions.yml
+++ b/docs/_data/product/atomic-definitions.yml
@@ -58,6 +58,17 @@ cursors:
     output: "cursor: not-allowed;"
     responsive: false
 
+visibility:
+  - class: v-visible
+    definition: The element visible
+    responsive: false
+  - class: v-visible-sr
+    definition: The element is visible only to screen readers
+    responsive: false
+  - class: v-hidden
+    definition: The element is invisible, but still affects layout as normal
+    responsive: false
+
 pointer-events:
   - class: pe-auto
     output: "pointer-events: auto;"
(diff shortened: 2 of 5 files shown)


balpha:
For the spinner in particular, why are we using CSS hacks instead of things like `aria-label`?

aaronshekey:
> For spinner in particular...

Good question,  @bnickel, is there an `aria`-only solution that makes sense here?

bnickel:
@balpha It had slightly better outcomes in my testing. https://codepen.io/bnickel/pen/ErPmWe


| | `.v-visible-sr`  | `aria-label` |
|---|---|---|
| Windows Chrome with NVDA | "alert loading with hidden text" | "loading with aria label alert" |
| Mac Safari with VoiceOver | "loading with hidden text" | "loading with aria label" |
| iOS Safari with VoiceOver | "loading with hidden text" | (nothing) |

aaronshekey:
Interesting that iOS Safari ignores the aria-label solution here. Gonna roll with this as is since we can always revisit our accessibility strategies (and should!)

Issue #236, opened 2019-01-24
Broken link in the Stack Overflow Design page


arulkumar4:
I see a broken link in one of the Stack Overflow Design page.

In the page [https://stackoverflow.design/email/base/typography](https://stackoverflow.design/email/base/typography), the first paragraph contains a link with "the main Stacks typography page". It is a broken link. 

Can you please fix the broken link?

aaronshekey:
Thanks for pointing this out! Something about our jekyll template was messing up our relative link. I've removed it for now.

PR #247, opened 2019-01-31
Create pattern for multi-column layouts

TedGoas:
This PR documents the multicolumn layouts present in our templates.

- [x] Two even columns
- [x] Two uneven columns
- [x] Two uneven columns, reverse stacking
- [x] Three even columns


ludwiczakpawel:
I don't know what your thoughts about 2+ columns layout are, but I personally think it basically doesn't make sense to build layouts using more than two columns for emails:

1. We have ~600px of width so it's not enough for more than 2 columns, especially when those include text. It's like <200px per each column, not even mention about spacing between.
2. I found it very tricky to make those responsive and good-looking at the same time.

That being said... I know you're focusing this PR on 2-columns layout but since it's called _**multi**-column_ maybe we could add small notification saying something like:
> Avoid building layouts based on more than 2 columns

Unless of course you disagree with my thinking.

Why am I bringing this up? When I was designing some emails I thought it would be cool to have three columns somewhere. But when I started actually writing code for that I quickly realized that I basically need to change my designs :). So I'm bringing this up because seeing this notification would save me some time.

TedGoas:
@ludwiczakpawel We made these templates `680px` wide so that we could make both *two* and *three* column hybrid layouts with padding. I've seen a few design explorations with three column layouts, but so far in practice, we only ever do two column layouts. One and two column layouts are definitely our go-to layouts.

But since we've already got the foundation in place to do three columns, I'd like to support that. But, as you say, we should document that somehow.

> Avoid building layouts based on more than 2 columns *unless ________*.

Make it known that 3 columns isn't as common. How does that sound?

ludwiczakpawel:
> Avoid building layouts based on more than 2 columns _unless you like pain_.

TedGoas:
@aaronshekey @balpha I placed each code example in an iFrame using [jQuery UI resizable](https://jqueryui.com/resizable/). I'm open to using something different or tweaking what's there.

Also for the life of me cannot figure out why `<a href="{{ "/email/base/responsiveness/" | relative_url }}"></a>` is not working in _some_ cases. It's working when I reference the iFrame src, but not in paragraph under the page title. I've copied/pasted `relative_url` from elsewhere in the site. 🤷‍♀️

aaronshekey:
> is not working in _some_ cases

That link comes from Jekyll's meta data description section, right? I've found that links don't work as designed in there. You'll have to figure out somewhere else to link to that page if it's truly important. See https://github.com/StackExchange/Stacks/issues/236

TedGoas:
That's strange, I swore [this](https://github.com/StackExchange/Stacks/issues/236) was working when I committed it. Ok, I'll find another place to include this link.

TedGoas:
@aaronshekey @balpha This is ready for formal review.

As discussed, I updated the examples to use the existing jQuery pattern we've used elsewhere. JS is not optimal in its current state, but works without pulling any new JS libraries into the site.

The hope is to merge this and balpha can rewrite this in stimulus when he has time.

PR #248, opened 2019-02-01
Make the "current page" highlight color for navigation sidebar widgets a configuration option

balpha:
This addresses (or rather, allows Q&A code to address) https://meta.stackexchange.com/questions/323309/currently-selected-indicators-colour-in-flag-summary-and-help-pages-can-be-the.

This splits up the CSS for `s-sidebarwidget` into a static and a dynamic file, since the vast majority of this CSS can be delivered in the static part and doesn't have to be compiled for each and every Q&A site.

screenshot with the highlight color blue 

screenshot with the highlight color orange, the default


balpha:
Yep, that change is already in: https://gh.stackoverflow.com/Core/StackOverflow/commit/cabf0b61a3cbd852ea1fe6d3a7287b419ace7ab2. This will just work (tm) once Stacks is updated there.

PR #253, opened 2019-02-06
Breadcrumbs

### LINKED ISSUE #234: Breadcrumbs
In an effort to further orient our users, certain pages can benefit from breadcrumbs. We ought to have a component for this.

---
cc: @ludwiczakpawel

aaronshekey:
This PR creates the minimum viable version of breadcrumbs. It closes #234 and is based on @ludwiczakpawel's awesome work in that issue.

Offering a dropdown to access children of parent and grandparent sections is out of scope for now, but something we can consider offering UI for when popovers are a mature component.

TODO:
- [x] Merge `@Svg.ArrowRightAltSm` into https://github.com/StackExchange/Stacks-Icons
- [x] Test responsive Less in Core / Careers
- [x] Kill the batman


aaronshekey:
These do what we want them to in Core.

image

image

aaronshekey:
This is ready for a sanity check from the crew. :v:

docs/product/components/breadcrumbs.html
```diff
@@ -0,0 +1,42 @@
...
+    <p class="stacks-copy">Breadcrumbs show users where they are within a site’s hierarchy. Breadcrumbs help orient the user and allow for navigation to previous page levels. It can be appropriate to use a breadcrumb when:</p>
+    <ul class="stacks-list">
+        <li>The user is many levels deep</li>
+        <li>The current page does not have its own navigation</li>
+        <li>The user needs to quickly go back to the previous parent page</li>
+    </ul>
+    <p class="stacks-copy">Breadcrumbs should be the first page element—placed directly above the page’s title. The page the user is on should not appear in the breadcrumb trail.</p>
+    <div class="stacks-preview">
+{% highlight html linenos %}
+<nav class="s-breadcrumb mb8 sm:mb4">
```
ludwiczakpawel: Shouldn't the component be called `s-breadcrumbs` instead of `s-breadcrumb`?

docs/product/components/breadcrumbs.html
```diff
@@ -0,0 +1,42 @@
...
+    <a class="s-breadcrumb--item" href="…">
+        @Svg.LogoGlyphXxs
```
ludwiczakpawel: Should we add a comment that it doesn't always have to be an icon? Sometimes it can be just a label (mostly for enterprise and other communities).

balpha:
One issue is that this currently doesn't handle long breadcrumb trails in narrow windows:

breadcrumbs-crumbling

It should probably look something like this?

grafik

That's doable, but will need a slightly different markup structure.

balpha:
On another note, we should probably add guidance (and include this in the examples) about aria attributes here -- because by the definition of breadcrumbs, all the links are conceptually parents of the current page, they should ususally be `aria-current="true"`. If the last crumb is a link to the current page itself, the value should be `"page"`.

ludwiczakpawel:
according to what @aaronshekey wrote on the docs (and I tend to agree with) last crumb shouldn't be a link to the current page itself.
> The page the user is on should not appear in the breadcrumb trail.

TedGoas:
How concerned are we with legibility? The text is `12px` and is `#9fa6ad`, which is lighter than `.fc-light`. This example looks good in relative isolation, but I worry these breadcrumbs might get lost when added to a full view.

ludwiczakpawel:
@TedGoas that's a fair concern. I don't have strong feelings, though I _personally_ like them being very subtle.

Here are couple screenshots with real content:
method_ sso
users
plan basic

docs/product/components/breadcrumbs.html
aaronshekey: In the systems that I surveyed, it goes both ways, but a singular breadcrumb is more common. I thought we should keep our internal logic of the component itself is just called a breadcrumb, while the pages referencing them are plural. This follows with "Avatars" being `s-avatar`, "Notices" being `s-notice`, and so on.

balpha: But isn't a "singular breadcrumb" that thing that you call `s-breadcrumb--item`?

aaronshekey: Yep, indeed, no arguing that! But I think the internal logic ought to apply here. Design systems are split on this, and I've chosen `s-breadcrumb` because all our components are each singular in name. "You'll need to add a breadcrumb there".

Happy to change to `s-breadcrumbs` if it's bugging everyone too much.

balpha: Counter-example: `s-tabs` 😄 

Yeah, my vote is for `s-breadcrumbs`

aaronshekey:
With the latest commit, I've written more CSS and markup to handle the wrapping @balpha's proposed. I think it looks good when we have an icon-only home button. Not sure I like the wrapping when it uses a larger full text name.

- Darkened up the text slightly
- Added better wrapping

image

Should we just have everything wrap and get rid of the logo's column like the following screenshot?

image

ludwiczakpawel:
yeap, i'd try to keep things simple and get rid of that _logo's column_. 

I've been also thinking about having this kind of _wrapping_ for mobile:
`[logo] » ... » Keeping a straight face`
Which tl;dr is basically collapsing all levels of crumbs into `...` except the one level above (if you know what i mean).

But that might be too weird.

aaronshekey:
> Collapsing all levels of crumbs into `...`

I'd like to punt on this. I think it'd be interesting to roll up sections into a dropdown so they don't just disappear behind `...`.

aaronshekey:
Calling this one good to merge, especially as an MVP. This has been tested in IE11, Edge, Chrome, Firefox, and Safari ❤️

PR #254, opened 2019-02-07
move the "click to make narrow" thing that's repeated on various pages into a stimulus controller

balpha:


balpha:
It's correct that this file is generated, but it shouldn't be commited. I added it to `.gitignore` but forgot to commit that particular change.

balpha:
IE fix is 👍, sorry for that.

PR #255, opened 2019-02-11
Add CSS grid features

### LINKED ISSUE #232: CSS Grid Support
I've done a little poking around at what it would take to support CSS Grid within Stacks while exploring a tiny grid applied to our [homepage](https://github.com/StackExchange/Stacks/pull/230).

There is some prior art out there for an atomic approach to CSS Grid. [Atto](https://github.com/joshpitzalis/atto), [Bulldog](https://vinceumo.github.io/atomic-bulldog-grid/)

aaronshekey:
This PR starts exploring adding proper CSS grid classes. Closes #232 

- [x] Rename existing `.grid` classes to `.flex`
- [x] Rename existing `.gs` classes to `.fs`
- [x] Design a way to interact with the API in an atomic way
- [x] Add IE11 fallbacks
- [ ] Flesh out CSS grid documentation pages


aaronshekey:
Having a look at IE11 support for CSS Grid seems like it might be a lot of work to support in a Stacks / atomic world. Each cell would need to be positioned manually since auto column placement doesn't work and will never work without JS applying CSS on the fly. If we drop IE11 support, this implementation would work as designed.

aaronshekey:
The latest set of commits introduces what it'll take to support IE11 atomically. What I've set out to design is a system where you'll apply `grid` to a container, with the number of columns you'd like to have automatically flow `grid__col4`. Optionally, you can apply a grid spacing class such as `gs16` that will give you proper gutters without any margin hacks. You can also add `md:grid__col2`, and your grid will flow differently based on the window breakpoint. 

For bonus extra points, it'd be amazing if on the children we could set `grid--cell__col-start2` and `grid--cell__col-end5`, or perhaps `grid--cell__span2` to change column spans. Awesome!

Where things make sense in all browsers, and columns wrap properly with all the proper gutters, things go haywire in IE11. IE11 doesn't support the following properties that we'd rely on in an atomic system:

- `grid-gap`
- `grid-column-start`
- `grid-column-end`

We end up having to write our way around IE11's gutter limitations. A common hack is to include the gutter as a column, and then position each `grid--cell` individually. Otherwise, they just all stack at the 0, 0 position.

https://github.com/StackExchange/Stacks/blob/0f27ac994c116a368e2e280e487343a847845ef4/lib/css/atomic/_stacks-grid.less#L152-L168

I'd argue we'd want to support all these rules at different breakpoints. This would lead us to a very big bundle size to support CSS Grids—since we'll have to write IE11-specific CSS for every permutation of column position, grid spacing, and column spans multiplied by our breakpoints.

Unless we can find workarounds to support the ideal scenario as described first in this PR comment, I think we should punt on supporting CSS Grid until we can officially drop IE11 support, whenever that is.

aaronshekey:
Punting on this until we've got a better IE11 strategy, or we drop support—whichever comes first.

PR #256, opened 2019-02-11
Fixes for IE11 modal blurriness

### LINKED ISSUE #244: .s-modal can be blurry on Internet Explorer 11
**Describe the bug**
Internet Explorer's implementation of `translate3d` results in s-modal--dialog being blurry for many widths of dialogs and browsers.

**To Reproduce**
1. Go to https://stackoverflow.design/product/components/modals in Internet Explorer
2. Click on "Launch example modal"
3. Adjust the browser to various widths

**Expected behavior**
Dialog is always crisp

**Screenshots**

*sometimes crisp*

screen shot 2019-01-30 at 9 46 31 am

*sometimes blurry*

screen shot 2019-01-30 at 9 43 21 am

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Internet Explorer
 - Version 11.253.17763.0

**Additional context**
You can see this on https://stackoverflow.com with the comment flagging dialog.

aaronshekey:
This PR closes #244.

It seems that any type of translation on IE11 is causing things to align to the subpixel blur. Simply removing those translations wasn't an option, because we had a double `position: fixed` situation happening. According to the comments, this was necessary for Edge, but I wasn't able to reproduce the scrolling issue.

So, I removed the second `position: fixed` and switched to `transform: none` for IE11. We'll still get some transitions, but things won't shrink and grow as they do in other browsers.

image

### DIFF
--- a/lib/css/components/_stacks-modals.less
+++ b/lib/css/components/_stacks-modals.less
@@ -67,8 +67,7 @@
         transform: translate3d(0, 0, 0) scale3d(1, 1, 1)
     }
 
-    //  IE11 and below don't correctly position the dialog box, so we need to set some additional
-    //  values to make sure it visibly centers.
+    //  IE11 and below aligns our dialog on subpixels causing a blurry layout. Let's just remove those transforms for IE11.
     @media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
         transform: none;
 
--- a/lib/css/components/_stacks-modals.less
+++ b/lib/css/components/_stacks-modals.less
@@ -48,7 +48,6 @@
 .s-modal--dialog {
     overflow-y: auto;
     visibility: hidden;
-    position: fixed; // We shouldn't NEED this, since the parent is fixed, but IE Edge needs it to properly scroll at small viewport sizes.
     z-index: @zi-hide; // Make sure it's also below everything so we can't interact with it.
     max-width: 600px;
     max-height: 100%;
--- a/lib/css/components/_stacks-modals.less
+++ b/lib/css/components/_stacks-modals.less
@@ -69,10 +69,10 @@
 
     //  IE11 and below aligns our dialog on subpixels causing a blurry layout. Let's just remove those transforms for IE11.
     @media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
-        transform: none;
+        transform: translate(0, 30%) scale(0.6, 0.6);
 
         .s-modal[aria-hidden="false"] & {
-            transform: none;
+            transform: translate(0, 0) scale(1, 1)
         }
     }
 }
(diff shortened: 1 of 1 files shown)


lib/css/components/_stacks-modals.less
```diff
@@ -68,15 +67,12 @@
         transform: translate3d(0, 0, 0) scale3d(1, 1, 1)
     }
 
-    //  IE11 and below don't correctly position the dialog box, so we need to set some additional
-    //  values to make sure it visibly centers.
+    //  IE11 and below aligns our dialog on subpixels causing a blurry layout. Let's just remove those transforms for IE11.
     @media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
```
balpha: Turns out that this isn't a good browser hack for IE11. It usually seems to work because `-ms-high-contrast: none` is no longer supported in Edge, but `active` is still one of the legal values there.

I would do this the other way around: Make `transform: none` the default, and only add actual transforms inside a "is not IE" check. The latter is pretty reliably doable by using a `@supports` feature query &ndash; and it doesn't matter which one. because IE doesn't support `@support` at all (yo dawg).

[Here's an example](https://github.com/StackExchange/Stacks/blob/26681d2bfa80293395802a733b79717482f5a4df/lib/css/components/_stacks-spinner.less#L29).

lib/css/components/_stacks-modals.less
```diff
@@ -48,7 +48,6 @@
 .s-modal--dialog {
     overflow-y: auto;
     visibility: hidden;
-    position: fixed; // We shouldn't NEED this, since the parent is fixed, but IE Edge needs it to properly scroll at small viewport sizes.
```
balpha: You added this in 97d1b1d6ad299a296ac3cd3aa3751ca2803b03f3 &ndash; do you happen to recall what precisely the scrolling issue was that you where referring to at the time?

lib/css/components/_stacks-modals.less
balpha: So I just played with this and it appears that the blurriness doesn't even happen if you use 2d transforms instead of 3d transforms. Since we're not actually doing anything with the third dimension anyway, we should just do that for the best of both worlds, giving IE users transtions *and* crisp fonts.

aaronshekey: The first thing I tried was switching from translate3d, but when that second `position: fixed` was there, this was still blurry. Removing the second `position: fixed` in 
https://github.com/StackExchange/Stacks/pull/256/commits/aa9d212c939b0ccf6914f164b570514464a642d4 lets us have a nice and crispy `translate`. Nice!

lib/css/components/_stacks-modals.less
aaronshekey: No, I don't remember, and I can't reproduce what the comment seems to be referring to 🤔. Is it possible that Edge has been patched in the meantime?

PR #262, opened 2019-02-20
Add login verb / noun section to grammar and mechanics

### LINKED ISSUE #259: Signin vs Sign in
`Signin` (noun) vs `Sign in` (verb)
`Login` (noun) vs `Log in` (verb)

It came up in chat and we should document that adding a space to these types of words changes it's part of speech. Perhaps in **Grammar and Mechanics**?

aaronshekey:
This PR closes #259. Since we don't actually use "Signin" or "Sign in", I've codified with "login". We use "sign up", but I don't think there is confusion there.


TedGoas:
Maybe worth including **Sign up**? You can **sign up** using the **signup** form.

PR #267, opened 2019-02-26
Email documentation edits from interviews w/ marketing

TedGoas:
I did a round of remote hallway research with a few folks on the marketing team who work with email. From my notes, I created following unabridged things we could do to make the documentation a little more clear and thorough for marketing folks.

- [x] Best practices means "design" best practices
- [x] Homepage should explain _WHO_ this is for (designers?) and why it exists.
- [x] Homepage: Components and patterns are "best practices are how we do things."
- [x] Templates, are they already in our ESPs or should I copy/paste what's here?
- [x] Accessibility: Add note about using `<strong>` and `<em>` tags.
- [x] Plain text - add notes about how ESPs do it for us but it's manual in our codebase.
- [x] Add FAQ about CSS inliners.
- [x] Preview text = need our own diagram / illustration 
- [x] Move [images FAQ](https://stackoverflow.design/email/guidelines/faq#how-should-images-be-prepared?) to images component page.
- [x] FAQ: Animated GIF = documentation on making the first frame good in case image is displayed but animation doesn’t play.
- [x] `.outlook-fix` is confusing [here](https://stackoverflow.design/email/base/mso). "Is this an actual class that does something?" (Narrator: No, it's not.) Let's rename to `.example-class` or something that's obviously a fake class to avoid confusion.
- [x] Templates: Add a line or two in each description that explains that these templates are guidelines, not rules. They're meant to be changed, remixed, and built upon. Gives a few examples of what things are typically changed (background colors, images, additional sections, etc.).


docs/email/components/images.html
```diff
@@ -90,6 +90,12 @@
             <img src="https://placehold.it/256" width="128" height="128" alt="alt_text" border="0" style="display: block;">
         </div>
     </div>
+    <div class="grid ai-start s-notice s-notice__info mb48">
+        <div class="grid--cell mr8">{% icon Alert %}</div>
+        <div class="grid--cell lh-lg">
+            <strong>Note:</strong> Every image that should <strong>scale down</strong> in small screens should have a <code class="stacks-code bg-powder-200">width: 100%; max-width: (desktop-width)px;</code> in it's <code class="stacks-code bg-powder-200">style=""</code> attribute. This allows images to scale down when their desktop width exceeds the width of its container.
```
ludwiczakpawel: I don't quite understand this _(desktop-width)px_ part even having frontend background. Although, I'm not sure how to better explain it.

docs/email/components/images.html
```diff
@@ -224,5 +230,16 @@
     <p class="stacks-copy">
         Images should be hosted within their respective ESP. Upload images in Iterable for Iterable emails, add images in GitHub for StackMail emails, etc. Images should not be referenced using services like Imgur or Dropbox, as they have a greater chance of being marked as spam.
     </p>
+    {% header h3 | Animated GIFs %}
+    <p class="stacks-copy">
+        Animated GIFs can help explain an interaction or add a bit of delight that isn’t possible with static email designs. However there a couple things to consider when designing and preparing animated GIFs.
```
ludwiczakpawel: Shouldn't that be _However there **ARE** ..._? My English is meh but this doesn't sound right even to me :DDD

docs/email/components/images.html
TedGoas: Hmm, I see how that can be confusing, I'll clarify what I mean by (desktop-width)px.

Issue #269, opened 2019-02-26
s-expandable-control enhancements


bnickel:
**Is your feature request related to a problem? Please describe.**
I have a button that shows and hides an editor panel.  I'm using `s-expandable-control` from the JavaScript package but have run into two issues.

1. There is no way to configure the button to indicate its state.
2. I want to be able to run a script every time it's about to show, in order to configure it.

**Describe the solution you'd like**
For 1, I'm planning on adding the `is-selected` to the source button.  It would like to toggle this class at the same time we're setting `aria-expanded`, along the lines of:

~~~javascript
if (this.element.classList.contains('s-btn')) {
    this.element.classList.toggle('is-selected', !newCollapsed);
}
~~~

For 2, I would like Stacks to emit events showing or hiding the element, something like `s-expandable-control:show` and `s-expandable-control:hide`.

**Describe alternatives you've considered**
For both 1 and 2, I could attach a click separate listener to the button to perform the changes, but I couldn't guarantee execution order so I'd have to use `setTimeout(_, 0)` or something. 

For 1, we could generalize the classes applied with `data-s-expandable-control-expanded-classes`, but that could be overkill.

balpha:
#270 adds the events as you suggested, which should allow you to achieve 1) and 2) for the time being.

I don't think toggling `is-selected` purely because the element happens to be `s-btn` is a safe idea. For now I would suggest that you do this in your event handler.

Something like what you suggest in your last paragraph sounds better, but I would postpone this until [Stimulus gets the `classes` API](https://github.com/stimulusjs/stimulus/pull/202), because that's addressing exactly this use case.

balpha:
We also added the `-toggle-class` parameter to #270, which has been merged now and released as part of [0.36.0](https://github.com/StackExchange/Stacks/releases/tag/v0.36.0).

PR #270, opened 2019-02-27
fire `s-expandable-control:show` and `s-expandable-control:hide` events

balpha:
Addresses #269.


aaronshekey:
I'd like to see an example of this documented, if possible. For a future PR, I'd like to see a more formal table of the javascript attributes similar to https://stackoverflow.design/product/components/tables#javascript-data-attributes

balpha:
Not sure what you mean -- I didn't add any attributes.

Also not sure what kind of examples you'd like to see.

aaronshekey:
> Not sure what kind of examples you'd like to see.

How about we toggle the class `is-selected` on this page's example? Or we can have another example further down the page:

image

We currently document that the event exists, but we don't answer the question "Ok, but how do I use it?"

> I didn't add any attributes

I think it would be helpful in a future PR to document these in the same way we did with the attributes on sortable tables. I know you didn't add any attributes.

lib/js/stacks.js
```diff
@@ -13,13 +13,25 @@
...
+    };
     StacksController.prototype.setElementData = function(element, key, value) {
         element.setAttribute("data-" + this.identifier + "-" + key, value);
-    }
+    };
     StacksController.prototype.removeElementData = function(element, key) {
         element.removeAttribute("data-" + this.identifier + "-" + key);
-    }
+    };
+    StacksController.prototype.triggerEvent = function(eventName, optionalElement) {
```
bnickel: It's not strictly needed here, but could we upgrade this to a custom event so we can pass data?

```
    StacksController.prototype.triggerEvent = function(eventName, detail, optionalElement) {
	        var event;
	        var namespacedName = this.identifier + ":" + eventName;
	        try {
	            event = new CustomEvent(namespacedName, {bubbles: true, detail: detail});
	        } catch (ex) {
	            // Internet Explorer
	            event = document.createEvent("CustomEvent");
	            event.initCustomEvent(namespacedName, true, true, detail);
	        }
	        (optionalElement || this.element).dispatchEvent(event);
	    };
```

I have a controller that wants to post a deletion event for another to read:

```
Stacks.addController('foo', {
   removeItem(e) {
    var el = e.currentTarget;
    var id = this.getElementData(el, 'id');
    el.parentNode.removeChild(el);
    this.triggerEvent('removed', { id: id });
  },
});

Stacks.addController('bar', {
   logRemoval(e) {
    var id = e.detail.id;
    this.element.textContent = `${id} removed`;
  },
});
```

```
<div data-controller="foo">
  <button data-foo-id="123" data-action="removeItem">Remove Me</button>
</div>
<div data-controller="bar" data-action="foo:removed@document->bar#logRemoval"></div>
```

balpha: Makes a lot of sense! Wanna commit that?

aaronshekey:
How's this one doing with Brian's changes, @balpha?

balpha:
Those changes are fine, I asked him to commit them (see above).

PR #272, opened 2019-03-04
Add color aliases

### LINKED ISSUE #187: Alias colors to danger/warning/info/success
We currently have [Notices](https://stackoverflow.design/product/components/notices) that come in different variations: info, success, warning, danger.

Sometimes, we want to indicate those states in other elements. For example, a table row could be marked with "warning" or "danger" if some data is outside a threshold. Or an SVG Icon might be colored like that.

Could we add some aliased [Colors](https://stackoverflow.design/product/base/colors) like fc, bg, bc that alias these? So e.g., bg-danger is the same background as the danger notice (#ffe7eb), fc-danger is the text color (#3b4045), and bc-danger is the border color (#e6d0d4)?

That allows for Theming without having to worry about notice colors being out of sync with elements meant to look like notices.

aaronshekey:
This PR adds color aliases for `.fc-success`, `.fc-danger`, `.fc-error`, and `.fc-warning`. The same aliases apply to background colors as well.

If I'm not mistaken, this should close #187. If it doesn't, I think these are still valuable as a slightly more abstracted case.

@mstum

### DIFF
--- a/docs/product/base/borders-color.html
+++ b/docs/product/base/borders-color.html
@@ -141,6 +141,7 @@
     </div>
 
     {% header h3 | Green %}
+    <p class="stacks-copy">Green borders can also be declared using our <code class="stacks-code">.bc-success</code> <a href="{{ '/product/base/colors#success' | relative_url }}">alias class</a>.</p>
     <div class="stacks-preview">
 {% highlight html linenos %}
 <div class="bb bc-green-3">Dark Green Border</div>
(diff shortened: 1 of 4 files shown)


mstum:
Yep, that addresses the main case in #187. Having the border colors would be neat as well, but the bg and fc are the important ones. Thanks!

mstum:
Awesome, that's great!

PR #281, opened 2019-03-18
Initial draft of background classes

Courtny:
Add in support for atomic background image utilities.

@aaronshekey Do we need all of these position classes? Can you think of a better way to do it?

I wanted to have just `bg-left`, `bg-right`, `bg-top`, `bg-bottom`, `bg-center`, but seeing as how the background position spec works I'm not sure I can go that atomic with them.

We could jazz up the examples a lot more too in the future for docs if you think it's helpful.

Also, does it make sense to chop all of the _background_ specific classes from the `_stacks-misc.less` into its own proper .less file?


Courtny:
Changes for review @aaronshekey <3

PR #282, opened 2019-03-20
Font stack adjustments

aaronshekey:
This PR considers using system fonts instead of Arial.

### What?

We're now specifying system fonts on Stack Overflow. On macOS and iOS, you'll see things set in San Francisco. On Windows, you'll see Segoe. On Android you'll get Roboto. Your Linux distro will show you its default as well.

### Why?

Mark Otto put it really well on his personal blog, documenting [GitHub's rationale](https://markdotto.com/2018/02/07/github-system-fonts/) for switching. Like GitHub, our original font stack used Arial across macOS, iOS, and Windows. Arial was first created in 1982, and has served the web well for _decades_. But technology moves on. Modern system typefaces look better on both new high DPI screens, and old screens alike. Apple's San Francisco and Microsoft's Segoe both look great on retina displays, have more expressive weights, and improve readability across all contexts. For example, in dark mode, we could [drop our font weights a bit for better readability](https://css-tricks.com/dark-mode-and-variable-fonts/).

### But what about consistency?

We originally chose Arial because it was available on the widest set of devices, allowing us to have a consistent experience wherever you were. However, Arial doesn't ship with Linux distros or Android devices. We're currently only achieving consistency between macOS, iOS, and Windows, leaving 2 platforms to choose their system font. We'd rather the consistency be shifted to the device itself—when viewing Stack Overflow on iOS, it'll feel that much more native in San Francisco. Same with Android, which is already being displayed in Roboto.

### Will my Stack Exchange site lose its custom font?

Nope! This change will not affect sites that have their own typefaces. Sites like Christianity and English will still be displayed in their custom fonts.

### Future possibilities

If consistency were the absolute goal, we'd ship our own custom font. I would love to be able to truly express our brand through a typeface, but that means users would have to download custom fonts and wait for them to display. Displaying custom fonts is getting more realistic every day at our scale, so we may revisit that choices at some point.

Additionally, now that lots of our site is built using our design system, we can store these font stacks as CSS variables. We could allow users to more easily specify dyslexia-friendly typefaces, or load their own entirely.

### TODO

- [x] Check that the design team that this change is desired
- [x] Vet this change with larger teams
- [x] [Communicate changes on meta](https://meta.stackexchange.com/questions/364048/were-switching-to-system-fonts-on-may-10-2021)
- [x] ~Figure out what to do about Segoe's slightly different baseline.~ We can live with it.


docs/product/base/typography.html
```diff
@@ -151,15 +151,15 @@
         <div class="stacks-preview--example grid ff-column-nowrap gs16 gsy my0 fs-subheading">
             <div class="grid--cell">
                 <p class="ff-sans mb0">Sans Serif</p>
-                <p class="ff-mono fs-caption fc-light">Arial, "Helvetica Neue", Helvetica, sans-serif</p>
+                <p class="ff-mono fs-caption fc-light">-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, "Noto Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji"</p>
```
Courtny, line 149: Do we have to reference Segoe at all? I know.. no native typeface for Windows... but given the baseline issue and not a clear way to deal with this in short term could we punt it for now?

ludwiczakpawel:
1. I think it's looking great but would be cool to steal some dev staging to test it in real life, not only Stacks.
2. Not sure if those are my eyes only but [`__mini` number badges](https://deploy-preview-282--stacks.netlify.com/product/components/badges#number-counts) look 1px off.

Please ship it.

docs/product/base/typography.html
aaronshekey: I’ve considered only adding `-apple-system, BlinkMacSystemFont` to the beginning of our stack as a bit of progressive enhancement for macOS and iOS users. Presumably Windows users would get "Helvetica Neue" if it’s installed, but most users would likely be shown Arial.

...But with that approach, we lose the tidy rationale of simply choosing platform-specific fonts. 🤔

PR #283, opened 2019-03-22
Popover javascript

aaronshekey:
This PR introduces and documents @bnickel's controller for popovers.

- [x] Convert jQuery-dependent code to vanilla JS


dist/js/stacks.js
```diff
@@ -1997,6 +1997,150 @@ Copyright © 2019 Basecamp, LLC
...
+
+        documentKeyup: function (e) {
+            var target = e.target;
+            if (e.which === 27 && this.isVisible && $(possibleChildSelector).length === 0) {
+                if (this.$popover.has(target).length > 0 || this.$popover.is(target)) { this.$source.focus(); }
+                this._toggle();
+            }
+        },
+
+        windowResize: function (e) {
```
aaronshekey: This doesn't seem to be firing in the context of the Stacks site 🤔

lib/css/components/_stacks-popovers.less
```diff
@@ -16,27 +16,25 @@
...
     border: 1px solid @black-100;
     background-color: @white;
     box-shadow: @bs-sm;
     color: @fc-dark;
     font-size: @fs-body1;
-    transition: opacity 100ms @te-smooth 0s, z-index 0s 100ms, visibility 0s 100ms; // Transition out
 
     &.is-visible {
-        visibility: visible;
+        display: block;
```
aaronshekey: This change will require some love in production to see if we're breaking things.

docs/product/components/popovers.html
```diff
@@ -26,21 +26,80 @@
...
+            data-action="s-popover#toggle"
+            aria-haspopup="true"
+            aria-expanded="false"
+            aria-controls="second-example-popover">
+            Show second example popover
+        </button>
+
+        <div class="s-popover ps-absolute" id="second-example-popover" role="menu" aria-hidden="true">
+            <div class="s-popover--arrow__lb"></div>
+            <div tabindex="0">
```
aaronshekey: This adds a nasty focus ring when interacting via the mouse. Is there a way we can only add the focus ring when the user tabs to the popover? And perhaps we put this on the popover itself. Happy to provide custom `:focus` classes for this case.

dist/js/stacks.js
aaronshekey: @bnickel Is there anything we can do in this function to make these popovers behave a bit better on smaller screens so we don't force horizontal scrolling?

docs/product/components/popovers.html
bnickel: We could just remove the `tabindex`  It was just meant as an alternative to putting dummy button or something in there.  I did choose a new div because I knew adding it to the root element would offend you more:

Screen Shot 2019-04-02 at 3 13 10 PM

aaronshekey:
I'd love to ship this right now, but I think the horizontal scrolling that these force at the smallest breakpoint is a show stopper. Is there _anything_ we can do to these at the smallest sizes?

bnickel:
`sm:d-none` on the source? :trollface:

There's not much we can do automatically because we don't really know where the source element will end up when the stuff and what's okay or not okay to cover up.

One simple idea is that we could add something like `data-s-popover-arrow-positions="sm:br"` to the source and have the controller manage arrow position as breakpoints are hit, making it incumbent on the user to test the different configurations.

aaronshekey:
Any approaches we can steal from https://popper.js.org?

bnickel:
Sure, it's just a bigger effort.  A key limitation we have at the moment is our arrows aren't designed for arbitrary placement, so we're limited to the 12.  I would need to be able to say "use top center but shift 5px to the right."

aaronshekey:
> Sure, it's just a bigger effort

I think it's worth the effort to make these as bulletproof as we can get them out of the gate. How can I help get the arrows in a state where they're arbitrarily positioned? I think we could swap classes on the fly while adjusting offsets of the arrow and the `s-popover` itself how it's currently implemented 🤔

aaronshekey:
Closing in favor of #319 :v:

PR #286, opened 2019-03-31
Documentation fixes

PeterMortensen:
Some copy editing (e.g. ref. https://en.wiktionary.org/wiki/set_up#Verb, https://en.wiktionary.org/wiki/holdout#Noun, and https://en.wiktionary.org/wiki/sure-fire#Adjective).


aaronshekey:
Thanks for catching these!

PR #288, opened 2019-04-01
Modal js

benjamin-hodgson:
for discussion: is putting the `aside` next to the `button` in the markup acceptable?


aaronshekey:
> Is putting the `aside` next to the `button` in the markup acceptable?

Ideally we'd be able to separate these out a bit more, but the modals are absolutely positioned, so they're outside the flow entirely. I worry there might be some applications where including the modal directly next to the spawning button is impossible within Core's architecture. For example, we may need to spawn an `s-modal` from within a markup helper, or somewhere we wouldn't be able to nest the modal.

Would something like this work? Or does it have to be a direct sibling?

``` html
<div class="random-parent" data-controller="s-modal">
  <div class="another-random-parent">
    <div class="s-btn-group">
      <button class="s-btn" data-action="s-modal#show">Launch Modal</button>
    </div>
  </div>
  <div class="s-modal"></div>
</div>

Instead of tying these to the nearest parent modal controller, can we have them communicate via explicit ID? The way our `s-popover` controller is designed right now uses `aria-controls="[s-popover id]"` in #283.

File: dist/js/stacks.js [hunk 1]
```diff
@@ -1997,6 +1997,32 @@ Copyright © 2019 Basecamp, LLC
...
+    Stacks.addController("s-modal", {
+        targets: ["content"],
+
+        show: function() {
+            this.contentTarget.setAttribute("tabindex", 0);
+            this.contentTarget.setAttribute("aria-hidden", false);
+            document.body.setAttribute("tabindex", -1);
+            document.body.setAttribute("aria-hidden", true);
+        },
+        hide: function() {
```
aaronshekey: How difficult is wiring the `esc` key to this `hide` function?

docs/product/components/modals.html
```diff
@@ -137,19 +186,31 @@ <h1 class="s-modal--header" id="modal-title">Example title</h1>
...
+            </a>
         </div>
-        <a class="s-modal--close s-btn s-btn__muted js-modal-close" href="#" aria-label="Close">
-            @Svg.ClearSm
-        </a>
-    </div>
-</aside>
+    </aside>
+</div>
 {% endhighlight %}
```
aaronshekey: I can handle this, but we'll want to add a table with all the data attributes explaining where to add them and what they do.

docs/product/components/modals.html
```diff
@@ -52,24 +37,88 @@ <h1 class="s-modal--header" id="modal-title">Example Danger Title</h1>
...
+                    <a class="s-modal--close s-btn s-btn__muted" href="#" aria-label="Close" data-action="s-modal#hide">
+                        {% icon ClearSm %}
+                    </a>
+                </div>
+            </aside>
+        </div>
+
+        <div class="grid--cell" data-controller="s-modal">
+            <button class="s-btn s-btn__filled" data-action="s-modal#show">Launch example modal w/ danger state</button>
+            <aside
```
aaronshekey: This whitespace pattern is so weird to me, but I understand why we do it. I might format this slightly differently.

benjamin-hodgson: No strong opinions re formatting, I just found this easier to read than having one massive line. Knock yourself out 😉

benjamin-hodgson:
> Would something like this work? Or does it have to be a direct sibling?

Your example should work fine, as long as both the button and the `target` are under the same `s-modal` controller. You just can't put two different modals inside the same controller - two different modals need two controllers. Makes sense?

dist/js/stacks.js
benjamin-hodgson: Can do

benjamin-hodgson:
Re explicit ID: makes sense, but it's not very idiomatic Stimulus. Stimulus wants you to use `target`s to make elements talk to each other

aaronshekey:
> Re explicit ID: makes sense, but it's not very idiomatic Stimulus. Stimulus wants you to use `target`s to make elements talk to each other

True, and what would we gain from using explicit IDs? Anything?

benjamin-hodgson:
I suppose there's a risk of styles cascading into the modal.

dist/js/stacks.js
benjamin-hodgson: 703a726

bnickel:
I've given some thought to how I would use s-modal with Stacks and I feel we should be going a very different route.

I currently have two methods for showing modals, StackExchange.helpers.showModal and StackExchange.helpers.loadModal.  They've been used in the following ways:

- I click a button on a page, it load HTML containing a modal.  I show it with the button still on the page.
- I click a button in a popover, it loads HTML possibly containing a modal.  If the modal exists, I show it, but... at this point the popover may be gone or exiting the screen.
- I click a button on a page and immediately show a modal that exists on the page.
- I click a button on a page, then I reset the contents of a modal to its initial values and display it. (In case it was reshown multiple times and values were edited.)


Inside the modal, several things can happen:
- Clicking outside, "Cancel", or "X (Dismiss)" hides the modal with no callback.
- Clicking a button inside the modal hides the modal and executes a callback.
- Clicking a button inside the modal triggers a request that could either hide the modal on completion or show an error toast in the modal.

Beyond that, hiding may be permanent, with the element getting fully discarded, or temporary, with the controller getting reshown after a subsequent interaction.

To me this speaks implies two requirements:

1. Showing a modal is something that should happen outside of the modal controller itself.  A separate controller or non-Stimulus code should be able to take a DOM element that may not yet live on the page and say "Show this modal".
2. Any modal controller we create will be incomplete as things not contained in the code will want to hide or remove the element on their timeline.  This could be done through composition (an element having multiple controllers that pass messages to eachother) or inheritance (Stacks.ModalController base class).

I personally would like to see a base class that could be inherited:

```typescript
class ModalController extends Controller {
	
	public static show(el: Element) {
		if (!el IS IN DOM) { APPEND TO DOM }
		el.dispatchEvent(new Event('s-modal:_show', {bubbles: false}));
	}

	public initialize() {
		this.element.addEventListener('s-modal:_show', () => this.show())
	}

	private show(e: Event) {
		...
	}

	public hide(e: Event) {
		...
	}

	...
}
```

Then I could have a controller that looks like:

```typescript
Stacks.addController('some-form-modal', class extends ModalController {
	public submit() {
		if (await submitForm()) {
			hide();
		} else {
			showError();
		}
	}

	...
});
```

The HTML looks like:

```html
<div class="s-modal" aria-hidden="true" data-controller="some-form-modal">...</div>
```

Then the code to show the dialog doesn't need to look much more complicated than this:

```javascript
function loadFormModal() {
	var html = await loadFormHTML();
	var $el = $(html);
	Stacks.ModalController.show($el.get(0));
}
```

bnickel:
Also, StackExchange.helpers.showModal has the following accessibility considerations I'd like to see adopted in some form:

1. The element with the ".js-first-tabbable" class cancels any <kbd>Shift</kbd><kbd>Tab</kbd> events it receives. (Typically the modal's title with tabindex="0" added.)
2. The element with the ".js-last-tabbable" class cancels any unshifted <kbd>Tab</kbd> events it receives. (Typically the "X" button)
3. The element with the "js-modal-initial-focus" class gets initial focus when the modal finishes appearing. (An appropriate button like an element in a radio list or a button.)
4. showModal takes an optional parameter: `{returnElements: $els}`.  When the dialog is dismissed, we focus the first element in that list that exists in the DOM.
5. Esc dismisses the dialog as covered in this design.

This provides sufficient support for keyboard users and some (but not all) screen readers.  Keyboard users enter and exit the dialog as it is shown and hidden and cannot move keyboard focus out of the dialog.  It works well in macOS's VoiceOver but cannot prevent focus changes in iOS's VoiceOver.  If we want to apply aria-hidden and tabindex=-1 to a page element we'll need to test it very thoroughly and make sure its always a sibling of and not a parent of the modal.

b-kelly:
Added clean implementation based on lessons learned from Popovers in PR #415. Recommending to close this one.

PR #289, opened 2019-04-02
Convert to TypeScript

benjamin-hodgson:
Might be a little bit tricky to review because of the file renames.

For the Stacks library:

* Set up `tsconfig.json` in `lib`
* Changed `Stacks` api to be more idiomatic TS:
    * Added `StacksController` base class
    * ~`Stacks.addController` now takes a `class`, not a prototype. This API is much closer to what Stimulus gives you out of the box, so folks following the Stimulus docs should have an easier time.~ I undid this. TS consumers can use `Stacks.application.register` and ES5 consumers can continue to use `addController`
    * ~These are breaking changes, so will affect the two other PRs currently in flight~
* Converted `controllers` to TS.
    * Convert `addController` calls to match the new API
    * Added type annotations
    * Stimulus auto-creates properties for each target in the static `targets` array. `fooTarget`, `fooTargets`, `hasFooTarget`. TypeScript doesn't know about these; you need to declare them. (Given that Stimulus is written in TypeScript I find it odd that it's not easier to consume.)
    * TypeScript caught a few existing type errors, which I fixed variously with assertions and `instanceof`/`null` checks
    * Added `stimulus.d.ts` to work around https://github.com/stimulusjs/stimulus/issues/238. (These two lines were by far the most time consuming task, haha.)
* Left `stacks.polyfills.js` as it is.
* Set up TS build:
    * added dev dependencies on `typescript` and `grunt-ts`
    * I had the compiler output intermediate JS files into a (new) `build` directory. The files are then fed into `concat`.

For the js on the docs site:
* I had to change `controllers/docs-resizer.js` because it was using the `Stacks.addController` api, which changed.
    * It didn't actually need any Stacks-specific code, so rather than make the TS compiler import Stacks it was simpler to just convert it to a regular Stimulus file.
* Set up a second `tsconfig`. I configured it to spit out a single `controllers.js` file.

To discuss:
* I don't really like that the build output is checked in to the source repo.
* If the docs JS does need to depend on Stacks in the future, we'll need to set up [project references](https://www.typescriptlang.org/docs/handbook/project-references.html).
* With the current setup, TypeScript's shims like `__extends` get spat out into each JS file, meaning there are duplicates in the concatted output. It'll work fine, but makes the file bigger than necessary. We should consider using TS's built-in concatenation.
* We should consider shipping Stacks as a UMD module.
    * TS's UMD output isn't actually 100% conforming (doesn't generate a UMD global) so we'd probably need to use Babel


aaronshekey:
Let's get some opinions from @bnickel and @aalear on this as well.

bnickel:
Overall, I love this.  I had been wanting to convert a controller to TypeScript a while back and found it impossible with the Stacks API.

That said, this is a breaking change for Stack Overflow since it uses the `Stacks.addController` public API as it exists today.  I think we should continue to allow both:

```typescript
export function addController(name: string, dataOrController: object | Stimulus.ControllerConstructor) {
    ...
    var Controller = typeof dataOrController === 'function' ? dataOrController : createController(name, data);
    application.register(name, Controller);
    ...
}

File: lib/ts/stacks.ts [hunk 1]
```diff
@@ -0,0 +1,48 @@
...
+                event.initCustomEvent(namespacedName, true, true, detail!);
+            }
+            (optionalElement || this.element).dispatchEvent(event);
+        };
+    }
+    
+    
+    type ControllersDictionary = { [identifier: string]: Stimulus.ControllerConstructor };
+    const _controllers : ControllersDictionary = {};
+    export const controllers : Readonly<ControllersDictionary> = _controllers;
```
benjamin-hodgson: @bnickel Do you know if `Stacks.controllers` is used? Suggest dropping, it's available as `Stacks.application.controllers` anyway

benjamin-hodgson:
@bnickel I propose a slightly different API. Keep `addController` as it was (for use by ES5-style controllers); TypeScript controllers can use `Stacks.application.register` directly. I can move the `hasPrefix` checking into a subclass of `Application`. What do you think?

benjamin-hodgson:
Alternatively, convert the ES5 things in Stack Overflow to do something like this instead: http://bdavidxyz.com/blog/stimulusjs-with-javascript/

lib/ts/stacks.ts
bnickel: I think we can remove it.  Not sure what it was originally included for but I don't see any references.

bnickel:
Moving the hasPrefix stuff to an Application subclass sounds reasonable.

aaronshekey:
> I don't really like that the build output is checked in to the source repo.

Neither do I, kill it.

> We should consider shipping Stacks as a UMD module.

I don't think this is a blocker, but I would love to play nicely in the larger ecosystem in the same way that we publish to NPM, even though we don't actually use Stacks as an actual NPM dependency anywhere.

aaronshekey:
When having a look at this locally after running `npm start` my branch is dirtied with:

`dist/js/stacks.js`
`dist/js/stacks.min.js`
`docs/assets/js/controllers.js.map`

lib/ts/stacks.ts
benjamin-hodgson: 21b3b56

benjamin-hodgson:
@bnickel 585e240 - undid the breaking change to `addController`

bnickel:
Is there any progress on this?  Turns out, I wish I had it today.

aaronshekey:
We were just discussing this in Slack this morning. This is low priority for @b-kelly, but the JS has firmed up enough that you or @benjamin-hodgson could take this over and get it across the finish line.

aaronshekey:
We're now planning on wrapping this work up by the end o' July.

bnickel:
Once this is ready, what's the best way to test this with the Stack Overflow codebase?

aalear:
@bnickel Grab `stacks.min.js` from the branch and replace your local version with it.

b-kelly:
Ready for re-review. Needed very little work from what was there before. I've manually checked every change to the migrated `.ts` files to make sure they were up to date with the `.js` files. In addition, I've migrated over `s-popover` too. Please ignore the `// @ts-ignore`s. I tried like crazy to do it "the right way", but gave up in the end due to the way Popper.js does its imports. This will be fixed / removed when we migrate to something that respects `import` instead of concat'ing the files.

NickCraver:
> With the current setup, TypeScript's shims like __extends get spat out into each JS file, meaning there are duplicates in the concatted output. It'll work fine, but makes the file bigger than necessary. We should consider using TS's built-in concatenation.

Do we have a net delta here on size? Given how many people we ship this to, we may need to address that before shipping to SO.

lib/ts/controllers/s-popover.ts
```diff
@@ -86,69 +104,69 @@
...
         /**
          * Forces the popover to hide if a user clicks outside of it or its reference element
          * @param {Event} e - The document click event
          */
-        _hideOnOutsideClick: function(e) {
+        _hideOnOutsideClick(e: MouseEvent) {
             // check if the document was clicked inside either the reference element or the popover itself
             // note: .contains also returns true if the node itself matches the target element
-            if (!this.referenceElement.contains(e.target) && !this.popoverElement.contains(e.target)) {
+            if (!this.referenceElement.contains(<Node>e.target) && !this.popoverElement.contains(<Node>e.target)) {
```
deanward81: Can we assign `<Node>e.target` to a const above? Easier to read

lib/ts/controllers/s-expandable-control.ts
```diff
@@ -141,17 +156,17 @@
...
             }
-        },
+        };
 
-        disconnect: function () {
-            this.events.forEach(function (e) {
-                this.element.removeEventListener(e, this.listener);
+        disconnect() {
+            this.events.forEach(e => {
+                this.element.removeEventListener(e, this.listener as EventListener);
```
deanward81: Just use a cast here, won’t it always be an `EventListener`?

lib/ts/controllers/s-expandable-control.ts
```diff
@@ -34,92 +32,109 @@
...
             this._dispatchShowHideEvent(!newCollapsed);
             this._toggleClass(!newCollapsed);
-        },
+        };
 
-        connect: function () {
-            this.events.forEach(function (e) {
-                this.element.addEventListener(e, this.listener);
+        connect() {
+            this.events.forEach(e => {
```
deanward81: Just use a cast here, won’t it always be an `EventListener`?

lib/ts/controllers/s-expandable-control.ts
```diff
@@ -34,92 +32,109 @@
...
 
-        listener: function(e) {
+        listener(e: UIEvent) {
             var newCollapsed;
             if (this.isCheckable) {
-                newCollapsed = !this.element.checked;
+                newCollapsed = !(<HTMLInputElement>this.element).checked;
             } else {
-                if (e.type == "keydown" && (e.keyCode != 13 && e.keyCode != 32)) {
+                if (e.type == "keydown" && ((<KeyboardEvent>e).keyCode != 13 && (<KeyboardEvent>e).keyCode != 32)) {
```
deanward81: Could use a type guard here?

lib/ts/controllers/s-expandable-control.ts
benjamin-hodgson: `x as Foo` and `<Foo>x` mean exactly the same thing in TS (type erasure)

lib/ts/controllers/s-expandable-control.ts
benjamin-hodgson: as above

lib/ts/controllers/s-expandable-control.ts
b-kelly: Updated `listener(e: Event)` to match the `EventListener` interface so a cast is no longer needed.

lib/ts/controllers/s-expandable-control.ts
b-kelly: Addressed as above

lib/ts/controllers/s-expandable-control.ts
b-kelly: Added `e instanceof KeyboardEvent &&` here as a type guard (as defined by the [TS docs](https://www.typescriptlang.org/docs/handbook/advanced-types.html#instanceof-type-guards)). Let me know if this was not what you were intending.

b-kelly:
Tested locally in Core. All components work as intended. Manually tested `popovers`, `table`, `expandable-control` in both the Stacks docs and where they appear in Core.

As far as size goes, I have not tested it minified, but the unminified bundle comes out to 1k less than before (gzipped).

NickCraver:
@b-kelly if those are the numbers of combined we're likely better and can't be much worse :+1: - LGTM

In case anyone doesn't have it - a quick way to test gzip is this VSCode extension - handy to enable for these scenarios: https://marketplace.visualstudio.com/items?itemName=logerfo.gzip-status

aaronshekey:
Ready to merge this one?

bnickel:
> Me: _strolls in late to the conversation_

Shouldn't there be a stacks.d.ts file in `dist/js` containing both the definitions for Stacks and Stimulus?  I'm not sure how I'd inherit the controller as a consumer.

b-kelly:
@bnickel I've added the `.d.ts` files. Please let me know if this is what you're looking for.

b-kelly:
> Ready to merge this one?

I'd like some testing on the Talent side of things before merge. I don't have any experience with that project, so someone else will need to take over on that.

bnickel:
@b-kelly I'd like to also see the Stimulus definitions in that the directory since stacks.js is also delivering Stimulus.  Otherwise, we can't register controllers.

bnickel:
I did an initial test with a few controllers I'd written in TypeScript, stubbing out definitions for the things I didn't have handy.  Seems to be working as described.  Once we have all the definitions, I think this is good to ship.

lib/ts/controllers/s-popover.ts
```diff
@@ -1,68 +1,86 @@
+
+
 (function () {
     "use strict";
-    Stacks.addController("s-popover", {
-        targets: [],
-        
+    Stacks.application.register("s-popover", class extends Stacks.StacksController {
+        static targets: [];
```
```suggestion
        static targets = [];
```

bnickel: Or nothing, since this isn't doing anything.

b-kelly:
> I did an initial test with a few controllers I'd written in TypeScript, stubbing out definitions for the things I didn't have handy. Seems to be working as described. Once we have all the definitions, I think this is good to ship.

Unfortunately this is all I get when turning declarations on for Typescript. This may need to be held off on until migrating to webpack / another less hacky build system (in progress).

Feel free to mess around with the config and commit any changes if you can get it to work.

lib/ts/controllers/s-popover.ts
b-kelly: Good eye! Suggestion committed.

b-kelly:
Got word that Talent has been tested and approved, so I'm looking to merge this in today. Any objections? Speak now or forever hold your peace!

@bnickel Regarding shipping Stimulus typings, I thought about it and came to the conclusion that it isn't Stacks' responsibility to include the types of another package. The new declarations file that I get when I use proper TS bundling (coming soon) actually adds `import * as Stimulus from 'stimulus'` at the top, so that should suffice if you're consuming the declarations correctly anyways.

bnickel:
@b-kelly I have to assume we're _not_ consuming declarations correctly.  Anyone consuming stacks.js is getting Stimulus from that file, not from a proper import system, so if they plan to use Stimulus from TypeScript, everything they need has to live in the dist/js folder.

I'm trying to see if Stimulus has a packaged definition somewhere, but I'm not having any luck yet.

b-kelly:
@bnickel This is a very fair point. I'm not sure what is common in this situation. You might be able to pull the typings from [here](https://github.com/stimulusjs/stimulus/blob/master/packages/stimulus/index.d.ts), but that really doesn't help us much. I'll look into best practices for shipping another package's typings with your own, but we're going to punt this issue for later.

bnickel:
@b-kelly It's common as far as Stack Overflow goes. 😉 The core product consumes stacks.js and has a bunch of individually built typescript files relying on bespoke definition files and no packager in sight.  We will be unable to use these enhancements without either Stacks providing the definitions or us finding a way to pull the definitions from somewhere else.  I'm trying to build the definitions by running tsc at https://github.com/stimulusjs/stimulus/tree/master/packages/%40stimulus/core, but I'm struggling to get my system up to date first.

PR #291, opened 2019-04-05
Block links

### LINKED ISSUE #268: `.s-sidebarwidget--item` but for `.s-popover`
**Is your feature request related to a problem? Please describe.**

I'm trying to implement the following with Stacks:

screen shot 2019-02-26 at 3 29 20 pm

I could some one-off atomic stuff here, but:

1. This is a pattern that'll probably get reused a lot.
2. It really looks a lot live like if `.s-popover` and `.s-sidebarwidget` had a baby.

**Describe the solution you'd like**

The same design patterns as `.s-sidebarwidget--item` but in a `.s-popover`.  A formal definition for horizontal separators would be nice too.

**Describe alternatives you've considered**

One-off code, but I'm not sure how I'd handle the theme-specific border-left-color.

aaronshekey:
This PR explores adding block links as their own component. These show up on Stack Overflow's left nav, and in the sidebar widgets.

image

This allows us to add a selected border to each direction, but perhaps we only allow for choosing left and right? Not sure about the practicality of top and bottom.

This closes #268

### DIFF
--- a/docs/product/components/sidebar-widgets.html
+++ b/docs/product/components/sidebar-widgets.html
@@ -381,7 +381,7 @@ description: Sidebar widgets are flexible containers that provide a lot of patte
                 </div>
                 <ul class="s-sidebarwidget--content s-sidebarwidget__items">
                     <li class="s-sidebarwidget--item"><a href="#">Edit Profile</a></li>
-                    <li class="s-sidebarwidget--item" aria-current="page"><a href="#">Developer Story Preferences</a></li>
+                    <li class="s-sidebarwidget--item s-block-link s-block-link__left is-active" aria-current="page"><a href="#">Developer Story Preferences</a></li>
                     <li class="s-sidebarwidget--item"><a href="#">Job Match Preferences</a></li>
                 </ul>
                 <div class="s-sidebarwidget--header">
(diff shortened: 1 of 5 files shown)


TedGoas:
This is cool but I almost wonder if we shouldn't provide all four options? Can we think of a use case for having a border on the bottom?

aaronshekey:
> Can we think of a use case for having a border on the bottom?

Not really, maybe an alternative to `s-tabs`, but I wanted to stub it out anyway. I was leaning toward deleting these as well.

docs/product/components/sidebar-widgets.html
```diff
@@ -381,7 +381,7 @@
                 </div>
                 <ul class="s-sidebarwidget--content s-sidebarwidget__items">
                     <li class="s-sidebarwidget--item"><a href="#">Edit Profile</a></li>
-                    <li class="s-sidebarwidget--item" aria-current="page"><a href="#">Developer Story Preferences</a></li>
+                    <li class="s-sidebarwidget--item s-block-link s-block-link__left is-active" aria-current="page"><a href="#">Developer Story Preferences</a></li>
```
bnickel: Is this meant to be here? It's incomplete / not working as expected.

Screen Shot 2019-04-05 at 1 10 28 PM

aaronshekey: Good catch. I was exploring refactoring those to use `s-block-link` but will most likely punt on that.

TedGoas:
What about admin / settings navigation?

Screen Shot 2019-04-05 at 4 20 01 PM

aaronshekey:
> What about admin / settings navigation?

This could be handled by `s-block-link` without any modifying class.

PR #292, opened 2019-04-09
Content: Error Messages

TedGoas:
# Problem:
We write a lot of error messages but don’t always put enough thought into their copy's utility and structure. We have our content principles and guidelines, but they're broad strokes that apply to all product copy.

We often have to write error messages while we’re in the midst of design or code. Sometimes we don't have ample headspace to think about error messages, and sometimes we don't have time. Our existing content guides are helpful, but sometimes you don't want to read pages of documentation just to write a single error message. Wouldn’t it be nice if we had something focused on the language and structure error messages so we could write good ones quicker and get back to design + code?

This PR attempts to solve provide that resource.

---

### 👋 Helped Wanted

Feedback welcome on everything, but I’m most interested in these areas:

1. More examples for each section. Right now there’s only one, but it’d be 💯 to have 5 or more.
2. I tried to define use cases for the various ways we have to display error messages under "Design, placement, and timing.” Please check my logic here.


donnachoi:
This looks great @TedGoas! Some thoughts:

* The questions that help you frame the error under "Understand the error" are useful. 
* +1 to "short, concise error messages."
* +1 to using active voice to accept responsbility. 
* The examples under System error and User error are great. I personally feel like while additional examples are nice, the two that you have illustrate the points quite well.
* The example "This file is broken..." - this to me feels too vague to be a good error message? I can't imagine where we would use a message like that but maybe I lack context. What about something more specific like, "We can't accept this file type. Upload a PNG or JPG file."
* Serious vs. non-serious error messages: I'm a bit leery of allowing non-serious error messages, or at least without a bit more thought on the guidance and where a lighter tone might fall flat. In the past I feel like our attempts at "fun" in UI copy haven't been very well received and are too easy to misinterpret. I would either remove the "non-serious" part for now or spend more time maybe defining what the right-type of non-serious  message is.

aaronshekey:
A quick note while I'm thinking of it. We're going to want to redirect the old URLs to the new locations if you're moving stuff. These 301s can be found in [netlify.toml](https://github.com/StackExchange/Stacks/blob/develop/netlify.toml) :v:

TedGoas:
@aaronshekey @donnachoi @phoebe-leung Thanks for your help y'all. Everyone's feedback is addressed and I think this is ready to merge.

PR #299, opened 2019-04-16
Email gallery

TedGoas:
**Problem:**
Every time we want to do a big email announcement, folks default to the Promotional Template and mold their content to fit it. Ideally, we'd write whatever we want and design around that, so I’d love a way to show folks that sometimes bespoke design is appropriate.

**I'd love to get this out soon** and add to it over time.


TedGoas:
Meeting notes:
Linking to a static image show **what was designed** but makes it difficult to see **how it was achieved**. So let's include the actual live code. We could grab the code from our ESPs and host it here, but leave the images hosted on our ESP's CDN.

We'd want to add the following note:

> The source for these emails is provided for historical purposes only. It's ok to grab **parts** of a design, but these templates should not be used as a starting point because the code likely outdated.

aaronshekey:
This is _great_ work. I'd rather have the live code, if outdated, than screenshots. These are pretty inspiring to see in a row. Really makes me want to add a fourth. 😛

One last little thing, let's replace the variables on these with real content.

PR #301, opened 2019-04-19
Dead link fix on getting-started.html

hteumeuleu:
The URL for the "Best practice" page seems to have changed, but the getting started page still links to the old URL.


TedGoas:
Oh, thanks for catching this Rémi!

PR #319, opened 2019-05-03
Add popover support via Popper.js

b-kelly:


aaronshekey:
I think we're much better off relying on [Popper.js](https://popper.js.org/) as a dependency to manage our popover positioning. Things from the spec this needs:

- [ ] Clicking outside the popover should close it
- [ ] Clicking the originating button again should close the popover.
- [ ] Ideally, we'd be able to specify the direction we'd prefer on initialization, allowing popper.js to manage the rest of the states if they change

b-kelly:
@aaronshekey Addressed all items on your checklist. I also added a basic hover example too (stimulus gives us this for free). Docs / wording might need some tweaking, but this is now feature complete.

aaronshekey:
I've popped this into an `s-btn-group` and was happy with the results.

image

Should we consider adding `is-selected` to the originating element?

aaronshekey:
Do we care about the `ERROR `/assets/js/popper.js.map' not found.` showing up in the logs? Something we can clean up there?

b-kelly:
> Do we care about the `ERROR `/assets/js/popper.js.map' not found.` showing up in the logs? Something we can clean up there?

Short answer, no we don't care about this. Not sure why exactly this is showing up though, but I'd be it has to do with the way we're concat'ing the file in. I'll take a look on Monday.

b-kelly:
> I've popped this into an `s-btn-group` and was happy with the results.
> 
> Should we consider adding `is-selected` to the originating element?

Looks excellent! Do you mean programatically adding the class on? Should be really easy to add in if so. I don't think I'd add any styles though, just let other components style it themselves maybe

aaronshekey:
> Do you mean programatically adding the class on?

Yeah, I think that we should probably wire up also toggling `is-selected` to the originating element, since there will likely be a lot of cases where we'll want to do that.

aaronshekey:
This one's ready for some review 🎉

aaronshekey:
@b-kelly There is some prior art with providing an optional list of classes on the [`.s-expandable`](https://stackoverflow.design/product/components/expandable). It accepts an arbitrary class list via `data-s-expandable-control-toggle-class="is-selected"`.

We could do the same thing with `data-s-popover-toggle-class`.

aaronshekey:
Had a look at IE11, and this isn't working yet. Chrome, Firefox, Safari, and Edge are all looking great!

docs/product/components/popovers.html
```diff
@@ -40,23 +26,122 @@
...
+            <p>There’s no data associated with your account yet. Please check back tomorrow.</p>
+        </div>
+    </div>
+
+    {% header h3 | Automatic %}
+    <p class="stacks-copy">If you’ve added the proper Stimulus controllers to your popover, positioning and arrow direction will be managed for you by <a href="https://popper.js.org">Popper.js</a>, a powerful popover positioning library we’ve added as a dependency.</p>
+
+    <div class="stacks-preview _preview-none">
+{% highlight html linenos %}
+<div class="s-popover"
```
bnickel: Popovers should appear immediately after buttons in the DOM so we can tab into them when open.

docs/product/components/popovers.html
```diff
@@ -40,23 +26,122 @@
...
+        </button>
+    </div>
+    <div id="popover-example-1" class="s-popover ws2" role="menu" aria-hidden="true">
+        <div class="s-popover--arrow"></div>
+        <div class="grid ai-center jc-space-between">
+            <div class="grid--cell tt-uppercase fs-fine fc-light">
+                Saved filters
+            </div>
+            <div class="grid--cell">
+                {% icon PlusSm | va-baseline %}
```
bnickel: This should be a button so we can test tabbing

lib/js/controllers/s-popover.js
```diff
@@ -0,0 +1,94 @@
...
+            this.popoverElement = document.getElementById(popoverId);
+
+            this.popper = new Popper(this.element, this.popoverElement, {
+                placement: this.data.get("placement") || "bottom",
+            });
+
+            // in order for removeEventListener to remove the right event, this bound function needs a constant reference
+            this._boundClickFn = this._hideOnOutsideClick.bind(this);
+
+            document.addEventListener("click", this._boundClickFn);
```
bnickel: balpha had concerns about having event listeners bound to every popover on the page, even if it's not active.

See https://gh.stackoverflow.com/Core/StackOverflow/blob/bbb139746376b76a8ac531ce1b37434e5178b220/StackOverflow/Content/Js/PartialJS/full-common/controllers/se-popover.js#L145 for how I'm binding and unbinding events between connect and toggle.

lib/js/controllers/s-popover.js
```diff
@@ -0,0 +1,94 @@
...
+        },
+
+        /**
+         * Forces the popover to hide if a user clicks outside of it or its reference element
+         * @param {Event} e - The document click event
+         */
+        _hideOnOutsideClick: function(e) {
+            // check if the document was clicked inside either the reference element or the popover itself
+            // note: .contains also returns true if the node itself matches the target element
+            if (!this.element.contains(e.target) && !this.popoverElement.contains(e.target)) {
```
bnickel: There are some nuanced cases to account for:

- What if I clicked a button in the popover and that button was removed e.target would have no parent?
- What if the popover displayed a modal and we click in the modal?

I have the following logic prevent popovers from hiding on a click event:

    var possibleChildSelector = '.s-modal:not([aria-hidden=false]), .s-toast:not([aria-hidden=false])';

        documentClick: function (e) {
            var target = e.target;
            if (!$.contains(document.documentElement, target)
                || !this.isVisible
                || this.$popover.has(target).length > 0 || this.$popover.is(target)
                || this.$source.has(target).length > 0 || this.$source.is(target)
                || $(target).parents(possibleChildSelector).length > 0 || $(target).is(possibleChildSelector)) {
                return;
            }

            this._toggle();
        },

lib/js/controllers/s-popover.js
```diff
@@ -0,0 +1,94 @@
...
+        _toggle: function(show) {
+            this.popoverElement.classList.toggle("is-visible", show);
+            this._toggleOptionalClasses(show);
+        },
+
+        /**
+         * Forces the popover to hide if a user clicks outside of it or its reference element
+         * @param {Event} e - The document click event
+         */
+        _hideOnOutsideClick: function(e) {
```
bnickel: We should also handle `ESC` keys, returning focus to the source element if focus was within the popover.  Here's my code for that:

    var possibleChildSelector = '.s-modal:not([aria-hidden=false]), .s-toast:not([aria-hidden=false])';

        documentKeyup: function (e) {
            var target = e.target;
            if (e.which === 27 && this.isVisible && $(possibleChildSelector).length === 0) {
                if (this.$popover.has(target).length > 0 || this.$popover.is(target)) { this.$source.focus(); }
                this._toggle();
            }
        },

bnickel:
Also, this is not a new issue but one we could hopefully solve here.

When the source element shifts above or into the top-bar space, the popover floats unexpectedly over it.  We have this on SO with CQL.  It would be cool if this controller detected that collision and responded by hiding the popover.

Screen Shot 2019-05-13 at 3 23 11 PM

aaronshekey:
> It would be cool if this controller detected that collision and responded by hiding the popover.

I think the easiest thing would be to make sure that our navigation is actually a higher z-index than the popover. Right now our scale assumes that we'd want a popover to be the top-most element. I don't think this assumption is true, and I don't think we have to be too clever with how we handle this.

lib/js/controllers/s-popover.js
abovedave: As an aside, have we considered touch events here?

aaronshekey: Ah, yes. Good catch. Touching outside on a mobile device doesn't work as expected.

bnickel: Touch events are tricky since they don't always mean what you think they will.  Most touch events are scrolling.  Similarly, if you make a change too early with a touch event, the click event will still fire in that exact same position, possibly on an entirely different element.

The specific problem we're experiencing here is that mobile browsers only fire `click` events if you touch an element with a `click` event, don't drag, then release focus.  I believe `click` also fires after `touchup`.  If you touch whitespace, there's no `click` event the event can't bubble up to `document`.  Touching a link or button on the other hand does work to hide the popover.

One solution _could_ be to move the dismiss event to the `body` element:

```javascript
document.querySelector('body').addEventListener("click", this._boundClickFn);
```

But I wouldn't do that. I would just do nothing. This kind of behavior is standard for mobile users and they are more than capable of clicking the button again.

aaronshekey:
> Make sure that our navigation is actually a higher z-index than the popover

Done in https://github.com/StackExchange/Stacks/pull/319/commits/b45842930c8928d24b631a4ab260e652a596b3d1

aaronshekey:
@b-kelly I've summarized the feedback into a todo list:

- [x] Fix z-index issues
- [ ] IE11
- [ ] Arrow placement edgecases
- [ ] Explore a way of wiring up the `esc` key

lib/js/controllers/s-popover.js
b-kelly: @bnickel brings up some excellent points on this. I'm punting on touch events for now because they're a pain. We can revisit later if we feel strongly about it.

b-kelly:
> I included a few minor things I'd like to see, some accessibility and edge cases I've seen while working with popovers. That said, there's a big issue in that Popper isn't doing the one thing i really want it to do, responsive arrow placement:

Looking at the demos, popper seems to explicitly support this. I'll take a look into this and see what I'm missing.

PR #324, opened 2019-05-06
Docs for using web fonts in email

### LINKED ISSUE #316: Document web-fonts in email
We've used web-fonts in email enough times that we should document how we do it.

* How and where to link an external font stylesheet.
* How to reference it the email body.
* How to plan a fallback for email clients that don't support web-fonts.
* Other considerations (which fonts we use, what typographic elements we typically use them for, etc.)

Past example using Roboto Slab as live text:

Screen Shot 2019-05-03 at 11 59 05 AM

TedGoas:
We've used web-fonts in email enough times, so I documented it.

1. Does this make sense?
2. Is this too detailed? Not detailed enough?
2. Is the wording clumsy?

---

This resolves #316

### DIFF
--- a/docs/email/base/typography.html
+++ b/docs/email/base/typography.html
@@ -212,7 +212,7 @@ console.log("request body: " + pm.globals.get("switch_json"));</code></pre>
             font-family: arial, sans-serif !important;
         }
     </style>
-    <!--<![endif]-->
+    <![endif]-->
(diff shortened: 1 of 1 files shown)


TedGoas:
✅ Brandon said this makes sense (in Slack), so I'll merge this.

Issue #317, opened 2019-05-03
`.s-btn__icon` too low on IE11


bnickel:
**Describe the bug**
Icons in `.s-btn__icon` are not vertically centered on IE11.  This problem gets worse the smaller the icon is.  The problem appears to be that IE11 does not support `vertical-align: initial` and the so the icons are falling back to the rule `vertical-align: bottom`

**To Reproduce**
1. Go to https://stackoverflow.design/product/components/buttons#icons on IE11
2. Look at the icons.

**Expected behavior**
Icons should be more-or-less vertically centered in the button.

**Actual behavior**
Icons are well anchored to the bottom padding of the button.

**Screenshots**

![Screen Shot 2019-05-03 at 9 45 01 AM](https://user-images.githubusercontent.com/662837/57152618-dbac6900-6d88-11e9-8716-6831c12f58ff.png)

![Screen Shot 2019-05-03 at 9 48 27 AM](https://user-images.githubusercontent.com/662837/57152623-de0ec300-6d88-11e9-96f7-5f81a7dd6cce.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Internet Explorer
 - Version 11.437.17763.0

PR #325, opened 2019-05-06
Additional button normalization

### LINKED ISSUE #300: Propose adding `display: inline-block` to the `s-btn` base class.
From [a discussion in Slack](https://stackexchange.slack.com/archives/CBK9R8D38/p1555524973318800):

When styling an `a` as an `s-btn` it's easy to forget to manually add `d-inline-block`. I would generally expect `s-btn` to style my element exactly like a `button`. I don't think I've ever deliberately styled a button with `display: inline`. If one day I do need it I can override it manually with `d-inline`.

aaronshekey:
This PR fixes #300 and #317 😎


aaronshekey:
I've run our backstop visual regression tests and adjusted a few bits in our documentation to compensate for things moving around.

PR #334, opened 2019-05-22
Remove position: relative on form elements

### LINKED ISSUE #333: Selecting text in an unfocused `s-input` fails in Firefox
**Describe the bug**
This issue comes from MSE but is reproducible in the Stacks documentation.  On Firefox (latest), a text area or input needs to be focused in order for you to select the text contained within.  E.g.

    <input class="s-input" value="hello world">

If you want to select "world" you can't just drag it.  You have to first click so it gets into a focus state, then highlight it.

Original bug with screenshots: https://meta.stackexchange.com/questions/328370/cant-select-from-a-text-field-without-focusing-the-text-field-first

The problem appears to be the following:

```less
.s-input:focus, .s-input:focus-within {
    position: relative;
}
```

Removing `s-input`, removing that rule or adding `ps-relative` to the element causes it to work as expected.  You can see this in Search on the Stacks docs (`style="position: relative"`) or on the Stack Overflow homepage hero (non-Stacks inputs)

**To Reproduce**
1. Go to https://stackoverflow.design/product/components/inputs#base-style
2. Enter text in "Full name"
3. Blur the text field by clicking somewhere else.
4. Try selecting the text you just wrote with your mouse in a single mousedown and drag.

**Expected behavior**
Text is selected

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows, MacOS, Linux
 - Browser: Firefox
 - Version 60, 66

aaronshekey:
Fixes #333 

I'm not sure why we included `position: relative` on our form elements, but this PR removes it to fix #333. I'll have to look back at the blame for some context there.

I need to do a bit of testing on IE and Edge for this.

### DIFF
--- a/lib/css/components/_stacks-inputs.less
+++ b/lib/css/components/_stacks-inputs.less
@@ -338,7 +338,6 @@ fieldset {
 //  Focus
 & {
     @focus-style: {
-        position: relative;
         border-color: @blue-300;
         box-shadow: 0 0 0 @su4 fade(@blue-500, 15%);
         outline: 0;


aaronshekey:
A little bit of context for this one: https://github.com/StackExchange/Stacks/commit/ac6f53034c4cf4d178920ec473128ed8dc802b01

In my testing in the docs, this doesn't seem to be necessary anymore 🤔

PR #348, opened 2019-06-27
Popover custom background color support

b-kelly:
Added support for inheriting popover arrow color from `fc-*` classes.

Reasoning for using `fc-*` classes instead of something more obvious:

> The way the popover is coded, the arrow uses a well known trick with borders to create a triangle (excellent browser support). Therefore, the arrow color is controlled by `border-color`.
>
> Stacks only supports so many border colors (like 3 per color. `bg-blue-400` corresponds to `bc-blue-2`). Rather than force Stacks to add a billion more border colors to bring it up to parity (or force devs to restrict themselves to what is available), I wanted to see if I could inherit from `bg-*`.
>
> Unfortunately, you can't inherit from other properties in CSS without having basically declared them first or using some CSS variables scheme (which isn't well supported by IE anyhow). What _is_ supported however is `currentColor`, which lets any color bearing property inherit from the `color` property.
>
> **So in conclusion**, we set `color` with `fc-*` to allow the border color to inherit this value via `currentColor`. Now we can allow the arrow to take advantage of the wide range of `fc-*` classes to match the popover's `bg-*.

### DIFF
--- a/docs/product/components/popovers.html
+++ b/docs/product/components/popovers.html
@@ -235,4 +235,23 @@ description: Popovers are small content containers that provide a contextual ove
             </div>
         </div>
     </div>
+
+    {% header h3 | background-color %}
+    <p class="stacks-copy">If you want to set the background color to anything other than the default, simply add your desired <code class="stacks-code">bg-*</code> class to the popover itself and a matching <code class="stacks-code">fc-*</code> class to the arrow element.</p>
+
+    <div class="stacks-preview">
+{% highlight html linenos %}
+<div class="s-popover bg-blue-400">
+    <div class="s-popover--arrow__lb fc-blue-400"></div>
+    …
+</div>
+{% endhighlight %}
+        <div class="stacks-preview--example bg-black-025">
+            <div class="s-popover ps-relative z-base bg-blue-400 is-visible">
+                <div class="s-popover--arrow__lb fc-blue-400"></div>
+                <p class="ff-mono mb8">.s-popover--arrow__lb</p>
+                <p class="mb0">Popover arrow appears left bottom</p>
+            </div>
+        </div>
+    </div>
 </section>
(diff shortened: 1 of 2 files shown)


bnickel:
More of a meta-question (not Meta question), but why does `.s-popover` use the border triangle hack instead of `transform: rotate(45deg)`?  I ask because on Stack Overflow, I originally implemented the New Feature Notice as arrow hack and an intrepid designer switched it to `rotate(45deg)`, which I then reused for the tag popover.

If viable, for `.s-popover`, the rotate solution has the advantages of supporting both custom background and border colors via the more natural `.bg-` and `.bc-` styles.

cc/ @aaronshekey 

Relevant SO commit: 8cb8004490245df8059005359968dd99129dbfc8

```less
    .-arrow {
        position: absolute;
        display: none;
        left: 50%;

        &:before {
            position: absolute;
            content: '';
            background: @white;
            border: 1px solid @black-200;
            border-bottom: none;
            border-right: none;
            top: -6px;
            left: -6px;
            width: @su12;
            height: @su12;
        }

        &.-top {
            display: block;
            transform: rotate(45deg);
            top: -1px;
        }

        &.-bottom {
            display: block;
            transform: rotate(225deg);
            bottom: -1px;
        }
    }
```

b-kelly:
@bnickel I wasn't a part of the original implementation, but if I had to guess the answer is probably just that is what the designer was familiar with and the solution is battle tested and well supported. In other words, no specific strong reason.

I'm definitely ok with migrating to a tilted box format (which allows us to add box shadows and the like too), but after soliciting feedback on our internal Slack, the answer was "do whatever is easiest, this won't come up often and the current solution is good enough".

If anyone feels strongly about this though, I'm happy to put in the effort to get it to a point where everyone is happy with the resulting implementation.

PR #352, opened 2019-07-11
Popover event support

b-kelly:
Add support for show/shown and hide/hidden events


b-kelly:
I don't think an event example is necessary since anyone who cares about them will know how to use them. But if we did want one, it could be something like this:

```html
<button id="example-button" data-controller="s-popover">...</button>

<script>
document.querySelector("#example-button").addEventListener("s-popover:shown", e => console.log("popover shown"));
</script>
```

Issue #297, opened 2019-04-12
Dark Theme for Stackoverflow


therenca:
Just a humble request, could we have a dark theme, for stackoverflow, some of us are afraid of the white, thanks

aaronshekey:
A dark theme isn't out of the question, but we're not ready to consider it seriously.

I think it can be solved in a way that allows us to swap all our color variables on the fly based on 1) user preference 2) OS-setting. But, organizationally, Stacks doesn't have the code coverage to enable this. Huge swathes of the product would be bright white even if dark mode is enabled.

It also introduces another another axis to design against:

1. Desktop
2. Responsive
3. Dark mode

I can very easily see us painting certain layouts into a corner by forgetting to test dark mode, especially when considering that we have 200 some themes to also theme.

bnickel:
A huge pre-requisite for this would probably be getting Stack Overflow completely onto Stacks for color, or at least to a point where the custom colors we do use are in a single less file.  Right now, there are thousands of matches for `color: #` in our codebase, not just in .css and .less files but also in .cshtml and .cs files. 😬

I would love to see Stacks support a dark theme just largely as an exercise to see what it takes and I also `prefers-color-scheme: dark`, but that its probably low on their priorities considering the high cost of auditing and transforming the whole Stack Overflow codebase.

samliew:
@aaronshekey This has been [brought up again](https://meta.stackoverflow.com/q/386150/584192) on Stack Overflow Meta, and I have quoted your comment.

aaronshekey:
I've got some thoughts about _how_ we'd design this.

We could possibly flip the colors by default when a user requests dark mode, getting us 60% there, and then offer dark mode-specific overrides when the default gets it wrong. Every color should have a best guess dark mode equivalent that is enabled by `prefers-color-scheme: dark` and a class on the body / html that's toggled by user preference or some combination of both.

Let's say you've got `fc-light` on `bg-white`. By default it could probably switch to `bg-black-900` and `fc-black-050`. Ultimately your markup might end up looking like `fc-light bg-white dark:bg-black-600 dark:fc-black-200` to accommodate situational overrides.

This would allow us to account for situations like what's illustrated in Steve Schoger's [really great tweet](https://twitter.com/steveschoger/status/1151160261170126850?s=20).

gbjbaanb:
I'm not afraid fo the white, but when I'm coding in the dark, with all other sites in dark modes... I open SO and it hurts. 

[...]

PR #356, opened 2019-07-26
Technical exploration of dark mode

aaronshekey:
This PR explores some of the ideas discussed in #297.

A note for Stack Overflow users: This does _not_ mean we're shipping dark mode. This PR is an exploration only.

Screen Shot 2020-01-13 at 9 22 19 PM

- [x] Generate light and dark-specific color scales
- [x] Implement each color as a native CSS variable
- [x] Refactor buttons to rely on native CSS variables
- [x] Build dark mode conditional atomic override classes
- [x] Figure out how to handle focus states
- [x] Figure out how to handle drop shadows
- [x] Fix notice code-coloring
- [x] Fix button group badge markup
- [x] Fix links component
- [x] Style modals appropriately
- [x] Fix badge progress bar presentation in dark mode
- [x] Style sidebar widgets appropriately
- [x] Style tags appropriately
- [x] Replace documentation border colors with more neutral treatment
- [x] Delete Less `@variable` color variables
- [x] Document dark mode override classes
- [x] Figure out production variables strategy
- [x] Figure out why some icons look bad in dark mode
- [x] Update colors documentation to be completely accurate
- [x] Make the scrollbar mixin darkmode-aware.


the-j0k3r:
May I ask why is it necessary to apply !important to all these rules?

lib/css/exports/_stacks-constants-colors.less
```diff
@@ -106,44 +106,50 @@
...
-@red-500: #d1383d;
+@red-500: @red;
 @red-600: #c02d2e;
 @red-700: #ac2726;
 @red-800: #942121;
 @red-900: #7a1819;
 
 //  $  BADGES
 //  ----------------------------------------------------------------------------
-@badge-gold:    #ffcc01;
```
b-kelly: This is a breaking change. Double check instances of this before updating any clients

lib/css/components/_stacks-inputs.less
```diff
@@ -69,6 +70,9 @@
     //  --  REMOVE EDGE ADDITIONS
     //  ------------------------------------------------------------------------
     &::-ms-clear { display: none; }
+
+    // --  STYLE SCROLLBARS
+    @scrollbar-styles();
```
b-kelly: This is potentially a breaking change too. Should see how this looks on scrollable inputs before merge.

lib/css/exports/_stacks-constants-colors.less
```diff
@@ -250,20 +256,20 @@ body {
...
-    --gold-bg:       @gold-bg;
-    --gold-border:   @gold-border;
-    --silver-dot:    @silver-dot;
-    --silver-bg:     @silver-bg;
-    --silver-border: @silver-border;
-    --bronze-dot:    @bronze-dot;
-    --bronze-bg:     @bronze-bg;
-    --bronze-border: @bronze-border;
+    // Gold
+    --gold: @gold;
```
b-kelly: Potential breaking change, but I doubt anyone is using our css variables in any capacity, much less these badge variables.

lib/css/exports/_stacks-constants-colors.less
```diff
@@ -276,27 +282,150 @@ body {
...
-@brc-lightest:  @black-050;
-@brc-lighter:   @black-075;
-@brc-light:     @black-100;
-@brc-dark:      @black-200;
-@brc-darker:    @black-300;
-@brc-darkest:   @black-350;
-@brc-orange:    @orange-400;
+//  --  Dark mode
+.dark-colors() {
+    --white: #2d2d2d;
```
b-kelly: Looks like we're not really caring to expose these colors as variables (eg `@dark-black-025`)? I think overall this is the right choice, since we want our users to ultimately consume these colors as `var(--black-025)` anyways and exposing as less variables would be time consuming and make migration harder in the future.

lib/css/components/_stacks-inputs.less
aaronshekey: Checked :v:

lib/css/exports/_stacks-constants-colors.less
aaronshekey: I'll make sure this rename goes into the production PR.

docs/_data/product/configurations.yml
```diff
@@ -93,15 +93,6 @@
     - variable: "@bs-lg"
       default: "0 @su4 @su12 fade(@black-800, 20%)"
       description:
-    - variable: "@bs-i-sm"
```
aaronshekey: Removing these is also a breaking change, but these classes aren't used anywhere in production.

PR #362, opened 2019-08-12
Migrate from Grunt to Webpack

b-kelly:
Exploration of migrating to webpack and proper JS bundling.

## Why?

The older process simply concat'd files together and hoped they compiled. Adding third party code was a pain.

The new process properly imports/bundles code, runs it through the TS compiler and runs it through Babel. On the less side, we now have PostCss in the compile pipeline for minification. We now have a lot of flexibility in what we can include in the future. This can potentially include `babel-polyfills` to only include the polyfills we need for only the browsers we support (browserlist support).

## Expectations

There's nothing inherently wrong with what we are using today. This is an exploration of potentially cleaning up the code a tad and adding support for proper bundling / publishing to make coding easier for devs and to make it easier for projects consume Stacks in the near future.

## Some changes to development / publishing:

* Need to run `npm run init-docs` once to prep the docs site
  * installs bundler/jekyll and preps the icons from npm instead of committing them to the repo
  * can just run `npm run update-icons` if you already have the ruby dependencies already installed
* The dist folder now bundles all files in the folder root instead of JS/Css folders
  * This includes all TS declaration files
* All published code is minified by default. No more `.min` files

## What needs tested:

- [x] Running tests. Did not test the `test` or `approve` scripts
- [x] Dropping the files into a prod env to test — `stacks/test-webpack`
- [x] Size deltas
- [ ] Tweaking package.json entry points so projects can automatically / easily import both js and less


aaronshekey:
Some naive size comparisons using Finder's preview as a benchmark. Seems our CSS gets smaller while JS gets larger.

`stacks.min.css` to `stacks.css` 214kb vs. 196kb
`stacks.min.js` to `stacks.js` 63kb vs. 68kb
`stacks.polyfills.min.js` to `stacks.polyfills.js` 24kb vs. 30kb

I think we should be delivering `.min` on these filenames regardless of how they're generated. I would expect, as a consumer of our library, that `stacks.js` is an unminified file. Changing the file name will also break our current CDN deliveries.

aaronshekey:
Tracking all of the `dist` folder is tricky since it'll be back to us having a bunch of merge conflicts. I really hate it. I think it's worth considering untracking the `dist` folder entirely and just relying on `npm` to deliver to unpkg, Core, and Careers.

b-kelly:
This is old and crusty. Closing in favor of #707

PR #363, opened 2019-08-13
Support for inactive table rows

### LINKED ISSUE #361: Inactive class for table rows
# Problem

We currently have a class to visually mark inactive table rows. (Eg. for users who have been deactivated). Problem is that our CSS is scoped to `.managed-users-table`. That's a bad classname and one that comes with a bunch of _other_ styles that we probably don't want in `s-table`.

Screen Shot 2019-08-09 at 3 07 25 PM

---

# Proposal

[This code](https://gh.stackoverflow.com/Core/StackOverflow/blob/master/StackOverflow/Content/Shared/Channels/_channels-admin.less#L85) seems to work pretty well.

```
tr.inactive {
        background-color: @black-025;

        td:not(.some-exception-class) {
            opacity: 0.3;
        }
    }
```
It visually marks the contents of the row as inactive but allows us to omit certain content from appearing inactive (Like say, a link to _reactivate_ the row).

Let's get a Stacks class for this!

TedGoas:
This PR adds proper CSS that visually marks the contents of table row as disabled/inactive while allowing us to omit certain content from appearing inactive (Like say, a link to reactivate the row).

Screen Shot 2019-08-13 at 5 03 17 PM

PR does not include any JS recommendation atm.

---

This PR resolves #361

### DIFF
--- a/lib/css/components/_stacks-tables.less
+++ b/lib/css/components/_stacks-tables.less
@@ -304,6 +304,17 @@
         padding-left: 0;
         width: 120px;
     }
+
+    //  $$  Disabled rows
+    //  ------------------------------------------------------------------------
+    tr.is-disabled {
+        background-color: @black-025;
+
+        th:not(.never-disabled),
+        td:not(.never-disabled) {
+            opacity: 0.3;
+        }
+    }
 }
 
 //  ============================================================================
--- a/docs/product/components/tables.html
+++ b/docs/product/components/tables.html
@@ -975,3 +975,67 @@ description: Tables are used to list all information from a data set. The base s
         </div>
     </div>
 </section>
+<section class="stacks-section">
+    {% header h2 | Disabled rows %}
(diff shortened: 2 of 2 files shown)


lib/css/components/_stacks-tables.less
```diff
@@ -304,6 +304,17 @@
         padding-left: 0;
         width: 120px;
     }
+
+    //  $$  Disabled rows
+    //  ------------------------------------------------------------------------
+    tr.is-disabled {
+        background-color: @black-025;
+
+        th:not(.never-disabled),
```
TedGoas: @aaronshekey I'm not happy about defining `<th>` and `<td>` separately. `:not` and `*:not` didn't work. However can `<tr>`'s have any other children other than `<th>` and `<td>`? I don't think so. I also can't think of another element that we'd want to apply this class to (eg. specific elements _within_ a table cell? Nah).

Issue #368, opened 2019-08-19
Label Badges: New, Beta, and make required less prominent


mstum:
The [Status badges for Labels](https://stackoverflow.design/product/components/labels) are Required and Optional. However, on a form with several fields, I find the "Required" badge a bit too prominent and propose to dial it back a bit.

Also, I'd like to add "New" and "Beta" labels. Will add a PR.

![image](https://user-images.githubusercontent.com/104127/63300508-61a3df80-c2a6-11e9-82b3-a3a62fe9eaf2.png)

phoebe-leung:
I like the idea of a dialed back label, but the white text on red-200 background makes it a bit hard to read. I would suggest something like what Atlassian is doing for their "subtle" tags - a red-600 text on a lighter, red-200 background.

Aaron gets final say when he's back though :) 

![Screen Shot 2019-08-20 at 10 44 25 AM](https://user-images.githubusercontent.com/47794687/63357639-bba0b580-c337-11e9-93d4-8927a0d7140e.png)

mstum:
Definitely not set on the colors, feel free to change them however you see fit.

mstum:
Closing as it's merged

PR #369, opened 2019-08-19
Label badges: Beta/New, make Required less prominent

mstum:
See #368


mstum:
Which branch is the correct in-progress branch? `master`?

As for the Required label, yeah, I think that Ted had the same comment, don't show the label until it's missing. But usually, we already make the form field red and display an error message anyway, so that might be double effort. Anyway, the idea is still to make it look a little bit less in-your-face :)

aaronshekey:
> Which branch is the correct in-progress branch?

We use `develop` for our main in-progress branch. We use `master` for release. If you clone the repo, branch from develop and submit a pull request from there. Now that you've got write access, there's no need to work with forks. :v:

PR #379, opened 2019-08-30
Label badges: Beta/New, make Required less prominent 

mstum:
See #368, replaces #369


mstum:
I like those changes, wondering if the Optional badge should be bold to match the others?

Issue #382, opened 2019-09-05
Input font size causes zoom in on mobile devices


aaronshekey:
Since our input / textarea font size is smaller than 16px, when a user interacts with them on mobile, the whole page zooms in.

We should find ways to mitigate this, either making the text size `>= 16px`, or finding a workaround.

aaronshekey:
Here's a good illustration of this issue:

![lame_zoom](https://user-images.githubusercontent.com/1369864/64380964-c1fd8580-cff7-11e9-9e8a-814d52a863ad.gif)

brettinternet:
Here is a [related post in meta](https://meta.stackoverflow.com/questions/389464/form-elements-should-be-at-least-16px-font-size-to-prevent-mobile-browsers-from).

aaronshekey:
Fixed in #383 🤙

PR #383, opened 2019-09-05
Explore input zooming workarounds

aaronshekey:
This PR explores fixes for #382


aaronshekey:
Using a metadata-only approach fixes the iOS use case, but disables Android's ability to pinch and zoom. Not ideal. Exploring further.

aaronshekey:
Simply boosting the inputs to `16px` makes everything look super chunky in every context.
 
image

bnickel:
I’m exploring something on the Ask a Question refresh project. When there’s a mouse down event I disable zooming and then reenable it a moment later. I’m nots sure how Android works so I’m currently just doing it on iOS user agents.

bnickel:
This approach looks good on the iPhone I tested on.  Honestly, the increased font size _feels_ good on an phone.

My only notes are 1) we'll want to redo the whole `.s-input__sm` etc sizes to be relative to 16 in this case, and 2) the inputs feel very tall and we may want to adjust down the vertical padding.

aaronshekey:
> the inputs feel very tall and we may want to adjust down the vertical padding

Good call! I'll do my best to adjust the padding to compensate for the larger font size.

aaronshekey:
> We'll want to redo the whole `.s-input__sm` etc sizes to be relative to 16 in this case

I've explored this and found it to be very tricky considering we'll likely need to combine a responsive media query with the mobile-specific `@supports`. I can tolerate the sizing overrides not being perfect since that's an edge case that comes up way less frequently than input zooming.

aaronshekey:
Tested on:

Windows: Internet Explorer 11, Edge, Chrome, Firefox
Mac: Safari, Chrome, Firefox
iOS 13: Safari, Chrome

PR #384, opened 2019-09-05
Patterns/Page Titles

TedGoas:
Our page titles in the some of our admin sections is inconsistent (screenshot below). 

- Some have breadcrumbs, some don’t
- Page titles are different text sizes / colors
- Description text is inconsistently designed
- Some have a bottom border (inconsistently colored), some don’t
- Some page titles don’t go full width
- One page doesn’t even have a page title 😐

Page titles are a small, but visible detail in our site. We shouldn’t have to think about their visual display. Can we standardize the design so we don’t have to?

**This PR standardizes the page titles used in various admin interfaces while giving us a few options to customize.**

---
current state of affairs:
Screen Shot 2019-09-05 at 4 47 53 PM

---

### To-Do:

- [x] Base styles
- [x] Examples
- [x] Responsive
- [x] Breadcrumbs
- [ ] Expand outside of admin
- [x] Decent copy


TedGoas:
This is far from done, but want to put this on your radar and have y'all start poking at this early. **I want to ship a small version of this soon** (Let's expand outside of admin screens in another PR).

lib/css/patterns/_stacks-page-titles.less
```diff
@@ -0,0 +1,35 @@
...
+
+.s-page-title {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    align-items: flex-end;
+    width: 100%;
+    padding-bottom: @su16;
+    border-bottom: 1px solid @black-100;
+    margin-bottom: @su16;
```
TedGoas: Is adding a bottom margin overly prescriptive here? I can't think of a time when we'd want *no* border... and it would be nice if the space between the page title and content was consistent on every page.

aaronshekey:
I don't think that a `Pattern` should have a corresponding class. In my opinion, a `Pattern` is a recipe of atomic class ingredients. This is currently being spec'd like a component. It shouldn't have _any_ CSS at all.

For example, a common pattern would be a three column grid. Our pattern would only include a page documenting

```
<div class="grid grid__allcells4">
    <div class="grid--cell">
```

It wouldn't be wrapped into a thing called `s-three-column-grid`.

TedGoas:
I'm confused. My goal here is to standardize (and lock down) certain parts of this pattern so that all instances of it are consistent. In the same way that buttons are defined with an `.s-btn` class and not a bunch of utility classes for border, background, and text color. We don't want folks rolling their own buttons.

I'm concerned that exposing all the classes for a pattern like would invite folks to make tiny tweaks that introduce inconsistency and leave us in the place that's not much better than where we are now.

Are you thinking of patterns more like "our recommendation for how to build a pattern?"

aaronshekey:
> My goal here is to standardize (and lock down) certain parts of this pattern so that all instances of it are consistent. In the same way that buttons are defined with an `.s-btn` class

Yep, Buttons are a component. Patterns are recipes. This should be a component called Title or something. :v:

TedGoas:
Ahh, I see. So components are locked down, whereas patterns are "our suggestion on how to build something but ultimately you can change the ingredients if you want."

If that's the case, I think this should be scoped as a component so that titles can't be bolded, spacing adjusted, borders removed, and other inconsistencies that we suffer from today.

I'll update!

docs/product/components/page-titles.html
```diff
@@ -96,7 +96,7 @@ <h1>Page title</h1>
                     <h1>Page title</h1>
                     <p>Optional description de Finibus Bonorum et Malorum with an <a href="#">optional link</a>.</p>
                 </div>
-                <div>
+                <div class="md:mt8">
```
TedGoas: @aaronshekey I don't love having part of the responsiveness in markup via this `.md:` class and part in the LESS file. I will try playing with padding on the parent `<div>`s but am open to other ideas.

TedGoas:
@aaronshekey @phoebe-leung @paintedbicycle This is ready for a more formal review. Please critique design, markup, accessibility, and documentation (words).

docs/product/components/page-titles.html
aaronshekey: This is where I get hazy on these _needing_ to be a component. This can just as easily be mocked up as a pattern using atomic classes for full flexibility. There are lots of contexts for this, and may lead to fighting an overly-prescribed component.

We also need to account for the overall Questions, Tags, Users page headers yet—and if we don't include those, we'll likely need to explain why.

docs/product/components/page-titles.html
```diff
@@ -0,0 +1,148 @@
...
+    </div>
+    {% header h3 | Title with description %}
+    <p class="stacks-copy">
+        A summary of a page’s content. Helpful for large pages with multiple actions.
+    </p>
+    <div class="stacks-preview">
+{% highlight html %}
+<div class="s-page-title">
+    <div>
+        <h1>Page title</h1>
```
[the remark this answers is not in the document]
TedGoas: A few options, all of which work and don't work on some level:

`<div class="s-page-title--title"></div>`
`<div class="s-page-title--actions"></div>`
This might work since the stuff on the right is almost always buttons or filters (but it would limit us to that).

`<div class="s-page-title--left"></div>`
`<div class="s-page-title--right"></div>`
😐

`<div class="s-page-title--primary"></div>`
`<div class="s-page-title--secondary"></div>`
This might work since the stuff on the right is optional, but is really vague and could be confusing if you only see the first div.

tscommand-4938160e.tmp.txt
```diff
@@ -0,0 +1 @@
+docs/assets/js/controllers/docs-resizer.ts docs/assets/js/controllers/stimulus.d.ts --sourcemap --declaration --removeComments --lib dom,es6,dom.iterable,scripthost --typeRoots "docs/assets/node_modules/@types,docs/assets/node_modules/@stimulus/core" --strict --target ES5 --out docs/assets/js/controllers.js
```
aaronshekey: We don't need this

TedGoas:
We decided to ship this now and add other instances later.

PR #388, opened 2019-09-16
Toggle popover display instead of visibility

aaronshekey:
As of #386, popovers are forcing an unnecessary horizontal scroll. This PR attempts to switch to toggling `display` instead of `visibility`.

This will kill our ability to easily transition between the hidden and `is-visible` state. It also introduces some intermittent scrollbars on Safari at small window sizes 🤔 

image


aaronshekey:
I've tested this locally in Internet Explorer 11, Edge, Firefox, Chrome, and Safari.

PR #398, opened 2019-09-27
Content/alt text

### LINKED ISSUE #395: Content: ALT text
Sometimes our designs include images. All images should have an `alt` attribute, but what goes in there is important for folks using assistive technologies or plain text preferences.

It would be helpful to have a framework for us to write consistent, helpful `alt` text. There are a few unofficial, undocumented best practices floating around, let's organize and document them.

TedGoas:
Sometimes our designs include images. All images should have an `alt` attribute, but what goes in there is important for folks using assistive technologies or plain text preferences.

This PR starts codifying guidelines on to write consistent, helpful `alt` text. This resolves #395


docs/content/examples/alt-text.html
```diff
@@ -0,0 +1,152 @@
+---
+layout: page
+title: Alt-text
+description: Everything we know about alt-texts, when to use them, and how to craft them.
```
paintedbicycle: Plural feels a little weird to me, though I'm sure it's correct, I've basically never seen plural used for alt-text. Feel free to ignore this nit.

docs/content/examples/alt-text.html
```diff
@@ -0,0 +1,152 @@
...
+---
+<section class="stacks-section">
+    {% header h2 | How to write alt-text %}
+    <p class="stacks-copy">
+        An alt-text is a description of an image that’s shown to people who can’t see the image. Alt-texts help people with little or no vision who use assistive technologies, people who have turned off images, and search engines.
+    </p>
+    {% header h3 | Describe the image concisely %}
+    <p class="stacks-copy">
+        It might sound obvious, but an alt-text should describe the image in case in image doesn’t display or someone has trouble seeing it. The goal of alt-text is to give the necessary information from the image <strong class="ws-nowrap">at a glance</strong>. It’s best to include only the necessary information.
```
paintedbicycle: "in case in" should likely be "in case an"

docs/content/examples/alt-text.html
```diff
@@ -0,0 +1,152 @@
...
+                </div>
+                <div class="grid--cell">
+                    <h2 class="grid--cell fs-body1 fw-bold fw-normal fc-dark mb4">TOP LEADERBOARD: 728X90</h2>
+                    <p class="grid--cell fs-body1 fc-medium m0">This ad unit is the most visible on the site.</p>
+                </div>
+            </div>
+        </div>
+    </div>
+    <p class="stacks-copy">
+        The nearby text here already explains what the graphic illustrates.
```
paintedbicycle: We can maybe add why. 

For example, changing "graphic illustrates." to "graphic illustrates. If there was alt-text in this scenario, screen readers, for example, would read nearly duplicate information to the user."

docs/content/examples/alt-text.html
```diff
@@ -0,0 +1,152 @@
...
+    </table>
+    {% header h3 | End with a period %}
+    <p class="stacks-copy">
+        End the alt-text with a period. This makes screen readers pause a bit after the last word in the alt-text, creating a natural pause before the next bit of text.
+    </p>
+</section>
+<section class="stacks-section">
+    {% header h2 | When not to write one %}
+    <p class="stacks-copy">
+        In most cases you should use an alt-text for images, but there are some exceptions where you should leave the alt-text blank.
```
paintedbicycle: Could add a bit more emphases on empty/blank vs non-existent.

For example, "leave the alt-text blank." could change to "leave the alt-text blank. Note that this does not mean omitting the alt attribute, it means leaving it empty"

docs/content/examples/alt-text.html
```diff
@@ -0,0 +1,152 @@
...
+            </tr>
+        </tbody>
+    </table>
+    {% header h3 | End with a period %}
+    <p class="stacks-copy">
+        End the alt-text with a period. This makes screen readers pause a bit after the last word in the alt-text, creating a natural pause before the next bit of text.
+    </p>
+</section>
+<section class="stacks-section">
+    {% header h2 | When not to write one %}
```
paintedbicycle: "When not to write one" doesn't seem to match the tone in the other headings.

Maybe "When not to include alt-text"?

docs/content/examples/alt-text.html
TedGoas: Holy moly, I forgot to include the entire section explaining this. Good catch!

paintedbicycle:
LGTM

PR #401, opened 2019-09-27
A more future-proof targeting of Safari in spinners

### LINKED ISSUE #290: -apple-trailing-word may not be a reliable way to detect WebKit in future
_stacks-spinner.less has:

@supports (-apple-trailing-word: auto) {
            box-shadow: 0 0 0 2px transparent;
}

WebKit is planning to remove support for -apple-trailing-word: https://bugs.webkit.org/show_bug.cgi?id=196525

so this will break at some point.

aaronshekey:
Closes #290


aaronshekey:
Tested on:

Windows: Chrome, Firefox, Edge, Internet Explorer 11
Mac: Safari, Firefox, Chrome

PR #410, opened 2019-10-30
Add "s-navigation__muted" secondary style for the Navigation component

juanrodriguezcen:
Stacks only support 1 style for the Navigation component, and trying to add 2 of these components on the same page can look confusing for the user. 

This PR introduces a "Muted" variation of the component that can be used as a secondary navigation on a page that already has a primary one. 

This is how it looks:

image

In order to use it, the class `s-navigation__muted` should be added to the parent element, using the same pattern that the muted variation for buttons use (`s-btn s-btn__muted`)

### DIFF
--- a/lib/css/components/_stacks-navigation.less
+++ b/lib/css/components/_stacks-navigation.less
@@ -8,9 +8,46 @@
 //
 //  TABLE OF CONTENTS
 //  • BASE STYLE
+//  • MUTED STYLE
+
+
+& { // create a single scope to load the config into, at least for now
+    #stacks-internals #load-config();
+
+//      MIXIN:
+//      Navigation item
 //
+//      HOW IT WORKS:
+//      Defines the s-navigation--item with the colors to be used. 
+//      The default values are the ones used by the base style.
+//      Style modifications can pass different colors as needed.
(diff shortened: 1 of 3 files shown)


lib/css/components/_stacks-navigation.less
```diff
@@ -8,9 +8,46 @@
...
+
+//      MIXIN:
+//      Navigation item
 //
+//      HOW IT WORKS:
+//      Defines the s-navigation--item with the colors to be used. 
+//      The default values are the ones used by the base style.
+//      Style modifications can pass different colors as needed.
+//  ----------------------------------------------------------------------------
+.s-navigation--item(@fc: @black-600, @hover-bg: @black-100, @hover-fc: @black-600, @selected-bg: @navigation-link-highlight, @selected-fc: @white){
```
aaronshekey: I don't think we need to be this clever. We can simply add the overrides with

```
&.s-navigation__muted .s-navigation--item {
    &:hover {
        // style
    }
}
```

This way, you aren't duplicating the entirety of the component's CSS, only the overrides... Unless I'm not understanding how Less compiles in this context. Every byte counts.

PR #414, opened 2019-11-03
Add CSS grid classes

aaronshekey:
This PR adds atomic CSS grid classes. Previous exploration #255. Page title will simply change to `grid` when #413 drops.

This is now ready for review. Do these class names make sense? What are we missing?


TedGoas:
This is a great start for the CSS classes. There's always room for more options. @paintedbicycle brought up a few good ones like:

1. Using stuff like `auto-fit`.
2. Using `fr` units. How do I say my first column spans 2 and then the second column just takes up "the rest". Or col 1 is 33%, col 2 is 66% (i.e. `1fr` `2fr`).

This is just the start, we'll get there.

---

I'd love to see us start layering in some **guidance** into the documentation sooner rather than later. Especially since we're not moving away from `Flexbox`.

So many *Flexbox vs CSS Grid* articles say "use both, they each have their own strengths!" Can we start to illustrate some Stack Overflow specific examples of when you'd reach for `CSS Grid` vs `Flexbox`? This would help cut down on support requests from someone trying to accomplish something using `CSS Grid` where `Flexbox` would have been more appropriate (and vice versa). A good example is bullet No.2 above.

Also if we start thinking about this now, it might always prevent us from building in something in `CSS Grid` that we don't need because it's already perfectly handled by `Flexbox`.

paintedbicycle:
Good point on the "this is just a start", @TedGoas - so maybe grid gets gets broken into more than one PR? This first one could go through with just the base classes to replace our current grid capabilities? And then a second one to add the more flexible stuff? 

Let me know what you think. cc/ @aaronshekey

aaronshekey:
No need to split this PR up, in my opinion. It's not _that_ complicated. I plan on designing the API, and then documenting classes first, _then_ writing about why and how you'd use CSS grid. I'll definitely add those 2 bullet points here :v: Great suggestions!

paintedbicycle:
The use of 
```
grid-column: x; 
grid-row: y
```
is nice for responsive. You can have the HTML built in a certain order, but easily change the order using these to something else at mobile.

First instance putting the right sidebar first on mobile in certain situations.

paintedbicycle:
It might be possible to add `auto-fit` using classes like:

`column-fit-100`
`column-fit-200`

which would output:

 `grid-template-columns: repeat(auto-fit, minmax(100px, 1fr));` and `grid-template-columns: repeat(auto-fit, minmax(200px, 1fr));`

Since you only need to declare that once, on the row (or entire grid) and then the browser will create columns out of the children. No matter how many children you give it, it will handle it using the min size, growing to fill the space.

Example: https://codepen.io/paintedbicycle/pen/mKEXaz

paintedbicycle:
Making a 3 or 4 column grid with `grid-template-column: 300px 1fr 400px` is super nice. Unfortunately the `fr` unit doesn't seem possible on the column itself and so creating atomic classes would be semi-impossible as there'd be too many permutations.

Some ideas:
- Create a few common ones? i.e. 2-column, 3 column, 4column layouts that we use a lot? i.e. Left nav, content area, right nav?
- Use CSS variables somehow?

docs/product/base/grid.html
```diff
@@ -0,0 +1,725 @@
...
+                    <div>.grid--row-end4</div>
+                </div>
+                <div class="bg-black-075 p12 ba bc-black-3 grid--col2">.grid--col2</div>
+                <div class="bg-black-075 p12 ba bc-black-3 grid--col2">.grid--col2</div>
+                <div class="bg-black-075 p12 ba bc-black-3 grid--col2">.grid--col2</div>
+                <div class="bg-black-075 p12 ba bc-black-3 grid--col2">.grid--col2</div>
+            </div>
+        </div>
+    </div>
+</section>
```
TedGoas: Additions I'd love to see:

- `justify-items` / `justify-self` / `justify-content`
- `align-items` / `align-self` / `align-content`
- `place-items` / `place-self` / `place-content`

aaronshekey: Skipping place-items since it's just shorthand. Adding `justify-items` and `justify-self`.

docs/product/base/grid.html
```diff
@@ -0,0 +1,808 @@
...
+                                {% icon Checkmark | fc-green-500 %}
+                            {% endif %}
+                        </td>
+                    </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+    </div>
+
+    {% header h3 | Alignment examples %}
```
TedGoas: Thanks for adding these classes. There's a ton of options in the table above and I like how we're trying to visualize some of it. This is helpful and I wish we'd do it on our flexbox page too.

However this example is a bit confusing. We're trying to visualize _some_ of the options above, but I'm not quite sure what's going on with these six boxes and why they're laying out they way they are.

I'm not sure what the answers is... maybe a bunch of tiny examples like on [CSS Tricks](https://css-tricks.com/snippets/css/complete-guide-grid/)? Or maybe no example?

aaronshekey: I like having just a small example so folks have it in their head what these classes do, but I draw a line at explaining every part of the CSS spec. I think this strikes the balance. Lemme know if I'm completely wrong 😛

TedGoas: Yea, I get that. I also don't think we should draw a ton of tiny diagrams. But I'm not sure if this example is helping much, maybe worth killing it and going back to words only?

docs/product/base/grid.html
```diff
@@ -0,0 +1,808 @@
...
+                </div>
+            </div>
+        </div>
+    </div>
+</section>
+
+<section class="stacks-section">
+    {% header h2 | Spacing %}
+    <p class="stacks-copy">The most powerful aspect of CSS grids are its native gutters. To add space between columns and rows add one of the following classes to your grid container.</p>
+    {% header h3 | Spacing classes %}
```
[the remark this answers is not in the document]
aaronshekey: Let's see if they're needed. I think `gsx` and `gsy` might not have been implemented if we'd had proper grid gutter APIs. I also have a hunch most grids will need equidistant spacing. Let's punt on making these independent and see if any requests come up. This'll keep the bundle smaller and the API a bit more simple.

PR #415, opened 2019-11-04
Add s-modal controller

b-kelly:
Adds new `s-modal` controller based on modified popover code. Also adds `Stacks.toggleModal` as a helper so existing legacy code can toggle modals outside of a controller.

Needs:

- [x] Docs pass
- [ ] Core cleanup?


docs/product/components/modals.html
```diff
@@ -3,31 +3,122 @@
...
+                    <td>Helper to manually show an s-modal element via external JS</td>
+                </tr>
+            </tbody>
+        </table>
+    </div>
+</section>
+
 <!-- Example base modal content -->
-<aside class="s-modal js-modal-overlay js-modal-close" id="modal-base" tabindex="-1" role="dialog" aria-labelledby="modal-title" aria-describedby="modal-description" aria-hidden="true">
+<aside class="s-modal js-modal-overlay js-modal-close" id="modal-base" tabindex="-1" role="dialog" aria-labelledby="modal-title" aria-describedby="modal-description" aria-hidden="true"
```
bnickel: Will there ever be a scenario where `data-controller="s-modal"` and `data-target="modal"` won't be the same element?  It probably makes sense to ditch `this.modalTarget` and just use `this.element` everywhere.

docs/product/components/modals.html
```diff
@@ -3,31 +3,122 @@
...
+    data-controller="s-modal" data-target="s-modal.modal">
     <div class="s-modal--dialog js-modal-dialog" role="document">
         <h1 class="s-modal--header" id="modal-title">Example title</h1>
         <p class="s-modal--body" id="modal-description">Nullam ornare lectus vitae lacus sagittis, at sodales leo viverra. Suspendisse nec nulla dignissim elit varius tempus. Cras viverra neque at imperdiet vehicula. Curabitur condimentum id dolor vitae ultrices. Pellentesque scelerisque nunc sit amet leo fringilla bibendum. Etiam feugiat imperdiet mi, eu blandit arcu cursus a. Pellentesque cursus massa id dolor ullamcorper, at condimentum nunc ultrices.</p>
         <div class="grid gs8 gsx s-modal--footer">
             <button class="grid--cell s-btn s-btn__primary js-modal-primary-btn" type="button">Save changes</button>
-            <button class="grid--cell s-btn js-modal-cancel-btn" type="button">Cancel</button>
+            <button class="grid--cell s-btn js-modal-cancel-btn" type="button" data-action="s-modal#toggle">Cancel</button>
```
bnickel: This would be good to remove all the `js-...` classes from this file for clarity.

docs/product/components/modals.html
```diff
@@ -57,30 +148,32 @@ <h1 class="s-modal--header" id="modal-title">Example danger title</h1>
...
     <div class="s-modal--dialog js-modal-dialog" role="document">
         <h1 class="s-modal--header" id="modal-title">…</h1>
         <p class="s-modal--body" id="modal-description">…</p>
         <div class="grid gs8 gsx s-modal--footer">
             <button class="grid--cell s-btn s-btn__primary js-modal-primary-btn" type="button">…</button>
-            <button class="grid--cell s-btn js-modal-cancel-btn" type="button">…</button>
+            <button class="grid--cell s-btn js-modal-cancel-btn" type="button" data-action="s-modal#toggle">…</button>
         </div>
-        <a class="s-modal--close s-btn s-btn__muted js-modal-close" href="#" aria-label="Close">
+        <a class="s-modal--close s-btn s-btn__muted js-modal-close" href="#" aria-label="Close" data-action="s-modal#toggle">
```
bnickel: Not specifically related to this PR, but it would be great if these were `<button type="button">` elements.

docs/product/components/modals.html
b-kelly: I don't have an example in the docs, but you could wrap a shared parent element so you can make a button toggle your modal, similar to how we do popovers (and how the other modal PR worked)

lib/ts/controllers/s-modal.ts
```diff
@@ -0,0 +1,88 @@
...
+        // if we're letting the class toggle, we need to figure out if the popover is visible manually
+        if (typeof toShow === "undefined") {
+            toShow = this.modalTarget.getAttribute("aria-hidden") === "true";
+        }
+
+        // show/hide events trigger before toggling the class
+        this.triggerEvent(toShow ? "show" : "hide");
+
+        this.modalTarget.setAttribute("aria-hidden", toShow ? "false" : "true");
+        this.modalTarget.setAttribute("tab-index", toShow ? "0" : "-1");
```
bnickel: Several things here. `tab-index` should be `tabindex` but... nothing this code does is really what you want it to do:

- `<div class="s-modal" tabindex="-1"><button></button></div>` does not stop the child button from receiving focus on tab.
- `<div class="s-modal" tabindex="0">` adds a useless focus stop on the entire modal element.

Is there some specific guidance somewhere that is telling us to use `tabindex="-1"` to hide the container from screenreaders?  Otherwise, the combination of `visibility:hidden` and `aria-hidden="true"` are sufficient for assistive technology and keyboard users.

lib/ts/controllers/s-modal.ts
```diff
@@ -0,0 +1,88 @@
...
+
+        if (toShow) {
+            this._bindDocumentEvents();
+        }
+        else {
+            this._unbindDocumentEvents();
+        }
+
+        // shown/hidden events trigger after toggling the class
+        this.triggerEvent(toShow ? "shown" : "hidden");
```
bnickel: 99% of my use-cases will involve removing the modal element from the DOM after hiding it, so I'd like the `shown` and `hidden` events to fire _after_ the animation completes so I can discard it appropriately.

Something along the lines of the following could help facilitate that:

```
connect() {
    this.element.addEventListener('transitionend', () => {
        this.triggerEvent(this.element.getAttribute('aria-hidden') === 'false' ? 'shown' : 'hidden');
    });
}
```

lib/ts/controllers/s-modal.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    // TODO should Stimulus bind this? Needed to keep TS error-free
+    private modalTarget!: HTMLElement;
+
+    private _boundClickFn!: any;
+    private _boundKeypressFn!: any;
+
+    connect () { };
+
+    toggle (e: Event) {
```
bnickel: Could we add `show` and `hide` for clarity? `data-action="s-modal#toggle"` feels weird on close buttons.

b-kelly: Added both and changed `toggle` to `hide` on the close buttons to make it more clear

lib/ts/controllers/s-modal.ts
b-kelly: Looks like many browsers don't have full support for this, most notably Chome: https://caniuse.com/#search=transitionend

Bootstrap ([for example](https://github.com/twbs/bootstrap/blob/master/js/src/util/index.js#L92)) deals with this by detecting whether the event was fired or not by setting a timeout (in js) that is exactly as long as the css transition length. I don't have the bandwidth right now to chase this down, but I'd be happy to accept a PR.

PR #418, opened 2019-11-22
Added a new tooltip controller for trivially enabling Stacks tooltips

bnickel:
This introduces a new controller `s-tooltip`, which allows for displaying `s-popover` elements as tooltips with minimal boilerplate.

From the documentation, you _can_ do a full tooltip pretty much the same way you would do a popover controller:

```html
<button class="s-btn" role="button"
        aria-describedby="tooltip-example"
        data-controller="s-tooltip"
        data-s-tooltip-placement="top-start">
    Upvote
</button>
<div class="s-popover s-popover__tooltip"
     id="tooltip-example"
     role="tooltip"
     aria-hidden="true">
    <div class="s-popover--arrow"></div>
    Click here if this answer helped you
</div>
```

Beyond the name, the key differences are:

1. We replace `aria-controls` with `aria-describedby` to support screen readers correctly.
2. We added a new `s-popover__tooltip` class which allows for narrower popovers.
3. We don't have to manually attach the hover events.

This also allows for tooltips and popovers to be placed on the same element, with the behavior that tooltips disappear when a `s-popover` controller is shown, and won't reappear until the `s-popover` controller is hidden.

Beyond the basic example, we allow for a much simpler transformation by dropping the need for a `s-popover` element in the HTML and instead generating it from the `title` attribute on first hover.

So the HTML can instead be:

```html
<button class="s-btn" role="button"
        title="Click here if this answer helped you"
        data-controller="s-tooltip"
        data-s-tooltip-placement="top-start">
    Upvote
</button>
```

This also allows for trivially converting existing elements to Stacks tooltips simply by throwing `data-controller="s-tooltip"` on the element.

There's also a couple scripted functions for adding tooltips to elements:

```javascript
Stacks.setTooltipText(el,
                      "Plain text tooltip",
                      {
                          placement: "top-start",
                          classNames: "ws1"
                      });

Stacks.setTooltipHtml(el,
                      "Tooltip <i>with</i> HTML",
                      {
                          placement: "top-start",
                          classNames: "ws1"
                      });
```


docs/product/components/popovers.html
```diff
@@ -263,42 +347,92 @@ <h3 class="fs-body2 mb2">Teams</h3>
...
+                id="tooltip-example"
+                role="tooltip"
                 aria-hidden="true">
                 <div class="s-popover--arrow"></div>
                 This popover will show content that isn’t intended to be interacted with directly.
             </div>
         </div>
     </div>
+
+    <p class="stacks-copy">In cases where no markup is needed in the tooltip, the popover element can be omitted and automatically generated using the <code class="stacks-code">title</code> attribute.</p>
```
aaronshekey: This is the killer feature. I think we lead with this?

docs/product/components/popovers.html
```diff
@@ -263,42 +347,92 @@ <h3 class="fs-body2 mb2">Teams</h3>
...
+                    data-controller="s-tooltip"
+                    data-s-tooltip-placement="top-start">
                 Hover popover
             </button>
-            <div class="s-popover"
-                id="popover-hover-example"
-                role="menu"
+            <div class="s-popover s-popover__tooltip"
+                id="tooltip-example"
+                role="tooltip"
```
aaronshekey: Let's load this example up with an _obnoxious_ amount of markup to really drive the point home. How about throwing an `s-tag` and an illustration in there?

lib/css/components/_stacks-popovers.less
```diff
@@ -30,6 +28,11 @@
     color: @fc-dark;
     font-size: @fs-body1;
 
+    &:not(.s-popover__tooltip) {
```
aaronshekey: I don't mind being verbose and getting rid of the `:not`. Something like:

```
&.s-popover__tooltip {
    min-width: initial; // or unset or whatever
    width: auto;
}

File: docs/product/components/popovers.html [hunk 4]
```diff
@@ -27,7 +27,7 @@
     </div>
 </section>
 <section class="stacks-section">
-    {% header h2 | JavaScript %}
+    {% header h2 | JavaScript (Interactive) %}
```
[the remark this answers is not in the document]
bnickel: Yeah, how to format all of this was a big question mark for me.  I mostly just wanted to get it all down and then deal with how to structure it.

docs/product/components/popovers.html
```diff
@@ -263,42 +347,92 @@ <h3 class="fs-body2 mb2">Teams</h3>
...
+                      {
+                          placement: "top-start",
+                          classNames: "ws1"
+                      });
+{% endhighlight %}
+        <div class="stacks-preview--example bg-black-025">
+            <button class="s-btn" role="button" id="scripted-tooltip">
+                Hover popover
+            </button>
+            <button class="s-btn" role="button" onclick="Stacks.setTooltipText(document.getElementById('scripted-tooltip'), 'Plain text tooltip', { placement : 'top-start' })">
```
b-kelly: Let's move these `onclick`s to a script tag

lib/ts/controllers/s-tooltip.ts
```diff
@@ -0,0 +1,141 @@
+
+class TooltipController extends Stacks.StacksController {
```
b-kelly: Instead of copy/pasting all the Popover code, can we instead inherit and extend it? Hopefully this isn't too difficult since we're using TS. Will re-review this file after all duped code is removed.

lib/ts/stacks.ts
```diff
@@ -76,7 +76,66 @@ namespace Stacks {
...
+        setTooltip(element, popover => popover.innerHTML = html, options);
+    }
+
+    export function setTooltipText(element: Element, text: string, options?: TooltipOptions) {
+        setTooltip(element, popover => popover.textContent = text, options);
+    }
+
+    var generatedTooltipIdNumber = 0;
+
+    function setTooltip(element: Element, fn: (popover: HTMLElement) => void, options?: TooltipOptions) {
```
b-kelly: Move all this into the tooltip controller. This should be handled by the controller itself

lib/ts/stacks.ts
```diff
@@ -76,7 +76,66 @@ namespace Stacks {
...
         application.register(name, createController(controller));
-    };
+    }
+
+    export interface TooltipOptions {
+        placement: string;
+        classNames: string;
+    }
+
+    export function setTooltipHtml(element: Element, html: string, options?: TooltipOptions) {
```
b-kelly: Having these wrapper functions is good, but let's have them wrap a "native" tooltip function call. I should be able to do all this from within the controller itself and not need outside JS support. See how we did it in the [modal controller](https://github.com/StackExchange/Stacks/pull/415/files#diff-6374cf76f80a96788d1735a9c3bc5efdR88)

lib/ts/stacks.ts
```diff
@@ -76,7 +76,66 @@ namespace Stacks {
...
+    export function setTooltipText(element: Element, text: string, options?: TooltipOptions) {
+        setTooltip(element, popover => popover.textContent = text, options);
+    }
+
+    var generatedTooltipIdNumber = 0;
+
+    function setTooltip(element: Element, fn: (popover: HTMLElement) => void, options?: TooltipOptions) {
+
+        var controllerText = element.getAttribute("data-controller") || "";
+        if (!/(^|\s)s-tooltip(\s|$)/.test(controllerText)) {
```
b-kelly: Is this for supporting space delimited controller strings? Use word boundaries instead:

`/\bs-tooltip\b/`

This will match all the following scenarios:

```
data-controller="s-tooltip"

data-controller="s-popover s-tooltip"

data-controller="s-popover s-tooltip s-something"

data-controller="s-tooltip s-popover"
```

lib/ts/stacks.ts
```diff
@@ -76,7 +76,66 @@ namespace Stacks {
...
+            element.setAttribute("data-controller", controllerText + " s-tooltip");
+        }
+
+        var popoverId = element.getAttribute("aria-describedby");
+        if (!popoverId) {
+            popoverId = "--stacks-s-tooltip-" + (generatedTooltipIdNumber++);
```
b-kelly: Let's stay away from `++`. Be more explicit

lib/ts/stacks.ts
```diff
@@ -76,7 +76,66 @@ namespace Stacks {
...
+
+        var popover = document.getElementById(popoverId);
+        if (!popover) {
+            popover = document.createElement("div");
+            popover.id = popoverId;
+            popover.className = "s-popover s-popover__tooltip";
+            popover.setAttribute("aria-hidden", "true");
+            popover.setAttribute("role", "tooltip");
+            // insertBefore inserts at end if element.nextSibling is null.
+            element.parentNode!.insertBefore(popover, element.nextSibling);
```
b-kelly: `!.` will still throw exceptions and cause the code to crash. Imo, if the parentNode doesn't exist, attach to the document instead.

lib/ts/stacks.ts
```diff
@@ -76,7 +76,66 @@ namespace Stacks {
...
+        }
+
+        const arrows = popover.getElementsByClassName("s-popover--arrow");
```
b-kelly: Nit: You can use `querySelector` to get only the first item and not need to allocate/index into the array

lib/ts/stacks.ts
b-kelly: Please add jsdocs here. I have no idea what the use of `fn` is without looking at calling code.

lib/ts/stacks.ts
```diff
@@ -76,7 +76,66 @@ namespace Stacks {
...
+    export interface TooltipOptions {
+        placement: string;
+        classNames: string;
+    }
+
+    export function setTooltipHtml(element: Element, html: string, options?: TooltipOptions) {
+        setTooltip(element, popover => popover.innerHTML = html, options);
+    }
+
+    export function setTooltipText(element: Element, text: string, options?: TooltipOptions) {
```
b-kelly: All exported functions need js docs added

docs/product/components/popovers.html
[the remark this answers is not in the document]
b-kelly: +1. I originally thought the buttons were broken when clicking

lib/ts/stacks.ts
bnickel: No can do: `/\bs-tooltip\b/.test('s-tooltip-foo') === true`

lib/ts/stacks.ts
bnickel: I can put it on the tooltip class but not use a native instance method because:

1. We may be dealing with an element that doesn't have a controller.
2. Even after we add the controller attribute, we won't have a controller until after Stimulus creates it.  If the element isn't yet in the DOM, it may never happen.

docs/product/components/popovers.html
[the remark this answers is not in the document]
bnickel: I'll explore something like the banner example when I get back on Thursday.

bnickel:
@b-kelly @aaronshekey I've resolved most of the code feedback and can work on the documentation bits next week.  Following Ben's suggestion of moving shared logic into a base controller brought in some improvements that could pay some dividends in the way we handle trickier problems in the future.  We now don't expect the popover to be fully defined until after the first `s-popover:show` event so we could have something novel like:

```javascript
el.addEventListener('s-popover:show', evt => {
    if (!el.hasAttribute('aria-controls')) {
        generatePopover(el);
    }
});
```

This would have helped in the case of share popover where we were generating all the popovers on page launch.

Or we could do something like this:

```javascript
el.addEventListener('s-popover:show', evt => {
    if (!el.hasAttribute('aria-controls')) {
        evt.preventDefault(); // Don't try showing the popover
        fetch('/popover-contents').then(html => {
            el.insertAdjacentHtml('afterend', html);
            el.setAttribute('aria-controls', someId);
            el.triggerEvent(new Event('showAfterLoad'));  // Now show the popover
        });
    }
});
```

This would have helped in the case of the reaction popover, where we fetch the popover HTML on the first interaction.

bnickel:
@aaronshekey @b-kelly Can you look at this again?  I'd like to start using this in production.

aaronshekey:
This may beyond the scope of _this_ PR, but is there a hack to allow dismissal of the hover popovers on mobile? They show on tap, but then can't be dismissed.

bnickel:
@aaronshekey Mobile bugs are definitely in scope!  I've pushed what I hope is a fix that just turns them off for pointer devices that don't hover.  I would argue that not showing tips on mobile is an expected behavior since title attributes function the same way.

docs/product/components/popovers.html
```diff
@@ -262,40 +260,227 @@ <h3 class="fs-body2 mb2">Teams</h3>
...
+            </button>
+        </div>
+    </div>
+
+    {% header h4 | JavaScript configuration %}
+    <p class="stacks-copy">In cases where the tooltip needs to display simple text or HTML, the popover can be configured using JavaScript.</p>
+    <div class="stacks-preview">
+{% highlight javascript %}
+Stacks.setTooltipHtml(el,
+    "Plain text tooltip",
```
aaronshekey: Do we even need this plain text tooltip example since we accept HTML and we killed `setTooltipText`?

b-kelly: Up to you. I left it because I liked that it showcased that it doesn't *need* to be html, but if you think it is redundant, then I'm ok with ditching it.

bnickel:
@b-kelly I brought back `setTooltipText` since it's needed for plain text strings that may contain HTML (think editor examples or possibly user reflected strings).  To support the new model, added a new attribute `data-s-tooltip-html-title` which can store HTML while we wait for the controller to initialize.  We could expose this in the docs as a way for users to expose simple HTML tooltips from HTML/Razor:

```
<button controller="s-tooltip" title="@_s("Simple tooltip")">...</button>

<button controller="s-tooltip" data-s-tooltip-html-title="@_m("***Fancy*** tooltip").ToHtmlString()">...</button>
```

docs/product/components/popovers.html
b-kelly: Resolved. Brian has added explicit separation of html/text content and the docs have been updated to reflect this.

PR #423, opened 2019-12-06
Add CSS Grid

aaronshekey:
This PR adds the CSS grid.

- [x] Add CSS Grid alongside the original flexbox grid without renaming anything
- [x] Rename all flex classes
- [x] Create a compiled shim for original flex grid classes


aaronshekey:
Most of the discussion happened in #414, so I think this is good to merge.

paintedbicycle:
Just a note that the Stack Editor repo will need to be updated with these class changes

docs/product/base/grid.html
```diff
@@ -1,277 +1,779 @@
...
             <tbody>
-                {% for item in grids.grid-cell %}
-                  <tr>
-                      <td><code class="stacks-code">{{ item.class }}</code></td>
-                      <td><code class="stacks-code">{{ item.all }}</code></td>
-                      <td><code class="stacks-code bg-white">{{ item.output }}</code></td>
-                  </tr>
+                {% for item in grid.column-templates %}
+                    <tr class="fs-caption">
+                        <th scope="row"><code class="stacks-code">.{{ item.class }}</code></th>
```
abovedave, line 270:
```suggestion
                        <th scope="row"><code class="stacks-code d-block">.{{ item.class }}</code></th>
```

abovedave:
Got to this as soon as I could... :trollface: 

Firstly, the API looks good. I do think what's missing is an **opinionated approach** on grids moving forward which we should all adopt. As a Stacks consumer, some clear guidance on which is best - is CSS Grid the future? Should I refactor to this approach? Or is this for some fancy layouts we haven't seen yet but Design should now know is possible?

I like [this definition](https://rachelandrew.co.uk/archives/2016/03/30/should-i-use-grid-or-flexbox/) which you allude to in the docs - perhaps we need a "Need layout help?" page or something.

> **Flexbox** is essentially for laying out items in a single dimension – in a row OR a column. 
> **Grid** is for layout of items in two dimensions – rows AND columns.

I think `grid`/`d-flex` is serving our needs and we'll end up confusing some folks as to what to use where, we might end up with pages which mix-and-match grid approaches, people using the wrong classes (e.g., I can see `ji-center` vs `jc-center` trolling), or some bugs which are in one but not the other.

**My opinion is the renaming and adding the feature are 2 separate actions**. If this were a feature in Node.js I'd have a deprecation warning saying `grid` is's being removed/changed in the next major version - everyone has fair warning and time to adapt – what's our equivalent? If we're sticking with Flexbox for the foreseeable then name-spacing this CSS Grid (as yuck as it is) so we can have access to it for those times flex is making things difficult then we can see where we are after some time? Like you said before we may even get better spacing features on flex soon.

Finally, I think this is probably the only time I would have reached for grid would have been a masonry style approach (`g-af-dense` in this world?), which we 'faked' with flex. Otherwise I'm happy with our grids as-is fwiw!

stackoverflow com_teams-march2021 png

PR #426, opened 2019-12-19
Update Modal example to match production

TedGoas:
Our examples for [Modals](https://stackoverflow.design/product/components/modals) do not match what's on the live site.

| **Stacks** | **StackOverflow.com** |
|:-:|---|
|Screen Shot 2019-12-19 at 7 24 49 AM|Screen Shot 2019-12-19 at 7 12 04 AM|

---

There's nothing wrong with the Stacks CSS/LESS, but the examples are picking up some unwanted CSS styles from the Stacks Documentation site that are not on our production sites.

This PR adds a few utility classes to the examples shown on the [Modals page](https://stackoverflow.design/product/components/modals) so they visually match how Stacks renders on our websites.

**Note:** This bug also appears in our Figma component for Modals.


TedGoas:
Nope. There's actually nothing wrong with the Stacks CSS, the Stacks site just doesn't benefit from having the Stack Overflow CSS reset file. This is an issue with the documentation site only.

For example, on StackOverflow.com we reset all headings to have the normal weight. That includes modal headings. However we don't reset this on Stacks docs site, so all headings in the docs are bold by default.

This makes the modal examples misleading because they have bold titles. Our Figma component is inaccurate and we also risk devs hacking titles to be bold after what they see in the Stacks docs.

lib/css/components/_stacks-modals.less
```diff
@@ -88,6 +88,7 @@
     margin-bottom: @su16;
     color: var(--fc-dark);
     font-size: @fs-headline1;
+    font-weight: normal;
```
TedGoas: @aaronshekey Is there an `@fw-normal`? I didn't see one.

TedGoas:
@aaronshekey Ping!

aaronshekey:
Closed in favor of #427

PR #428, opened 2020-01-14
s-tooltip: fix a variety of small bugs

bnickel:
This fixes a few bugs encountered during deployment:

1. The tooltip position is sometimes slightly off on first rendering.  Scheduling an update appears to fix it without being noticable to the user.
2. Tooltip's updated while visible weren't recentering on the content
3. If the tooltip is too wide for the browser window and obscures the content, we get a flickering as mouseover and mouseout rapidly fire.  This requires a more thorough examination, but can be mitigated for now by removing the popover's pointer events.  This should only really come up when we have a browser at mobile width but with a pointer device.


lib/ts/controllers/s-popover.ts
```diff
@@ -78,6 +78,11 @@ namespace Stacks {
             this.popper.enableEventListeners();
 
             this.popoverElement!.classList.add("is-visible");
+
+            // If the popover has undergone a content change while hidden, it may need to update its position slightly.
+            // Scheduling an update will cause the change to happen at the next animation frame.
+            this.popper.scheduleUpdate();
```
b-kelly: We're calling `popper.update` above, which *should* be taking care of the positioning. It's possible that it doesn't position right until being made visible though... Needs a second look. I'd either remove this line and move the `.update` down to here, or remove the `.update` above so we don't double up on the positioning calls.

lib/ts/controllers/s-popover.ts
```diff
@@ -174,6 +179,12 @@ namespace Stacks {
 
             this.popoverElement = popoverElement;
         }
+
+        protected updateIfVisible() {
```
b-kelly: I'd rename this to simply `update` and use it in place of wherever we'd call `popper.update` instead. Please add JSDocs to newly added functions.

lib/ts/controllers/s-tooltip.ts
```diff
@@ -82,7 +82,7 @@ namespace Stacks {
             if (!popover) {
                 popover = document.createElement("div");
                 popover.id = popoverId;
-                popover.className = "s-popover s-popover__tooltip";
+                popover.className = "s-popover s-popover__tooltip pe-none";
```
b-kelly: I don't anticipate needing to interact with a tooltip's contents, so this looks good to me! Hovering on regular popovers is... not great at the moment, so when we revisit that we can investigate this at the same time.

lib/ts/controllers/s-tooltip.ts
```diff
@@ -108,6 +108,8 @@ namespace Stacks {
                 popover.insertAdjacentHTML("beforeend", `<div class="s-popover--arrow"></div>`);
             }
 
+            this.updateIfVisible();
```
b-kelly: Good fix! I didn't anticipate this, but it's obvious in hindsight.

lib/ts/controllers/s-popover.ts
bnickel: I was running into the issue with the following markup, possibly because of line wrapping.

When the popover would first appear, it would be a smidge too high, then when I moved the popover would move to the correct position.  Just calling `scheduleUpdate()` seems to be sufficient without any visual movement of the element.

```
<button class="grid--cell s-btn s-btn__unset" data-controller="s-tooltip" data-s-tooltip-placement="right" title="Downvote this answer if you find it unclear or not useful"><svg aria-hidden="true" class="svg-icon m0 iconArrowDownLg" width="36" height="36" viewBox="0 0 36 36"><path d="M2 10h32L18 26 2 10z"/></svg></button>
```

lib/ts/controllers/s-popover.ts
bnickel: Doing `scheduleUpdate` since that seems to be all we need.

PR #433, opened 2020-01-27
Various improvements to modals

aaronshekey:
This PR fixes some in-page JS, renames `.has-danger` to `.s-modal__danger` to be more consistent with our naming. It also removes some additional IE11 workarounds.

With Webkit, I've also experimented with blurring our backdrop.

Screen Shot 2020-01-27 at 1 59 49 PM

### DIFF
--- a/docs/product/components/examples/banners.html
+++ b/docs/product/components/examples/banners.html
@@ -1,5 +1,5 @@
 <!-- Additional javascript -->
-<script src="{{ "/assets/js/feature.banners.js" | relative_url }}"></script>
+<script src="{{ site.baseurl }}/{{ site.js }}/feature.banners.js" defer></script>
 
 <!-- Override stacks classes for demos -->
 <style>
--- a/docs/product/components/modals.html
+++ b/docs/product/components/modals.html
@@ -57,7 +57,7 @@ description: Modals are dialog overlays that prevent the user from interacting w
 </aside>
 {% endhighlight %}
         <div class="stacks-preview--example ps-relative grid jc-center hs3">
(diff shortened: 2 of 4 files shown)


bnickel:
Do we know if this kind of blurring may trigger a doorway effect for the user?  In cases where we use modals for confirmation, it is helpful for the user to see past the modal to what they are confirming, and blurring that out might require us to resurface more information in the modal.

Related: We have a use-case on SO where we explicitly remove the backdrop from the modal.  The moderator action über-modals are configured as follows:

```
<aside class="s-modal pe-none bg-transparent">
    <div class="s-modal--dialog pe-auto">...</div>
</aside>
```

The specifics are that moderators need to copy links from a page into a text field to provide justification for an action.  With `pe-none`, they can right click and copy without the modal disappearing.  If we go with this blur option (or even if we don't) it would be helpful to have `s-modal__transparent` to make sure this functionality still works.

aaronshekey:
This is frustrating since, in my opinion, we shouldn't be using a modal for this situation at all.

> Modals are dialog overlays that prevent the user from interacting with the rest of the website until an action is taken or the dialog is dismissed. Modals are purposefully disruptive and should be used thoughtfully and sparingly.

Something I need to ponder. 🤔

bnickel:
I agree in principle, but I Stack Overflow has quite a few modals (both `.s-modal` and `.popup`) that are extremely heavy and contextual and won't go away without some radical rethinking of the experience. (Looking at you, close dialog.)

Maybe, we need an `.s-drawer` which is a modal that acts like those iOS UI things.

aaronshekey:
Punt on blurring until we can build and document a transparent modal, if that's something we we want to do, philosophically.

PR #437, opened 2020-02-06
Use codepen template instead of direct link

bnickel:
Right now, if you click on the Codepen link on pages, you're taken to a pen owned by stackoverflowstacks which can be edited but not saved.  You have to fork in order to save and that's not very intuitive.

This tweaks the link format so you're placed in a new savable Codepen using the existing pen as a template.

Try for yourself here:

- Old: https://codepen.io/stackoverflowstacks/pen/PMRpVb?editors=1000
- New: https://codepen.io/pen?template=PMRpVb&editors=1000


aaronshekey:
Hero.

PR #438, opened 2020-02-09
Cleanup docs icons in git

b-kelly:
## What?

This PR removes from git the icons that have been copied from the `stacks-icons` dependency into the docs `_includes`. This was originally done in #362, but was easily extracted out into its own PR. I've also added a short script to init the docs and updated the install documentation / netlify deploy config to reflect this.

I've also updated the Gemfile.lock (fresh install updated it). Feel free to trash this change.

## Why?

Removing the icons from the docs folder gives us a few advantages:
* Smaller initial pull size. We already pull in the icons via npm, why commit them to git as well?
* Decouples the dependencies from the docs. This helps us to potentially migrate to a new build system in the future (see #362)

## Disadvantages

* Icons now need to be copied at every deploy.
  * I don't see this as a big issue. A straight directory copy is fairly cheap
* Icons need to be copied on first pull
  * I've created a shortcut script that initialized the docs entirely. It is now *less* steps for the user than before, despite technically having more steps in the background

## Breaking changes

* Devs with an existing setup will need to run `npm run update-icons` before they build the site again. New devs will install according the the new instructions.
* No other breaking changes :)


aaronshekey:
Thanks for tackling this chore. Looks good to me!

PR #439, opened 2020-02-10
Fix expandable again

bnickel:
This was fixed in #367 but was merged into master instead of develop. 🤦‍♂

Merging into the correct place this time.


aaronshekey:
Oh, hahaha. We run a tight ship here.

PR #440, opened 2020-02-10
Add code coloring variables

aaronshekey:
This PR stubs out the variables necessary to introduce new code coloring. Tons of room for refinement, but let's get them into Stacks first.

https://codepen.io/aaronshekey/pen/jOPEVyO


aaronshekey:
Punted 🏈

PR #444, opened 2020-02-17
Switch to 11ty for docs

aaronshekey:
This PR switches from Jekyll to 11ty. There are a ton of benefits, but the biggest one is getting rid of ruby as a dependency. It also gets us Less styling and live reload.

Todo:

- [x] Product
- [x] Email
- [x] Content
- [x] Brand


aaronshekey:
Bye :wave:
image

PR #451, opened 2020-03-01
Revisit s-modal controller

b-kelly:
Reopens the work done in #415 that was accidentally left unmerged into develop.

In addition to porting over the previous code / docs, I've added a few more requested features to bring the accessibility/features up to par with prod's `showModal` function.

The modal now traps user tabbing/focus to within the modal and supports a `return-element` selector so we can refocus elements on modal close. I've also addressed the `transitionend` request from the last PR now that all major browsers support it (Feb 2020).

This is *almost* up to parity with `showModal`. I've punted on the following items:

* supporting `js-first-tabbable`, `js-last-tabbable` (auto detected)
* `closeOthers` support

I recommend reviewing this PR with **whitespace off** since I've moved some items around so we can mimic they layout on our Popovers page.


lib/ts/controllers/s-modal.ts
```diff
@@ -0,0 +1,222 @@
...
+         * Validates the modal settings and attempts to set necessary internal variables
+         */
+        _validate() {
+            // check for returnElement support
+            var returnElementSelector = this.data.get("return-element");
+            if (returnElementSelector) {
+                this.returnElement = <HTMLElement>document.querySelector(returnElementSelector);
+
+                if (!this.returnElement) {
+                    throw "Unable to find element by return-element selector: " + returnElementSelector;;
```
aaronshekey:
```suggestion
                    throw "Unable to find element by return-element selector: " + returnElementSelector;
```

Just a double semicolon.

docs/product/components/modals.html
```diff
@@ -3,75 +3,127 @@
...
+                    </td>
+                    <td>Helper to manually show an s-modal element via external JS</td>
+                </tr>
+            </tbody>
+        </table>
     </div>
+
+    {% header "h2", "Accessibility" %}
+    <p class="stacks-copy">
+        Models are designed with accessibility in mind by default. When a model is open, elements navigated by pressing <kbd>Tab</kbd> will be constrained to only those within the modal itself.
```
aaronshekey:
```suggestion
        Modals are designed with accessibility in mind by default. When a modal is open, navigation with the keyboard will be constrained to only those elements within the modal.
```

docs/product/components/modals.html
```diff
@@ -3,75 +3,127 @@
...
+        To ensure maximum compatibility, all <code class="stacks-code">a</code> tags must have <code class="stacks-code">href</code> attributes and any default focusable items you don't want focusable must have their <code class="stacks-code">tabindex</code> set to <code class="stacks-code">-1</code>.
```
aaronshekey:
```suggestion
        To ensure maximum compatibility, all <code class="stacks-code">a</code> tags must have <code class="stacks-code">href</code> attributes and any default focusable items you don’t want focusable must have their <code class="stacks-code">tabindex</code> set to <code class="stacks-code">-1</code>.
```

docs/_data/modals.json
```diff
@@ -0,0 +1,44 @@
+{
+  "classes": [
+    {
+      "class": ".s-modal",
+      "applies": "N/A",
+      "description": "Base parent container for modals"
+    },
+    {
+      "class": ".has-danger",
```
aaronshekey:
```suggestion
      "class": ".s-modal__danger",
```

We recently renamed this one

lib/ts/controllers/s-modal.ts
```diff
@@ -0,0 +1,222 @@
...
+                }
+            });
+        }
+
+        /**
+         * Binds tab presses on tabbable items such that tabbing only works within the modal
+         */
+        _bindTabFocusTrap() {
+            // get all tabbable items
+            var allTabbables = Array.from(this.modalTarget.querySelectorAll("[href], input, select, textarea, button, [tabindex]"))
```
bnickel: If we're doing this right, we should move `allTabbables` et al inside `_boundTabTrap`.  That way we capture elements being added and removed.

It also means we don't need a whole `_bindTabFocusTrap` method.

lib/ts/controllers/s-modal.ts
```diff
@@ -0,0 +1,222 @@
...
+    /**
+     * Helper to manually show an s-modal element via external JS
+     * @param element the element the `data-controller="s-modal"` attribute is on
+     * @param show whether to force show/hide the modal; toggles the modal if left undefined
+     */
+    export function toggleModal(element: HTMLElement, show?: boolean | undefined) {
+        var controller = Stacks.application.getControllerForElementAndIdentifier(element, "s-modal") as ModalController;
+
+        if (!controller) {
+            throw "Unable to get s-modal controller from element";
```
bnickel: We should follow a similar pattern to s-tooltip.  If the controller is null and we're showing, `setAttribute("data-s-modal-show-on-connect", "true")`.  Then check and clear that data in `connect` and show the modal.

This is useful for circumstances where a modal was fetched from the server or created from a template.

lib/ts/controllers/s-modal.ts
```diff
@@ -0,0 +1,222 @@
...
+            this._toggle(false);
+        }
+    }
+
+    /**
+     * Helper to manually show an s-modal element via external JS
+     * @param element the element the `data-controller="s-modal"` attribute is on
+     * @param show whether to force show/hide the modal; toggles the modal if left undefined
+     */
+    export function toggleModal(element: HTMLElement, show?: boolean | undefined) {
```
bnickel: I think we really just need `showModal(el)` and maybe `hideModals()`.  When would we want to indeterminately toggle a modal from code?

Since we explode if `return-element` is not set, it would make sense to have:

```
export function showModal(element: HTMLElement, returnElement?: HTMLElement) {

    if (returnElement && !returnElement.id) {
        returnElement.id = generateId();
        element.setAttribute('data-s-modal-return-element', '.' + returnElementId);
    }
    ...
```

lib/ts/controllers/s-modal.ts
```diff
@@ -0,0 +1,222 @@
...
+            // note: .contains also returns true if the node itself matches the target element
+            if (!this.modalTarget.querySelector(".s-modal--dialog")!.contains(target)) {
+                this._toggle(false);
+            }
+        }
+
+        /**
+         * Forces the popover to hide if the user presses escape while it, one of its childen, or the reference element are focused
+         */
+        _hideOnEscapePress (e: KeyboardEvent) {
```
bnickel: If we have a popover inside a modal, `Esc` will kill the outer rather than the inner.  This needs to be more sophisticated if we're binding at the document level.

lib/ts/controllers/s-modal.ts
```diff
@@ -0,0 +1,222 @@
...
+            this.triggerEvent(toShow ? "show" : "hide");
+
+            this.modalTarget.setAttribute("aria-hidden", toShow ? "false" : "true");
+
+            if (toShow) {
+                this._bindDocumentEvents();
+            }
+            else {
+                this._unbindDocumentEvents();
+                this._focusReturnElement();
```
bnickel: Situations on the ground may vary.  The comment flagging dialog could could have a return element inside the comment row but then the comment row gets deleted.

We should do something along the lines of the following, so clients can adjust the return element on `s-modal:hide`:

```
var event = this.triggerEvent("hide", { returnElement: this.returnElement });
this.modalTarget.setAttribute("aria-hidden", "true");
this._unbindDocumentEvents();
this._focusReturnElement(event.data.returnElement);

File: lib/ts/controllers/s-modal.ts [hunk 10]
```diff
@@ -0,0 +1,222 @@
...
+
+        /**
+         * Binds tab presses on tabbable items such that tabbing only works within the modal
+         */
+        _bindTabFocusTrap() {
+            // get all tabbable items
+            var allTabbables = Array.from(this.modalTarget.querySelectorAll("[href], input, select, textarea, button, [tabindex]"))
+                .filter((el: Element) => el.matches(":not([disabled]):not([tabindex='-1'])"));
+
+            //TODO support prod's js-[first,last]-tabbable overrides?
```
bnickel: No need.  This is fundamentally better.
```suggestion
```

lib/ts/controllers/s-modal.ts
```diff
@@ -0,0 +1,222 @@
...
+
+            // check for transitionend support
+            var supportsTransitionEnd = (<HTMLElement>this.modalTarget).ontransitionend !== undefined;
+
+            // shown/hidden events trigger after toggling the class
+            if (supportsTransitionEnd) {
+                // wait until after the modal finishes transitioning to fire the event
+                this.modalTarget.addEventListener("transitionend", () => {
+                    //TODO this is firing waaay to soon?
+                    this.triggerEvent(toShow ? "shown" : "hidden");
```
bnickel: I get that the focus trap brings focus into the modal on tab, but I'd still like the ability to automatically focus bring focus into the modal when it appears.

e.g. `data-target="s-modal.initialFocus"`...

Focusing would look like `if (this.initialFocusTargets.length > 0) { this.initialFocusTargets[0].focus(); }`, assuming that the targets array is in DOM order.

lib/ts/controllers/s-modal.ts
```diff
@@ -0,0 +1,222 @@
...
+     * @param show whether to force show/hide the modal; toggles the modal if left undefined
+     */
+    export function toggleModal(element: HTMLElement, show?: boolean | undefined) {
+        var controller = Stacks.application.getControllerForElementAndIdentifier(element, "s-modal") as ModalController;
+
+        if (!controller) {
+            throw "Unable to get s-modal controller from element";
+        }
+
+        controller._toggle(show);
```
bnickel: We should also check if the element is in the DOM.  If not, we should insert it at the end of `body`.

lib/ts/controllers/s-modal.ts
bnickel: I'd also like to see `data-s-modal-remove-when-hidden` which removes the element from the DOM after triggering the final event.  A lot of our modals are short-lived, getting generated by the server, doing their job once, and then disappearing forever.

lib/ts/controllers/s-modal.ts
```diff
@@ -0,0 +1,222 @@
...
+         * Hides the modal
+         */
+        hide () {
+            this._toggle(false);
+        }
+
+        /**
+         * Validates the modal settings and attempts to set necessary internal variables
+         */
+        _validate() {
```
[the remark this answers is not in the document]
b-kelly: Good point. Definitely a throwback and I'm not sure why I kept it this way originally. Will update 👍

lib/ts/controllers/s-modal.ts
b-kelly: Very good point. I'm not sure off-hand how I'd do it other than manually checking if the modal has esc closeable items... I'm going to punt on this now so we can get this in, but I'll be thinking about it over the next few days.

lib/ts/controllers/s-modal.ts
b-kelly: I was just talking about this today with another dev (the focus bit). I can add that easily enough. The second item should also be fairly easy to add as well and would save some boilerplate code across the platform that listens on `hidden` and then removes.

lib/ts/controllers/s-modal.ts
b-kelly: See above comment. I'm going to take your earlier advice to add support for setting the return element in `show/hideModal` instead since that seems like a more friendly implementation to me

lib/ts/controllers/s-modal.ts
```diff
@@ -0,0 +1,222 @@
...
+                    throw "Unable to find element by return-element selector: " + returnElementSelector;
+                }
+            }
+        }
+
+        /**
+         * Toggles the visibility of the modal element
+         * @param show Optional parameter that force shows/hides the element or toggles it if left undefined
+         */
+        _toggle (show?: boolean | undefined) {
```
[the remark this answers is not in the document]
b-kelly: I'm keeping it as-is for right now. The `_toggle` function is currently fairly symmetric and I don't want to tear it apart since we're trying to merge + release. I did add in the code shown to not change the modal state if the current state matches.

lib/ts/controllers/s-modal.ts
b-kelly: We still need the `_bindFocusTrap` method so I can unbind it from the document on close. I agree on adding `allTabbables` into the method so we can capture changes, but I'm going to punt this for now. I'm going to revisit s-modal very soon for a second round of changes and this'll be one of 'em.

lib/ts/controllers/s-modal.ts
b-kelly: Added both features! If `data-s-modal-remove-when-hidden="true"` is set, then it'll remove on `hidden` is triggered. If `initialFocus` is set, then that element will be focused on show, else the first focusable element will be focused instead.

lib/ts/controllers/s-modal.ts
b-kelly: Punting for now. Will add in my next round of changes.

lib/ts/controllers/s-modal.ts
b-kelly: Punting for now. Will add in my next round of changes.

b-kelly:
@bnickel Thanks for your review! I've added in most of the requested features. I had to temporarily punt on a few though. I do like those suggestions and I do plan to add them in, but I'd like to punt so we can cut a release with this in it as a v1.

lib/ts/controllers/s-modal.ts
```diff
@@ -0,0 +1,279 @@
...
+            var allTabbables = Array.from(this.modalTarget.querySelectorAll("[href], input, select, textarea, button, [tabindex]"))
+                .filter((el: Element) => el.matches(":not([disabled]):not([tabindex='-1'])"));
+
+            if (!allTabbables.length) {
+                return;
+            }
+
+            var initialFocus = <HTMLElement>allTabbables[0];
+
+            if (this.hasInitialFocusTarget) {
```
bnickel: :+1: We have some use patterns where multiple elements may have `data-target="s-modal.initialFocus"` because we're looping or rendering conditionally.  I've double checked and this code won't blow up and will successfully match the first element in the list. https://codepen.io/bnickel/pen/abOVBmX?editors=1011

lib/ts/controllers/s-modal.ts
```diff
@@ -0,0 +1,279 @@
...
+            }
+
+            // show/hide events trigger before toggling the class
+            var triggeredEvent = this.triggerEvent(toShow ? "show" : "hide");
+
+            // if this pre-show/hide event was prevented, don't attempt to continue changing the modal state
+            if (triggeredEvent.defaultPrevented) {
+                return;
+            }
+
```
bnickel:
```suggestion
            var triggeredEvent = this.triggerEvent(toShow ? "show" : "hide", { returnElement: this.returnElement });

            // if this pre-show/hide event was prevented, don't attempt to continue changing the modal state
            if (triggeredEvent.defaultPrevented) {
                return;
            }

            this.returnElement = triggeredEvent.detail.returnElement;
```

This is the change I was talking about.  It sends the current return element to the event listeners, and then applies any changes they make back to the property.  See https://codepen.io/bnickel/pen/oNXoYdX?editors=1011

PR #453, opened 2020-03-03
Rename body classes to theme-*

aaronshekey:
This PR renames our theming classes to something a bit more sensible from the outside:

`theme-light` nothing is wired up to this, but we'll be applying it by default to our body in production for clarity.
`theme-dark` currently called `has-darkmode__forced`
`theme-system` automatically shifts based on the media query, currently called `has-darkmode`

`theme-light has-highcontrast` a future flag on top of our themes
`theme-dark has-highcontrast`
`theme-system has-highcontrast`

### DIFF
--- a/lib/css/exports/_stacks-constants-colors.less
+++ b/lib/css/exports/_stacks-constants-colors.less
@@ -423,12 +423,12 @@ body {
     --scrollbar: rgba(255, 255, 255, .2);
 }
 
-body.has-darkmode {
+body.theme-system {
     @media (prefers-color-scheme: dark) {
         .dark-colors();
     }
 }
 
-body.has-darkmode__forced {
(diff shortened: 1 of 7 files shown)


lib/css/exports/_stacks-constants-colors.less
```diff
@@ -423,12 +423,12 @@ body {
...
 
-body.has-darkmode {
+body.theme-system {
     @media (prefers-color-scheme: dark) {
         .dark-colors();
     }
 }
 
-body.has-darkmode__forced {
+body.theme-dark
```
NickCraver: Missing a `{` here - then g2g

aaronshekey: Definitely!

PR #455, opened 2020-03-09
Add box-shadow xtra small: bs-xs

paintedbicycle:
This PR adds a more subtle box-shadow option: `bs-xs`

Screen Shot 2020-03-09 at 6 44 58 PM


aaronshekey:
Actually, merging as is and cleaning up in a separate PR :v:

aaronshekey:
I've beefed up the shadow so it's readable on 1x screens but I'm finding that these look pretty dang close to a border that we may want to consider using a border instead. 🤔 

There's also a possibility of splitting `--bs-xs` into two definitions behind a media query.

paintedbicycle:
Yeah we can likely get close enough with a border for most applications. It may make sense to add a smaller shadow down the line, but with `1x` issues and issues between chrome / safari, it's likely not worth fighting all that right now

PR #459, opened 2020-03-16
Preferences content guidance

aaronshekey:
This PR adds some guidance on writing good preferences copy.


TedGoas:
I dig it. We touch upon this in [voice and tone](https://stackoverflow.design/content/guidelines/voice-and-tone/#positive), but preferences is a big enough topic to warrant its own section.

tenor

---

I'd love to expand this to cover **Settings** as well and am happy to provide some examples.

aaronshekey:
What else should go in here? Maybe some guidance on checkboxes vs. radio?

TedGoas:
More examples of how to avoid negative language when creating labels or form fields in general:

| ❌ | ✅ |
|:-:|:-:|
| Disable  | Enable |
| Prevent | Allow |
| Hide | Show |
| Omit | Include |
| Reject | Accept |

PR #462, opened 2020-03-19
Add helper functions to support scripting popovers

bnickel:
This adds three helper functions to support popovers as they are scripted on Stack Overflow today.

Two are fairly basic:

- `Stacks.showPopover`:  Shows a popover.
- `Stacks.hidePopover`:  Hides a popover.

To facilitate these actions on popovers that don't yet have connected controllers, I've added the `data-s-popover-show-on-connect` attribute.  This has two advantages over setting `is-visible` on the `.s-popover` directly:

1. We don't have to recreate the logic for finding the popover when working with the element.
2. It avoids the "brief flash of misplaced popover" you can get when a popover has `is-visible` but is not yet positioned.

The last is much more advanced and deals with the fairly common use-case of attaching a popover to an element on demand:

`Stacks.attachPopover(element: HTMLElement, popover: HTMLElement | string, options?: PopoverOptions)`

Examples of this pattern on Stack Overflow:
- The share sheet controller takes a set of configuration options and configures its own share sheet element that it inserts into the DOM.
- The Teams reaction controller doesn't load its popover until the user clicks the button the first time.  Then it loads the popover HTML once for the page and inserts it into the DOM for the clicked element and shows it.
- An upcoming project receives a popover as a response for an action and want to display it on the button as an action.

In each of these cases,  the pattern would look very much like the example included in the docs.

There's a fourth pattern on SO not covered here which is basically "Show when the button is in the visible viewport" which could alternatively be "Don't hide until the popover has been in the visible viewport" but we can handle that at a future point if needed.


bnickel:
I also added functions for manipulating an element's `data-controller`, `data-target` and `data-action` attributes, since we're doing this fairly inconsistently on SO.

lib/ts/controllers/s-popover.ts
```diff
@@ -275,6 +284,162 @@ namespace Stacks {
...
+    }
+
+    /**
+     * Attaches a popover to an element and performs additional configuration.
+     * @param element the element that will receive the `data-controller="s-popover"` attribute.
+     * @param popover an element with the `.s-popover` class or HTML string containing a single element with the `.s-popover` class.
+     *                If the popover does not have a parent element, it will be inserted as a immediately after the reference element.
+     * @param options an optional collection of options to use when configuring the popover.
+     */
+    export function attachPopover(element: Element, popover: Element | string, options?: PopoverOptions)
```
b-kelly: Is this going to be used to attach attributes to existing markup or to inject a popup into the dom as a string? Stimulus does dom mutation observation, so if you're adding in a popover via an ajax call (or similar), then you'd just render the attributes into the string as usual and let Stimulus hook up the element on insertion. No interference needed.

Can we hang sometime to go over this? I'm having trouble envisioning how this will be used.

docs/product/components/popovers.html
```diff
@@ -261,6 +267,99 @@ <h3 class="fs-body2 mb2">Teams</h3>
...
+
+    {% header "h4", "JavaScript configuration" %}
+    <p class="stacks-copy">Situations may also arise where popovers need to be attached to an element after the document is rendered.  For example, a button could have a contextual menu that is too expensive to serve up on every page load.</p>
+    <p>Popovers can be attached to an element after the fact using <code class="stacks-code">Stacks.attachPopover</code>.</p>
+
+    <p>This method takes three parameters, the element to attach the popover to, the popover either as an element or an HTML string, and optional options for displaying the popover.</p>
+
+    <div class="stacks-preview">
+{% highlight javascript %}
+Stacks.application.register("actions", class extends Stacks.StacksController {
```
b-kelly: I'd write this type of scenario a bit more simply:

```html
<button id="test1">Show popover</button>
<div id="container1"></div>
<script>
document.querySelector('#test1').addEventListener('click', function(e){
    var popoverElement = document.querySelector('#container1 [data-controller="s-popover"]');

    if (popoverElement) {
        Stacks.showPopover(...);
    }
    else {
        //TODO fetch the popover, append to container1 (letting stimulus do the hookup on mutation) and then calling show
    }
});
</script>
```

lib/ts/controllers/s-popover.ts
```diff
@@ -194,6 +194,15 @@ namespace Stacks {
         private boundHideOnOutsideClick!: any;
         private boundHideOnEscapePress!: any;
 
+        connect() {
+            super.connect();
+
+            if (this.data.get("show-on-connect") === "true") {
```
b-kelly: I'd move this into the `BasePopoverController` and drop the `connect` function here. It'll likely never be used on tooltip, but it gives us the option and will work for other future inheritors

lib/ts/controllers/s-popover.ts
```diff
@@ -194,6 +194,15 @@ namespace Stacks {
...
+                this.data.delete("show-on-connect");
```
b-kelly: Why delete the data after showing? You'd want it to re-run if the controller was ever reconnected.

docs/product/components/popovers.html
```diff
@@ -78,6 +78,11 @@
                     <td>Controller element</td>
                     <td class="p8 lh-lg">Dictates where to place the popover in relation to the reference element. By default, the placement value is <code class="stacks-code">bottom</code>. Accepted placements are <code class="stacks-code">auto</code>, <code class="stacks-code">top</code>, <code class="stacks-code">right</code>, <code class="stacks-code">bottom</code>, <code class="stacks-code">left</code>. Each placement can take an additional <code class="stacks-code">-start</code> and <code class="stacks-code">-end</code> variation.</td>
                 </tr>
+                <tr>
+                    <th scope="row"><code class="stacks-code">data-s-popover-show-on-connect="true"</code></th>
```
b-kelly: Let's change to something like `data-s-popover-auto-show`. `connect` is more technically correct, but doesn't mean much to anybody who hasn't written a Stimulus controller and doesn't know that the lifecycle step is called `connect`.

lib/ts/controllers/s-popover.ts
```diff
@@ -275,6 +284,162 @@ namespace Stacks {
             });
         }
     }
+
+    function getPopover(element: Element) : [boolean, PopoverController | null, Element | null, Element | null] {
```
b-kelly: Docs please, especially when the return type is so complex

lib/ts/tokenlist.ts
```diff
@@ -0,0 +1,80 @@
+namespace Stacks {
+
+    /**
+     * A token list similar to DOMTokenList used for manipulating whitespace separated attributes.
+     * When this list is modified, the associated attribute will also be changed.
+     */
+    export class TokenList {
```
b-kelly: This structure seems a little overkill to me. The majority of the time you're going to have a single value anyways and the rest of the time you just need to call `.split(' ')`. I don't foresee the need for `remove` either in any case that doesn't just involve blanking out the entire target attribute.

lib/ts/controllers/s-popover.ts
```diff
@@ -275,6 +284,162 @@ namespace Stacks {
...
+    }
+
+    /**
+     * Helper to manually show an s-popover element via external JS
+     * @param element the element the `data-controller="s-popover"` attribute is on
+     */
+    export function showPopover(element: Element) {
+        const [isPopover, controller, _] = getPopover(element);
+
+        if (controller) {
```
b-kelly: In what cases will `controller` be empty, but `isPopover` set? The only one I can think of is when the element exists, but hasn't yet been attached to the dom, in which case setting the `show-on-connect` doesn't help much

lib/ts/controllers/s-popover.ts
```diff
@@ -275,6 +284,162 @@ namespace Stacks {
...
+    }
+
+    /**
+     * Helper to manually hide an s-popover element via external JS
+     * @param element the element the `data-controller="s-popover"` attribute is on
+     */
+    export function hidePopover(element: Element) {
+        const [isPopover, controller, _, popover] = getPopover(element);
+
+        if (controller) {
```
b-kelly: See question above regarding usage / dom status. In this case, Do we care much about removing the `is-visible` class? If the element isn't in the dom, it won't be visible anyways.

lib/ts/tokenlist.ts
bnickel: You would be so... so.. surprised here.  We have produced the following scenarios **TWICE** in the last few months.

A button has a popover that announces its new functionality.  When the popover is dismissed it is discarded and the element's natural tooltip starts functioning as normal.  When the button is clicked, its action loads a new popover to be displayed immediately.

So, the button transitions through the following three states:

```html
1. <button data-controller="s-popover s-tooltip" aria-controls="intro" />
2. <button data-controller="s-tooltip" />
3. <button data-controller="s-tooltip s-popover" aria-controls="details" />
```

Even without the third step, we need a mechanism to cleanly remove `data-controller="s-popover" aria-controls="intro"` while preserving and hopefully being unaware of `data-controller="s-tooltip"`

lib/ts/controllers/s-popover.ts
bnickel: That is exactly the case, and that is why we need `show-on-connect` (and why I brought this up when we were doing `s-modal`).

Imagine on Stack Overflow you have the following code:

```javascript
StackExchange.ready(() => {
    Stacks.showPopover(document.querySelector('#my-popover"));
});
```

That function block will execute _after_ Stacks has loaded but before Stimulus has connected any controllers.  In that case `isPopover` would be true, `controller` would be `null`, and the existing production code would throw an error.

When we detect that exact configuration, we insert the `data-s-popover-show-on-connect` attribute.  Milliseconds later, Stimulus creates the controller and calls `connect`, which in turn reads the attribute and shows the popover.

This pattern was directly inspired by usage.  If you look at `toggleStacksPopover` in the SO codebase, it waits 50ms before showing the popover with this comment `// needs some delay because we initialize the popover controller on the fly and need to give stimulus a chance to take over`

lib/ts/controllers/s-popover.ts
bnickel: This one is mostly about good housekeeping.  If we're hiding, we're hiding.  I just know we will have some case in our codebase where we do `showPopover` and `hidePopover` in the same breath, and if we don't set users up for success, we're going to have a bad time.

lib/ts/controllers/s-popover.ts
bnickel: I could rename it to `show-on-first-connect` to reduce ambiguity.  This attribute is meant mainly for `show` and when you call `show`, you indend for it to fire once, not many times.

lib/ts/controllers/s-popover.ts
bnickel: That's reasonable.  Will do.

docs/product/components/popovers.html
bnickel: I prefer the technical precision but am fine with that.

lib/ts/controllers/s-popover.ts
bnickel: The goal is to go from having the following in the DOM:

```html
<button>React</button>
```

And the following payload from the server:

```
<div class="s-popover">...</div>
```

To having the following in the DOM:

```
<button data-controller="s-popover" data-action="s-popover#toggle" aria-controls="{uniqueId}">React</button>
<div class="s-popover is-visible" id="{uniqueId}">...</div>
```

Stimulus is smart enough that if the payload were the final state and I simply replaced the button with it, it would wire it up and work.  Unfortunately, in practice what we end up with is what I showed above, a button on the page and a popover in a JSON response.

The demo shows exactly what I want to do but for a real-world example, look at:

https://gh.stackoverflow.com/Core/StackOverflow/blob/master/StackOverflow/Content/Js/PartialJS/full/04_Vote.js#L908

That function could be reduced to:

```javascript
loadPopover() {

    if (Stacks.getControllerList(this.buttonTarget).contains("s-popover")) {
        return;
    }

    loadTemplate().done(html => {
        Stacks.attachPopover(this.button, html, { showOnConnect: true, placement: 'right", toggleOnClick: true });
        this.updateUI();
    });
}
```

docs/product/components/popovers.html
bnickel: That's a huge `//TODO` right there.  In all my scenarios, the button you'd click is the same as the s-popover's reference element.  Per the current implementation, it can't have `data=controller="s-popover"` until the popover is loaded or it will throw errors.  That means we need code to add the controller to the button after it loads.

lib/ts/controllers/s-popover.ts
b-kelly: That's fair enough. I was more worried about the `remove, then readd to dom` circumstance, but... that's not likely to happen often and even then, I'm ok with not auto showing

lib/ts/controllers/s-popover.ts
b-kelly: Ah, race conditions. Fun. [Stimulus](https://github.com/stimulusjs/stimulus/blob/master/packages/%40stimulus/core/src/application.ts#L74) fires on `DOMContentLoaded`, which fires *after* scripts are parsed (unless marked `async`). This is all starting to make sense now... Thank you for the excellent description of the issue.

I'm going to pop a little comment in the source code explaining this for future devs.

lib/ts/controllers/s-popover.ts
b-kelly: Reviewers note: See above comment by bnickel with the explanation. tl;dr: race conditions

I'm cool leaving these in now that I know the background behind it

docs/product/components/popovers.html
b-kelly: Ah, we can't have the popover missing at validation time, right. Hmm... We'd definitely want to support this case natively without hacks, but if we don't have a named popover target, we attempt to generate it, which will fail for the base popover. I'd love to suppress that error in this case and just let the consumer keep writing the controller like they're used to.

An alternative would be to have a "loading"/placeholder popover as a dummy to bypass validation, then load the real one in on click later.

Maybe we should add `s-popover-async` and `s-modal-async` (loadModal replacement) controllers for something like this. Seems common enough.

Let's punt on this for now and revisit soon. Resolving.

lib/ts/controllers/s-popover.ts
```diff
@@ -275,6 +280,172 @@ namespace Stacks {
...
+        if (existingPopover) {
+            throw `element already has popover with id="${existingPopover.id}"`
+        }
+
+        if (!referenceElement) {
+            throw `element has invalid data-s-popover-reference-selector attribute`
+        }
+
+        if (typeof popover === 'string') {
+            const elements = document.createRange().createContextualFragment(popover).children;
```
b-kelly: PR Note: we do this exact type of string to dom insert in `tooltip`. Potential for abuse/xss was already discussed then, so we should be good here.

No action needed.

lib/ts/controllers/s-popover.ts
b-kelly: I'd like to improve this approach later, but I'll approve for now so we can get this into the next release. It should be fairly simple to clean up any usages if we decide to pivot later.

lib/ts/tokenlist.ts
b-kelly: Are both situations with attaching and detaching a popover? I think you've mostly handled this in the new `attach/detachPopover` helpers you've added. Do you see this being used more often? This seems like a fairly unique situation where a user is dynamically attaching multiple controllers to a single element, which I feel shouldn't come up that often.

lib/ts/controllers/s-popover.ts
```diff
@@ -40,6 +40,11 @@ namespace Stacks {
                 // just call initialize here, not show. This keeps already visible popovers from adding/firing document events
                 this.initializePopper();
             }
+
+            if (this.data.get("auto-show") === "true") {
+                this.data.delete("auto-show");
+                this.show();
```
b-kelly: Just tested locally and there's a slight issue. I'd expect the `auto-show` to work like our current `is-visible` popovers so we can deprecate that method (this is *far* superior).

The issue I found is that with our current default visible popovers we don't close them on outside clicks until you've interacted with them at least once. We do this by simply not attaching the event handlers at connect and letting them stay shown. Your current implementation causes any auto-shown popovers to hide themselves on document click.

I'd recommend *not* removing the `auto-show` attribute here, check for it in `show`, use it to not trigger `show` event, and then bail after the popper has been initialized, but before `shown()` is called, removing the attribute before bail.

bnickel: This would mean `showPopover` has different behavior if it's called before or after the controller is wired up, which I'd be strongly against.

I think the real problem is that `is-visible` is being used to do something novel and short-circuiting the show logic (I thought it was a bug at first) to keep itself around.  What we really need is something like `data-s-popover-dismiss-on-outside-click="false"`

lib/ts/tokenlist.ts
bnickel: Only one controller needs to be dynamically added or removed to introduce a risk of things breaking. (Again, the need to attach or remove a popover from an element that has another controller has come up several times in practice since we've added the popover controller.  It's far from a one-off experience and there are several places in code where we would have benefitted from this feature.)

Imagine for example, we have the following code `<button class="js-my-button" data-controller="s-popover">Some button</button>`. We need to remove the popover after an interaction and call `$('.js-my-button').removeAttr('data-controller');`.

Later the button gets a second controller and becomes `<button class="js-my-button" data-controller="s-popover my-button-controller">Some button</button>`. Now when we call `$('.js-my-button').removeAttr('data-controller');`, we remove two controllers instead of one.

Not having a standard surgical way of manipulating the controller list means we're accumulating complexity in our codebase that will lead to bugs and testing complexity.

Alternatives I considered were:
- Using the native `DOMTokenList` that is used for `Element.classList`.  This wasn't feasible because there is no public constructor.
- Using the token functionality that exists in Stimulus.  This wasn't feasible because Stimulus only listens to mutation events and doesn't surface anything to do the actual mutations.
- Having simple functions like `Stacks.addToControllerAttribute(element, value)`.  This would result in 6 to 9 methods and a lot of the same underlying complexity.
- Having simple functions like Stacks.addTokenToAttribute(element, 'data-controller', value)`.  This was also pretty verbose.

I ended up with this because it wasn't too verbose and it mirrored existing patterns in the DOM.

One step further would be to put helpers in `Element.prototype` so you could do `myElement.stimulusControllerList.remove('s-popover')` but that seemed like a bridge too far at the moment.

bnickel:
@aaronshekey @b-kelly I added back support for the "only hide on X button" behavior by introducing a new property `data-s-popover-hide-on-outside-click="never"`.

I also used that opportunity to catch the last common Stack Overflow use-case "don't hide if the user's never seen it" with `data-s-popover-hide-on-outside-click="if-in-viewport"`. (I'm open to suggestions on the name.)

We'll use that for cases where we have an onboarding popover below the fold, like on an answer or at the bottom of a long question.

lib/ts/controllers/s-popover.ts
b-kelly: The issue here is that we *want* the popover to dismiss on outside click, just not until the user interacts with it the first time. If the user chooses to dismiss it, then later shows it, we'd want it to act "normal".

I can see a compromise here though where setting `auto-show` or `is-visible` also sets `dismiss=false` automatically (which is where the real regression is here) and just deal with the event not being bound at all and making the user manually dismiss if the popover is reshown (unlikely).

@aaronshekey Thoughts?

b-kelly:
@bnickel I really appreciate the addition of `hide-on-outside-click` (especially with the extra work for `if-in-viewport`). I've already received a similar request and I know this will be needed at large eventually, but if we can, please refrain from adding additional features to this PR since it's already fairly large / complicated. Every new feature needs to go through another round of reviews and testing and we're hoping to get this in so we can cut a new release soon.

lib/ts/controllers/s-popover.ts
bnickel: @b-kelly Maybe I'm misunderstanding.  The current behavior of `is-visible` at https://stackoverflow.design/product/components/popovers/ is that the "Create team" popover can only be dismissed by clicking the dismiss button.  This is achieved by not binding the document level events.

If that's the requirement, adding a setting `data-s-popover-dismiss-on-outside-click` makes things a lot clearer than having a `auto-show` or `is-visible` overloaded with a second unexpected behavior, especially since we are using `auto-show` inside `showPopover` and we want it to behave exactly like it would if the controller instance already existed.

bnickel:
I get that.  I can pull the `if-in-viewport` part out into a second PR if it helps.  The inclusion of `hide-on-outside-click` was meant to fix the regression that Aaron and you had mentioned and I wanted to flesh out the story a bit because if we just did `true|false` now, it wouldn't be as extensible.

PR #465, opened 2020-03-23
Update Popper 1.x to version 2.x

b-kelly:
## What?

This PR updates our Popper.js dependency to the [2.x release line](https://popper.js.org/docs/v2/) so we can take advantage of new positioning functionality and just generally keep all our dependencies up to date.

## What changed?

Updated popperjs to v2.1.1 which has some minor repercussions:

* Instead of disabling all events on show/hide, we now just destroy/create the popper itself. They removed the api to easily disable events, so this was easier and more in line with their new [performance recommendations](https://popper.js.org/docs/v2/tutorial/#performance).
* some css had to change. `x-placement` is now `data-popper-placement`. We shouldn't have any consumers relying on this
* Margins are now ignored for placement, so I had to hardcode the value of `@su8` (10px) into the JS file

## Breaking changes?

None. :)


b-kelly:
> It would be cool if we had a test suite for this, but so little has changed on our side and all the docs work as expected so it seems fine.

+1 on a test suite eventually. I'm semi-actively looking into this for another project that'll need it in the near future, so hopefully I can bleed some of that knowledge into Stacks soon.

aaronshekey:
> > It would be cool if we had a test suite for this, but so little has changed on our side and all the docs work as expected so it seems fine.
> 
> +1 on a test suite eventually. I'm semi-actively looking into this for another project that'll need it in the near future, so hopefully I can bleed some of that knowledge into Stacks soon.

We have visual regression testing in [backstop](https://github.com/garris/BackstopJS). We'd have to add a few steps to https://github.com/StackExchange/Stacks/blob/develop/backstop.json in `develop` and then compare in this branch, but absolutely possible.

PR #466, opened 2020-03-23
Fix documentation bug when script executes prior to library loading

bnickel:
This fixes an issue where the script-generated tooltips fail to appear because the script is evaluated before the Stacks JavaScript is loaded.


aaronshekey:
I think the thing that's trolling us the most is that I'm loading subsequent pages via AJAX and swapping the `#content`. This is so we can maintain scroll position on our navigation, but also messes up in-page JS when we have it. I've thrown `defer` on other JS links, but that requires a separate file.

🤔

PR #467, opened 2020-03-23
Prevent modals and popovers from disappearing when a clicked element is removed

bnickel:
There's a not-uncommon scenario in modals and popovers where an element inside the modal will be removed on click.  For example, when an `.s-tag--dismiss` element is clicked.  Since the internal event will trigger prior to the document level event, our "outside click" event will incorrectly treat these orphaned elements as outside clicks because no element contains them.

See https://codepen.io/bnickel/pen/LYVgEzq?editors=1011 for a repro.

This applies two fixes to prevent that issue from occuring:

1. `s-popover` and `s-modal` now listen for `mousedown` events instead.  These events fire even on mobile devices (https://codepen.io/bnickel/pen/vYOVEXw) and fire before `click` events.  This means it fires before most standard code does anything, so click handlers can delete elements with impunity.

2. Outside click handlers no longer consider elements detached from the document body to be "outside clicks", since they are ambiguous.


b-kelly:
You have a merge conflict here on the dist/minified file. Merge master, then recompile, then commit again so we can ensure your new changes make it in. This is good to go otherwise.

bnickel:
@b-kelly Done

b-kelly:
Fixed merge conflicts and merging since all reviewers gave their approval.

PR #469, opened 2020-03-27
Add dispatching element to the popover and modal's event details

b-kelly:
## What?

Adds a `dispatcher` property to an event's `detail` object that is set to the element that originally caused that event to occur. This allows users to customize the controller's show/hide behavior based on the circumstances around the event. Applies to all events in `s-popover`, `s-tooltip`, `s-modal`.

I've also added documentation for this addition on both the popover and modal pages. In addition, I've also updated the modal show/hide events to match the popover documentation to better illustrate that the events are cancelable.

## Why?

We've had an influx of requests recently wanting to customize the conditions a modal is hidden and pretty much all of them start with "if I click X, I don't want to popover to hide". We already supported canceling a modal/popover's show/hide, but we didn't expose what was the actual cause of the event which makes these requests really hard to fulfill without resorting to ugly hacks.

## Real world uses

Consider a modal with a third-party autocomplete inside. The autocomplete's "dropdown" selection is appended to the `body` tag for maximum positioning compatibility. Clicking inside this dropdown causes the modal to hide because it is considered "outside" the modal. We can now simply add an event listener to `s-modal:hide` and check if the dispatcher was inside the autocomplete dropdown to cancel hiding the modal.

## Sample code

```html
<body>

    <!-- pretend the rest is here ;) -->
    <div data-controller="modal" ...>
        ...
        <input class="js-autocomplete" />
        ...
    </div>

    <!-- autocomplete items dynamically added by the .js-autocomplete element -->
    <div class="js-autocomplete-items"></div>
</body>

<script>
document.querySelector('[data-controller="s-modal"]').addEventListener('s-modal:hide', (e) => {
    let autocompleteItems = document.querySelector('.js-autocomplete-items');    

    // if the autocomplete items was clicked, then we don't want to consider this "outside" the modal
    if (autocompleteItems.contains(e.detail.dispatcher)) {
        e.preventDefault();
    }
});
</script>
```

## Breaking changes / Regressions

None :)

## For consideration

I'm not married to the term `dispatcher`. This could probably be improved. Open to suggestions here.

Docs need to be checked over for tone, readability, usefulness, etc.


bnickel:
I personally would just make the signatures `show(sourceEvent?: Event)` and `hide(sourceEvent?: Event)` and the payload would contain `{ sourceEvent: sourceEvent }`.  We don't currently have examples of passing an `Element` directly, and I'm struggling to imagine a case where it would both make sense to pass an element that wouldn't have an associated with an event.

The event would give more context into what happened (if it was an escape press versus an outside click) while still exposing the target element.

Another thing that I would find useful would be including the popover and reference elements, both in show/hide and shown/hidden.  The design does not lend itself to trivially accessing these elements outside of the controller, so having a quick avenue to access them would be helpful.

b-kelly:
@bnickel I originally did that for the flexibility, but during implementation I realized it was not really necessary. I'm 100% on board for changing the signatures to just take `Event|null`. Also agreed in exposing the source event and the reference/popover elements there too.

I'm going to go ahead and merge so we can get this into the release, but I'll follow up with these changes this weekend since your suggestions won't cause a breaking change in the external api as-is.

PR #470, opened 2020-03-29
Let badge__mini be used with any coloration

b-kelly:
## What?

Renames `s-badge__mini` to `s-badge__sm` and moves it further up in the file so it's specificity is lower than the other classes. This allows it to be used with any other class, not just `important`.

## Example

```html
<span class="s-badge s-badge__important s-badge__mini">Mini important!</span>
<span class="s-badge s-badge__gold s-badge__mini">Mini gold!</span>
```
Actually looks like:

image

With these changes in place it becomes:

```html
<span class="s-badge s-badge__important s-badge__sm">Mini important!</span>
<span class="s-badge s-badge__gold s-badge__sm">Mini gold!</span>
```

image

## Breaking changes

`s-badge__mini` has been renamed to `s-badge__sm`. In addition, it no longer gives a `blue-600` background. Migrations will need to both rename `mini` to `sm` and add `s-badge__bounty` to regain the blue background.

### DIFF
--- a/docs/_data/badges.json
+++ b/docs/_data/badges.json
@@ -83,17 +83,17 @@
     "description": "Badge which notes reputation points taken away."
   },
   {
-    "name": "Mini Number Count",
-    "class": "s-badge__mini",
-    "html": "s-badge s-badge__mini",
-    "label": "99+",
-    "description": "Badge which notes activity."
-  },
-  {
-    "name": "Mini Number Count, Important",
+    "name": "Number Count, Important",
     "class": "s-badge__important",
     "html": "s-badge s-badge__mini s-badge__important",
     "label": "99+",
     "description": "Badge which notes important activity."
+  },
+  {
+    "name": "Badge, Small",
+    "class": "s-badge__sm",
+    "html": "s-badge s-badge__sm",
+    "label": "99+",
+    "description": "Decreases the badge size."
   }
 ]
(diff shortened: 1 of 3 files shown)


aaronshekey:
Big 👍 . Would we wanna go further and rename this to t-shirt sizes?

b-kelly:
> Big 👍 . Would we wanna go further and rename this to t-shirt sizes?

I'd say yes, just to be consistent across the framework, but we'd need to either alias `mini` for backwards compat or do a find/replace across the product suite. I'm not at my work laptop right now, so I couldn't say how many of these there actually are.

If we'd go so far as to replace all instances, I'd even just remove that default blue background and add a new class for it. That way `mini`/`sm` will just be for sizing.

aaronshekey:
I'm happy to do that. I'd guess we've got less than 10 instances of this. I could be wrong though 😛

b-kelly:
Sounds good to me. Agreed that we probably have very few instances of this. Should I do the changes here or in a follow up PR?

aaronshekey:
> Should I do the changes here or in a follow up PR?

Let's do it on this PR with `s-badge_sm` replacing `s-badge__mini`. We can get blue from `bounty`. `important` should also be stripped of sizing overrides as well.

This component might have been a little rushed :)

b-kelly:
@aaronshekey Done! Updated the docs as well to have a new section showcasing the newly separated size class.

PR #471, opened 2020-03-30
Fix modal being immediately hidden on toggle

b-kelly:
## What?

When a modal is *wrapped* in an `s-modal` controller, then toggled via `data-action`, it is shown, then immediately hidden again. This is because the toggling button is considered to be "outside" of the modal and causes the `hideOnOutsideClick` handler to fire.

## Reproduction

```html
<div data-controller="s-modal">
    <button data-action="s-modal#show">toggle</button>
    <aside class="s-modal" data-target="s-modal.modal" aria-hidden="true">...</aside>
</div>
```

## Breaking changes

None :)


bnickel:
https://github.com/StackExchange/Stacks/pull/467 would also fix for this.

bnickel:
The issue exists for showing modals on any click, not just inside controllers.

For example, the following code `.addEventListener('click', e => { Stacks.showModal(modal) });` also fails to produce a dialog per https://codepen.io/bnickel/pen/XWbOoEj?editors=1010, the only think that works is separating the presentation of the dialog from the click event that reached the document (either by moving the document event listener to `mousedown`, or using `setTimeout`, or `stopPropagation`).

This is demonstrated with this Codepen: https://codepen.io/bnickel/pen/XWbOoEj?editors=1010

Since #467 already addresses this, I think we should move forward with that PR.

b-kelly:
@bnickel I had forgotten that #467 wasn't merged yet and thought this was a new issue. I've merged that PR, so I'm going to close this one.

PR #474, opened 2020-04-01
Feature/toggle switches

### LINKED ISSUE #366: Component: On/Off Button Toggle
We have an On/Off Button Toggle in a few places and the pattern is spreading, so let's codify this as a Stacks component and write some guidelines on when and how it should be used. Here's what the pattern looks like in the email preferences section of the user profile and settings:

Screen Shot 2019-08-16 at 3 07 41 PM

---

**Let's document:**
1. How they should be coded (HTML+JS)
2. When they should be used (vs. checkboxes or radio buttons)
3. How the labels should be worded
4. How the form submission works (the change should be applied immediately with [a notice confirming](https://stackoverflow.design/product/components/notices/), regardless of other `Save` buttons on the page)

TedGoas:
This PR formalizes the markup for toggle switches, and provides guidance on when they should be used and how they should behave.

Screen Shot 2020-04-01 at 4 51 36 PM

It's meant to replace `f-radio-toggle`, `s-option-toggle`, and switches built using `s-btn`.

---
This PR resolves #366


TedGoas:
@b-kelly 👋 Any chance we can hook up some [success notices](https://stackoverflow.design/product/components/notices/) to these toggle switch examples do demonstrate the intended UX?

TedGoas:
Do comments get to Slack now?

b-kelly:
> @b-kelly 👋 Any chance we can hook up some [success notices](https://stackoverflow.design/product/components/notices/) to these toggle switch examples to demonstrate the intended UX?
> 
> My thought: Each toggle switch would trigger a floating success notice. For demo purpose, can we the same one.

Just for the demo/doc pages and not as a whole Stimulus controller? No prob!

Just so I know exactly what you're looking for, it's just that changing the toggle state shows a toast saying something like `Toggle value: [VALUE] selected`?

lib/css/components/_stacks-toggle-switches.less
```diff
@@ -0,0 +1,89 @@
...
+        background-color: @white;
+        color: var(--black-500);
+        font-size: @fs-body1;
+        font-weight: 400;
+        line-height: 1;
+        text-align: center;
+        white-space: nowrap;
+        cursor: pointer;
+          -webkit-user-select: none;
+             -moz-user-select: none;
```
b-kelly, line 39: We can drop `-moz` and `-ms`. The only browser that needs a prefix is Safari

https://caniuse.com/#feat=user-select-none

b-kelly:
@TedGoas I slapped in a quick toast shamelessly ripped from `features.notices.js`. It will show on `change` for any radio under `.s-toggle-switch` with a `value` attribute present. The toast itself is up at the top of the file, so please change the text to something that doesn't suck.

docs/product/components/toggle-switch.html
```diff
@@ -0,0 +1,249 @@
...
+    <div class="stacks-preview">
+    {% highlight html %}
+<div class="grid ai-center">
+    <div class="s-toggle-switch mr8">
+        <input type="radio" name="group" id="one" checked>
+        <label for="one">On</label>
+        <input type="radio" name="group" id="two">
+        <label for="two" class="s-toggle-switch--label-off">Off</label>
+    </div>
+    <label class="s-label">…</label>
```
aaronshekey, line 50: Is there a way to get these labels wired up to flipping? If not, we should get rid of `cursor: pointer`

docs/product/components/toggle-switch.html
```diff
@@ -0,0 +1,249 @@
...
+                    <input type="radio" name="example1" id="example1-off" value="Off">
+                    <label for="example1-off" class="s-toggle-switch--label-off">Off</label>
+                </div>
+                <label class="s-label">Label</label>
+            </div>
+        </div>
+    </div>
+</section>
+<section class="stacks-section">
+    {% header "h3", "3+ options" %}
```
aaronshekey, line 67:
```suggestion
    {% header "h3", "3 or more options" %}
```

Way better slug this way.

docs/product/components/toggle-switch.html
```diff
@@ -0,0 +1,249 @@
...
+                    <div class="grid ai-center mb12">
+                        <div class="s-toggle-switch wmx1 mr8">
+                            <input type="radio" name="dont-2-1" id="dont-2-1-on" value="On" checked>
+                            <label for="dont-2-1-on">On</label>
+                            <input type="radio" name="dont-2-1" id="dont-2-1-off" value="Off">
+                            <label for="dont-2-1-off" class="s-toggle-switch--label-off">Off</label>
+                        </div>
+                        <label class="s-label">Feature name</label>
+                    </div>           
+                    <label class="s-label d-block mb4">Feature URL</label>
```
aaronshekey, line 162: Let's get the `for` attributes wired up here. I guarantee you someone will copy and paste from these and grab the markup without it 😋

docs/product/components/toggle-switch.html
```diff
@@ -0,0 +1,249 @@
...
+    {% header "h2", "Guidelines" %}
+    {% header "h3", "Apply the change immediately" %}
+    <p class="stacks-copy">Toggles should apply the result immediately and confirm it with <a href="/product/components/notices/#classes">a success notice</a>. Toggle switches should never require users to press a “Save” button to apply the change. When you require users to press a submit button with a toggle, you may confuse them because it’s not the expected next step. In those cases, use checkboxes instead.</p>
+    <table class="s-table s-table__lg fs-body2 va-top w100 mb24">
+        <thead>
+            <th class="btw0 bxw0 bbw3 bbw3 bc-black-3 bg-transparent fs-body1 s-table--cell3">Context</th>
+            <th class="btw0 bxw0 bbw3 bbw3 bc-red-2 fc-red-500 bg-transparent fs-body1">{% icon "Clear", "mtn1 mbn1" %} Don’t</th>
+            <th class="btw0 bxw0 bbw3 bc-green-2 fc-green-500 bg-transparent fs-body1">{% icon "Checkmark", "mtn1 mbn1" %} Do</th>
+        </thead>
+        <tbody>
```
aaronshekey, line 123: This table is trolling our responsive layout. Let's wrap it like the other ones in `overflow-x-auto`.

docs/product/components/toggle-switch.html
```diff
@@ -0,0 +1,249 @@
...
+                        <label class="s-label">Feature name</label>
+                    </div>           
+                    <label class="s-label d-block mb4">Feature URL</label>
+                    <input class="s-input mb12" placeholder="https://...">
+                    <button class="s-btn s-btn__sm s-btn__primary">Save settings</button>
+                </td>
+                <td class="p24">
+                    <div class="grid ai-center mb24 pb24 bb bc-black-4">
+                        <div class="s-toggle-switch wmx1 mr8">
+                            <input type="radio" name="do-2-1" id="do-2-1-on" value="On" checked>
```
aaronshekey, line 169: I'd argue that this could be written as an affirmative checkbox.

lib/css/components/_stacks-toggle-switches.less
```diff
@@ -0,0 +1,87 @@
...
+        font-size: @fs-body1;
+        font-weight: 400;
+        line-height: 1;
+        text-align: center;
+        white-space: nowrap;
+        cursor: pointer;
+          -webkit-user-select: none;
+                  user-select: none;
+        -webkit-touch-callout: none;
+        transition-property: background-color, border-color, color;
```
aaronshekey, line 41: I'm ok with killing the transitions here since we don't have transitions anywhere else.

lib/css/components/_stacks-toggle-switches.less
```diff
@@ -0,0 +1,87 @@
...
+//
+//  ============================================================================
+//  $   BASE STYLE
+//  ----------------------------------------------------------------------------
+.s-toggle-switch {
+    display: flex;
+    align-items: stretch;
+
+    //  First hide the radio buttons
+    input[type="radio"] {
```
aaronshekey, line 17: Should we guard these with an explicit class name? Longterm it might be trouble to target the elements in this fashion.

lib/css/components/_stacks-toggle-switches.less
```diff
@@ -0,0 +1,87 @@
...
+        position: absolute;
+        left: -999em;
+        opacity: 0;
+    }
+
+    //  Next modifications to option cells
+    label {
```
aaronshekey, line 24: Same comment. Should we call these `s-toggle-switch--label` or something?

lib/css/components/_stacks-toggle-switches.less
TedGoas: My thought here is to minimize the number of things someone has to remember. This is well scoped and should avoid collisions. I'm willing to be wrong here, though.

TedGoas:
> I'm having some mild existential questions here. I'm not fully internalizing why one would use a checkbox vs. a toggle switch. Would we be better off converting any toggles into check boxes, and recommending toggle switches are only used if there are more than 2 options?! Why not just radios at that point?

In my mind, the thing that makes toggle switches unique is the immediacy of the change. If you want to turn something on or off, but it's part of a larger form or otherwise should be applied in a separate action, then traditional checkboxes or radio buttons are the way to go. But if you want to flip a switch and have the change take immediately, that's what these toggles are for. That's not the expected behavior of traditional checkboxes or radio buttons.

I've tried to play this up in the docs. Happy to discuss.

docs/product/components/toggle-switch.html
TedGoas: Are they not flipping for you??

aaronshekey:
One last overarching bit of feedback. I think our toggles are backwards.

image

image

image

image

PR #476, opened 2020-04-05
Badge minimal

paintedbicycle:
This adds `s-badge__minimal` which allows the badge component to be used in user cards or other places the user avatar is shown.

I hope to use this to build a `s-usercard` component so I can use those in an eventual `s-question` component.

Screen Shot 2020-04-05 at 8 33 33 PM


paintedbicycle:
Let's chat about this because it's not the exact spacing currently being used on the usercard. I hesitated to do too many overrides to make it perfect, so tried to find a middle ground. Happy to go further if we decide it makes sense.

aaronshekey:
The insanely annoying thing about our approach to documenting these badges is that they aren't actually embedded images. For theming, they're delivered as svg sprite backgrounds. As they stand, we aren't able to actually use these badge components across the Stack Exchange network. I explored switching these to inlined SVG, but we couldn't quite justify it from a performance standpoint. 🤔 

Though, all that said... I would slap `fc-gold-darker` on the respective text colors here.

paintedbicycle:
Yeah I noticed that the way we display the badge SVGs here in Stacks and in the codebase are different. I really only changed the style of the wrapper -> so we could use `s-badge` in the upcoming `s-usercard` I'm working on, instead of other markup. I think that part would still apply, regardless of how we output the `SVG`, no?

Happy to switch this up to whatever is needed.

paintedbicycle:
Oh I think I know what you mean -> it's applied by a CSS sprite currently? So yeah it's quite different than this markup. Got it.

The badge dot itself isn't important "content" in terms of needing to be in the markup.

We could:
- recreate the dots before the numbers using a CSS pseudo element
- or (worse) make one out of an empty span with a bg colour
- or maybe there is a circle unicode character for this

paintedbicycle:
Closing in favour of first updating badges in general

PR #479, opened 2020-04-07
Fix outside click race conditions

b-kelly:
## Pre-merge Todos

- [ ] remove ugly / out of place testing link on the popovers documentation

## What?

This PR introduces some changes to `s-modal` and `s-popover` controllers to automatically prevent event bubbling when toggling visibility. It also updates the docs to showcase using this with our `Stacks.*` helpers.

I've also updated the modals documentation to have one modal triggered by helper and another triggered by controller action. This is not apparent to the viewer, but helps us during manual testing.

## Why?

We have some issues right now with users wrapping `Stacks.showModal` inside a click event. The click shows the modal, then bubbles up to the document level, registers as an "outside" click, then closes the modal. This is also an issue in general with other controllers that listen for "outside" clicks,  such as popovers. This is *not* an issue for items toggled with `data-action`, but we can't force this usage in many cases (ajax loaded modals for instance).

## Reproduction

https://codepen.io/bk-SO/pen/MWwNpVG?editors=0010

## Breaking changes

Modal / popover outside click triggering has been moved from `mousedown` back to `click`. The reasoning here is that this drastically decreases support requests since `click` is the default for basically everyone when binding events and I'm already getting tired of having to say "no, bind to mousedown instead, just because. But only in this specific case".


bnickel:
Why would we need to tell people to bind things to `mousedown` events today?  The point of binding dismiss events to `mousedown` was so users could continue using `click` as-is.

b-kelly:
> Why would we need to tell people to bind things to `mousedown` events today? The point of binding dismiss events to `mousedown` was so users could continue using `click` as-is.

@bnickel We need to tell them to bind to mousedown so they can cancel the propagation, so the event doesn't hit `document`, triggering outside click. Please see the attached codepen for an example.

Do we have a specific reason to continue using `mousedown` over `click`? I thought it was thrown in there as an "extra" to protect us in a hypothetical scenario where we'd want the outside click handler to run before a click event happened.

bnickel:
The bug that was triggered here is caused by a somewhat undefined behavior.

When an outside click occurs, the popover is supposed to close per the design.  The code that you shared depends on using `return false` on the click button to prevent the event from bubbling up to the document and causing the event handler from being called.  That's something we don't explicitly declare support for.

A more explicit mechanism, that works with the existing code can be seen at https://codepen.io/bnickel/pen/NWGvXYO?editors=0011

    popover.on("s-popover:hide", (e) => {
        if ($('#toggle1').get(0).contains(e.detail.dispatcher)) {
            e.preventDefault();
        }
    });

Basically, we recreate the logic from `hideOnOutsideClick` to encompass the external button.

We don't currently surface the cause of `s-popover:hide` in the event, so we're relying on the fact that `toggle` was called without a parameter, but if we did, the code could look something like:

    popover.on("s-popover:hide", (e) => {
        if (e.detail.reason = Stacks.ToggleReason.outsideClick 
            && $('#toggle1').get(0).contains(e.detail.dispatcher)) {
            e.preventDefault();
        }
    });

That said, I'm generally skeptical of that code because it is explicitly calling `toggle`, something I would consider to be a controller protected method.  I'm also curious about the design pattern here, since I'm having trouble envisioning a use case where we would want to _toggle_ a popover from outside itself.  Better understanding the use-case would help in determining a long-term solution.

PR #481, opened 2020-04-20
Migrate rep badges to use award-bling

paintedbicycle:
This PR is a WIP.

It moves some badges to use `s-reputation` and adds the badge via pseudo element instead of a SVG.

1) One thing that is interesting is the text colour on the reputation. If I set it on the rep itself, it overrides the colour in a badge (this could mean in our examples, inside a badge, the text colour would be the same gold, silver, etc, where we'd want it to be fc-medium. We do however, want text to be that colour in certain places (i.e. in the user card). This is unfinished and may just have to be an atomic class in the user card.

2) I'm also not at all sold on the use of spans and divs here. Maybe we leave this to the devs in product? But we should be using the specific HMTL element where appropriate and avoiding the use of spans if we can.

Open to thoughts.

### DIFF
--- a/docs/product/components/badges.html
+++ b/docs/product/components/badges.html
@@ -106,7 +106,7 @@ description: Badges are labels used for flags, earned achievements, and number t
                     </span> 
                 </span>
                 <span class="grid--cell s-badge s-badge__silver">
-                    <span class="award-bling award-bling__gold">
+                    <span class="award-bling award-bling__silver">
                         7624
                     </span> 
                 </span>
--- a/docs/product/components/badges.html
+++ b/docs/product/components/badges.html
@@ -21,7 +21,7 @@ description: Badges are labels used for flags, earned achievements, and number t
                         <td>
                             {% if item.repType %}
                                 <span class="{{ item.html }}">
-                                    <span class="s-reputation s-reputation__{{ item.repType }}">
+                                    <span class="award-bling award-bling__{{ item.repType }}">
                                         {{ item.label }}
                                     </span>
                                 </span>
@@ -43,17 +43,17 @@ description: Badges are labels used for flags, earned achievements, and number t
     <div class="stacks-preview">
 {% highlight html %}
 <span class="s-badge">
-    <span class="s-reputation s-reputation__gold">
+    <span class="award-bling award-bling__gold">
         Great Question
     </span> 
 </span>
 <span class="s-badge">
-    <span class="s-reputation s-reputation__silver">
+    <span class="award-bling award-bling__silver">
         Favorite Question
     </span> 
 </span>
 <span class="s-badge">
-    <span class="s-reputation s-reputation__bronze">
+    <span class="award-bling award-bling__bronze">
         Altruist
     </span> 
 </span>
@@ -61,17 +61,17 @@ description: Badges are labels used for flags, earned achievements, and number t
         <div class="stacks-preview--example">
             <div class="grid gs4 fw-wrap">
                 <span class="grid--cell s-badge">
-                    <span class="s-reputation s-reputation__gold">
+                    <span class="award-bling award-bling__gold">
                         Great Question
                     </span> 
                 </span>
                 <span class="grid--cell s-badge">
-                    <span class="s-reputation s-reputation__silver">
+                    <span class="award-bling award-bling__silver">
                         Favorite Question
                     </span> 
                 </span>
                 <span class="grid--cell s-badge">
-                    <span class="s-reputation s-reputation__bronze">
+                    <span class="award-bling award-bling__bronze">
                         Altruist
                     </span> 
                 </span>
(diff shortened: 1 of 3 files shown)


aaronshekey:
This seems like a sane approach to me. Haven't dug into the text color issue you've referenced though.

I haven't been too concerned with semantic elements, but generally we should guard for `span` vs `div` in our CSS. For example, if we want these to be `display: block` we should force that so a `span` with these classes behaves as expected.

NickCraver:
FWIW, this is odd because badges are not reputation, they're just counts. Semantically this would be weird (and...confusing) all over the codebase. We display something like this on all properties with the top bar, so it'd be in many codebases.

paintedbicycle:
@NickCraver Yeah, what I tried to accomplish was:

1. You have badges, they are generic
2. You can add numbers inside badges
3. You can add reputation inside badges

So, `s-reputation` isn't used on the badge, but inside. Is that still weird? Did I misunderstand?

NickCraver:
I'm not sure I follow, let me try to phrase differently: badges have nothing to do with reputation is what I'm trying to convey, they are distinct concepts. Badges have counts whereas reputation is a single number. Calling the counts for a badge category "reputation" just isn't the right naming and will add a lot of confusion here. To us, they're just "badge counts".

paintedbicycle:
Ok, I'll revise it. Thank you

paintedbicycle:
OK I've changed `s-reputation` to `award-bling` so we can discuss.

1. I first tried `award-count` but we use this on named awards as well (as this micro component really only puts a bit of bling before the item it encases...it could be numeric or it could be a word like in "Great Question"
2. I kept the badges that have `award-bling` inside them (rather than using this PR to delete the badges that are using bling) because I found a place in product where we are using it in this style. It's inside the profile
3. I didn't require `award-bling` to be inside `s-badge` as this was it is available to be used in other places like a user card.

I'm open to other names for `award-bling`, it's a pretty trivial change

lib/css/components/_stacks-badges.less
```diff
@@ -105,3 +105,38 @@
...
+//  $$  Award Count
+//  ---------------------------------------------------------------------------
+.award-bling {
+    display: flex;
+    align-items: center;
+    color: inherit;
+
+    &:before {
+        content: "";
+        margin-top: -1px;
```
aaronshekey, line 118:
```suggestion
        margin-top: -1px;
```

I think we can kill this. They both look wrong to me, but this looks wrong in a way that's centered more perfectly in its container.

lib/css/components/_stacks-badges.less
```diff
@@ -105,3 +105,38 @@
...
+        margin-right: 3px;
```
aaronshekey, line 119:
```suggestion
        margin-right: 4px;
```

Looking a little tight with the standard HTML spacing.

lib/css/components/_stacks-badges.less
```diff
@@ -105,3 +105,38 @@
 .s-badge__important {
     .badge-styles(transparent, var(--red-600), var(--white));
 }
+
+//  $$  Award Count
+//  ---------------------------------------------------------------------------
+.award-bling {
```
aaronshekey, line 111: Let's still namespace these.

```suggestion
.s-award-bling {
```

docs/product/components/badges.html
```diff
@@ -70,25 +83,37 @@
...
             <div class="grid gs4 fw-wrap">
                 <span class="grid--cell s-badge s-badge__gold">
-                    <img class="s-badge--image" src="/assets/img/BadgeGold.svg" aria-hidden="true"> 635
+                    <span class="award-bling award-bling__gold">
+                        635
+                    </span> 
                 </span>
                 <span class="grid--cell s-badge s-badge__silver">
-                    <img class="s-badge--image" src="/assets/img/BadgeSilver.svg" aria-hidden="true"> 7624
+                    <span class="award-bling award-bling__gold">
```
aaronshekey, line 109:
```suggestion
                    <span class="award-bling award-bling__silver">
```

docs/product/components/badges.html
```diff
@@ -70,25 +83,37 @@
...
+                        7624
+                    </span> 
                 </span>
                 <span class="grid--cell s-badge s-badge__bronze">
-                    <img class="s-badge--image" src="/assets/img/BadgeBronze.svg" aria-hidden="true"> 8234
+                    <span class="award-bling award-bling__gold">
```
aaronshekey, line 114:
```suggestion
                    <span class="award-bling award-bling__bronze">
```

lib/css/components/_stacks-badges.less
paintedbicycle: @aaronshekey What's the change here? I think something happened with the review comment..it's not showing a change

aaronshekey: Just get rid of `margin-top` entirely.

PR #490, opened 2020-05-21
Code block component & highlighting

b-kelly:
## What?

This is an exploration of what an opinionated syntax highlighting theme would look like as a code block component.

## Tasks

- [x] New theme
- [x] Documentation
- [x] Light theme colors
- [x] Accessibility concerns
- [x] Test other popular languages


lib/css/syntax-highlighting/highlightjs.less
```diff
@@ -0,0 +1,85 @@
...
+        color: var(--highlight-variable);
+    }
+
+    .hljs-emphasis {
+        font-style: italic;
+    }
+
+    .hljs-strong {
+        font-weight: bold;
+    }
```
aaronshekey: For consistency, we probably shouldn't have bolding / italic in one highlighting and none in another. Considering just removing these.

b-kelly: These classes specifically are required for markdown syntax highlighting:

image

b-kelly: I forgot to add these into the non-`highlightjs` themes since they aren't used there, but I can give it a go pretty easily.

b-kelly: Removed the non-markdown `bold` styling for all themes

b-kelly:
@aaronshekey Updated the branch with my [highlightjs fork](https://github.com/b-kelly/eleventy-plugin-highlightjs) of `@11ty/eleventy-plugin-syntaxhighlight` and remove all prism references.

lib/css/components/_stacks-code-blocks.less
```diff
@@ -0,0 +1,102 @@
+pre[class*="language-"],
```
aaronshekey: These selectors need to be cleaned up. They're a little gnarly, and probably not how I'd do it from scratch.

lib/css/components/_stacks-prose.less
```diff
@@ -341,7 +341,7 @@
     // ============================================================================
     //  $ CODE STYLES
     // ----------------------------------------------------------------------------
-    code {
+    code:not(.s-code-block) {
```
aaronshekey, line 344: This is for unhighlighted code elements within `s-prose`.

lib/css/components/_stacks-prose.less
```diff
@@ -350,13 +350,16 @@
 
     pre {
         margin: 0 0 1.2em 0; // This is increased from 1em for better optical spacing
+    }
+
+    pre:not(.s-code-block) {
```
aaronshekey, line 355: This is for unhighlighted pre elements within `s-prose`.

docs/.eleventy.js
```diff
@@ -78,11 +79,20 @@ module.exports = function(eleventyConfig) {
...
-  eleventyConfig.addPlugin(syntaxHighlight);
+  eleventyConfig.addPlugin(syntaxHighlight, { className: "s-code-block" });
 
   // Add markdown shortcode
   eleventyConfig.addPlugin(markdownShortcode, {
-    html: true
+    html: true,
+    highlight: function (str, lang) {
+      if (lang && hljs.getLanguage(lang)) {
+        return '<pre class="language-' + lang + ' s-code-block"><code class="language-' + lang + ' s-code-block">' +
```
aaronshekey, line 89: This matches the output from our custom `eleventy-highlight` plugin, but it's probably a little weird.

docs/.eleventy.js
```diff
@@ -1,4 +1,5 @@
-const syntaxHighlight = require("@11ty/eleventy-plugin-syntaxhighlight");
+const hljs = require("highlight.js");
```
aaronshekey, line 1: Not sure why this works, lol

docs/.eleventy.js
b-kelly: I'm not 100% sure what this plugin actually does (embed markdown inside other template types?), but our highlight plugin should support markdown templates already. See the docs for the [official 11ty plugin](https://www.11ty.dev/docs/plugins/syntaxhighlight/#markdown-highlighter) that I ripped off of for examples

b-kelly: Also, this looks *suspiciously* like [markdown-it](https://github.com/markdown-it/markdown-it#syntax-highlighting) 😉. The "official" approach is very similar to the one you took here.

lib/css/components/_stacks-code-blocks.less
b-kelly: Technically speaking, this doesn't even cover all the bases (in practice). In prod we use `lang-*` classes instead... That being said, the post contents bake all the code blocks as bare `pre>code` and we add in the highlighting classes client-side, so I have absolutely no problem with this approach.

In addition, the [html spec](https://html.spec.whatwg.org/#the-code-element) states:

> There is no formal way to indicate the language of computer code being marked up. Authors who wish to mark code elements with the language used, e.g. so that syntax highlighting scripts can use the right rules, can use the class attribute, e.g. by adding a class prefixed with "language-" to the element.

So, we're adhering to best practices there too.

lib/css/exports/_stacks-constants-colors.less
```diff
@@ -291,6 +291,32 @@ body,
...
+    --highlight-attribute: #0b558a; // ~powder-700 AAA 7.25
+    --highlight-literal: #ab5a0d; // ~orange-700 AA 4.63
+    --highlight-symbol: #803378; // Custom purple AAA 7.30
+    --highlight-keyword: #0b558a; // ~powder-700 AAA 7.25
+    --highlight-variable: #3d5d01; // Custom green AAA 7.03
+    --highlight-addition: @green-700; // AA 5.59
+    --highlight-deletion: @red-600; // AA 5.31
+}
+
+@supports (-webkit-overflow-scrolling: touch) {
```
b-kelly: Was this meant to be added to this PR? Looks like it might be unrelated

lib/css/components/_stacks-code-blocks.less
```diff
@@ -0,0 +1,102 @@
...
+    border-radius: @br-md;
+    max-height: 24rem;
+    margin: 0;
+    padding: @su8 @su12;
+    overflow: auto;
+
+    @scrollbar-styles();
+}
+
+code[class*="language-"] {
```
b-kelly, line 17: I suggest also supporting the case where `code` is bare, but the enclosing `pre` has the `s-code-block` class attached. This is going to be required for prod since we attach the language class to the `pre` instead and switching that to `code` is a major pain (I know, I tried...)

```suggestion
pre.s-code-block > code,
code[class*="language-"] {
```

lib/css/exports/_stacks-constants-colors.less
```diff
@@ -419,6 +445,20 @@ body,
 
     // Scrollbars
     --scrollbar: rgba(255, 255, 255, .2);
+
+    // Syntax highlighting
+    --highlight-bg: #1c1b1b;
+    --highlight-color: @white; // AAA 17.19
```
b-kelly, line 439: Having these stats available to point to is awesome! Do we want to add a little blurb to the `Code block` docs page saying something like

> All our colors are designed against WCAG best practices and have a contrast value of at least blah blah blah

b-kelly:
Here are some other themes for comparison. Large image warning! I've hidden them behind spoilers so I don't junk up the PR.

Comparatively speaking, this theme looks really good. The only items I wish were "better" are being detected as "text" by highlight.js, so that's more of a shortcoming of the highlighter itself, not the theme. If the function/type detection by highlight.js was _slightly_ better, I'd be happy using this exact theme in a code editor for professional purposes. Excellent work!

<details>
<summary>Prettify / prod theme</summary>

prettify-samples

</details>

<details>
<summary>VS Code default theme in markdown file; some syntaxes not supported due to not having the right language extensions</summary>

vscode-samples

</details>

b-kelly:
Rendered with Chrome's `Emulate vision deficiencies` tools, both light and dark themes. Large image warning! Adding behind spoilers so I don't junk up the PR.

Tritanopia looks good to my eyes - the colors are all distinguishable. Both Deuteranopia and Protanopia suffer from `--highlight-symbol` and `--highlight-keyword` appearing as basically indistinguishable, *however* these are only used together in less/css, where the visuals are acceptable in my opinion (class name vs property keys are the same-ish color).

<details>
<summary>Tritanopia</summary>

tritanopia-light

tritanopia-dark

</details>

<details>
<summary>Deuteranopia</summary>

deuteranopia-light

deuteranopia-dark

</details>

<details>
<summary>Protanopia</summary>

protanopia-light

protanopia-dark

</details>

lib/css/components/_stacks-code-blocks.less
aaronshekey: Ok, so the `pre[class*="language-"]` selector can be removed entirely?

b-kelly: I'm 100% down with that. I feel that explicitly scoping these to `s-code-block` is preferred. We can remove the `code[...]` selector too, in favor of my added suggestion.

docs/.eleventy.js
aaronshekey: I tried this out. We don't want to highlight the syntax of Markdown here. Instead, we want to render all the markdown as HTML and shove it into our documentation. Like so:

```
<div class="s-prose">
    {% markdown '/_includes/markdown-sandbox.md' %}
</div>
```

I'm sure there's an opportunity to clean this up, but in the nearterm, this matches if I'd just written all that HTML inside of the `s-prose` content, including the syntax highlighting.

b-kelly: Ah, I get ya. It isn't a markdown 11ty template, it's straight up a raw markdown file getting read and rendered. I don't feel strongly about how we're doing this, just figured I'd mention it. If we migrate that file to an 11ty markdown template in the future we could take advantage of the built in highlighter, but until then this is more than good enough.

aaronshekey:
> that's more of a shortcoming of the highlighter itself, not the theme

I triple checked this was the case. Frustrating, especially in Ruby and Java.

lib/css/exports/_stacks-constants-colors.less
aaronshekey: Done!

docs/product/components/code-blocks.html
```diff
@@ -0,0 +1,419 @@
+---
+layout: page
+title: Code block
+description: Stacks provides styling for code blocks with syntax highlighting provided by highlight.js. Special care was taken to make sure our light and dark themes felt like Stack Overflow while maintaining near AAA color contrasts and still being distinguishable within tritanopia, deuteranopia, and protanopia.
```
b-kelly: Suggestion: drop the medical terms and make it so people like myself who had no idea what those terms meant get it without thinking about it.

```suggestion
description: Stacks provides styling for code blocks with syntax highlighting provided by highlight.js. Special care was taken to make sure our light and dark themes felt like Stack Overflow while maintaining near AAA color contrasts and still being distinguishable for those with a color vision deficiency.
```

PR #491, opened 2020-05-25
Improve notice buttons presentation

aaronshekey:
This PR tints the stock `s-btn` appropriately within an `s-notice`.

### DIFF
--- a/docs/_data/notices.json
+++ b/docs/_data/notices.json
@@ -6,6 +6,11 @@
       "description": "Default style used to separate notices from the main content."
     },
     {
+      "class": ".s-notice--btn",
+      "applies": "N/A",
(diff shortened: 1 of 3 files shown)


bnickel:
`.s-notice .s-btn` has higher priority than `s-btn__danger`, so this currently breaks staying for something like:

```
<aside class="s-notice">
    <p>This question is a duplicate</p>
    <button class="s-btn s-btn__danger">Delete question</button>
</aside>
```

aaronshekey:
No doubt! I've actually been thinking on this, but not sure how to adjust. I suppose I could re-style every button state. I was just testing what kinds of buttons we have in practice within `s-notice`.

aaronshekey:
This would also mess with theming. Perhaps there's `s-notice--btn` for these specific overrides.

PR #497, opened 2020-06-07
Add responsive tables classes

paintedbicycle:
This PR adds responsive classes to tables.

1. It adds a `s-table__responsive` class that allows tables to scroll
1. It adds a `s-table__responsive--sticky` class that sticks the first column
2. It adds `s-table__responsive--sticky-col-[size]` classes allowing for 4 different widths of the sticky first column


lib/css/components/_stacks-tables.less
```diff
@@ -315,6 +314,23 @@
...
+    //  $$  Responsive
+    //  ------------------------------------------------------------------------
+    &.s-table__responsive {
+        display: block;
+        overflow-x: auto;
+        @scrollbar-styles(); // Style the scrollbars appropriately
+    }
+
+    &.s-table__sticky {
+        th:first-child {
```
aaronshekey, line 321: I've found a potential show stopper with this approach. We lose a lot of borders in Firefox.

image

docs/product/components/tables.html
```diff
@@ -376,6 +376,150 @@
...
+                        <td>Design</td>
+                        <td><div class="grid ai-center"><div class="fw-bold mr8" title="reputation score " dir="ltr">611</div><div class="grid ai-center"><img class="s-badge--image" src="{{ "/assets/img/BadgeBronze.svg" | relative_url }}" aria-hidden="true"> <span>11</span></div></td>
+                        <td>July 23 at 18:47</td>
+                        <td>Aug 10 at 14:57</td>
+                    </tr>
+                    <tr>
+                        <th><a>Piper Lawson</a></th>
+                        <td>New York City</td>
+                        <td>Design</td>
+                        <td><div class="grid ai-center"><div class="fw-bold mr8" title="reputation score " dir="ltr">25</div><div class="grid ai-center"><img class="s-badge--image" src="{{ "/assets/img/BadgeBronze.svg" | relative_url }}" aria-hidden="true"> <span>3</span></div></td>
```
aaronshekey: I'll switch these to `s-award`.

lib/css/components/_stacks-tables.less
paintedbicycle: Firefox mobile? Or just desktop shrunk down?

paintedbicycle: Ah nevermind I was looking in the wrong branch. I recreated.

There is actually another issue too - we need to set the bg colour (which I'm doing) but it's not dark mode respecting right now

aaronshekey: I switched the background color to `var(--white)` which makes that column appear darker in dark mode, since the rest of the table is actually transparent. This means we can't reliably choose something like `background-color: inherit`. We could be more explicit on the entire table (or each `td` and `th`) by setting their background colors `var(--white)`, but we'd still have the issue with the borders disappearing.

I wonder if instead of using borders, we actually just use box shadows 🤔

paintedbicycle: Yeah I can investigate. 

This too: https://stackoverflow.com/questions/266877/disappearing-css-table-cell-borders-in-gecko-based-browsers

lib/css/components/_stacks-tables.less
```diff
@@ -315,6 +314,23 @@
             opacity: 0.3;
         }
     }
+
+    //  $$  Responsive
+    //  ------------------------------------------------------------------------
+    &.s-table__responsive {
```
aaronshekey: When applying a background color to `s-table` I also discovered a side effect of applying `display: block` here... struggling to get the non overflowed table to take up 100% of its parent's width. Tables 😖

aaronshekey:
Closing in favor of #499

PR #498, opened 2020-06-07
s-post-summary

paintedbicycle:
First pass at a question (in list) component

Comes in 2 versions: 
1. The "slim" version seen on "home" page and profile pages
2. "Full" version seen in listing pages

Note: There is a "search" version that's not here. Do we need it or can search (long term) conform to one of these?

Note: Mobile isn't finished

Still needs quite a bit of tightening, just wanted to get it jump started


bnickel:
A few notes based on existing usage.

It would be good to have examples of embellishments that can occur on SO, i.e., the bounty indicator, the private indicator, and the "new content" indicator.

- The bounty indicator can be found on https://stackoverflow.com/questions?tab=Bounties. Right now we just use `float:left` on it, to inline it with wrapping, but it would be nice if we could have a formal pattern with `s-badge`.
- The private indicator is visible on any private team's question lists.  I believe it is `just display:inline-block`.
- The new content indicator is visible on private team question lists with new activity, and user question lists with new activity.  It is a red dot.  This one may best be served as its own class like `s-question__activity-indicator`.

These components can be highlighted a few ways:
- "Interesting", if a question has one of your watched tags, it gets a light yellow background.
- "ignored", if a question has one of your ignored tags, it has an opacity applied.
- "deleted", users which high enough rep can see deleted questions in search/question lists.  In this case, a red background is applied.

Search as it exists today would be fine using this and probably only needs minor changes to work.  

- Search currently used Q and A in the front of the title to indicate post type.  If we figure out something nifty for indicators, it would be cool to have this as an icon.
- Answers in search just show vote count and color it green if it's accepted.  This would be a fourth variant like `.s-question--is-accepted-answer` that just does color but not border.

And to bike shed the name, I feel like `s-question` has a few problems:

1. This means we can't use that class name for actual questions.
2. We may want to use this component for other posts like documentation articles, answers, etc.
3. The name "question" is really Q&A specific and shifts away from Stacks being an (at least in theory) general purpose framework.  We're definitely going to walk that tightrope as we bring more and more of SO's components into Stacks, but I think we can avoid that ATM.

I personally would call it something like `.s-article-listing` although that's fairly verbose.   In code today, we call it a "question summary" so `.s-post-summary` could be another choice.

paintedbicycle:
Thanks @bnickel great points, all this needs to be in there and I'll chip away at it.

The naming thing did come up for me too. I know @aaronshekey is thinking of `s-post` for the full (single) version, so maybe `s-post-summary` would work for this component.

paintedbicycle:
Closing in favo(u)r of https://github.com/StackExchange/Stacks/pull/547

PR #499, opened 2020-06-09
Table responsiveness

aaronshekey:
This PR explores switching to box shadows for table cell border rendering to allow for sticky responsive columns. Yeesh.

Seems there's a rendering issue where we lose table borders on tables wrapped in overflow, since the borders render outside the container as shadows.

image


paintedbicycle:
I really like this approach. It looks great across light and dark mode and I really don't see any issue with the last cell not having a border. Doesn't bother me.

Also great work with the "auto width" first sticky column. Solves my issue around sizing. 

Any reason this couldn't land? Looks pretty solid to me

PR #500, opened 2020-06-09
Change avatar component modifier classes to pixel values instead of t-shirt sizes

aaronshekey:
This was proving to be unwieldy in production, so let's just change these to their actual pixel values.

This will become `0.52.1`.


paintedbicycle (approved):
👍

b-kelly (approved):
lgtm! Did a quick project-wide find and looks like you got all instances of `s-avatar__`.

There are some specific changes we'll need to make to our core codebase to match up, but nothing extreme or blocking.

PR #506, opened 2020-07-01
Add moderator and admin icons to user card

paintedbicycle:
Screen Shot 2020-08-05 at 9 46 45 AM


lib/css/components/_stacks-user-cards.less
```diff
@@ -35,6 +35,20 @@
...
+.s-user-card__admin {
+    &::after {
+        content: "⬢";
+        padding-left: @su2;
+    }
+}
+
+.s-user-card__moderator {
+    &::after {
+        content: "♦";
```
gtsiolis: Although using just symbols to convey meaning for permissions (moderator, admin) could be difficult and these vary when they are rendered in difference browsers and operating systems, maybe it's worth considering using characters like  ▲ and △, WDTY?

| Before | After  |
| ------------- | ------------- |
| image | image  |

gtsiolis (commented):
Hi @paintedbicycle! I've added a small note to improve the icons (characters) used but I could be missing something like previous or relevant specs. 🏀

lib/css/components/_stacks-user-cards.less
paintedbicycle: Hi @gtsiolis! Thank you. This PR was to bring this component into parity with what exists currently on SO. I missed the admin + mod icons when I made the user cards. These symbols are from the current design. Having said that, I do agree they could be improved. 

However if I were to do that, I'd likely go more of a route of using a badge with a word to increase understanding for new users. So `Mod` and `Admin` beside the displayname instead of an icon at all. I'll have a think about that, but I may have to proceed with this for now so it can be used in current scenarios without disrupting other parts of the product. I'll ponder a bit. Thanks!

gtsiolis: @paintedbicycle —

> This PR was to bring this component into parity with what exists currently on SO. I missed the admin + mod icons when I made the user cards. These symbols are from the current design.

Indeed, I've just noticed the moderator symbol on SO but haven't seen the admin symbol anywhere.

> Having said that, I do agree they could be improved.

**Question:** Out of curiosity, do you have a specific workflow for discussing potential updates of design specifications for a component like this?

> However if I were to do that, I'd likely go more of a route of using a badge with a word to increase understanding for new users.

Agree! 💯 

>  I'll have a think about that, but I may have to proceed with this for now so it can be used in current scenarios without disrupting other parts of the product.

Sounds good. ⛳

paintedbicycle:
Revisiting this, I've change the mod and admin icons to badges and added a staff badge.

Screen Shot 2020-07-31 at 3 12 48 PM

paintedbicycle:
Cool, added a couple more examples, LGTM!

PR #508, opened 2020-07-14
Move things to devDependencies

aaronshekey:
These dependencies should only be necessary for development, so moving these to `devDependencies`. This should simplify what's installed further down the line.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,10 @@
     "approve": "backstop approve"
   },
   "license": "MIT",
+  "dependencies": {
+    "stimulus": "^1.1.1",
+    "@popperjs/core": "^2.4.0"
+  },
   "devDependencies": {
     "@11ty/eleventy": "^0.11.0",
     "@11ty/eleventy-plugin-syntaxhighlight": "^3.0.1",
@@ -37,9 +41,7 @@
     "grunt-contrib-watch": "^1.1.0",
     "grunt-shell": "^3.0.1",
     "grunt-ts": "^6.0.0-beta.22",
-    "@popperjs/core": "^2.4.0",
     "@stackoverflow/stacks-icons": "^2.8.0",
-    "stimulus": "^1.1.1",
     "typescript": "^3.9.5"
   }
 }
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "version": "0.52.1",
   "files": [
     "dist",
-    "lib/css/exports/*.less"
+    "lib/css"
   ],
   "unpkg": "dist/css/stacks.min.css",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -23,6 +23,8 @@
   "devDependencies": {
     "@11ty/eleventy": "^0.11.0",
     "@11ty/eleventy-plugin-syntaxhighlight": "^3.0.1",
+    "backstopjs": "^5.0.1",
+    "eleventy-plugin-nesting-toc": "^1.2.0",
     "grunt": "^1.1.0",
     "grunt-cli": "^1.3.2",
     "grunt-concurrent": "^3.0.0",
@@ -35,13 +37,9 @@
     "grunt-contrib-watch": "^1.1.0",
     "grunt-shell": "^3.0.1",
     "grunt-ts": "^6.0.0-beta.22",
-    "typescript": "^3.9.5"
-  },
-  "dependencies": {
     "@popperjs/core": "^2.4.0",
-    "@stackoverflow/stacks-icons": "^2.6.0",
-    "backstopjs": "^5.0.1",
-    "eleventy-plugin-nesting-toc": "^1.2.0",
-    "stimulus": "^1.1.1"
+    "@stackoverflow/stacks-icons": "^2.4.0",
+    "stimulus": "^1.1.1",
+    "typescript": "^3.9.5"
   }
 }
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "version": "0.52.1",
   "files": [
     "dist",
-    "lib/css"
+    "lib"
   ],
   "unpkg": "dist/css/stacks.min.css",
   "scripts": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -301,8 +301,7 @@
     "@popperjs/core": {
       "version": "2.4.0",
       "resolved": "https://registry.npmjs.org/@popperjs/core/-/core-2.4.0.tgz",
-      "integrity": "sha512-NMrDy6EWh9TPdSRiHmHH2ye1v5U0gBD7pRYwSwJvomx7Bm4GG04vu63dYiVzebLOx2obPpJugew06xVP0Nk7hA==",
-      "dev": true
+      "integrity": "sha512-NMrDy6EWh9TPdSRiHmHH2ye1v5U0gBD7pRYwSwJvomx7Bm4GG04vu63dYiVzebLOx2obPpJugew06xVP0Nk7hA=="
     },
     "@stackoverflow/stacks-icons": {
       "version": "2.8.0",
@@ -314,7 +313,6 @@
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/@stimulus/core/-/core-1.1.1.tgz",
       "integrity": "sha512-PVJv7IpuQx0MVPCBblXc6O2zbCmU8dlxXNH4bC9KK6LsvGaE+PCXXrXQfXUwAsse1/CmRu/iQG7Ov58himjiGg==",
-      "dev": true,
       "requires": {
         "@stimulus/mutation-observers": "^1.1.1"
       }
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
   "version": "0.52.1",
   "files": [
     "dist",
-    "lib/css/exports/*.less"
+    "lib/css"
```
b-kelly: This will export all files under `lib/css`. I'm ok with this in general (over-exporting is less of an issue than under-exporting), but we have to keep in mind that changing exported apis is technically a breaking change for Semver.

Not so much an issue for us, but just mentioning it so we can think about it pre-1.0 launch.

package.json
```diff
@@ -35,13 +37,9 @@
...
-    "typescript": "^3.9.5"
-  },
-  "dependencies": {
     "@popperjs/core": "^2.4.0",
     "@stackoverflow/stacks-icons": "^2.8.0",
-    "backstopjs": "^5.0.1",
-    "eleventy-plugin-nesting-toc": "^1.2.0",
-    "stimulus": "^1.1.1"
+    "stimulus": "^1.1.1",
+    "typescript": "^3.9.5"
```
b-kelly, line 45: `typescript` needs to be a `devDependency` instead

b-kelly (changes requested):
Just need to change `typescript` to be a `devDependency` and this will be good to go.

package.json
aaronshekey: Got it. I want those Less files to be available to Core and Careers if we deliver these things via `npm` there. :v:

package.json
aaronshekey: Everything is now a devDependency. The diff looks a little weird but there are no `dependencies` anymore.

package.json
b-kelly: Sounds good to me. We might want to rethink how we export singular mixins, etc. I remember once wanting to use the button mixing and changing some colors around for some custom marketing pages, but was completely unable to do so because of some internal weirdness in our dependency chain.

Not a problem for right now, but closer to a 1.0 launch.

package.json
b-kelly: Oh, yeah.... that's correct. We want to instead move anything that gets included in the final build into `dependencies`. So, popperjs, icons, stimulus, etc all need to be dependencies.

aaronshekey: Why? Popper, Stimulus, etc. are all compiled into JS, right? Anything consuming Stacks will only be using the compiled JS.

aaronshekey: As I understand it, you only need those dependencies when developing _for_ Stacks, not consuming it.

b-kelly: It's the opposite logic. There's two major ways to consume a package:

1. using the fully compiled script in your browser
    * dependencies don't matter at all because, hey, it's all compiled for you
2. importing into another project for use
    * `devDependencies` says "these are for developing ON Stacks"
    * `dependencies` says "these are needed to USE Stacks"
    * Anything missing from `dependencies` will not get included into the user's project and Stacks will fail to compile

Think of it as: "will Stacks work without this package?" and anything answered **no** will go into `dependencies`.

b-kelly: So, I'm actually partially wrong: `stacks-icons` is not actually used in Stacks, but *only* in the docs, so it would go into `devDependencies`. Missed that one when writing them out.

aaronshekey: I think I'm missing something with our JS side of things. We don't deliver JS source, only compiled `stacks.js` and that's by design. We can't compile JS even if we include Popper or Stimulus, because we don't include JS source files.

https://unpkg.com/browse/@stackoverflow/stacks@0.52.1/

I'm happy with the path forward being "Here's compiled everything, CSS and JS. Also here are Less files if you need to compile the CSS yourself :v:" This redesigned `package.json` to reflects this.

b-kelly: I see your point, but this is the "correct" way of doing things. It gives a clean separation of "what goes into the bundle and what doesn't". Other than my listed reasons, there are many other reasons why this is a good idea, including (but not limited to): 

* security
  * immediately know what we're shipping to consumers (notably, stackoverflow.com!)
* code debt / cleanliness
  * same as security, but so we ship less obsolete bytes
* typings
  * allows consumers to use the typescript types from our library
* future proofing / flexibility
  * include the whole library, but bundle it inside another larger bundle
  * possibly split out common dependencies so you don't ship them twice
  * eventually, someone will want to import single components

I can probably come up with some more if you press me hard enough 😛

aaronshekey: "This" meaning, including all JS source files and adding `Popper` and `Stimulus` as `Dependencies`?

b-kelly: We don't need to include all JS source files at the moment (it would be nice to have them later), but you will need to:

* move both stimulus and popper into dependencies
* run `npm i` to update `package.lock`

package.json
```diff
@@ -8,7 +8,7 @@
   "version": "0.52.1",
   "files": [
     "dist",
-    "lib/css/exports/*.less"
+    "lib"
```
b-kelly, line 11: 👍 This change looks ok to me. I ran `npm publish --dry-run` to see what was included and it all seems on the ball. Pasting here in case anyone is curious:

<details>
<summary>npm publish --dry-run results</summary>

```
> npm publish --dry-run

package: @stackoverflow/stacks@0.52.1
=== Tarball Contents ===
274.8kB dist/css/stacks.css
226.6kB dist/css/stacks.min.css
166.4kB dist/js/stacks.js
72.7kB  dist/js/stacks.min.js
1.5kB   package.json
364B    lib/tsconfig.json
4.8kB   lib/css/components/_stacks-avatars.less
3.9kB   lib/css/components/_stacks-badges.less
2.6kB   lib/css/components/_stacks-banners.less
707B    lib/css/components/_stacks-blank-states.less   
1.1kB   lib/css/base/_stacks-body.less
10.3kB  lib/css/atomic/_stacks-borders.less
908B    lib/css/components/_stacks-breadcrumbs.less
2.7kB   lib/css/components/_stacks-button-groups.less
17.4kB  lib/css/components/_stacks-buttons.less
1.3kB   lib/css/components/_stacks-cards.less
5.8kB   lib/css/components/_stacks-collapsible.less
4.3kB   lib/css/atomic/_stacks-colors.less
9.1kB   lib/css/base/_stacks-configuration-dynamic.less
3.1kB   lib/css/base/_stacks-configuration-static.less
10.8kB  lib/css/exports/_stacks-constants-colors.less
5.1kB   lib/css/exports/_stacks-constants-helpers.less
2.6kB   lib/css/exports/_stacks-constants-type.less    
679B    lib/css/exports/_stacks-exports.less
7.3kB   lib/css/atomic/_stacks-flex.less
9.9kB   lib/css/atomic/_stacks-grid.less
551B    lib/css/base/_stacks-icons.less
17.8kB  lib/css/components/_stacks-inputs.less
10.4kB  lib/css/base/_stacks-internals.less
3.6kB   lib/css/components/_stacks-link-previews.less
3.9kB   lib/css/components/_stacks-links.less
14.4kB  lib/css/atomic/_stacks-misc.less
4.0kB   lib/css/exports/_stacks-mixins.less
4.1kB   lib/css/components/_stacks-modals.less
1.9kB   lib/css/components/_stacks-navigation.less
4.8kB   lib/css/components/_stacks-notices.less
1.5kB   lib/css/components/_stacks-page-titles.less
1.5kB   lib/css/components/_stacks-pagination.less
8.0kB   lib/css/components/_stacks-popovers.less
5.5kB   lib/css/components/_stacks-progress-bars.less
1.6kB   lib/css/base/_stacks-reset-meyer.less
8.2kB   lib/css/base/_stacks-reset-normalize.less
468B    lib/css/base/_stacks-reset.less
6.5kB   lib/css/atomic/_stacks-spacing.less
3.4kB   lib/css/components/_stacks-spinner.less
9.0kB   lib/css/components/_stacks-tables.less
4.9kB   lib/css/components/_stacks-tags.less
2.6kB   lib/css/components/_stacks-toggle-switches.less
8.0kB   lib/css/atomic/_stacks-typography.less
2.4kB   lib/css/components/_stacks-user-cards.less
1.1kB   lib/css/components/_stacks-widget-dynamic.less
9.7kB   lib/css/components/_stacks-widget-static.less
10.2kB  lib/css/atomic/_stacks-width-height.less
1.3kB   lib/css/stacks-dynamic.less
2.1kB   lib/css/stacks-static.less
463B    lib/css/stacks.less
1.1kB   LICENSE.MD
2.6kB   README.md
31B     lib/ts/finalize.ts
7.0kB   lib/ts/controllers/s-expandable-control.ts
12.0kB  lib/ts/controllers/s-modal.ts
11.3kB  lib/ts/controllers/s-popover.ts
10.8kB  lib/ts/controllers/s-table.ts
8.4kB   lib/ts/controllers/s-tooltip.ts
816B    dist/js/stacks.d.ts
3.7kB   lib/ts/stacks.ts
222B    lib/ts/stimulus.d.ts
=== Tarball Details ===
name:          @stackoverflow/stacks
version:       0.52.1
package size:  183.4 kB
unpacked size: 1.1 MB
shasum:        097fd088d87e4af138df160dad3054a6daa429c2
integrity:     sha512-HW5qHdEcaqSUX[...]N0CeMEO4BcE+A==
total files:   67

+ @stackoverflow/stacks@0.52.1
```

</details>

No action needed.

b-kelly (approved):
lgtm! Thanks for making the changes for me. I'll try importing the Stacks' `ts` files in a project later and see what's missing from that (officially unsupported) scenario. We'll probably need to add a `main` and a `types` property to `package.json`, but I'm not worried about that for this PR.

I've attached the output of `npm publish --dry-run` in my comment below. Everything included looks ok to me, but I'd recommend taking a second look to have another pair of eyes on it.

PR #511, opened 2020-07-21
Add official browserslist to package.json

b-kelly:
Currently, this has absolutely no effect on building/bundling/anything, but it would be nice to get an official "source of truth" / place to point to when asked what browsers we support.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,5 +44,21 @@
     "grunt-ts": "^6.0.0-beta.22",
     "@stackoverflow/stacks-icons": "^2.8.0",
     "typescript": "^3.9.5"
-  }
+  },
+  "browserslist": [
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -44,5 +44,21 @@
...
+    "not baidu > 0",
+    "not and_ff > 0",
+    "not and_qq > 0",
+    "not kaios > 0",
+    "not op_mob > 0",
+    "not ie_mob > 0",
+    "not and_uc > 0",
+    "not Samsung > 0",
+    "not Android > 0",
+    "edge 18"
```
aaronshekey, line 62: I'm coming for you, Edge 18. You better watch yourself.

aaronshekey (approved):

PR #512, opened 2020-07-21
Use word-break: break-all instead of break-word

paintedbicycle:
**For consideration.**

`Break-all` will break at any character, while `break-word` only breaks on words. Ideally, we'd use `break-word`. However there are situations where `break-word` won't trigger - i.e. when using `nbsp;` between words. We recently ran into this situation with cut-and-pasted copy into the editor

### DIFF
--- a/lib/css/components/_stacks-prose.less
+++ b/lib/css/components/_stacks-prose.less
@@ -14,7 +14,7 @@
     // Base styling
     font-size: @fs-body2;
     line-height: @lh-lg;
-    word-wrap: break-word;
+    word-break: break-all;
 
     pre {
         word-wrap: normal;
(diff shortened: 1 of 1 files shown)


aaronshekey (changes requested):
Let's ship this without the automatic prettifyin'

paintedbicycle:
Oops, sorry, I'll redo

paintedbicycle:
Reverted and fixed. I'm interested in testing this against your sandbox to see if it breaks inside too many words and how it feels

aaronshekey:
In that case, just ~throw a bunch of trash into~ modify any of those `.md` files within `docs/_includes/`

lib/css/components/_stacks-prose.less
```diff
@@ -14,7 +14,7 @@
     // Base styling
     font-size: @fs-body2;
     line-height: @lh-lg;
-    word-wrap: break-word;
+    word-wrap: break-all;
```
aaronshekey, line 17: This could end up having some weird ramifications by breaking words we don't want to be broken, but you're way closer to the problem than I am. Poke around at it, and if you feel good, feel free to merge.

aaronshekey (approved):

lib/css/components/_stacks-prose.less
paintedbicycle: Yep, agreed. I'm playing around in the branch preview now

paintedbicycle: Already seeing stuff I don't like

paintedbicycle:
I think we need another approach

PR #515, opened 2020-07-22
Allow removing borders responsively

aaronshekey:
@abovedave This should handle your use in production, right?


abovedave (approved):
Yep that'll do it, thank you @aaronshekey

PR #518, opened 2020-07-28
Bump dependencies

aaronshekey:
This bumps all our dependencies. The biggest change is Popper `v2.4.4` in our library. This will require a Stacks version bump.

Tested locally, and everything seems to be working as designed.


b-kelly (approved):
lgtm! Popper changes are only bugfixes, nothing that will affect us. For devDependencies... if it builds, it builds.

PR #521, opened 2020-08-05
Dynamically import icons/spots during docs build step

b-kelly:
## What?

This PR removes all "manual" (via grunt) icon/spot file copying and instead uses the values exported from the `@stackoverflow/stacks-icons` package.

## Why?

I got tired of always forgetting to run `update-icons` after a fresh pull and having my build crash every time. 😛

## Major changes

* .svg files and .json files are no longer copied from `node_modules` into the project
* committed .json data files eleventyIcons and eleventySpots have been removed (and can be removed from stacks-icons upstream)
* the docs will now use whatever version of stacks-icons you have installed and will no longer crash due to versioning mismatches between the icons and the json files


aaronshekey (approved):
You know I love this. What an improvement, and a dumb chore. Thank you for taking care of this!

PR #522, opened 2020-08-06
Grid rename

aaronshekey:
As discussed today, here's a single PR that simply renames our grid cells to `flex`, allowing us to do the rename before introducing new CSS grid classes.


aaronshekey:
This is the approach, but we can do better in a fresh PR.

PR #524, opened 2020-08-12
Add a display delay to the s-popover__tooltip component

aalear:


aalear:
@aaronshekey I played around with CSS transition properties for a while here for various combinations of opacity, visibility, and height. But IMO ultimately none of it worked out too great and losing the DOM behavior we have with `display` is kind of a big deal. So I took a stab at a Typescript-based approach instead. This seems to be working pretty well, but assuming you're cool with it in the first place, should probably get @b-kelly's eyes as well.

bnickel:
I wouldn't put this in the base controller because it muddies things up (base controller knowing about a specific subclass).

Since `s-tooltip` is already responsible for binding mouseover and mouseout events, we could make the changes in `TooltipController` instead:

1. Add a method called `scheduleShow` which does what you're suggesting, `setTimeout(() => this.show(dispatcher), 300)`
2. Overriding `hide` to cancel the timeout and then just call `super.hide()`
3. Updating [`bindMouseEvents`](https://github.com/StackExchange/Stacks/blob/ad0091d42802afa03d5686317917eff748fd8653/lib/ts/controllers/s-tooltip.ts#L154) and `unbindMouseEvents` to use `boundScheduleShow` instead of `boundShow`.

bnickel:
I'm also glad to see the typescript approach here even though its less fancy.  `visibility: hidden` was historically the source of a lot of drama with popovers.

bnickel (approved):
This looks really good and the delay feels natural on the docs page.

aaronshekey (approved):
I still have mixed feelings about this, but I'm glad this solution is consistent. Thank you for jumping in on this, @aalear!

@b-kelly One last sanity check? I think `300ms` is feeling pretty good.

lib/ts/controllers/s-tooltip.ts
```diff
@@ -50,6 +51,23 @@ namespace Stacks {
...
+         */
+        scheduleShow(dispatcher: Event | Element | null = null) {
+            this.activeTimeout = setTimeout(() => this.show(dispatcher), 300);
+        }
+
+        /**
+         * Cancels the scheduled tooltip popover display and hides it if already displayed
+         */
+        hide(dispatcher: Event | Element | null = null) {
+            clearTimeout(this.activeTimeout);
```
b-kelly: Sanity checking: [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/clearTimeout#Notes) states that it is 100% ok to pass invalid ids here (such as `null` if `hide` is called before `scheduleShow`).

Optional: Personal preference. Commit if desired.

```suggestion
            window.clearTimeout(this.activeTimeout);
```

lib/ts/controllers/s-tooltip.ts
```diff
@@ -50,6 +51,23 @@ namespace Stacks {
             super.show(dispatcher);
         }
 
+        /**
+         * Sets up a tooltip popover show after a delay.
+         */
+        scheduleShow(dispatcher: Event | Element | null = null) {
+            this.activeTimeout = setTimeout(() => this.show(dispatcher), 300);
```
b-kelly: Recommended to clear an existing `activeTimeout` on call. That way, if this is somehow called again before the delay is finished, we won't get multiple `show` calls (not that it really matters in the end, since it is programmed to return early if shown).

Optional: Personal preference. Commit if desired.

```suggestion
            window.clearTimeout(this.activeTimeout);
            this.activeTimeout = window.setTimeout(() => this.show(dispatcher), 300);
```

b-kelly (approved):
lgtm! Nice and simple. Added two optional code suggestions to prepend `window` onto your bare global function calls. No specific technical reason, mostly just personal preference.

Thanks for submitting a PR for this.

aalear:
Thanks for the reviews, everyone! All the tweaks and suggestions have been applied, so I believe this is ready to roll at your convenience, @aaronshekey.

aaronshekey:
Pew pew!

PR #525, opened 2020-08-17
Tweak s-prose styles

b-kelly:
A PR in haiku:

```
To align with prod
tweak the styles to match what's there
(for compat reasons)
```

```
Headlines shift one size
instead of the two we had
for lower impact
```

```
Change keyboard margins
back to their prior value.
Not sure on this one
```

```
Reduce the padding
of lists within other lists
since they look too tall
```

```
When they are empty
a spoiler's height will collapse,
so add a min-height
```

### DIFF
--- a/lib/css/components/_stacks-prose.less
+++ b/lib/css/components/_stacks-prose.less
@@ -66,34 +66,34 @@
 
     h1,
     h1 code {
-        font-size: @fs-headline2;
+        font-size: @fs-headline1;
         margin-bottom: .2em;
     }
 
     h2,
     h2 code {
-        font-size: @fs-headline1;
+        font-size: @fs-title;
         margin-bottom: .3em;
     }
 
     h3,
     h3 code {
-        font-size: @fs-title;
+        font-size: @fs-subheading;
     }
 
     h4,
     h4 code {
-        font-size: @fs-subheading;
+        font-size: @fs-body3;
     }
     
     h5,
     h5 code {
-        font-size: @fs-body3;
+        font-size: @fs-body2;
     }
     
     h6,
     h6 code {
-        font-size: @fs-body2;
+        font-size: @fs-body1;
     }
 
     // ============================================================================
@@ -167,6 +167,7 @@
         > ul,
         > ol {
             padding-top: 0.5em;
+            margin-bottom: 0;
         }
     }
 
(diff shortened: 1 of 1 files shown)


lib/css/components/_stacks-prose.less
```diff
@@ -66,34 +66,34 @@
...
     h5,
     h5 code {
-        font-size: @fs-body3;
+        font-size: @fs-body2;
     }
     
     h6,
     h6 code {
-        font-size: @fs-body2;
+        font-size: @fs-body1;
```
aaronshekey, line 96: The thing that annoys me about this is that this header is now smaller than our body text 🤔

lib/css/components/_stacks-prose.less
```diff
@@ -280,6 +281,7 @@
         border-radius: @br-md;
         color: var(--black-800);
         cursor: pointer;
+        min-height: 48px;
```
aaronshekey:
```suggestion
        min-height: @su48; // TODO: Let's find a solution that doesn't have a magic number
```

Dang ol' magic numbers. Don't have the energy to look into a way of avoiding this right now, ship it.

lib/css/components/_stacks-prose.less
```diff
@@ -325,7 +327,7 @@
     kbd {
         display: inline-block;
         margin: 0 .1em;
-        padding: .1em .4em;
+        padding: .1em .6em;
```
aaronshekey, line 330: 👍

lib/css/components/_stacks-prose.less
```diff
@@ -167,6 +167,7 @@
         > ul,
         > ol {
             padding-top: 0.5em;
+            margin-bottom: 0;
```
aaronshekey, line 170: Looks good in our sandbox. 👍

aaronshekey (approved):

lib/css/components/_stacks-prose.less
b-kelly: Agreed. Should we set to the same size as `h5` or just not worry about it? `h6` isn't really supported on prod anyways, so I'm not *that* worried about the styles for it.

aaronshekey: Something to revisit, for sure. Not holding up the show. Just want to make the point that this could consume any markdown source, and not _just_ markdown from our editor, ya know?

PR #526, opened 2020-08-18
Add small variation to s-prose

aaronshekey:
image


aaronshekey:
Lots of conversation was had via Slack on this one. Sorry for anyone following along 😛

PR #527, opened 2020-08-19
Increasing contrast in sidebar widget header

PiperLawson:
Super simple. Just trying to increase sidebar header copy color from 500 to 600. I think it makes quite a difference in readability.


aaronshekey:
Total no-brainer, thanks!

aaronshekey (approved):

PR #528, opened 2020-08-20
Bump dependencies

aaronshekey:
Looks like Typescript compiles a little differently here. Sanity check?


dist/js/stacks.js
```diff
@@ -4570,11 +4570,12 @@ var Stacks;
         TooltipController.prototype.scheduleShow = function (dispatcher) {
             var _this = this;
             if (dispatcher === void 0) { dispatcher = null; }
-            this.activeTimeout = setTimeout(function () { return _this.show(dispatcher); }, 300);
+            window.clearTimeout(this.activeTimeout);
```
b-kelly, line 4573: This code is "all new". Looks like the tooltip PR got merged with without the latest version of the compiled code. Not an issue since the added code is "optional", but it's something we'll need to make sure to check in the future.

b-kelly (approved):
lgtm! The changes in the compiled JS from the new TS update are equivalent to what was there before.

b-kelly:
Forgot to mention: Looking at the changelog for TS 4.0, none of the [breaking changes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#breaking-changes) affect us.

PR #529, opened 2020-08-21
Updating time stamp type color and rep color on minimal card

PiperLawson:
Just trying for some better contrast in the user card. Minimal as well since it's being used on new answer UI


aaronshekey (approved):
Punch it!

PR #530, opened 2020-08-25
Add company email signatures

abovedave:
Add the option for our more client-facing colleagues to be consistent.


aaronshekey (approved):
I removed some unintentional changes and made sure things were looking right in dark mode.

PR #531, opened 2020-08-26
List nesting refinements

aaronshekey:
This adds a bit of refinement to the following test case:

```
* List item

  List paragraph
* List item

  List paragraph
  * Subitem
  * Subitem

  List paragraph
```

I also snuck some last minute header margin refinements since we changed some font sizes at the last minute.


b-kelly (approved):
Gave it a quick visual pass. lgtm!

PR #535, opened 2020-08-28
s-code-block style tweaks

b-kelly:
Just a few styling tweaks for `s-code-block`:

* remove `max-height` since vertical scrolling is not desired in prod
  * re-added it for Stacks documentation *only*
* fix `s-prose` accidentally adding the inline `code` styles to our code blocks
  * docs do not exhibit this behavior because the syntax highlighting plugin (that I wrote...) adds the `s-code-block` class to *both* the `pre` and `code` elements, while prod only adds it to the `pre` element


aaronshekey (approved):
Nice catch on both of these!

b-kelly:
Well, looks like these made their way into dev so... closing 🎉

aaronshekey:
I still love it!

b-kelly:
Reopened this PR because apparently I'm not firing on all cylinders today.

PR #536, opened 2020-08-31
Tweak link-preview code styling

b-kelly:
Tweaks link-previews to use `s-code-block` in all the docs / prod. Also, added a linenumbers plugin that is very similar to what we'll be using in prod. Requires some slightly funky syntax to enable, but it's the best I can do.

Also, tweaked s-code-block padding to match that used in `s-prose pre` so we don't get style flickers on highlight.

Creating as a draft since it will likely require a few tweaks (and some testing in Core).

### DIFF
--- a/lib/css/components/_stacks-code-blocks.less
+++ b/lib/css/components/_stacks-code-blocks.less
@@ -99,3 +99,18 @@ code[class*="language-"] {
         font-weight: bold;
     }
 }
+
+pre.s-code-block .s-code-block--line-numbers {
+    float: left;
+    color: var(--black-300);
+    text-align: right;
+    border-width: 0;
+    border-style: solid;
+    border-color: var(--black-100);
+    border-right-width: 1px;
+    margin: -12px;
+    margin-right: 12px;
+    padding: 12px;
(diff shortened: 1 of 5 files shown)


lib/css/components/_stacks-code-blocks.less
```diff
@@ -99,3 +99,18 @@ code[class*="language-"] {
         font-weight: bold;
     }
 }
+
+pre.s-code-block .s-code-block--line-numbers {
+    float: left;
```
aaronshekey, line 104: Did we want to avoid floating here? Probably more appropriate to punt this to later.

b-kelly: Yeah, most definitely, but I need to do a little exploration on how to get this to work on both our docs (where the block itself is not accessible at render time to check for `linenums` class name) and in prod where we highlight a block directly (and can check its classes).

aaronshekey (approved):

b-kelly:
@aaronshekey Is this good to go or are you still tweaking? Go ahead and merge when you're satisfied with the result. No need to wait on me.

PR #537, opened 2020-08-31
Add pointer events to spoilers

aaronshekey:
Addresses https://meta.stackexchange.com/questions/353670/reveal-spoiler-label-blocks-mouse-interaction-after-revealing-the-spoiler


b-kelly:
I think you may have misunderstood the meta post. The issue is that the label is sitting over the content (invisibly) and is blocking clicks to items underneath it. Non-`is-visible` spoiler content already doesn't respond to click events because it is set to `visibility: hidden`. I've committed a fix to this branch for the underlying meta issue and tested locally.

aaronshekey:
Honestly, when have I ever understood anything? 😛

PR #539, opened 2020-09-04
Popover automatic arrow placement

aaronshekey:
This PR leans more heavily on Popper to position the arrows within our popovers. This maintains the manually positioned arrows, and documents adding `data-popper-arrow`. Practically, this means our arrows are way less likely to mis-aligned as popovers shift around to accommodate the viewport.

We'll need to add `data-popper-arrow` to each instance of `s-popover--arrow` in production for Popper to manage the placement.


b-kelly (commented):
I'll take another look this weekend, but it looks like we can set which element is the arrow without needing to go add `data-popper-arrow` to all our popovers:

https://popper.js.org/docs/v2/modifiers/arrow/

b-kelly (commented):
Just committed a change to this branch that uses our `.s-popover--arrow` selector for the arrow so we don't have to add the `data` attribute to all our existing popovers

aaronshekey:
Beautiful!

aaronshekey:
This has been tested in production and has also passed visual regression tests :v:

PR #541, opened 2020-09-09
Fix user-select inconsistency with buttons and links

### LINKED ISSUE #540: More consistent button vs link selectability
**Describe the bug**
Stacks provides support for links and button components. However, not every Stacks button is a HTML `<button>` and not every Stacks link is a HTML `<a href="">`.

Now, this causes inconsistency issues in Firefox. When I have a HTML `<a>`, the element's text can be selected by the user, too. When I have a HTML `<button>`, the element's text cannot be selected by the user. However, this behavior doesn't neccessarily match the styling provided by Stacks.

For example, in the post action button list, the new "follow" link cannot be selected, while all other elements can:

image showing the aforementioned bug of most but one post actions being user selectable

While the use of a `<button>` element is likely correct here, it would be better if the whole thing were consistent.

Furthermore, things that look like buttons are actually links (for example the "Ask Question" btn) and should behave like such.

image showing the ask question button being selectable.

This should be consistency-fied. I propose to make

* Stack buttons, that look like buttons, to be `user-select: none`, and
* Stack links, that look like links, to be `user-select: all`.

Buttons that look like buttons are IMHO everything with `.s-btn`, **except** for those with `.s-btn__unset` or `.s-btn__link`. Links that look like links are all links without `.s-btn`.

**Reproduced on desktop Firefox**
 - OS: Windows 10
 - Browser firefox
 - Version 80.0.1 (latest)

**Not an issue on desktop Chrome**

luap42:
Closes #540 by adding appropriate user-select parameters to buttons and links to cause consistent behavior:

* buttons **cannot** be selected,
* links **can** be selected.

This is my first PR here, so if I have forgotten something, please tell me; especially when I need to build the LESS foöes (couldn't find a build product, hence I didn't try). :)

### DIFF
--- a/lib/css/components/_stacks-links.less
+++ b/lib/css/components/_stacks-links.less
@@ -20,6 +20,7 @@ a,
     text-decoration: none;
     color: @link-color;
     cursor: pointer;
+    user-select: auto;
 
     &.s-link__visited:visited {
         color: @link-color-visited;
@@ -101,6 +102,7 @@ button.s-link {
     border: 0;
     padding: 0;
     line-height: inherit;
+    user-select: auto;
 
     &:focus {
         outline: none;
--- a/lib/css/components/_stacks-buttons.less
+++ b/lib/css/components/_stacks-buttons.less
@@ -540,7 +540,7 @@
     background: none;
     box-shadow: none;
     cursor: default;
-    user-select: all;
+    user-select: auto;
 }
 
 .s-btn__unset:focus {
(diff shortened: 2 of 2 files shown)


lib/css/components/_stacks-links.less
```diff
@@ -20,6 +20,7 @@ a,
     text-decoration: none;
     color: @link-color;
     cursor: pointer;
+    user-select: auto;
```
aaronshekey, line 23: I don't think we _need_ this, but it's not hurting anything.

aaronshekey (approved):
Thanks for this! Sorry it took so long to review!

PR #543, opened 2020-09-16
Add full borders classes

aaronshekey:
Let's add the rest of the border color classes and see what it'll cost us in bytes. cc: @benjamin-hodgson 

This deprecates the `bc-black-3` naming scheme and keeps it consistent with other classes `bc-black-100`.

### DIFF
--- a/docs/product/base/borders.html
+++ b/docs/product/base/borders.html
@@ -197,29 +197,39 @@ description: Stacks provides atomic classes for borders.
     {% header "h3", "Black" %}
     <div class="stacks-preview">
 {% highlight html %}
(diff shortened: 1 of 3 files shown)


benjamin-hodgson (approved):
Didn't wanna codegen these?

I wouldn't expect there to be much impact on bundle size post-zipping.

aaronshekey:
> Didn't wanna codegen these?

Nahhh, I feel we generally get too optimized on that side. We don't need to get too clever. This allows me to keep everything super searchable and obvious. I'd rather just see it, you know?

PR #548, opened 2020-10-16
Add 1 pixel position coordinate classes

aaronshekey:
Sometimes you need to move things by a single pixel and we should allow that.


lib/css/atomic/_stacks-spacing.less
```diff
@@ -140,23 +140,23 @@
...
     '.r', { .template(@value) { right: @value !important; } },
-    0 @su2 -@su2 @su4 -@su4 @su6 -@su6 @su8 -@su8 @su12 -@su12 @su16 -@su16 @su24 -@su24 @su32 -@su32 @su48 -@su48 @su64 -@su64 @su96 -@su96 @su128 -@su128 50% -50% 100% -100%
+    0 1px @su2 -1px -@su2 @su4 -@su4 @su6 -@su6 @su8 -@su8 @su12 -@su12 @su16 -@su16 @su24 -@su24 @su32 -@su32 @su48 -@su48 @su64 -@su64 @su96 -@su96 @su128 -@su128 50% -50% 100% -100%
 );
 
 //  $$ Bottom
 #stacks-internals #build-classes(
     '.b', { .template(@value) { bottom: @value !important; } },
-    0 @su2 -@su2 @su4 -@su4 @su6 -@su6 @su8 -@su8 @su12 -@su12 @su16 -@su16 @su24 -@su24 @su32 -@su32 @su48 -@su48 @su64 -@su64 @su96 -@su96 @su128 -@su128 50% -50% 100% -100%
+    0 1px @su2 -1px -@su2 @su4 -@su4 @su6 -@su6 @su8 -@su8 @su12 -@su12 @su16 -@su16 @su24 -@su24 @su32 -@su32 @su48 -@su48 @su64 -@su64 @su96 -@su96 @su128 -@su128 50% -50% 100% -100%
```
b-kelly: nit: You've got 1 and -1 on either side of the `@su2` instead of next to each other. No effect on the output, but it bugs me :trollface:

aaronshekey: hahah you're 100% right.

PR #551, opened 2020-11-10
Highlight.js update + new theme classes

b-kelly:
This is a draft PR exploring how we might take advantage of the new classes / syntax fixes that the most recent version of highlight.js has added.

Something I'd like to explore in this PR are some possible changes to the LaTeX theme. The tex community was not terribly happy with the new syntax highlighting/theme, so they took it upon themselves to contribute better highlighting detection upstream. I'd love to support these new classes so their hard work doesn't go to waste. They have a post [here](https://tex.meta.stackexchange.com/questions/8688/adapting-the-highlighting-colors-for-the-new-highlight-js-version) with suggestions on how to incorporate some of the changes into a new theme.

### Changes

I've explicitly added highlight.js as a devDependency so we can tell what version our styles are being built against. I've also added a few of the unused classes to the `_stacks-code-block` with a TODO comment on them, as well as a LaTeX code example to the docs.

### DIFF
--- a/docs/product/components/code-blocks.html
+++ b/docs/product/components/code-blocks.html
@@ -350,6 +350,54 @@ namespace MyApplication
         </div>
     </div>
 
+    {% header "h3", "LaTeX" %}
+    <div class="stacks-preview">
+{% highlight html %}
+<pre class="s-code-block language-csharp">
+    …
+</pre>
+{% endhighlight %}
+        <div class="stacks-preview--example">
+{% highlight latex %}
+% !TeX encoding = utf8
+\documentclass{article}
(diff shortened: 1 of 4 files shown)


aaronshekey:
Awesome! Excited to dig into the new classes. I think this all looks super good, and TeX is improved out of the box already.

schtandard:
Is this still being pursued? We (meaning the TeX.sx community) might have gone quiet about this but I don't think it's because we don't want our old color scheme back but rather out of resignation. If you are working on something and need feedback, I'm sure we'll be happy to assist.

PR #554, opened 2020-11-20
Add aria-label to icon-only breadcrumb

bnickel:
The icon-only variant needs an aria-label on it's link.

> *Aside*: I am floored at how closely the design here conforms to the [wai recommendations](https://www.w3.org/TR/wai-aria-practices-1.1/examples/breadcrumb/index.html). :100:


aaronshekey (approved):
Looks good!

PR #557, opened 2020-12-01
Fix the s-notice documentation

bnickel:
The accessibility documentation around s-notice was largely incorrect and based on assumptions made when `s-notice` and `s-toast` were the same thing. 

This makes things more consistent with the reality of how it's used and what does and does not need to be included.  I do thing there's more to be done in refining the component, given the number of overrides being used in the toast example, but that's outside of the scope here.

I did this all in GItHub so let's see if it event deploys. :)

### DIFF
--- a/docs/_data/notices.json
+++ b/docs/_data/notices.json
@@ -53,27 +53,21 @@
   "access": [
     {
       "item": "aria-describedby=\"[id]\"",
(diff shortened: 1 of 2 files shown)


aaronshekey (approved):
👏

PR #559, opened 2020-12-03
add s-list__horizontal and docs

benjamin-hodgson:
Screenshot 2020-12-03 at 3 12 00 PM

### DIFF
--- a/lib/css/components/_stacks-lists.less
+++ b/lib/css/components/_stacks-lists.less
@@ -21,8 +21,8 @@
     display: flex;
     flex-wrap: wrap;
 
-    > li:not(:last-child)::after {
         content: '\2022';  // U+2022 BULLET
+    > li:not(:last-child):after {
         margin: 0 .4em;
     }
 }
--- a/lib/css/components/_stacks-lists.less
+++ b/lib/css/components/_stacks-lists.less
@@ -19,7 +19,6 @@
     padding: 0;
 
     display: flex;
-    flex-direction: row;
     flex-wrap: wrap;
 
     > li:not(:last-child)::after {
--- a/lib/css/components/_stacks-lists.less
+++ b/lib/css/components/_stacks-lists.less
@@ -1,0 +1,29 @@
+//
+//  STACK OVERFLOW
+//
+//  This CSS comes from Stacks, our CSS & Pattern library for rapidly building
+//  Stack Overflow. For documentation of all these classes and how to contribute,
+//  visit https://stackoverflow.design/
+//
+//  TABLE OF CONTENTS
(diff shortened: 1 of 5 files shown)


lib/css/components/_stacks-lists.less
```diff
@@ -0,0 +1,29 @@
...
+.s-list.s-list__horizontal {
+    list-style: none;
+    margin: 0;
+    padding: 0;
+
+    display: flex;
+    flex-direction: row;
+    flex-wrap: wrap;
+
+    > li:not(:last-child)::after {
```
benjamin-hodgson: @aaronshekey Do you prefer `:not(:last-child)::after` or `:not(:first-child)::before` for the bullets? It'd affect text styling applied to the `li`s and also the wrapping behaviour when it goes to a second line

lib/css/components/_stacks-lists.less
```diff
@@ -0,0 +1,29 @@
...
+        content: '\2022';  // U+2022 BULLET
+        margin: 0 @su2;
```
benjamin-hodgson: @aaronshekey I suppose we'll need a way to control the spacing?

aaronshekey:
The organization / naming of this is a little weird. This is the first component we have that also has atomic classes of the same name. There's a `List` page in both the base classes and components 🤔 

Would calling this something like separators make sense? It's mildly weird that this component only has the horizontal modifier. Would we move all list styling over here? Or can this be a tiny component within our base styles. This one has got me thinking too much.

Additionally, on the core side of things I haven't quite found any examples of these separators. We've gotten rid of lots of `|` throughout the product, instead opting for simple white space and font color changes. Should we get with @PiperLawson to consider alternate treatments?
image

Sometimes the best component is no component.

lib/css/components/_stacks-lists.less
```diff
@@ -21,8 +21,8 @@
     display: flex;
     flex-wrap: wrap;
 
-        content: '\2022';  // U+2022 BULLET
     > li:not(:last-child):after {
+        content: '•';
```
benjamin-hodgson, line 25: I don't recommend using Unicode in code. This is liable to behave unpredictably depending on whatever default encoding the useragent has. Stick with the `\u` literal

aaronshekey (changes requested):
As discussed in Slack, I'm taking the official stance that instead of horizontal lists with bulleted dividers, we should be using white space instead. This doesn't need to be a component, and we should shy away from this Talent-only aesthetic moving forward.

aaronshekey:
Closing in favor of whitespace-only in our horizontal lists. 😞

PR #563, opened 2020-12-09
Missing Bottom Border on rowspan = total rows. 

koide:
If I can't fix it, at least document it!

### DIFF
--- a/docs/product/components/tables.html
+++ b/docs/product/components/tables.html
@@ -91,7 +91,7 @@ description: Tables are used to list all information from a data set. The base s
     <div class="grid s-notice s-notice__info mt16">
         <div class="grid--cell mr8">{% icon "Alert" %}</div>
         <div class="grid--cell lh-xl">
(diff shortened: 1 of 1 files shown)


docs/product/components/tables.html
```diff
@@ -88,6 +88,12 @@
             </table>
         </div>
     </div>
+    <div class="grid s-notice s-notice__info mt16">
+        <div class="grid--cell mr8">{% icon "Alert" %}</div>
+        <div class="grid--cell lh-xl">
+                    <b>Note:</b> Given the current implementation for default table borders, cells with a rowspan equal to the total amount of rows won't have a bottom border. In order to fix this shortcoming you will need to attach a <code class="stacks-code bg-powder-300">border-bottom: 1px solid var(--black-100)</code> style to those cells. Sorry for the inconvenience!
```
aaronshekey:
```suggestion
                    <b>Note:</b> Given the current implementation for default table borders, cells with a rowspan equal to the total amount of rows won’t have a bottom border. As a workaround you will need to add a border manually to those cells with <code class="stacks-code bg-powder-300">bb bc-black-100</code>.
```

I like how polite you were 😛

aaronshekey (approved):

PR #564, opened 2020-12-10
Prevent override of user-card--link style

paintedbicycle:
When using a usercard in a sidebar widget, we want to preserve the link styling.

**BEFORE**
Screen Shot 2020-12-09 at 7 06 07 PM

**AFTER**
Screen Shot 2020-12-09 at 7 06 19 PM

### DIFF
--- a/lib/css/components/_stacks-widget-static.less
+++ b/lib/css/components/_stacks-widget-static.less
@@ -28,7 +28,7 @@
     font-size: @fs-body1;
     background-color: var(--white);
 
-    &:not(.s-anchors) a:not(.button):not(.s-btn):not(.post-tag):not(.s-sidebarwidget--action) {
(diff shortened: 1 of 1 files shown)


lib/css/components/_stacks-widget-static.less
```diff
@@ -28,7 +28,7 @@
     font-size: @fs-body1;
     background-color: var(--white);
 
-    &:not(.s-anchors) a:not(.button):not(.s-btn):not(.post-tag):not(.s-sidebarwidget--action) {
+    &:not(.s-anchors) a:not(.button):not(.s-btn):not(.post-tag):not(.s-sidebarwidget--action):not(.s-user-card--link ) {
```
aaronshekey, line 31: I think our sidebar widget is a bit too opinionated here. These overrides seem like an anti-pattern. Not your problem 🤗 Noting it for me in the future.

aaronshekey (approved):

PR #568, opened 2020-12-16
[WIP] Added sidebarmenu control

dusanrad:
This PR adds sidebarmenu control that can be used for sidebar menus like the main SO navigation or the on the company edit page in talent.

image

### DIFF
--- a/docs/_data/site-navigation.json
+++ b/docs/_data/site-navigation.json
@@ -251,6 +251,10 @@
               "url": "/product/components/sidebar-widgets/"
             },
             {
+              "title": "Sidebar menu",
+              "url": "/product/components/sidebar-menu/"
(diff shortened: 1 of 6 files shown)


benjamin-hodgson:
What do you think about calling this `s-navigation__vertical`?

aaronshekey:
> What do you think about calling this `s-navigation__vertical`?

Big time. We actually just synced on this in a hangout. Instead of making our left navigation a repeated component, we're going to leave it as a one off. Instead, we're going to beef up our existing `s-navigation` component with a treatment like this:

image

I like it for naming, theming, and other bits of flexibility. At some point, we may want to consider re-styling our global leftnav like pictured above, but that's for another day.

PR #571, opened 2020-12-21
Fix appearance of button.s-navigation--item

bnickel:
A pattern we're exploring on Stack Overflow is using `s-navigation` to represent in-page tabs.  When doing so, it makes more sense to use `button.s-navigation--item` than `a.s-navigation--item`, but that currently has a few rendering glitches:

Screen Shot 2020-12-21 at 11 43 13 AM

This PR applies the following changes:

- It applies various unsetting styles from `.s-btn__unset` to `.s-navigation--item`.
- It adds focus styling to `.s-navigation--item` which matches the active style.  Mouse-only users shouldn't notice a difference here because they shouldn't see a button that's `:focus` that wasn't already `:active`.  This will be helpful for buttons, however because we're removing `outline`.
- I added `tabindex` to the doc previews so the elements could be focused.

Open questions:

- What would a focus style look like for `is-selected`?  Currently it's impossible to tell that a selected item is focussed.
- How should this look in the docs?  I just changed one `<a>` to a `<button>`.  I do have a full-fledged tab controller I could add to Stacks that we could document separately with `s-navigation`.


lib/css/components/_stacks-navigation.less
```diff
@@ -31,12 +31,20 @@
 
     .s-navigation--item {
         padding: @su6 @su12;
+        border: none;
```
aaronshekey, line 60: We are _so_ close to being able to use [all](https://caniuse.com/css-all) here. We would have to remove support for Edge 18, but I would much prefer that to all these resets. Perhaps I'll turn this into a mixin in the meantime.

```
.s-navigation--item {
    all: unset;
}
```

lib/css/components/_stacks-navigation.less
```diff
@@ -31,12 +31,20 @@
...
+        cursor: pointer;
+        user-select: auto;
         border-radius: 1000px;
         margin: @su2;
         color: var(--black-600);
 
         &:hover,
-        &:active {
+        &:active,
+        &:focus {
```
aaronshekey: I think I prefer a true "outline" style. I'll add this in a follow-up commit.

aaronshekey (commented):
> * How should this look in the docs?  I just changed one `<a>` to a `<button>`

I think we should call out that we can use anchors or buttons interchangeably here. I'll be committing a few changes to this branch. Sit tight!

aaronshekey (approved):

PR #572, opened 2020-12-21
s-navigation-tablist controller

bnickel:
This depends on style changes from #571.

This introduces a controller for interactive tab lists using `.s-navigation` using best practices from https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role. I guess we could call it the `.s-navigation` controller but I named it for the accessibility role.

One open question is what to do about MDN's recommendation for each tab panel to have `tabindex="0"`.  The panels will generally be plain `<div>` elements and `<div tabindex="0">` will receive a browser-native focus ring when users click inside.  We could do `outline-none`, `f:outline-ring` or consider omitting the suggestion for `tabindex="0"` entirely since it's not clear that it's suggested in the WAI-ARIA Authoring Practices.


bnickel:
`class="s-navigation f:bs-ring bar-pill px2 d-inline-block"` 🤔

Screen Shot 2020-12-22 at 8 13 57 AM

lib/ts/controllers/s-navigation-tablist.ts
```diff
@@ -0,0 +1,121 @@
...
+        /**
+         * Handles click events on individual tabs, causing them to be selected.
+         */
+        selectTab(event: MouseEvent) {
+            this.switchToTab(<HTMLElement>event.currentTarget);
+
+            // We need to be very agressive about selecting the current tab on click.
+            // Mouse click will not always achieve this and failing to do so will prevent
+            // keyboard navigation from working.
+            this.selectedTab?.focus();
```
aaronshekey: Oh I just hate it though ☹️ Is there any way we can avoid a hard `focus` on these? I really don't want there to be the regression of killing `:focus` styling. I'd love to rely on `:focus-visible` which is why I went on that tangent.

aaronshekey (changes requested):
The thing that's tripping me up is the super aggressive focus states on these. I can't wait to have this controller though!

lib/ts/controllers/s-navigation-tablist.ts
bnickel: Maybe we can?  The issue was trying to avoid was users clicking on a tab and then being unable to keyboard navigate it, something that affects Firefox and Safari but not Chrome.  Looking at https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role, their example has the exact same issue so maybe it's fine to only give keyboard controls to users who keyboard into it.

lib/ts/controllers/s-navigation-tablist.ts
```diff
@@ -95,7 +95,7 @@ namespace Stacks {
         /**
          * Switches the tablist to the provided tab, updating the tabs and panels
          * to reflect the change.
-         * @param selectedTab The tab to select.  If `null` is provided or the element
```
bnickel: This wasn't a typo tho.

aaronshekey: This guy and his double spaces.

lib/ts/controllers/s-navigation-tablist.ts
aaronshekey: I think that'd be a worthy tradeoff. The thing I want to avoid is focus states for simple clicking, but of course, I want focus states for those keyboard folks.

lib/ts/controllers/s-navigation-tablist.ts
```diff
@@ -0,0 +1,121 @@
...
+            for (let tab of this.tabTargets) {
+                tab.removeEventListener("click", this.boundSelectTab);
+                tab.removeEventListener("keydown", this.boundHandleKeydown);
+            }
+        }
+
+        /**
+         * Gets all tabs within the controller.
+         */
+        get tabTargets() {
```
b-kelly, line 31: I'm torn here between using the built in Stimulus `data-target` functionality vs rolling our own. Honestly, I think adding `data-target` to all these would be a pain in the rear, so I'm ok with doing this.

optional: cache the value to a private property so we don't fetch over and over. This will break if tabs are added dynamically, so... there's some tradeoffs. Not a blocker, just a suggestion

lib/ts/controllers/s-navigation-tablist.ts
b-kelly: From a JS viewpoint, this looks fine to me. If styling is the only blocker, then I'd recommend taking a note to come back to this when we flip over to `:focus-visible`

lib/ts/controllers/s-navigation-tablist.ts
```diff
@@ -0,0 +1,121 @@
...
+            if (event.key === "ArrowRight") {
+                tabIndex++;
+            } else if (event.key === "ArrowLeft") {
+                tabIndex--;
+            } else {
+                return;
+            }
+    
+            if (tabIndex < 0) { tabIndex = tabs.length - 1; }
+            if (tabIndex >= tabs.length) { tabIndex = 0; }
```
b-kelly, line 62: I love that you thought through and did this. The circular selection feels really good in the implementation. I'd recommend adding a very succinct comment explaining:

```suggestion
            // if the user selection goes off either end of the array, loop back around to the other side
            if (tabIndex < 0) { tabIndex = tabs.length - 1; }
            if (tabIndex >= tabs.length) { tabIndex = 0; }
```

lib/ts/controllers/s-navigation-tablist.ts
```diff
@@ -0,0 +1,121 @@
...
+
+            tabElement = <HTMLElement>tabs[tabIndex];
+            this.switchToTab(tabElement);
+            this.selectedTab?.focus();
```
b-kelly, line 68: optional: We can probably move this to the bottom of `switchToTab`, since it is used everywhere that method is called. That also removes it from `selectTab`, where it is more fully documented than it is here.

lib/ts/controllers/s-navigation-tablist.ts
```diff
@@ -0,0 +1,121 @@
+namespace Stacks {
+    export class TabListController extends Stacks.StacksController {
```
b-kelly, line 2: nit: I'd even just call this `TabController` since we could use it for lots of different things. Pagination comes to mind immediately and we could extend this in the future (like we did for Popovers>Tooltips) to handle collapsible and others.

lib/ts/controllers/s-navigation-tablist.ts
```diff
@@ -0,0 +1,121 @@
...
+                    tab.classList.remove('is-selected');
+                    tab.setAttribute('aria-selected', 'false');
+                    tab.setAttribute('tabindex', '-1');
+                    panel?.classList.add('d-none');
+                }
+            }
+        }
+    }
+}
+Stacks.application.register("s-navigation-tablist", Stacks.TabListController);
```
b-kelly, line 119: nit: See above regarding `TabController`. I'd call this simply `s-tab` because it could be used for lots of different things.

docs/product/components/navigation.html
```diff
@@ -223,6 +224,150 @@
     </div>
 </div>
 
+<section class="stacks-section">
+    {% header "h2", "Interactive tabs" %}
+
+    <p class="stacks-copy">Stacks provides a Stimulus controller for interactive tab-style navigation within a single document.</p>
+
+    {% header "h3", "Attributes" %}
+    <div class="overflow-x-auto mb32">
```
b-kelly, line 233: Docs look good overall. The table of "add all these things" looks a little long to me, though. I think we could easily transition to a paragraph saying "make sure you add all these aria attributes and make your markup look like this", leaving the controller values in the table.

nit: I don't see a section for the fired events, like we have for [popovers](https://stackoverflow.design/product/components/popovers/#interactive-events).

b-kelly (approved):
Overall, looks pretty good to me. I appreciate the attention to detail in regards to accessibility.

There are a few very items that could be tweaked, but nothing blocking overall. Nice work!

b-kelly:
Regarding the tab panels themselves having `tabindex=0`, I'm with you in that I'm not really a fan. Clicking gives them a focus state and you can tab to them. I'm not really sure what benefits giving them a tabindex gives us, but I do see a few drawbacks. I'm down to remove them entirely, rather than making them appear as if they aren't focused (due to keyboard/TAB interactions).

lib/ts/controllers/s-navigation-tablist.ts
aaronshekey: This already has the `:focus-visible` styling in it. The problem I'm seeing with calling `.focus()` is that it adds focus styling on the element regardless of the browser's default behavior. We really only want focus styling if the user used their keyboard.

aaronshekey (approved):

lib/ts/controllers/s-navigation-tablist.ts
bnickel: My inspiration here is `s-expandable`, `s-popover`, et al where we're willing to out own query selectors if we're doing them on ARIA annotations.

lib/ts/controllers/s-navigation-tablist.ts
bnickel: I removed it from the other location, so I added an explanation for the selection here.

lib/ts/controllers/s-navigation-tablist.ts
bnickel: I prefer `TabList` here because it is intended to function on a `TabList` element and makes assumptions / adheres to the prescribed behavior.  I could imagine doing a similar controller for pagination but it would likely function very differently and have very different niche considerations.

lib/ts/controllers/s-navigation-tablist.ts
bnickel: Same as above, the name here means it is intended to be used exclusively for a `s-navigation` element (because it applies things like `is-selected` rather than having something like `data-s-tablist-selection-class-name="is-selected"`) and because it is intended only for elements that are acting as a `tablist`.  Theoretically it could be used for a button group, but I would rather add `Stacks.application.register("s-btn-group-tablist", Stacks.TabListController);` as a second actor which allows for a potential subclass and differentiation than try making all of that property driven at this juncture.

docs/product/components/navigation.html
bnickel: Done

aaronshekey:
This looks good, merge it!

PR #575, opened 2021-01-04
Add focus-visible polyfill

aaronshekey:
This PR explores adding the https://github.com/WICG/focus-visible polyfill, and swaps our approach to button focus. This _should_ allow us to switch to using focus-visible instead of focus, meaning focus states are only added when using keyboard navigation.


bnickel:
This looks good overall and is a welcome improvement.  One bug I spotted is that `.s-btn.s-btn__muted.s-btn__filled` loses its focus style in dark mode.  Just your average selector priority and sequence battles.

aaronshekey:
> Just your average selector priority and sequence battles.

Good looking out! I caught all those in light mode.

This is a result of our stock button being super weird and not having borders.

aaronshekey:
@b-kelly Any reservations in using this polyfill and including it in our JS?

b-kelly:
I'll take a look at this today. Thanks for the reminder!

b-kelly (changes requested):
Taking a quick peek at the source code for `focus-visible`, it adds a handful of document listeners, including binding then unbinding bunch of mouse listeners on page load. I'm a little skeptical of adding a bunch of js overhead to every page load. I'm pumping the brakes on this for now until I can do some additional review/testing to see what our tradeoffs are.

lib/css/exports/_stacks-mixins.less
```diff
@@ -29,6 +29,27 @@
...
+    }
+
+    // Adds a strong focus indicator for keyboard focus.
+    .js-focus-visible & {
+        &.focus-visible {
+            outline: none;
+            @rules();
+        }
+    }
+}
```
bnickel, line 52:
```suggestion

    &:focus-visible {
        outline: none;
        @rules();
    }
}
```

Since Chrome supports this and Firefox is on the way, should we include actual support for the selector?

lib/css/exports/_stacks-mixins.less
```diff
@@ -29,6 +29,27 @@
             appearance: @all;
 }
 
+//  ===========================================================================
+//  --  FOCUS VISIBLE
+//      Powered by https://github.com/WICG/focus-visible while we wait for better
+//      :focus-visible support
+//  ---------------------------------------------------------------------------
+.focus(@rules) {
```
bnickel, line 37: Should this be `focus-visible` or some variant thereof? `focus` already has its own different meaning.

bnickel (changes requested):
I'd like to add the actual selector so it's not a huge regression for people who consume the CSS and not the JS.

Also, could we package focus styles for `.s-navigation--item` in this PR or should that be its own thing after this ships?

lib/css/exports/_stacks-mixins.less
b-kelly: +1. [caniuse](https://caniuse.com/css-focus-visible) reports that Safari is the last holdout (Edge legacy too, I guess). Maybe we can figure out a progressive enhancement scenario for those two?

aaronshekey:
I'll just roll with proper `:focus-visible` styles and leaving off the focus-visible polyfill. It won't be perfect for those browser that don't support it, but it'll be better than showing focus styling on click.

PR #576, opened 2021-01-06
Call out when exactly to use `aria-pressed`

bnickel:
I've seen a few PRs where people are applying `aria-pressed="false"` to buttons that really shouldn't have it.

For context, `<button>Ask question</button>` is read as something like "Ask question, button ... To click this button press Control-Option-Space". `<button aria-pressed="false">Ask question</button>` is read as something like "Ask question, toggle button ... To select or deselect this checkbox, press Control-Option-Space".

This removes the attributes from the examples and adds a section demonstrating how to use `.is-selected` (We didn't already.) and how to use it accessibly.

### DIFF
--- a/docs/product/components/buttons.html
+++ b/docs/product/components/buttons.html
@@ -58,7 +58,7 @@ description: Buttons are user interface elements which allows users to take acti
                                     </div>
                                 </td>
                                 <td class="va-middle ta-center px4"><button class="ws-nowrap s-btn {{ class.class }}" type="button">Ask question</button></td>
-                                <td class="va-middle ta-center px4"><button class="ws-nowrap s-btn {{ class.class }} is-selected" type="button">Ask question</button></td>
+                                <td class="va-middle ta-center px4"><button class="ws-nowrap s-btn {{ class.class }} is-selected" type="button" aria-pressed="true">Ask question</button></td>
                                 <td class="va-middle ta-center px4"><button class="ws-nowrap s-btn {{ class.class }}" type="button" disabled>Ask question</button></td>
                             </tr>
                             {% endfor %}
@@ -105,7 +105,7 @@ description: Buttons are user interface elements which allows users to take acti
                                     </div>
                                 </td>
                                 <td class="va-middle ta-center px4"><button class="ws-nowrap s-btn {{ class.class }}" type="button">Ask question</button></td>
-                                <td class="va-middle ta-center px4"><button class="ws-nowrap s-btn {{ class.class }} is-selected" type="button">Ask question</button></td>
+                                <td class="va-middle ta-center px4"><button class="ws-nowrap s-btn {{ class.class }} is-selected" type="button" aria-pressed="true">Ask question</button></td>
                                 <td class="va-middle ta-center px4"><button class="ws-nowrap s-btn {{ class.class }}" type="button" disabled>Ask question</button></td>
                             </tr>
                             {% endfor %}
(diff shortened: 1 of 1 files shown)


docs/product/components/buttons.html
```diff
@@ -418,14 +418,41 @@
...
                 </tr>
                 {% endfor %}
             </tbody>
         </table>
     </div>
 </section>
 
+<section class="stacks-section">
+    {% header "h2", "Toggle buttons" %}
+    <p class="stacks-copy">Each button class has a selected state which can be visially activated by applying the <code class="stacks-code">.is-selected</code> class. When a button can switch between selected and unselected states, it is important to also annotate the button with the <code class="stacks-code">aria-pressed</code> attribute for accessibility. A <code class="stacks-code">title</code> attribute may also be appropriate to describe what will happen when pressing the button.</p>
```
aaronshekey:
```suggestion
    <p class="stacks-copy">Each button class has a selected state which can be visually activated by applying the <code class="stacks-code">.is-selected</code> class. When a button can switch between selected and unselected states, it is important to also annotate the button with the <code class="stacks-code">aria-pressed</code> attribute for accessibility. A <code class="stacks-code">title</code> attribute may also be appropriate to describe what will happen when pressing the button.</p>
```

docs/product/components/buttons.html
```diff
@@ -418,14 +418,41 @@
...
+<button class="s-btn is-selected" type="button" aria-pressed="true" title="…">…</button>
+{% endhighlight %}
+{% highlight javascript %}
+toggleButton.addEventListener('click', () => {
+    let wasSelected = toggleButton.getAttribute('aria-pressed') === 'true';
+    let isSelected = !wasSelected;
+    toggleButton.classList.toggle('is-selected', isSelected);
+    toggleButton.setAttribute('aria-pressed', isSelected.toString());
+    …
+});
```
aaronshekey:
```suggestion

<script>
    toggleButton.addEventListener('click', () => {
        let wasSelected = toggleButton.getAttribute('aria-pressed') === 'true';
        let isSelected = !wasSelected;
        toggleButton.classList.toggle('is-selected', isSelected);
        toggleButton.setAttribute('aria-pressed', isSelected.toString());
        …
    });
</script>
```

This docs presentation was a little wack, so let's just wrap this in script tags within the HTML block.

docs/product/components/buttons.html
```diff
@@ -418,14 +418,41 @@
...
+{% endhighlight %}
+        <div class="stacks-preview--example">
+            <div class="grid gs4 fw-wrap">
+                <button class="grid--cell s-btn js-toggle-button" type="button" aria-pressed="false" title="Click to toggle this button's state">Initially unselected toggle button</button>
```
aaronshekey:
```suggestion
                <button class="grid--cell s-btn s-btn__primary js-toggle-button" type="button" aria-pressed="false" title="Click to toggle this button’s state">Initially unselected toggle button</button>
```

docs/product/components/buttons.html
```diff
@@ -418,14 +418,41 @@
...
+                <button class="grid--cell s-btn js-toggle-button is-selected" type="button" aria-pressed="true" title="Click to toggle this button's state">Initially selected toggle button</button>
```
aaronshekey:
```suggestion
                <button class="grid--cell s-btn s-btn__primary js-toggle-button is-selected" type="button" aria-pressed="true" title="Click to toggle this button’s state">Initially selected toggle button</button>
```

aaronshekey (changes requested):

aaronshekey (approved):

PR #577, opened 2021-01-07
s-user-card improvements

bnickel:
I was looking at a reusable component for `s-user-card` and ran into some issues:

- `s-user-card--info` doesn't reorient itself horizontally in small/minimal configurations, meaning we can't use the same structure for all sizes.
- `s-user-card--info` is `display: grid` so it shouldn't be used to hold deleted user usernames, in case we want to do markup annotations.
- `s-user-card__small` lacks margins provided by all other variants.

This addresses each by:

- Making `s-user-card--info` a `display: flex` in small/minimal, updating documentation to use `s-user-card--info`.  If we want to keep the info section optional here, I think we should explicitly say so.
- Making the documentation use `--link` for usernames even if they are deleted.
- Copying margins from minimal to small.

I would also argue that we should:

- Deprecate `s-user-card__minimal` since the only difference between small and minimal is that we unbold some text and decreases a margin by 2px.
- Rename `s-user-card--link` to `s-user-card--name` since it should always be a name but it might not always be a link.
- Radically, consider getting rid of `s-user-card__deleted` in place of `.s-user-card--name:not(:link) { color: var(--black-500); }` If people we're worried about selector priority and people want to go bananas with formatting they could do `<div class="s-user-card--link"><span class="s-btn s-btn__primary">🍌</span></div>`

## Preview

Identical cards with different size classes and avatar size classes:

Screen Shot 2021-01-07 at 2 49 19 PM

### DIFF
--- a/docs/product/components/user-cards.html
+++ b/docs/product/components/user-cards.html
@@ -239,12 +239,14 @@ description: User cards are a combination of a user and metadata about the user 
     <a href="…" class="s-avatar s-avatar__24 s-user-card--avatar">
         <img class="s-avatar--image" src="…" />
     </a>
-    <ul class="s-user-card--awards">
-        <li class="s-user-card--rep">…</li>
-        <li class="s-award-bling s-award-bling__gold">…</li>
-        <li class="s-award-bling s-award-bling__silver">…</li>
-        <li class="s-award-bling s-award-bling__bronze">…</li>
-    </ul>
+    <div class="s-user-card--info">
+        <ul class="s-user-card--awards">
+            <li class="s-user-card--rep">…</li>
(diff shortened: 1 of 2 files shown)


aaronshekey (approved):
Logic is sound! Thank you for making the markup a bit more consistent here.

> * Rename `s-user-card--link` to `s-user-card--name` since it should always be a name but it might not always be a link.

Wanna go ahead and do this since we're likely to have to do markup changes in production anyway?

@paintedbicycle, we're using these on "For You" right? I'll need your help making sure those views look good :v:

paintedbicycle:
It is, yeah. It's also used in other places on prod including Articles and Collections

PR #578, opened 2021-01-07
Better darkmode override mixin and popover arrows

aaronshekey:
This PR started as better syntax for dark-mode overrides, but morphed into some fixes for popover arrows as well. 

image

This:

- Moves the mixin to `mixins.less` so they're more widely available
- Switches to the `whatever &` syntax, so we can more easily use them within existing blocks
- Refines popover arrows to use stacked shadows instead of hard borders which allows us to remove some dark mode overrides, actually
- Breaks how manual arrows are marked up, by requiring `s-popover--arrow` first, but saves us lots of CSS.

@abovedave Will you double check that these popovers don't look worse to you?
@b-kelly Can you make sure I'm not committing anything stupid here?

### DIFF
--- a/docs/assets/less/stacks-documentation.less
+++ b/docs/assets/less/stacks-documentation.less
@@ -26,11 +26,11 @@
     background-image: url("../img/illustration-crane.svg"), url("../img/illustration-field.svg");
     background-position: top -35% left 15%, bottom 10% right 8%;
     background-repeat: no-repeat, no-repeat;
-}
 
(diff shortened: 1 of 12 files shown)


abovedave (dismissed):
@aaronshekey nice, a bit softer and looks better where the borders meet.

Something amiss with the centring (horizontal and vertical) though:

## Existing

Screenshot 2021-01-08 at 10 27 20

## This PR

Screenshot 2021-01-08 at 10 27 13

PR #583, opened 2021-01-25
Fix typo in auto-margin docs

gkoscky:
Left margin should be `.ml-auto`, not `.mr-auto`


aaronshekey (approved):
Good catch, thank you!

PR #584, opened 2021-01-27
Fix typos on `s-sidebarwidget__small-bold-text`

gkoscky:
The documentation mistakenly lists it as `...small-bold_text` (with an underscore instead of hyphen) in a couple of places
It actually propagated to SO code in a few spots 😄


bnickel:
These doc bugs sure love to propagate.

image

gkoscky:
Clearly the solution here is to write less useful docs

aaronshekey:
Working toward that goal every day 😛

PR #587, opened 2021-02-03
User card links should have overflow-wrap:break-word

yellis:
- Addresses: https://meta.stackexchange.com/a/360268/51
- Applicable in other situations where we have a username that is many characters long with no spaces - as it is right now, this can extend the user name (inside `.s-user-card--link`) to overlap with content to the right (badge count, etc).


bnickel:
:+1: This is consistent with how we handle long names in `.user-details`.

aaronshekey:
I dig it! Thanks for getting this upstream 🤗

PR #588, opened 2021-02-03
Ensure dispatcher is set on shown/hidden events

b-kelly:
When calling `shown/hidden` from `show/hide`, we weren't passing the `dispatcher`, which caused the `e.details.dispatcher` to always be undefined.


bnickel (approved):
lgtm

PR #590, opened 2021-02-11
Circular progress bars

aaronshekey:
[](https://www.youtube.com/watch?v=9GPstH1lUYY)

image

### DIFF
--- a/lib/css/components/_stacks-progress-bars.less
+++ b/lib/css/components/_stacks-progress-bars.less
@@ -221,7 +221,8 @@
 //  ---------------------------------------------------------------------------
 .s-progress__circular {
     --s-progress-value: 0;
-    @circleCircumference: (2 * pi() * 14); // 2πr, r = 14.
+    @s-progress-circle-circumference: (2 * pi() * 14); // 2πr, r = 14.
+    @s-progress-stroke-width: 4;
 
     background: none;
     width: @su32;
@@ -231,9 +232,8 @@
         // Make everything originate from the top of the circle
         transform: rotate(270deg);
 
-        @strokeWidth: 4;
         circle {
-            stroke-width: @strokeWidth;
+            stroke-width: @s-progress-stroke-width;
             fill: none;
             stroke-linecap: round;
         }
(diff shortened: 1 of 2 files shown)


aaronshekey:
Still gotta test this one in Legacy Edge, which Windows keeps deleting without asking me.

lib/css/components/_stacks-progress-bars.less
```diff
@@ -215,3 +215,56 @@
         }
     }
 }
+
+//  ===========================================================================
+//  $   CIRCULAR
+//  ---------------------------------------------------------------------------
+.s-progress__circular {
+    --s-progress-value: 0;
+    --s-circle-circumference: 88; // 2πr, r = 14.
```
b-kelly: optional: If you wanna get real picky, you could do this math [in Less](http://lesscss.org/functions/#math-functions-pi), setting to a local less variable. It would also prevent attempts to override it on the client side. I don't foresee that happening, but I've seen less likely things before...

b-kelly (commented):
This is a pretty darn clean approach. Much better than what I was able to come up with. This would pair well with a Razor helper, such as our `Svg.*` helpers.

Side note, I think the math might be a little off. Here's a value of `0.05` compared to a value of `0.95`:

image

lib/css/components/_stacks-progress-bars.less
```diff
@@ -215,3 +215,56 @@
...
+        circle:nth-of-type(2) {
+            stroke: currentColor;
+            stroke-dasharray: var(--s-circle-circumference);
+
+            // Multiply the circle circumference by an the opposite percentage of what we want to display
+            // For example 70%, represented as a decimal is 0.7
+            // The opposite of that is 0.3 so ( 1 - 0.7 )
+            stroke-dashoffset: calc( ((1 - var(--s-progress-value)) * var(--s-circle-circumference)) * 1px); // Multiply everything by 1px since Safari and Firefox require these to be in pixels
+        }
+    }
```
b-kelly: I figured out why 5% looks so much larger than 95%. The [round linecap](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-linecap#round) will

> be extended by a half circle with a diameter equal to the stroke width

To compensate for this, I moved the stroke width out into a variable and re-added it in the final calculation to compensate. Additionally, I moved the pi calculation into less as I suggested above.

image


```suggestion
.s-progress__circular {
    --s-progress-value: 0;
    @circleCircumference: (2 * pi() * 14); // 2πr, r = 14.

    background: none;
    width: @su32;
    height: @su32;

    .s-progress-bar {
        // Make everything originate from the top of the circle
        transform: rotate(270deg);

        @strokeWidth: 4;
        circle {
            stroke-width: @strokeWidth;
            fill: none;
            stroke-linecap: round;
        }

        circle:nth-of-type(1) {
            stroke: currentColor;
            opacity: .4;
        }

        circle:nth-of-type(2) {
            stroke: currentColor;
            stroke-dasharray: @circleCircumference;

            // Multiply the circle circumference by an the opposite percentage of what we want to display
            // For example 70%, represented as a decimal is 0.7
            // The opposite of that is 0.3 so ( 1 - 0.7 )
            // Additionally, since the "round" linecap adds (width / 2)px length, adjust the total offset to compensate
            stroke-dashoffset: calc( (((1 - var(--s-progress-value)) * @circleCircumference) + @strokeWidth) * 1px); // Multiply everything by 1px since Safari and Firefox require these to be in pixels
        }
    }
```

PR #591, opened 2021-02-18
Explore switching to native gap for flex grid

### LINKED ISSUE #328: Grid spacing classes force horizontal scroll when grid items wrap
When using our grid spacing classes responsively, they force a horizontal scroll when wrapping. This is the smallest example that will trigger this.

```
<div class="grid gs64 fw-wrap">
    <div class="grid--cell">
        This cell has dynamic content that ought to wrap with the window
    </div>
    <div class="grid--cell">
        This cell has dynamic content that ought to wrap with the window
    </div>
    <div class="grid--cell">
        This cell has dynamic content that ought to wrap with the window
    </div>
    <div class="grid--cell">
        This cell has dynamic content that ought to wrap with the window
    </div>
    <div class="grid--cell">
        This cell has dynamic content that ought to wrap with the window
    </div>
</div>
```

At smaller window sizes, this forces a horizontal scroll.

### Workaround
You can add an `overflow-x-hidden` to your parent for now, but we ought to solve the underlying issue.

aaronshekey:
Native `gap` support is coming to flex layouts in Safari. It's already shipped in Chrome and Firefox. This pull request explores removing our margin workaround and closes #328. We'll need a strategy for Edge Legacy 18. Perhaps it's time to do what needs to be done there.

Size with native gap: original size: `255140` / gzipped size: `36515`
Size with margin workaround: original size: `255631` / gzipped size: `36571`


aaronshekey:
This is now available in Safari 14.1

aaronshekey:
This approach could safely merge in Q2 2022, given adoption rates of macOS and iOS. A majority of our users would be on the latest 2 versions of Safari by then.

aaronshekey:
The biggest difference I see here is `fw-wrap` wraps way more aggressively when using native gaps vs. our negative margin hacks. When using margins, you were guaranteed that the columns would fill the available space. Gap, not so much.

**Margin**
image

**Gap**
image

dancormier:
I love me some grid, but resolving the conflicts in this branch would be more work than they're worth though. We'll _eventually_ pursue this in another branch. 

Closing!

PR #592, opened 2021-02-19
Toggle switches

aaronshekey:
Let's revisit our toggle switches. This adds the single, compact toggle switch while clarifying how to use our older multiple switch. We'll need to add `s-toggle-switch__multiple` to any existing toggles in production.

It also rounds the current switches to better match these single option switches.

image


lib/css/components/_stacks-toggle-switches.less
```diff
@@ -10,76 +10,142 @@
...
-        position: absolute;
-        left: -999em;
+    input[type="checkbox"] {
+        // "Hide" it by moving it out of sight
+        // This allows users to still "tab" to it, interact via keyboard, etc
         opacity: 0;
-        margin: 0;
-    }
+        position: absolute;
+        left: -99999999px;
```
bnickel: This line provides a weird and inconsistent visual artifact while navigating with VoiceOver.

- Chrome has a visual indicator that sits at the left edge of the browser.
- Safari has a visual indicator extending from the left edge of the browser to the left edge of the **screen**.
- FireFox has a visual indicator surrounding the whole browser window.

Screen Shot 2021-02-19 at 11 14 03 AM

 An alternative would be:

```less
.s-toggle-switch {
    position: relative;

    input[type="checkbox"] {
        position: absolute;
        left: 0;
        top: 0;
        width: 44px;
        height: 24px;
    }
}
```

This positions the invisible checkbox exactly under the visual checkbox, giving it the following visual indicators:

- Around the toggle + label in Safari
- Around the toggle in Chrome/Firefox

Screen Shot 2021-02-19 at 11 36 19 AM
Screen Shot 2021-02-19 at 11 35 44 AM

Placing the input directly under the graphic also has a practical benefit...

lib/css/components/_stacks-toggle-switches.less
```diff
@@ -10,76 +10,142 @@
...
+
+        + label {
```
bnickel: I don't really like the idea of <s>Milhouse having two spaghetti meals in one day</s> an input having two labels, one real and the other empty.  We can get into a weird positional thing where if the real label is after the empty one, the description isn't announced.  Then the input would need an `aria-labeledby`.  This is an edge case but one we could avoid with the previous suggestion, we could replace:

```less
+ label {
   ...
}
```

with

```less
+ .s-toggle-switch--indicator {
    pointer-events: none;
}
```

This would let users click directly on the checkbox itself and leave the page with a single label.

```html
<div class="grid ai-center">
    <label class="grid--cell s-label" for="toggle-example-1">…</label>
    <div class="grid--cell s-toggle-switch">
        <input id="toggle-example-1" type="checkbox">
        <div class="s-toggle-swich--indicator"></div>
    </div>
</div>
```

bnickel (commented):
Just two thoughts to add wrt accessibility.  Everything is _probably_ fine, but could be finer.

PR #594, opened 2021-02-22
Usercards: Move padding to main card

paintedbicycle:
This moves padding to the main .user-card class instead of only on the highlighted card. This allows a highlighted and non highlighted card to sit side-by-side with the text aligned 

Screen Shot 2021-02-22 at 1 38 00 PM

### DIFF
--- a/docs/product/components/user-cards.html
+++ b/docs/product/components/user-cards.html
@@ -129,22 +129,47 @@ description: User cards are a combination of a user and metadata about the user 
                 </div>
             </div>
 
-            <div class="mt16 s-user-card s-user-card__highlighted wmx2">
-                <time class="s-user-card--time">3 minutes ago</time>
-                <a href="#" class="s-avatar s-avatar__32 s-user-card--avatar">
-                    <img class="s-avatar--image" src="https://picsum.photos/64/64" />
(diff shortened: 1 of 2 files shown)


bnickel:
This impacts `.s-usercard__small` and `.s-usercard__minimal`, both of which I feel should still have zero padding.

paintedbicycle:
@bnickel I've reset the padding on those variants for you to test. It feels _a little_ weird that it's not across them all, but I do agree there's little chance of needing the padding on those variants. LMK what you think of this update

bnickel:
@paintedbicycle Looks good.  The reason I raised the small/mini case is that we use both of them on https://stackoverflow.com/review/suggested-edits, one in a post summary and one as part of a sentence.  It may make sense to add them back some padding if we ever want to highlight those, but we'll need to review all usages.

paintedbicycle:
@bnickel Makes sense to me. Let's move ahead with this, then? No padding makes sense in most situations

aaronshekey (approved):
We're going to fight alignment issues on this for the rest of our lives, but I understand the approach.

image

bnickel (approved):

PR #595, opened 2021-02-26
Upgrade to Less 4.0

aaronshekey:
Hey neat, Less is now on [4.x](https://github.com/less/less.js/releases/tag/v4.0.0) as of December 2020, so this PR bumps the `grunt-contrib-less` version and fixes some of our math as a result. I opted to add parens on division on the Stacks side of things since that's the new default for Less. I don't want to config our Less to use the old math. :v:

I diff'd the output of `stacks.css` before this latest version and after and they're the same.


lib/css/components/_stacks-widget-static.less
```diff
@@ -12,7 +12,7 @@
 //  _stacks-widget-dynamic.less.
 
 
-@s-sidebarwidget-content-padding: @su16 - 1px; // subtract 1px for border
+@s-sidebarwidget-content-padding: @su16 - 1; // subtract 1px for border
```
bnickel: No units?  What is this?

aaronshekey: Yeah, I don't know. It's at least consistent without the units. We do stuff like `@su2 / 2` all over Stacks.

PR #597, opened 2021-02-26
Bump Popper to 2.8.4

aaronshekey:
Bumpity bump. Had a gander at our examples and everything seems in order.


b-kelly (approved):
lgtm! I looked over the changelogs for the range we are updating from/to. Nothing there looks amiss, mostly bugfixes and new features that don't affect us.

bnickel (approved):

PR #598, opened 2021-02-26
Segmented Progress Bars

jbubriski:
This is working perfectly in Chrome, but it looks like there is an issue with Firefox.

This is similar to the circular progress bars but with a little twist. Currently limited to 20 segments given the way I needed to accomplish this. Relevant comment from the code about this:

`Dasharrays are pattern based and the line keeps going "infinitely".
As such, the only way to create a line with a big empty space at the end is to make a LONG pattern
that is long as you could possibly want, and then add an equivalent spacing to the end.
Then we can offset the array to hide the starting dashes.`

Would love some input on how to make this work in Firefox too. It looks like it should work, haven't tried any Firefox specific tests yet.

Chrome:

2021-02-26 12_53_36-Progress bars - Stacks

Firefox:

2021-02-26 12_57_04-Progress bars - Stacks — Mozilla Firefox

### DIFF
--- a/docs/product/components/progress-bars.html
+++ b/docs/product/components/progress-bars.html
@@ -276,12 +276,22 @@ description: A component that visually communicates the completion of a task or 
             </svg>
         </div>
     </div>
+</div>
 
     <p class="stacks-copy">By default, the segmented progress bars are 256px wide. These dimensions can be modified by adding
         t-shirt sizing classes to <code class="stacks-code">.s-progress__segmented</code>: <code
             class="stacks-code">.s-progress__sm</code>, <code class="stacks-code">.s-progress__md</code>, and <code
             class="stacks-code">.s-progress__lg</code></p>
 
+<div class="stacks-preview">
+{% highlight html %}
+<div class="s-progress s-progress__segmented s-progress__sm" style="--progress-value-max: 4; --progress-value: 1;">
+    <svg viewBox="0 0 100 3">
+        <line x1="1" y1="1" x2="100" y2="1" />
+        <line x1="1" y1="1" x2="100" y2="1" />
+    </svg>
+</div>
+{% endhighlight %}
     <div class="stacks-preview--example">
         <div class="s-progress s-progress__segmented s-progress__sm" style="--progress-value-max: 4; --progress-value: 1;">
             <svg viewBox="0 0 100 3">
@@ -311,6 +321,7 @@ description: A component that visually communicates the completion of a task or 
             </svg>
         </div>
     </div>
+</div>
 </section>
 
 <section class="stacks-section">
--- a/docs/product/components/progress-bars.html
+++ b/docs/product/components/progress-bars.html
@@ -232,6 +232,88 @@ description: A component that visually communicates the completion of a task or 
 </section>
 
 <section class="stacks-section">
+    {% header "h2", "Segmented" %}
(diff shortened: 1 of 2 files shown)


jbubriski:
@aaronshekey any thoughts?

lib/css/components/_stacks-progress-bars.less
```diff
@@ -270,3 +270,61 @@
...
+        stroke-dashoffset: calc((21 - var(--progress-value)) * (var(--progress-value-width) + var(--progress-value-spacer-width)));
+    }
+
+    &.s-progress__sm {
+        width: 128px;
```
aaronshekey, line 320: I think with the exception of the circular progress bar, these should just take the width of the parent, so `width: 100%`. No need for t-shirt sizing on these, since there are very few layouts where these sizes will actually make sense.

docs/product/components/progress-bars.html
```diff
@@ -231,6 +231,99 @@
...
+        <div class="s-progress s-progress__segmented s-progress__md" style="--progress-value-max: 4; --progress-value: 1;">
+            <svg viewBox="0 0 100 3">
+                <line x1="1" y1="1" x2="100" y2="1" />
+                <line x1="1" y1="1" x2="100" y2="1" />
+            </svg>
+        </div>
+
+        <div class="s-progress s-progress__segmented s-progress__lg" style="--progress-value-max: 4; --progress-value: 1;">
+            <svg viewBox="0 0 100 3">
+                <line x1="1" y1="1" x2="100" y2="1" />
```
aaronshekey, line 319: The thing that's weird to me about these being built off of SVG is that they scale proportionately, unlike our other line-based progress bars. Is there a way to build these with actual markup and flex? Perhaps we build them with `grid` and reach for `gap`?

Another unfortunate side-effect of the SVG is that they're blurry on 1x screens like mine 🤔

docs/product/components/progress-bars.html
```diff
@@ -231,6 +231,99 @@
...
+<section class="stacks-section">
+    {% header "h2", "Segmented" %}
+    <p class="stacks-copy">For task tracking, switching to a segmented progress bar might be more appropriate. Our segmented
+        progress bar inherits the parent’s text color. You and your designer will need to choose an appropriate font
+        color. You’ll also need to pass the <code class="stacks-code">--s-progress-value-max</code> CSS variable the total number
+        of segments, and the <code class="stacks-code">--s-progress-value</code> CSS variable the number of completed segments..</p>
+
+<div class="stacks-preview">
+{% highlight html %}
+<div class="s-progress s-progress__segmented" style="--progress-value-max: 8; --progress-value: 4">
```
aaronshekey, line 243: This API is nice and super flexible, but I've got a few issues with this approach down below.

docs/product/components/progress-bars.html
jbubriski: I'm pretty sure we could do it with an HTML grid. I was trying to avoid that because I think it means we'll have x number of divs nested inside... 1 for each segment. Any idea on how to do it with having potentially 20 extra divs?

aaronshekey:
After a bit of discussion yesterday, we're gonna change the design here to more closely align with the progress bars we already have. https://codepen.io/aaronshekey/pen/Rwoqxva?editors=1100

PR #599, opened 2021-02-26
Fix `aria-expanded` in expandable example

kylejrp:
The example initial state is not expanded, but `aria-expanded` was set to `true`.

### DIFF
--- a/docs/product/components/expandable.html
+++ b/docs/product/components/expandable.html
@@ -31,7 +31,7 @@ description: An expandable, sometimes called an accordion, is an element that ca
                 <button class="s-btn s-btn__filled mb8"
                     data-controller="s-expandable-control"
                     data-s-expandable-control-toggle-class="is-selected"
-                    aria-expanded="true"
+                    aria-expanded="false"
                     aria-controls="expandable-example">
                         Show first paragraph
                 </button>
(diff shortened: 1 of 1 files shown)


docs/product/components/expandable.html
```diff
@@ -31,7 +31,7 @@
                 <button class="s-btn s-btn__filled mb8"
                     data-controller="s-expandable-control"
                     data-s-expandable-control-toggle-class="is-selected"
-                    aria-expanded="true"
+                    aria-expanded="false"
```
bnickel, line 34: I think this is the right idea in the wrong place.  `aria-expanded="true"` should always be paired with `s-expandable is-expanded` (and in this case `s-btn is-selected`).

Right now this code is behaving as expected, but the code above between `{%highlight html %}` and `{% endhighlight %}` is wrong for the reasons I described above.  They would be fixed either by setting `aria-expanded="false"` there, or applying the two additional classes to the appropriate elements.

bnickel (changes requested):

docs/product/components/expandable.html
bnickel: Oh carp, my bad.  This is correct, but we should update the code snippet as well. 🤦

bnickel (approved):
lgtm, good catch!

PR #600, opened 2021-02-26
Add v-truncate-fade classes

kylejrp:
Adds `v-truncate-fade-{sm|md|lg}` that adds a vertical fade out at a height of `100px`, `200px`, or `400px` using [`mask-image`](https://caniuse.com/mdn-css_properties_mask-image).

Some browsers require the `-webkit-mask-image` prefixed version for now.

The copy isn't anything special so feel free to request changes there!

image

### DIFF
--- a/docs/product/base/typography.html
+++ b/docs/product/base/typography.html
@@ -302,20 +302,20 @@ description: Stacks provides atomic classes to override default styling of typog
     <p class="stacks-copy">Alternatively, you can use a vertical fade that will set <code class="stacks-code">max-height</code> and sets a vertical <code class="stacks-code">mask-image</code>:</p>
     <div class="stacks-preview">
         {% highlight html %}
-<p class="v-truncate-fade-sm">…</p>
-<p class="v-truncate-fade-md">…</p>
-<p class="v-truncate-fade-lg">…</p>
+<p class="v-truncate-fade v-truncate-fade__sm">…</p>
+<p class="v-truncate-fade">…</p>
+<p class="v-truncate-fade v-truncate-fade__lg">…</p>
         {% endhighlight %}
         <div class="stacks-preview--example">
-            <div class="v-truncate-fade-sm wmx3 mb16">
+            <div class="v-truncate-fade v-truncate-fade__sm wmx3 mb16">
                 Regardless of length, this text will be this text will be truncated to 100px, vertically. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
             </div>
 
-            <div class="v-truncate-fade-md wmx3 mb16">
+            <div class="v-truncate-fade wmx3 mb16">
                 Regardless of length, this text will be this text will be truncated to 200px, vertically. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
             </div>
 
-            <div class="v-truncate-fade-lg wmx3 mb16">
+            <div class="v-truncate-fade v-truncate-fade__lg wmx3 mb16">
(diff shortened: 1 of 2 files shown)


kylejrp:
This has dark mode support, so we can fix these two places once this is merged and Stack Overflow is on the next release:

Mod dashboard for an answer:
image

Tour page:
image

docs/product/base/typography.html
```diff
@@ -299,6 +299,27 @@
             </div>
         </div>
     </div>
+    <p class="stacks-copy">Alternatively, you can use a vertical fade that will set <code>max-height</code> and sets a <code>vertical-mask</code>:</p>
```
aaronshekey:
```suggestion
    <p class="stacks-copy">Alternatively, you can use a vertical fade that will set <code class="stacks-code">max-height</code> and sets a vertical <code class="stacks-code">mask-image</code>:</p>
```

docs/product/base/typography.html
```diff
@@ -299,6 +299,27 @@
...
+    <div class="stacks-preview">
+        {% highlight html %}
+<p class="v-truncate-fade-sm">…</p>
+<p class="v-truncate-fade-md">…</p>
```
aaronshekey: Do we make one of these the default and have sizing as a modifier?

`v-truncate-fade` is medium. `v-truncate-fade v-truncate-fade__sm` for small and `v-truncate-fade__lg`? Might be a bit more consistent with other components.

lib/css/atomic/_stacks-typography.less
```diff
@@ -151,6 +151,26 @@ h1, h2, h3, h4, h5, h6, p {
...
+    .truncation-fade(1)
+}
+
+.v-truncate-fade-lg {
+    .truncation-fade(2)
+}
+
+.truncation-fade(@count) {
+    -webkit-mask-image: linear-gradient(180deg,#000 (pow(2, @count) * 100px - 50px),transparent); 
+    mask-image: linear-gradient(180deg,#000 (pow(2, @count) * 100px - 50px)px,transparent);
```
aaronshekey: How's our Legacy Edge 18 support? 😬

lib/css/atomic/_stacks-typography.less
```diff
@@ -151,6 +151,26 @@ h1, h2, h3, h4, h5, h6, p {
...
+.v-truncate-fade-md {
+    .truncation-fade(1)
+}
+
+.v-truncate-fade-lg {
+    .truncation-fade(2)
+}
+
+.truncation-fade(@count) {
+    -webkit-mask-image: linear-gradient(180deg,#000 (pow(2, @count) * 100px - 50px),transparent);
```
aaronshekey:
```suggestion
    -webkit-mask-image: linear-gradient(180deg, #000 (pow(2, @count) * 100px - 50px), transparent); 
```

aaronshekey: This is clever, but we may want to tweak things a bit to make the bottom edge of each fade more consistent. See how you can see more or less of the last row of text on each of those variations? It'd be nice if they were more consistent. My favorite fade is on the small variation :v:

aaronshekey (changes requested):
These'll be super useful, thank you for contributing! I might split off Truncation into its own section from Typography once these drop. 🔥

lib/css/atomic/_stacks-typography.less
kylejrp: 👌

image

I haven't tested it myself, but I know a guy that has a Legacy Edge VM set up... 😉

lib/css/atomic/_stacks-typography.less
kylejrp: I don't have a great trick to make it fade nicely like the small variation. We would need to make sure that `max-height` is set to some multiple of our line height, but the use case I'm imagining is user content with headings, images, tables, etc.

Maybe setting the 100/200/400px to values that are some multiple of our line height so our base case is okay? I'm not sure what font size we'd want to base this off of, though. 🤔 Maybe we can set up different `max-height`/`content-mask` for different sizes? Eg.

```
.v-truncate-fade-sm {
    .truncation-fade(0, 13px)
}

.v-truncate-fade.fs-body2 {
    .truncation-fade(0, 14.3px)
}

.v-truncate-fade.fs-body3 {
    .truncation-fade(0, 15.4px)
}

}

.truncation-fade(@count, @font-size){
   mask-image: linear-gradient(180deg, #000 (@font-size * 12 * @count), transparent);
}
```

Does this spark any ideas for you @aaronshekey?

aaronshekey: We could try monkeying with `ems` which is based on the font size. Unfortunately, this is a fork and not a branch, so I can't easily pull it down to commit. You've got write access to the repo, by the way, for future stuff.

```
  .v-truncate-fade {
    mask-image: linear-gradient(180deg, #000 4em, transparent);
}
```
Otherwise, I'm totally cool with some magic numbers here if it comes down to that 😛

kylejrp: `em` was a good choice, I combined it with the `@lh-md` class so that I could make `max-height` based off of number of lines. We could mess with more modifiers so that there's different versions for different line heights, but that feels like too early of an optimization for me!

```
-webkit-mask-image: linear-gradient(180deg, #000 @lh-md * 1em * 9, transparent); 
mask-image: linear-gradient(180deg,#000 @lh-md * 1em * 9,transparent); 
max-height: @lh-md * 1em * 12;
```

image

lib/css/atomic/_stacks-typography.less
```diff
@@ -151,6 +151,33 @@ h1, h2, h3, h4, h5, h6, p {
     overflow: hidden;
 }
 
+.v-truncate-fade {
+    overflow: hidden; 
+    -webkit-mask-image: linear-gradient(180deg, #000 @lh-md * 1em * 9, transparent);
```
aaronshekey: `9em` is `1em * 9` already, so you should just be able to go with that.

```suggestion
    -webkit-mask-image: linear-gradient(180deg, #000 @lh-md * 9em, transparent); 
```

lib/css/atomic/_stacks-typography.less
```diff
@@ -151,6 +151,33 @@ h1, h2, h3, h4, h5, h6, p {
...
+    }
+
+    &.v-truncate-fade__lg {
+        -webkit-mask-image: linear-gradient(180deg, #000 @lh-md * 1em * 21, transparent); 
+        mask-image: linear-gradient(180deg,#000 @lh-md * 1em * 21,transparent); 
+        max-height: @lh-md * 1em * 24;
+    }
+}
+
+.v-truncate-fade-lg {
```
aaronshekey: Do these go away now?

lib/css/atomic/_stacks-typography.less
kylejrp: In my head it was (one line height) * (9 lines), where "one line height" = `@lh-md * 1em`, so it didn't make sense to combine. I'll use a local variable so my mental model is a little more clear:

```
@one-lh: @lh-md * 1em;
-webkit-mask-image: linear-gradient(180deg, #000 @one-lh * 9, transparent); 
mask-image: linear-gradient(180deg,#000 @one-lh * 9,transparent); 
max-height: @one-lh * 12;
```

lib/css/atomic/_stacks-typography.less
kylejrp: Yup, good catch

aaronshekey (approved):

PR #603, opened 2021-03-04
Don't double grey-out `.s-description`

bnickel:
`.s-description` is pretty much always hosted inside a `.s-label`, so greying them both out results in small text at 25% opacity.  This fixes that.


aaronshekey (approved):
Interesting that we don't have this documented in Stacks. I may have to pop this in there at some point :v:

PR #605, opened 2021-03-06
Menu component

aaronshekey:
This PR adds a menu component, since we're starting to see a lot of these within our popovers. Time to formalize and provide some guidance!

image


aaronshekey:
This needs a disabled state. Adding now :v:

bnickel:
Some thoughts from a quick scan:

- The examples really draw the eye and it took me a moment to realize that the `s-popover` in the box was not part of `.s-menu`'s rendering.  We may want to include the enclosing popover in the example HTML and/or include rendered examples without popovers.
- `.is-selected` should have an `aria-current="page"` or similar.
- Why are we doing `.is-disabled` on a button instead of `disabled`?  It would probably make more sense on a link, but we'd probably also want to add `aria-disabled="true"`.
- This is very similar to `.s-sidebarwidget--content.s-sidebarwidget__items`.  Would there be any point in trying to deprecate `s-sidebarwidget__items` and make this component work in there?

kylejrp:
+1 on including `s-popover` in the example HTML and for using the `disabled` attribute on a button where we can for accessibility

aaronshekey:
I go back and forth on disabled options at all. You'll notice that both buttons and anchors can be used interchangeably in these menus, but what the heck is a disabled link? I think there might be less value in showing disabled options than we think.

> The examples really draw the eye and it took me a moment to realize that the s-popover in the box was not part of .s-menu's rendering. We may want to include the enclosing popover in the example HTML and/or include rendered examples without popovers.

The interesting thing about this component is it's practically never used outside of a popover. We do have the notion of a post menu, and could be introduced as a variant of `s-menu` as `s-menu__horizontal`. I think if we were to build a post today from scratch, we'd throw all those options behind a button and have them contained in a menu within a popover. 🤔 

I'll add a menu example or two outside of a popover.

These are the examples I'm trying to formalize first, before we grow too organically beyond them. For each of these examples, there are about 3 or 4 different contexts where they appear. 

image

As for the sidebar widget, that was an interesting first exercise in components at Stacks, and its design choices are a bit different than the majority of our other components in that it's a bit more prescriptive. You can tell by how many lines of CSS are on it. Most of our components are much smaller in scope. A single button, a single card, a single menu, a single popover. Sidebar widgets could probably be split, or could be variations of a card + menu, etc. 🤔

aaronshekey:
Punted on documenting the disabled styling on the block links. It's a mildly larger project that I'd like to revisit while addressing some `button` changes.

PR #608, opened 2021-03-11
Various docs improvements

aaronshekey:
This PR does a couple of things to the docs side of things.

1. Fixes the description meta data on each page
2. Brings in a few libraries that we were loading via CDN, bumps their versions, and adds them as proper dependencies—jQuery, docsearch, and list.js
3. Uses the minified version of Stacks and puts it behind a cachebreaker

This PR stares longingly at #362.


b-kelly (approved):
lgtm. Tested locally and everything seems to be working as intended.

PR #611, opened 2021-03-11
Drop support for Edge 18

aaronshekey:
You know, there really isn't a ton in Stacks to consider here. Edge 18 doesn't seem to be holding us back on any particular feature we're trying to ship. ¯\_(ツ)_/¯ 

I think the only thing this would mean is we'd stop testing in Edge 18, specifically.

### DIFF
--- a/lib/css/atomic/_stacks-colors.less
+++ b/lib/css/atomic/_stacks-colors.less
@@ -14,14 +14,9 @@
 
 & {
     .color(@color-name) {
-        // need to define focus-within as a separate rule for IE compat.
-        // If a selector contains focus-within, Edge ignores the whole rule,
-        // including any other selectors combined using comma!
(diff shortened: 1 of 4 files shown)


aaronshekey:
Features we'd gain if we dropped Edge 18:

- `all` This would allow us to more easily override different properties, eg. links styled as buttons, buttons styled as links, etc.
- Native `gap` support in flexbox.

aaronshekey:
Pulled some usage statistics. In versions of Legacy Edge 18 and below, we're at `15,230` users over the last 7 days. Edge represents `6.13%` of all our users, with `1.02%` of that amount using 18 or below.

For full context, We have 23,333 Internet Explorer users, a browser we don't support _at all_. Legacy Edge is a smaller user share than that.

PR #612, opened 2021-03-12
Use CSSnano and postcss instead of cssmin

aaronshekey:
Can you tell I'm procrastinating? This PR looks at using CSSnano instead of grunt's built-in cssmin. We could save about 300 bytes on the output, and it wouldn't even effect production since we do our own compilation there.


aaronshekey:
Closing in favor of #707

PR #613, opened 2021-03-12
Add an "Edit on GitHub" button

kylejrp:
edit-on-github

Adds an "Edit on GitHub button" to the top and bottom of the page

For those changes that you don't want to pull up an IDE for. Note that with the GitHub editor you'll have the option to commit directly or make a new branch/PR:

image

### DIFF
--- a/docs/.eleventy.js
+++ b/docs/.eleventy.js
@@ -1,7 +1,6 @@
 const hljs = require("highlight.js");
 const syntaxHighlight = require("eleventy-plugin-highlightjs");
 const pluginTOC = require("eleventy-plugin-nesting-toc");
-const pluginDate = require('eleventy-plugin-date');
 const markdownShortcode = require("eleventy-plugin-markdown-shortcode");
 const { default: Icons, Spots } = require("@stackoverflow/stacks-icons");
 const { version } = require("../package.json");
@@ -150,9 +149,6 @@ module.exports = function(eleventyConfig) {
       return ''; // use external default escaping
     }
   });
-
-  // Add date formatting
-  eleventyConfig.addPlugin(pluginDate);
 
   // Add submenu generation
   eleventyConfig.addPlugin(pluginTOC, {tags: ['h2', 'h3'], wrapper: 'nav aria-label="Table of contents"'});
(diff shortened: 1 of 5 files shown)


aaronshekey:
Neat trick! Should we limit this to a footer on each page with the possibility of a "Last edited" date?

``` html
<div class="grid jc-space-between fc-light my48 pt24 bt bc-black-050">
    <div class="grid--cell">Last edited Friday, March 12th</div>
    <a href="https://github.com/StackExchange/Stacks/edit/develop/docs/{{ page.inputPath }}" class="grid--cell s-btn s-btn__muted">
        {% icon "Pencil" %}
        Edit on GitHub
    </a>
</div>
```

kylejrp:
Thanks! I like just having it in the footer. I've modified my PR to include your markup. 

It has some weird vertical alignment that I'm not quite sure how to fix:
image

I'm also reading up on [Content Dates on 11ty](https://www.11ty.dev/docs/dates/) and it seems like it uses the creation date unless you manually assign dates, which might be a little tricky. 🤔 Date formatting also seems tricky. 🤷‍♂️ This is the format we get unless we add some formatting plugins (I think)
image

There's enough weirdness that I'm kinda willing to drop the date stuff for now. 😕 What do you think?

aaronshekey:
Just pop an `ai-center` on the `grid` and it'll look great.

aaronshekey:
> There's enough weirdness that I'm kinda willing to drop the date stuff for now.

DISAGREE :trollface:

aaronshekey:
Oh shoot, I was too busy being a smartass to see what this would do on Netlify. Netlify builds everything from scratch so the date is actually the _build_ and not what we see locally. Hmmmmmmmmm. 🤔

kylejrp:
Yeah, when I was testing it on my PC, all the dates were the same because I had just cloned that day. Not sure how we get dates other than manually adding it in the header of the file.

> ᵗʰᵉʳᵉ'ˢ ᵉⁿᵒᵘᵍʰ ʷᵉⁱʳᵈⁿᵉˢˢ ᵗʰᵃᵗ ⁱ'ᵐ ᵏⁱⁿᵈᵃ ʷⁱˡˡⁱⁿᵍ ᵗᵒ ᵈʳᵒᵖ ᵗʰᵉ ᵈᵃᵗᵉ ˢᵗᵘᶠᶠ ᶠᵒʳ ⁿᵒʷ

aaronshekey:
Sorry about the back and forth. I think it belongs in the footer, but it looks a little weird without any friends down there. I can deal with it!

I've cleaned up my mess and this is ready to ~ship~ shrimp 🦐

aaronshekey (approved):

PR #614, opened 2021-03-18
Change modal <a href="#"> to <button>

kylejrp:
Use `<button>` elements instead of doing `<a href="#">` for better semantic HTML and accessibility

https://css-tricks.com/use-button-element/
https://stackoverflow.com/a/25350722/3023052

I didn't check to see if this breaks the JS - I'm assuming it doesn't but I'm willing to be yelled at


aaronshekey (approved):
I was thinking of doing this as well! Glad we're on the same page. These should be safe since we're only change the documentation, and presumably, only the future modals people add. We can do more work in production to swap these for buttons, but that can be the Stacks team's concern. :v:

bnickel (changes requested):
We probably want `type="button"` on these.  There will be times that we have `<form class="s-modal--dialog">` and `<button>` is a loaded gun for random form submits. (Ask me how I know. 😭)

Even if we don't use forms in the example, a lot of Stacks example code ends up verbatim in the codebase so always be vigilant.

bnickel:
Ping.  I'd like to see this merged. We _just_ had an issue where this `href="#"` caused a `popstate` event listener to assert.

bnickel (approved):

PR #615, opened 2021-03-19
Popover utils and auto-show

b-kelly:
This is a fork of #462 which has some incredibly useful content, but has been sitting too long over an implementation regression.

Added features:

* `Stacks.showPopover`, `Stacks.hidePopover`, `Stacks.attachPopover`, `Stacks.detachPopover` helper methods
* `data-s-popover-auto-show` attribute
* updated Popover documentation page

95% of the code from this PR has been straight copied from the existing PR. The only changes are as follows:

* auto-show now doesn't hook up document events like .is-visible
  * This is less than ideal (especially for new functionality), but I'll have a quick follow-up PR to add the `hide-on-outside-click` functionality from the previous PR
* minor readability tweaks to `getPopover` internal function
* Didn't carry over `hide-on-outside-click` or `TokenList`
  * both of these are coming in a follow-up PR (I wanted to get this smaller set of changes out first)


lib/ts/controllers/s-popover.ts
```diff
@@ -315,6 +324,177 @@ namespace Stacks {
...
+    /**
+     * Removes the popover controller from an element and removes the popover from the DOM.
+     * @param element the element that has the `data-controller="s-popover"` attribute.
+     * @returns The popover that was attached to the element.
+     */
+    export function detachPopover(element: Element) {
+        const { isPopover, controller, referenceElement, popover } = getPopover(element);
+
+        // Hide the popover so its events fire.
+        if (controller) { controller.hide(); }
```
```suggestion
        controller?.hide();
```

Maybe shorthand here?

lib/ts/controllers/s-popover.ts
```diff
@@ -315,6 +324,177 @@ namespace Stacks {
...
+
+        // Remove the popover if it exists
+        if (popover) { popover.remove(); }
```
```suggestion
        popover?.remove();
```

lib/ts/controllers/s-popover.ts
```diff
@@ -315,6 +324,177 @@ namespace Stacks {
...
+
+        // Remove the popover controller and the aria-controls attributes.
+        if (isPopover) {
+            element.removeAttribute("data-controller");
```
bnickel: I know we're planning to handle this separately, but there's a lot of risk associated with this line.  (E.g. killing `data-controller="s-popover s-tooltip"`.)  We could do a slapdash `element.setAttribute("data-controller", element.getAttribute("data-controller")?.replace(/\bs-popover\b/, ''))` or warn about it in the documentation.

lib/ts/controllers/s-popover.ts
```diff
@@ -315,6 +324,177 @@ namespace Stacks {
...
+
+        if (!existingId) {
+            referenceElement.setAttribute("aria-controls", popoverId);
+        }
+
+        if (!popover.parentElement && element.parentElement) {
+            referenceElement.insertAdjacentElement("afterend", popover);
+        }
+
+        element.setAttribute("data-controller", "s-popover");
```
bnickel: Same

lib/ts/controllers/s-popover.ts
```diff
@@ -315,6 +324,177 @@ namespace Stacks {
...
+        if (!popover.classList.contains('s-popover')) {
+            throw `popover should have the "s-popover" class but had class="${popover.className}"`;
+        }
+
+        if (existingId && existingId !== popoverId) {
+            throw `element has aria-controls="${existingId}" but popover has id="${popoverId}"`;
+        }
+
+        if (!popoverId) {
+            popoverId = "--stacks-s-popover-" + Math.random().toString(36).substring(2, 10);
```
bnickel, line 421: We have a similar function in `TooltipController.generateId()`.  Maybe we could add a base method on StacksController:

```
export function generateId() {
    return `--stacks-${this.identifier}-{Math.random().toString(36).substring(2, 10)}`;
}
```

bnickel (approved):
This looks good _but_ the `removeAttribute("data-controller")` stuff will continue to be a deal-breaker for cases where we're doing a temporary popover on top of an element that otherwise has a tooltip.  (These are the main cases where we would be using `detachPopover` today.)

I know you're looking at token list separately, but could we have a stopgap here? E.g., 

```typescript
function toggleController(el: Element, name: string, include: boolean) {
  var controllers = new Set(el.getAttribute('data-controller')?.split(/\s+/));
  if (include) { controllers.add(name); } else { controllers.delete(name); }
  el.setAttribute('data-controller', [...controllers.values()].join(' '))
}
```

docs/product/components/popovers.html
```diff
@@ -281,6 +287,97 @@
...
+                        </div>
+                    </div>
+                </div>
+                <p class="fc-light">Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla et metus molestie nulla luctus sodales ac luctus justo. Aenean iaculis ac ante sit amet aliquam. Duis dolor velit, imperdiet sed mauris eu, sollicitudin egestas nisl. Ut vitae nulla eu risus iaculis semper sit amet vitae tortor. Sed convallis lacus quis libero placerat finibus. Phasellus pulvinar vel nunc eu tempor. Sed pharetra magna a felis egestas placerat. Sed imperdiet dui a sem fermentum, eget consectetur elit feugiat. Phasellus non condimentum orci. Nam id molestie elit, ut gravida metus. Vivamus vel nunc risus. Maecenas posuere sit amet tellus vel laoreet. Nulla lacus mauris, rhoncus eu laoreet quis, mattis vehicula nisl. Integer efficitur quam et nisi luctus scelerisque. Mauris efficitur lectus ac malesuada congue.</p>
+                <button class="s-btn s-btn__link" data-action="section#showHelp">Click here for help</button>
+            </div>
+        </div>
+    </div>
+    <script>
+        window.addEventListener("load", () => {
```
bnickel: Ugh... This is only firing if you land on the page directly, not navigating to it.  I forget what the right approach is here.

docs/product/components/popovers.html
```diff
@@ -281,6 +287,97 @@
...
+        this.loaded = true;
+    }
+});
+{% endhighlight %}
+    <div class="stacks-preview--example bg-black-025">
+        <button class="s-btn s-btn__outlined" data-controller="actions" data-action="actions#load">Show actions for this post</button>
+    </div>
+</div>
+<script>
+    window.addEventListener("load", () => {
```
bnickel: Same.  Should these move into a .js file?

docs/product/components/popovers.html
aaronshekey: Ah, right right right. Yep. Toss this into a JS file and load it with `defer`. It's really great as a consumer of the docs to load everything via AJAX but pretty much universally trolls every new feature demo we build, lol.

lib/ts/controllers/s-popover.ts
b-kelly: I'm punting on this one for now. I'd have to make the method static (since it is accessed in a helper here), in which case it won't have access to `this.identifier` and I'd have to hardcode the controller name somewhere for both types

lib/ts/controllers/s-popover.ts
b-kelly: I took the cheap way out with set/removeAttribute, but you're totally right. I implemented your toggleController code as a stopgap for a more full featured solution.

docs/product/components/popovers.html
b-kelly: Added to `feature.popovers.js`

docs/product/components/popovers.html
b-kelly: Added to `feature.popovers.js`

b-kelly:
@bnickel I addressed your concerns, especially the ones concerning the `data-controller` attributes. I already have a green check from you, but would you like to give it another lookover?

lib/ts/controllers/s-popover.ts
```diff
@@ -315,6 +324,193 @@ namespace Stacks {
...
+       toggleOnClick?: boolean;
+        /**
+         * When set, `data-s-popover-placement` will be set to this value on the controller element.
+         */
+        placement?: string;
+
+        /**
+         * When true, the popover will appear immediately when the controller connects.
+         */
+        autoShow?: boolean;
```
aaronshekey, line 379: How difficult would it be to enable:

``` js
/**
* When set, `data-s-popover-toggle-class` will be set to this value on the controller element.
*/
toggleClass?: string;
```

I gave it a whirl. It didn't work on the first click, but on subsequent clicks, it allowed adding `is-selected` as expected on the originating element.

bnickel (approved):

lib/ts/controllers/s-popover.ts
b-kelly: This one is tricky-ish. I left the original behavior intact for when the visible class was on a popover at initialization. In this case, the popover did _not_ have the toggle class added. That _should_ be fixed in #616, where we purposefully introduce the regression along with the new `hide-on-outside-click` interaction settings. I'm punting here purposefully and will double check it on #616.

PR #616, opened 2021-03-19
Popover hide-on-outside-click support

b-kelly:
Adds `data-s-popover-hide-on-outside-click` support from #462.

95% of the code is from the previous PR, with the exception of:
* fixing typo in the name (`hide-on-ouside-click`)
* adding new `after-dismissal` value to mimic what `.is-visible` does currently

Additionally, I made `auto-show` behave like a regular popover, in regards to outside clicking. **This is a departure from the current is-visible behavior**. `.is-visible` itself is unaffected (no regressions in behavior), but I suggest we use this opportunity to officially deprecate using it in this manner and migrate all instances to using `auto-show`. If desired, I can add a check in the controller to throw an error when `is-visible` is present on an initialized popover.


aaronshekey:
Should we be changing default behavior though? At the very least, it's breaking our persistent popover example here:

image

Should `never` be the default? Or do we add `never` to that example and call it a day?

Clicking anywhere will dismiss this popover, even though we have the explicit close button. For default popovers, this works, but until these changes, you could have more than one popover open on a page. We'll need to check if there are any cases we break, but changing this assumption. Popovers within popovers?

b-kelly:
@aaronshekey I'm ok with changing the default behavior, provided we migrate everything beforehand. That's the primary reason why I opened this as a separate PR from the previous one. For `is-visible` popovers, the default behavior should probably be either `never` or `after-dismissal`. The one in the docs should probably be `never` since it can never be reopened, but I made it toggleable so I could showcase `after-dismissal` (which is the current behavior).

aaronshekey:
Couple of cross-browser issues:

This menu doesn't work in Safari. Also, can we switch to the `s-menu` component when shown?

image


This action doesn't work in Safari either. The close button in the popover probably shouldn't be there, or should be wired up to something. I could go back on that close button, especially in the case where you've shown it by clicking something other than the originating element.
image

aaronshekey:
One more thought. The originating element should maybe toggle like elements within the controller?

image

Clicking this re-shows the popover each time instead of a toggle on / off.

docs/product/components/popovers.html
```diff
@@ -191,9 +204,11 @@
                 data-controller="s-popover"
                 data-s-popover-reference-selector="#reference-element"
                 data-s-popover-placement="auto"
-                data-s-popover-auto-show="true">
+                data-s-popover-auto-show="true"
+                data-s-popover-hide-on-outside-click="if-in-viewport">
```
bnickel: Maybe make this "never" to match the existing behavior?  It seems like "never" is going to be a far more common use-case than "if-in-viewport".

lib/ts/controllers/s-popover.ts
```diff
@@ -30,6 +33,36 @@ namespace Stacks {
...
 
+        /**
+         * Gets whether the element is visible in the browser's viewport.
+         */
+        get isInViewport() {
+            const element = this.popoverElement;
+            if (!this.isVisible || !element) { return false; }
+
+            // From https://stackoverflow.com/a/5354536.  Theoretically, this could be calculated using Popper's detectOverflow function,
+            // but it's unclear how to access that with out current configuration.
```
```suggestion
            // but it's unclear how to access that with our current configuration.
```

bnickel:
I wouldn't want "never" to be the default case.  There are four types of popovers I can think of currently on the site:

- Marketing/onboarding popovers that appear on page load.
- Menu popovers that are toggled on click.
- The "You just did a thing" popover that appears after following a question.
- Tooltips.

Ignoring tooltips, the "default" popover in my opinion is the menu one, and those should almost always close on outside click.  The only one that really shouldn't is marketing/onboarding, and those can be handled via the non-brief `data-s-popover-auto-show="true" data-s-popover-hide-on-outside-click="???"` pair.

bnickel:
As for "Click here for help", here's our options:

1. We could prevent it from disappearing by simply listening for `s-popover:hide`, checking if the button is/contains the dispatcher, and preventing default in that case.
2. We could make it toggle by checking for the popover's visibility on click and doing the appropriate thing *BUT* we would run into trouble because hiding the popover would be indistinguishable from an outside click to the `s-popover:hide` callback.  We could address this by:
    a. Having `hide` be `hide(dispatcher, reason)`, calling `hide(dispatcher, "outside-click")`, and passing that into the `s-popover:hide` event.  The logic from 1 would then only prevent default if it was the dispatcher and the reason was `outside-click`.
    b. Having a temporary variable `togglingHelp` in the help controller, it would set it to true before toggling visibility and false afterwards.  The logic from 1 would then only prevent default if `togglingHelp` was false.

I generally don't think there's a good use case where a button outside of the popover would function as a toggle.  (Showing on click elsewhere seems pretty flimsy as well but I needed an example.)  It's also dramatically simpler if we just go with 1.

bnickel (approved):
lgtm

docs/product/components/popovers.html
aaronshekey: Yep, co-signed on this one.

aaronshekey (changes requested):
~Safari isn't playing along on this one. Works great in Chrome / Brave / Firefox.~ Turns out this was a total lie. No idea what I was seeing there.

docs/product/components/popovers.html
b-kelly: Yup, can make this change. I set it to `if-in-viewport` purely to have a place to test it and forgot to change it back.

aaronshekey (approved):

PR #627, opened 2021-04-09
Remove `.s-btn__lg` and `.s-btn__xl`

aaronshekey:
These button variations are not used in the product, and they're impractically / comically huge. This removes them in favor of just using `s-btn__md` when we need a little extra spacing.

### DIFF
--- a/docs/_data/buttons.json
+++ b/docs/_data/buttons.json
@@ -239,16 +239,6 @@
       "size": "Medium",
       "fs": "17px",
       "class": "s-btn__md"
-    },
-    {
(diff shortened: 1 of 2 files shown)


double-beep:
They're actually used in [Tour](https://stackoverflow.com/tour) and this PR broke them.

aaronshekey:
That's right! I meant to switch those to `s-btn__md`! Thanks for the reminder!

aaronshekey:
Fixed in prod. Hopefully it gets deployed today :v:

PR #631, opened 2021-04-13
Explicitly load the stacks-internals configuration

aaronshekey:
This allows us to use things like `s-tag__muted` as a mixin elsewhere in our Less files.

### DIFF
--- a/lib/css/components/_stacks-tags.less
+++ b/lib/css/components/_stacks-tags.less
@@ -11,8 +11,6 @@
 //      Any tag we use will always modify the same items, so to save time,
 //      here's a mixin to state all that CSS.
 //  ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


lib/css/components/_stacks-tags.less
```diff
@@ -11,8 +11,6 @@
 //      Any tag we use will always modify the same items, so to save time,
 //      here's a mixin to state all that CSS.
 //  ---------------------------------------------------------------------------
-& { // create a single scope to load the config into, at least for now
-#stacks-internals #load-config();
 
 .tag-styles(@border: transparent, @background: transparent, @color: inherit) {
```
benjamin-hodgson: Suggest renaming this to `.s-tag-styles`

benjamin-hodgson (approved):
LGTM

PR #634, opened 2021-04-20
Add appearance:button to s-btn

benjamin-hodgson:
`<input type="submit" class="s-btn s-btn__primary">` looks like this on my iphone:

Untitled

Looks like it's coming from the default `-webkit-appearance: push-button` in the user agent stylesheet.

### DIFF
--- a/lib/css/components/_stacks-buttons.less
+++ b/lib/css/components/_stacks-buttons.less
@@ -34,7 +34,9 @@
     text-decoration: none;
     cursor: pointer;
     user-select: none;
-    
+
+    // Guard against the difference in configurable resets
+    // Eric Meyer's reset does not include this, while normalize.css does
     -webkit-appearance: button;
     -moz-appearance: button;
     appearance: button;
--- a/lib/css/components/_stacks-buttons.less
+++ b/lib/css/components/_stacks-buttons.less
@@ -34,6 +34,10 @@
     text-decoration: none;
     cursor: pointer;
     user-select: none;
+    
+    -webkit-appearance: button;
+    -moz-appearance: button;
+    appearance: button;
 
     // Override for buttons having inline-block by default
     &.grid {


aaronshekey:
What version of iOS are you running? I'm not reproducing this one in:

https://codepen.io/aaronshekey/pen/mdRoBQx

aaronshekey:
Is this overly-specific CSS in `Careers` that's trolling you?

benjamin-hodgson:
No I don't think so. I don't see anything like that in the inspector. Overriding the `-webkit-appearance` does fix the bug. Plus this is Jobs so most of the CSS comes from SO.

benjamin-hodgson:
I'm on iphone 12 pro, ios 14.4.2. Happens in both Chrome and Safari (although that's not surprising since they're basically the same browser)

aaronshekey:
Does that Codepen have the same issue? ☝️

benjamin-hodgson:
Codepen looks correct, but in the inspector I see a rule for `[type="submit"] { -webkit-appearance: button; }`, which wasn't there on Jobs. Looks like the code is here: https://github.com/StackExchange/Stacks/blob/3877584721bc8ffd1ad9d07f39ffae176eb2bf56/lib/css/base/_stacks-reset-normalize.less#L282. Why is that rule missing in production?

aaronshekey:
> Why is that rule missing in production?

Oh no, lol. I was just talking about this today. Stacks, for some historical reason, allows for configuration of which reset it uses. I believe that Jobs & Talent ships with `reset.css` and not `normalize.css`. By default, Stacks chooses `normalize.css`, but I'm actively exploring [not including either](https://github.com/StackExchange/Stacks/pull/609) as it's 2021.

Ok, happy to add this in. Thanks for helping me pin down what was causing this.

PR #635, opened 2021-04-23
Untrack compiled JavaScript

aaronshekey:
This adds the compiled JS to our .gitignore and removes from the repo to reduce the number of conflicts we have. The source files will still be included in `/lib/` and compiled files will still be included in the npm package.


b-kelly (approved):
👍

PR #636, opened 2021-04-25
Switch to lists in navigations

### LINKED ISSUE #607: s-navigation should work cleanly when there's a list inside the nav
The [MDN docs for `nav`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav) have an `ol` inside the `nav`, and each navigation link inside an `li`. [The docs for `s-navigation`](https://stackoverflow.design/product/components/navigation/) omit the list; if I do use a list I have to manually use `list-reset` on the list and `d-inline-block` on the list items. It also breaks `s-navigation__scroll`.

Per [a discussion in chat](https://stackexchange.slack.com/archives/C27RWNQN9/p1615403488108900), it sounds like the version with the list is better for accessibility, so we should at least make sure that works with `s-navigation`, and ideally with minimal additional markup.

**Example**:

```html
<nav class="s-navigation s-navigation__scroll">
    <ol>
        <li><a class="s-navigation--item">Link 1</a></li>
        <li><a class="s-navigation--item">Link 2</a></li>
        <li><a class="s-navigation--item">Link 3</a></li>
    </ol>
</nav>
```

**Expected behaviour**:
1. (Nice-to-have) The nav is styled as a horizontal nav, with no bullets, without having to use `list-reset` on the `ol` and `d-inline-block` on the `li`s. Likewise, `s-navigation__vertical` behaves correctly
2. `s-navigation__scroll`'s overrides work correctly (when container is narrow, items don't wrap to a second line and scrollbar appears).
3. The docs are updated to recommend (or require??) the use of an `ol` or `ul` inside the `nav`.

aaronshekey:
Closes #607 and switches the docs to be list-based.

### DIFF
--- a/docs/product/components/navigation.html
+++ b/docs/product/components/navigation.html
@@ -35,31 +35,37 @@ description: Our navigation component is a collection of pill-shaped buttons tha
     {% header "h3", "Default" %}
     <div class="stacks-preview">
 {% highlight html %}
-<nav class="s-navigation">
-    <a href="…" class="s-navigation--item is-selected">…</a>
-    <a href="…" class="s-navigation--item">…</a>
-    <a href="…" class="s-navigation--item">…</a>
-    <a href="…" class="s-navigation--item">…</a>
-    <a href="…" class="s-navigation--item">…</a>
+<nav>
+    <ul class="s-navigation">
+        <li><a href="…" class="s-navigation--item is-selected">…</a></li>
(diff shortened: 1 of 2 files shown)


benjamin-hodgson:
LGTM, and looks like it should be backwards compatible too

aaronshekey:
Indeed it is. You can see the main navigation on the Stacks docs was untouched and still presents exactly the same.

PR #640, opened 2021-05-07
Add support for multiple expandables from a single controller

gkoscky:
Allow a single `s-expandable-control` button to control the visibility of multiple `s-expandable` elements by supporting a space-separated list of Ids in `aria-controls`

```html
<button class="s-btn s-btn__filled mb8"
    data-controller="s-expandable-control"
    data-s-expandable-control-toggle-class="is-selected"
    aria-expanded="false"
    aria-controls="expandable-example1 expandable-example2">
        Show paragraphs
</button>
<div class="s-expandable" id="expandable-example1">
    <div class="s-expandable--content">
        <p class="mb8">Lorem ipsum dolor sit amet</p>
    </div>
</div>

<div class="s-expandable" id="expandable-example2">
    <div class="s-expandable--content">
        <p class="mb8">consectetur adipiscing elit</p>
    </div>
</div>
```

I didn't want to mess with the documentation too much, but I'm happy to expand on it if needed.


lib/ts/controllers/s-expandable-control.ts
```diff
@@ -84,7 +84,8 @@
             if (!attr) {
                 throw "couldn't find controls"
             }
-            const result = document.getElementById(attr);
+            const controlledElements = `#${attr.replace(/ /g, ", #")}`;
+            const result = Array.from(document.querySelectorAll(controlledElements));
```
bnickel: https://www.w3.org/TR/2017/REC-wai-aria-1.1-20171214/#typemapping leads to https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#space-separated-tokens which leads to https://infra.spec.whatwg.org/#ascii-whitespace

> **ASCII whitespace** is U+0009 TAB, U+000A LF, U+000C FF, U+000D CR, or U+0020 SPACE.

So, we need to protect against:

```
aria-controls="   foo
    bar"
```

`attr.split(/\s+/g).filter(s => s.length)` can get a list of ids.

We should also be careful because not everything that is a valid id makes for a valid query selector.  E.g., `aria-controlls="foo.bar"`

The following should give the same results.

```suggestion
            const result = attr.split(/\s+/g).map(s => document.getElementById(s)).filter(e => e);
```

lib/ts/controllers/s-expandable-control.ts
```diff
@@ -127,7 +128,9 @@
                 }
             }
             this.element.setAttribute("aria-expanded", newCollapsed ? "false" : "true");
-            this.controlledCollapsible.classList.toggle("is-expanded", !newCollapsed);
+            for (let controlledElement of this.controlledCollapsible) {
```
bnickel: We should make the property plural.

lib/ts/controllers/s-expandable-control.ts
```diff
@@ -69,8 +69,8 @@
         // for non-checkable elements, the initial source of truth is the collapsed/expanded
         // state of the controlled element (unless the element doesn't exist)
         _isCollapsedForClickable() {
-            var cc = this.controlledCollapsible;
-            return cc ? !cc.classList.contains("is-expanded") : this.element.getAttribute("aria-expanded") === "false";
+            var cc = this.controlledCollapsibles;
+            return cc ? !cc.every(element => element.classList.contains("is-expanded")) : this.element.getAttribute("aria-expanded") === "false";
```
bnickel:
```suggestion
            return cc.length > 0 ? !cc.every(element => element.classList.contains("is-expanded")) : this.element.getAttribute("aria-expanded") === "false";
```

`this.controlledCollapsibles` will never be null.  The equivalent to the before behavior is checking if we're empty.  That said, we never hit that condition before because we just threw within `controlledCollapsibles`

lib/ts/controllers/s-expandable-control.ts
```diff
@@ -79,12 +79,12 @@
...
 
-        get controlledCollapsible() {
+        get controlledCollapsibles() {
             const attr = this.element.getAttribute("aria-controls");
             if (!attr) {
                 throw "couldn't find controls"
             }
-            const result = document.getElementById(attr);
+            const result = attr.split(/\s+/g).map(s => document.getElementById(s)).filter((e): e is HTMLElement => !!e);
             if (!result) {
```
bnickel:
```suggestion
            if (result.length == 0) {
```

Like the above, this will never fail anymore, and we could preserve the existing behavior by throwing on length.

lib/ts/controllers/s-expandable-control.ts
```diff
@@ -145,12 +147,14 @@
             // attribute; for checkable controls this also means setting the `is-collapsed` class
             this.element.setAttribute("aria-expanded", this.isCollapsed() ? "false" : "true");
             if (this.isCheckable) {
-                var cc = this.controlledCollapsible;
+                var cc = this.controlledCollapsibles;
                 if (cc) {
                     var expected = !this.isCollapsed();
-                    var actual = cc.classList.contains("is-expanded");
+                    var actual = cc.some(element => element.classList.contains("is-expanded"));
                     if (expected !== actual) {
```
bnickel:
```suggestion
                    if (cc.some(element => element.classList.contains("is-expanded") !== expected)) {
```

We want to fix this if _any_ element doesn't match our expectations.  As implemented, it would work for collapsing if something is out of sync but not expanding.

bnickel (changes requested):
Some small suggestions.  I've regression tested and everything is working for the non-broken single element code that in the docs.  It would be good to have an example of multiple expandables in the docs just since it's the easiest place to test, but I'm fine without it.

gkoscky:
@bnickel I just added a new section to the documentation page 👍🏼
I Initially just wasn't sure if this was something noteworthy enough to warrant a new example.

bnickel (approved):
Screen Shot 2021-05-13 at 1 08 51 PM

docs/product/components/expandable.html
```diff
@@ -44,16 +44,66 @@
             </div>
         </div>
     </div>
+
+    {% header "h3", "Multiple expandable elements" %}
+    <p class="stacks-copy">It's also possible to control the state multiple elements from a single  <code class="stacks-code">s-expandable-control</code> by listing all the <code class="stacks-code">ids</code> for the controlled elements in the <code class="stacks-code">aria-controls</code> attribute.</p>
```
b-kelly: nit: typo

```suggestion
    <p class="stacks-copy">It's also possible to control the state of multiple elements from a single  <code class="stacks-code">s-expandable-control</code> by listing all the <code class="stacks-code">ids</code> for the controlled elements in the <code class="stacks-code">aria-controls</code> attribute.</p>
```

lib/ts/controllers/s-expandable-control.ts
```diff
@@ -79,13 +79,13 @@
...
-        get controlledCollapsible() {
+        get controlledCollapsibles() {
             const attr = this.element.getAttribute("aria-controls");
             if (!attr) {
                 throw "couldn't find controls"
             }
-            const result = document.getElementById(attr);
-            if (!result) {
+            const result = attr.split(/\s+/g).map(s => document.getElementById(s)).filter((e): e is HTMLElement => !!e);
+            if (result.length == 0) {
```
b-kelly: optional: very minor code cleanup. Also, nice use of type guards here!

```suggestion
            const result = attr.split(/\s+/g)
                .map(s => document.getElementById(s))
                .filter((e): e is HTMLElement => !!e);
            if (!result.length) {
```

lib/ts/controllers/s-expandable-control.ts
```diff
@@ -79,13 +79,13 @@
         };
 
 
-        get controlledCollapsible() {
+        get controlledCollapsibles() {
             const attr = this.element.getAttribute("aria-controls");
             if (!attr) {
                 throw "couldn't find controls"
```
b-kelly: While we're in here making changes, I'm going to change this error message a tad.

```suggestion
                throw `[aria-controls="targetId1 ... targetIdN"] attribute required`;
```

lib/ts/controllers/s-expandable-control.ts
```diff
@@ -145,12 +147,13 @@
             // attribute; for checkable controls this also means setting the `is-collapsed` class
             this.element.setAttribute("aria-expanded", this.isCollapsed() ? "false" : "true");
             if (this.isCheckable) {
-                var cc = this.controlledCollapsible;
+                var cc = this.controlledCollapsibles;
                 if (cc) {
```
b-kelly: `cc` will always be truthy here:

```suggestion
                if (cc.length) {
```

lib/ts/controllers/s-expandable-control.ts
```diff
@@ -69,8 +69,8 @@
         // for non-checkable elements, the initial source of truth is the collapsed/expanded
         // state of the controlled element (unless the element doesn't exist)
         _isCollapsedForClickable() {
-            var cc = this.controlledCollapsible;
-            return cc ? !cc.classList.contains("is-expanded") : this.element.getAttribute("aria-expanded") === "false";
+            var cc = this.controlledCollapsibles;
+            return cc.length > 0 ? !cc.every(element => element.classList.contains("is-expanded")) : this.element.getAttribute("aria-expanded") === "false";
```
b-kelly, line 74: nit: adding comment for myself here. Commit as desired

```suggestion
            // the element is considered collapsed if *any* target element is collapsed
            return cc.length > 0 ? !cc.every(element => element.classList.contains("is-expanded")) : this.element.getAttribute("aria-expanded") === "false";
```

lib/ts/controllers/s-expandable-control.ts
```diff
@@ -145,12 +147,13 @@
...
             this.element.setAttribute("aria-expanded", this.isCollapsed() ? "false" : "true");
             if (this.isCheckable) {
-                var cc = this.controlledCollapsible;
+                var cc = this.controlledCollapsibles;
                 if (cc) {
                     var expected = !this.isCollapsed();
-                    var actual = cc.classList.contains("is-expanded");
-                    if (expected !== actual) {
-                        cc.classList.toggle("is-expanded", expected);
+                    if (cc.some(element => element.classList.contains("is-expanded") !== expected)) {
```
b-kelly: This looks fine to me, but the inconsistent usage of some (here) vs every (isCollapsed) messed with my brain a little, so I added some comments. Commit as desired:

 ```suggestion
                    // if any element does not match the expected state, set them all to the expected state
                    if (cc.some(element => element.classList.contains("is-expanded") !== expected)) {
```

b-kelly (approved):
lgtm, nice work. Thanks for picking this up!

lib/ts/controllers/s-expandable-control.ts
gkoscky: Thanks! All credit goes to Sublime's TypeScript plugin telling me "object is possibly null" in places until I figured it out what type guards even were 😅

aaronshekey (approved):

PR #642, opened 2021-05-10
Additional font tweaks

aaronshekey:
Rolling up any tweaks we may need to the font stack as a hotfix.

These changes can be [previewed on the Stacks documentation](https://deploy-preview-642--stacks.netlify.app/).

### DIFF
--- a/lib/css/base/_stacks-body.less
+++ b/lib/css/base/_stacks-body.less
@@ -23,10 +23,6 @@ body {
     font-family: var(--theme-body-font-family);
     font-size: @fs-base;
     line-height: @lh-base;
-
-    // As of macOS 11.3, you can no longer disable font smoothing.
-    // This does it for us at the CSS-level.
-    -webkit-font-smoothing: antialiased;
 }
 
 #stacks-internals #screen-sm({
--- a/lib/css/exports/_stacks-constants-type.less
+++ b/lib/css/exports/_stacks-constants-type.less
@@ -22,7 +22,7 @@
 @ff-mono:
     ui-monospace, // San Francisco Mono on macOS and iOS
     "Cascadia Mono", "Segoe UI Mono", // Newer Windows monospace fonts that are optionally installed. Most likely to be rendered in Consolas
-    "Liberation Sans Mono", // Linux
+    "Liberation Mono", // Linux
     Menlo, Monaco, Consolas, // A few sensible system font choices
     monospace; // The final fallback for rendering in monospace.
 
--- a/lib/css/exports/_stacks-constants-type.less
+++ b/lib/css/exports/_stacks-constants-type.less
@@ -16,6 +16,7 @@
 @ff-sans:
     -apple-system, BlinkMacSystemFont, // San Francisco on macOS and iOS
     "Segoe UI", // Windows
+    Arial, // For Linux machines, let's continue to specify Arial and let their machine choose the equivalent, status quo < 0.64.0
     sans-serif; // The final fallback for rendering in sans-serif.
 @ff-serif: Georgia, Cambria, "Times New Roman", Times, serif;
 @ff-mono:
--- a/lib/css/exports/_stacks-constants-type.less
+++ b/lib/css/exports/_stacks-constants-type.less
@@ -16,14 +16,12 @@
 @ff-sans:
     -apple-system, BlinkMacSystemFont, // San Francisco on macOS and iOS
     "Segoe UI", // Windows
-    "Ubuntu", // Ubuntu
     "Roboto", "Noto Sans", "Droid Sans", // Chrome OS and Android with fallbacks
     sans-serif; // The final fallback for rendering in sans-serif.
 @ff-serif: Georgia, Cambria, "Times New Roman", Times, serif;
 @ff-mono:
     ui-monospace, // San Francisco Mono on macOS and iOS
     "Cascadia Mono", "Segoe UI Mono", // Newer Windows monospace fonts that are optionally installed. Most likely to be rendered in Consolas
-    "Ubuntu Mono", // Ubuntu
     "Roboto Mono", // Chrome OS and Android
     Menlo, Monaco, Consolas, // A few sensible system font choices
     monospace; // The final fallback for rendering in monospace.
--- a/lib/css/exports/_stacks-constants-type.less
+++ b/lib/css/exports/_stacks-constants-type.less
@@ -14,7 +14,7 @@
 //  $   FONT FAMILY (ff-)
 //  ----------------------------------------------------------------------------
 @ff-sans:
-    system-ui, -apple-system, BlinkMacSystemFont, // San Francisco on macOS and iOS
+    -apple-system, BlinkMacSystemFont, // San Francisco on macOS and iOS
     "Segoe UI", // Windows
     "Ubuntu", // Ubuntu
     "Roboto", "Noto Sans", "Droid Sans", // Chrome OS and Android with fallbacks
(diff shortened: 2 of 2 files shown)


MatthiasPortzel:
I think it's worth noting, `ui-monospace` is only supported by Safari [[1]](https://caniuse.com/extended-system-fonts). This means macOS users on browsers other than Safari (the majority) don't get SF Mono. Unless they have another one of these fonts installed, they'll fall-back to Menlo.

aaronshekey:
> Unless they have another one of these fonts installed, they'll fall-back to Menlo.

This is what I've found as well. Currently, there doesn't seem to be a way outside of that progressively-enhanced `ui-monospace` CSS property to actually get San Francisco Mono to display. I've spec'd that in the hopes that other browsers on macOS will adopt that. Thanks for calling it out.

LunarWatcher:
Please don't give us Arial back everywhere: https://meta.stackexchange.com/questions/364048/we-are-switching-to-system-fonts-on-may-10-2021?cb=1#comment1214996_364048 https://meta.stackexchange.com/questions/364048/we-are-switching-to-system-fonts-on-may-10-2021?cb=1#comment1215982_364097

I'm starting to see the value more and more in just letting the browser determine the font - readable is subjective enough for no single font stack to universally work.

fireattack:
@aaronshekey  Can't leave comments in https://meta.stackexchange.com/a/364291/998555 due to reputation, so I think I would leave a comment here.

The reason you see discrepancy between Firefox and Chrome is because Firefox [doesn't support](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#browser_compatibility) `system-ui` keyword at all. It simply ignores it. 

Also, the poor result in non-English Windows is the exactly reason why GitHub, Twitter, etc. have tried and then immediately reverted their change of font stack of including system-ui.

aaronshekey:
> Also, the poor result in non-English Windows is the exactly reason why GitHub, Twitter, etc. have tried and then immediately reverted their change of font stack of including system-ui.

Fair enough! I feel pretty confident leaving this off now.

aaronshekey:
> Please don't give us Arial back everywhere

Based on the order of things, this should only affect Linux distros, and lets us maintain the status quo across all Linux installs. One person's "Arial" will be "Liberation Sans" on Ubuntu. It'll be "Roboto" on Android, etc.

aaronshekey:
After my testing in Linux distros, here are the following font mappings with and without Arial. Looks like Chrome actually installs Liberation Sans, at least it wanted to on Debian, but I failed to complete the install.

So, the logic in including `Arial` here, is for some normalization across Linux distros between Chrome and Firefox ¯\_(ツ)_/¯ 

| OS | Firefox with Arial | Firefox without Arial | Chrome with Arial | Chrome without Arial |
| --- | --- | --- | --- | --- |
| **Ubuntu** | Liberation Sans | DejaVu Sans | Liberation Sans | Liberation Sans |
| **Fedora** | Liberation Sans | DejaVu Sans | Liberation Sans | Liberation Sans | 

I plan on testing Debian, and CentOS, but I'm struggling to get Chrome installed on Debian, and CentOS doesn't seem to like my network adapter.

aaronshekey:
This branch is now spec'ing "Liberation Sans" and "Liberation Mono" for a wide variety of Linux desktop distros.

Elliot2560:
Actually I think the substitution for Arial on Linux depends on a few other things too. If you have Arimo installed, which you would if you installed all the Google web fonts, or if you have the `fonts-croscore` package on Ubuntu, then Arimo takes precedence. This is my experience on Arch. I'm not entirely sure how it works on Chrome, but it seems to me that Firefox uses the substitution rules set by fontconfig, and the precedence goes `Arimo > Liberation Sans > Albany > Albany MT`, I think. (Check out [this conf file](https://github.com/freedesktop/fontconfig/blob/master/conf.d/30-metric-aliases.conf) where it's specified.)

Though in the end these could all be reasons why it makes (more) sense to specify Liberation Sans and Liberation Mono. I can also confirm that Liberation Sans is a dependency of both Firefox and Chromium.

aaronshekey:
@LunarWatcher Care to chime in or are you good with 👎 each commit? 😛

LunarWatcher:
I've come to fundamentally dislike Arial and clones for their awful readability when contextual reading fails. That often happens on abbreviations, or more relevant to SO, on certain posts with enough typos for every letter to be in question. (Or just whenever my brain decides to nope out and crash contextual reading) 

That being said, it seems that opinion isn't a mainstream one (people on meta are practically begging to have Arial back), and it's therefore unlikely to lead anywhere. So for now, I'm good with the 👎s and silently disappearing into userstyles and whatever layout breaking that entails.

pxeger:
Please just stop trying to be clever. Use `sans-serif`. Just `sans-serif`. It exists for a reason - the browser does it very damn best to make sure it's accessible, quick to load, and looks good. And it's customisable on most browsers without needing unreliable user styles.

fireattack:
That's something I never understand about all these "font stacks" either. I honestly feel they're just re-inventing the wheels.

Sure, `sans-serif` may not be the "system fonts" on some OSs by default, but all the modern browsers have the ability to let the user adjust the "default" font, which is exactly what `sans-serif` would choose to use. 

IMO this is just straight up superior than forcing the "system font" using hard-coded rule (if one really want to use system font, they can always set it as their browser's default font. Not to mention, in some OSs like Android/iOS, the default font is already system font).

aaronshekey:
As I explained in my [post on meta](https://meta.stackexchange.com/questions/364048/we-are-switching-to-system-fonts-on-may-10-2021):

> Apple’s San Francisco and Microsoft’s Segoe both look great on retina displays, have more expressive weights, and improve readability across all contexts. With more weights, we can have better hierarchy. We can also get more expressive. For example, in dark mode, we could drop our font weights a bit for better readability.

If I choose `sans-serif`, it will render as Helvetica on iOS and macOS. It will render as Arial on Windows. These typefaces have none of those benefits I mentioned, none of the font hinting built into these new typefaces. We'd rather not wait until the operating systems update their underlying definition of `sans-serif`. Additionally, especially on Linux, it varies per browser.

> Accessible, quick to load, and looks good

Yep, these are our goals. This font stack is designed to choose the newest, most accessible _screen native_ typeface as possible. :v:

fireattack:
>It will render as Arial on Windows. These typefaces have none of those benefits I mentioned, none of the font hinting built into these new typefaces

Like I said, It will render as Arial because that's the default font of the *browser*: 

Chrome: 

image

Firefox:

image

It has nothing to do with Windows. These default choices are hard-coded into browser due to historical reasons (there are tickets on both Chromium and Firefox's bug trackers about changing them, but no enough traction yet.)

Users can change it to Segoe if they don't like Arial. 

`Sans-serif` simply uses these settings (same for `serif` and `monospace`). System's built-in font fallback will only be engaged if it can't display certain glyphs. 

 I assume it would be similar for MacOS, but I don't use it. 

You do have a case for iOS though.

Edit: by the way, all these font stacks have the drawback that seldom got talked about: if someone happened to have installed some of these fonts in the rule on their OS, it has the chance to mess up the display even further by showing a font that is not from their OS, which invalidates the "native look" point.

aaronshekey:
I hear you on browser-default fonts. But remember, we want to modernize our fonts without too much user intervention and feel a bit more native to their OS.

stamminator:
What I hear from @aaronshekey is that the hesitation of browser vendors to default their browsers to more modern fonts (with all the advantages they entail) should not prevent SO from benefitting from those fonts in as "native", per-platform a way as possible. Provided SO can work out any kinks in a timely manner, I don't see why this is a bad take.

PR #643, opened 2021-05-11
Bump the base font size by 1px

aaronshekey:
This PR beefs up our base font size by `1px`. Our fonts are comically small across the site, and now that we've changed fonts, we should be able to get away with exploring this.

Also deprecates `.fs-category`

You can preview these changes in our [staging environment](https://deploy-preview-643--stacks.netlify.app/).


bnickel:
This is going to break _something_. Maybe tightly spaced internationalize buttons, maybe some meticulously measured marketing, but definitely something.  Only time will tell, I suppose.

aaronshekey:
No doubt. What's nice is that on the Stacks side, everything is based on `rem` and `em`, so things generally scale to the base font size. Right now, I've locked everything to the existing site width, but it could be powered entirely by the font size. Kinda neat, and worth exploring.

We could even expose this to the user.

dancormier:
Old branch with a boatload of conflicts 😬 Closing to revisit at another time.

PR #645, opened 2021-05-12
Switch to CSS variable on monospace font variables

aaronshekey:
Let's quit hardcoding the fonts and render them at runtime.


b-kelly (approved):
lgtm

PR #647, opened 2021-05-19
Fixes for #619

### LINKED ISSUE #619: Button disabled states should allow for interaction
In current versions of Stacks, we disable pointer events with our disabled button states. This is so we don't have to override a bunch of hover / active / focused states. In doing so, we disallow any interaction. We may want to show our users things like a tooltip as to why the button is disabled, so we shouldn't completely remove all interaction there.

aaronshekey:
This makes sure we can still interact with `disabled` buttons (within reason). Often, we want to add a popover explaining a button's disabled state, and this should allow us to do so.

Originally, I'd added `pointer-events: none` to avoid the hover state, but then, of course, that doesn't allow us to interact with the button.

Closes #619

### DIFF
--- a/lib/css/components/_stacks-buttons.less
+++ b/lib/css/components/_stacks-buttons.less
@@ -52,7 +52,7 @@
     //  Default Style (Clear)
     color: @button-color;
 
-    &:hover,
+    &:hover:not([disabled]):not(.is-selected),
     &:focus,
     &:active {
         color: @button-hover-color;
@@ -71,7 +71,7 @@
 
     &[disabled] {
         opacity: .5;
-        pointer-events: none;
+        cursor: default;
         box-shadow: none !important;
     }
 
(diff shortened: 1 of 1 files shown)


bnickel:
I don't think this will actually fix things.  Disabled buttons don't have `mouseover` or `mouseout` events so you can't use `s-tooltip` directly.

The solution would still have to be something like:

```html
<span data-controller="s-tooltip" title="You just can't, that's why.">
    <button disabled>Delete all the things</button>
</span>

bnickel:
If this fix doesn't work, we should add some documentation and workaround examples on `s-tooltip`.

aaronshekey:
I think that’s absolutely the approach and this is definitely HTML acting as designed. Was past me was correct about lazily relying on pointer-events: none in the context of disabled buttons? Possibly!

aaronshekey:
Confirmed that the standard click popover doesn't work either. Will approach this differently via a documentation change.

PR #650, opened 2021-05-20
s-expandable: Prevent click after keydown on Firefox

### LINKED ISSUE #649: `s-expandable-controller` instantly collapses on Firefox with spacebar
This is reproducible with https://stackoverflow.design/product/components/expandable/#basics

Go to that page and tab to Show first paragraph.  Clicking <kbd>Enter</kbd> expands or collapses the element.  Clicking <kbd>Spacebar</kbd> on the other hand starts, then undoes the action, while also unfocusing the button.

This is a user-agent specific bug triggered by https://bugzilla.mozilla.org/show_bug.cgi?id=1487102 and the logic in the code:

https://github.com/StackExchange/Stacks/blob/42a5825cb3501bb636c6671030faf9983dd89aea/lib/ts/controllers/s-expandable-control.ts#L117-L127

When a keyboard user clicks space or Enter, the browser will emit a `keydown` event followed by a `click` event.  When a mouse user clicks, we just get `click`.  The above code uses that knowledge to try to prevent focus rings on the button when we click with the mouse while keeping them if we click with a keyboard.  When a matching `keydown` occurs, we perform the action and call `preventDefault` so `click` never fires.  When `click` occurs, we know it was from a mouse so we are free to blur the element.

Unfortunately, https://bugzilla.mozilla.org/show_bug.cgi?id=1487102  means `preventDefault` doesn't fire and we have a second event listener undo the first.

One solution would be to use a debouning timestamp.  E.g., if we have a click event within 300ms or so of a `keydown` event, we suppress it.  We unfortunately can't just debounce events in general because the delay between `keydown` and `click` are as high as two click events.

bnickel:
Closes #649.

The 300ms delay was chosen as an arbitrary number higher than the 160ms delay I observed in Firefox. If you keyboard click and then quickly mouse click, you'll see your mouse click get rejected (i.e., do nothing) but that's not typical user behavior.

### DIFF
--- a/lib/ts/controllers/s-expandable-control.ts
+++ b/lib/ts/controllers/s-expandable-control.ts
@@ -51,6 +51,7 @@
         private events!: string[];
         private isCheckable!: boolean;
         private isRadio!: boolean;
+        private lastKeydownClickTimestamp: number = 0;
 
         initialize() {
             if (this.element.nodeName === "INPUT" && ["radio", "checkbox"].indexOf((<HTMLInputElement>this.element).type) >= 0) {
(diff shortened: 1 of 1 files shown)


lib/ts/controllers/s-expandable-control.ts
```diff
@@ -120,8 +121,20 @@
...
+                // Prevent "click" events from toggling the expandable within 300ms of "keydown".
+                // e.preventDefault() should have done the same, but https://bugzilla.mozilla.org/show_bug.cgi?id=1487102
+                // doesn't guarantee it.
+                if (e.type == "keydown") {
+                    this.lastKeydownClickTimestamp = Date.now();
+                } else if (e.type == "click" && Date.now() - this.lastKeydownClickTimestamp < 300) {
+                    return;
+                }
+
+
```
b-kelly: nit: double newline

```suggestion

```

b-kelly (approved):
Oh, this is a "fun" one. This approach looks fine to me. It's simple and gets the job done. If we _really_ wanted to complicate matters, we could probably use the same binding/unbinding techniques we use in modals/popovers, but that would probably be overkill here. Thank you for putting comments in w/ links to the browser bug!

lgtm

PR #651, opened 2021-05-20
s-modal: Fix keyboard focus on complex views

bnickel:
Right now, we emit show/hide events on `this.element` but listen for them on `this.modalTarget`.  If those two properties aren't the same element, we fail to bring focus into the modal or return focus when hiding the modal (and probably other things).

This fixes it by emitting the event from `modalTarget`.  We could alternatively fix it by binding the events on `element`, but if we made this mistake it's only natural to think consumers would too.

### DIFF
--- a/docs/product/components/modals.html
+++ b/docs/product/components/modals.html
@@ -85,22 +85,22 @@ description: Modals are dialog overlays that prevent the user from interacting w
             <tbody class="fs-caption">
                 <tr>
                     <th scope="row"><code class="stacks-code">s-modal:show</code></th>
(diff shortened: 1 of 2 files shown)


b-kelly (changes requested):
I'm ok with changing the emitter to the modal target. This should not be a breaking change for those listening to the controller element as it is either A) the same element or B) a parent of the modal target and will receive the event when it bubbles.

My only requested change would be to update the docs to match. See the `Events` table [here](https://stackoverflow.design/product/components/modals/#events).

PR #652, opened 2021-05-21
Configure and run stylelint

aaronshekey:
This PR configures stylelint and then runs it against our existing `.less` files.


.stylelintrc
```diff
@@ -0,0 +1,19 @@
...
+    "at-rule-empty-line-before": null,
+    "declaration-empty-line-before": null,
+    "selector-pseudo-element-colon-notation": "single",
+    "no-descending-specificity": null,
+    "selector-disallowed-list": [
+      "/^[.#]js-/",
+      {
+        "message": "js-* selectors can’t have visual styling. https://stackoverflow.design/product/guidelines/javascript/#javascript-classnames"
+      }
+    ],
```
b-kelly: nit: Drop trailing comma in json files

```suggestion
    ]
```

PR #664, opened 2021-05-25
Add `.is-selected` to tags

### LINKED ISSUE #663: Add selected styles to `s-tag`.
Currently, the only way to add a selected treatment to an `s-tag` is via atomic classes. We should introduce a similar API to `is-selected` on the `s-btn` component, and handle the proper theming. 

cc: @benjamin-hodgson

aaronshekey:
Closes #663.

This adds an `.is-selected` state to `s-tag` for added interactivity e.g. managing tag autocomplete. Also makes the tag configuration variables more consistent with buttons, so we'll have to babysit this change in production. @taiwei426 does this satisfy the highlighted ads use case?

### DIFF
--- a/docs/_data/tags-component.json
+++ b/docs/_data/tags-component.json
@@ -29,6 +29,11 @@
       "class": ".s-tag--sponsor",
       "applies": "N/A",
       "description": "A child element within <code class=\"stacks-code\">.s-tag</code> that correctly positions a tag’s sponsor logo."
+    },
+    {
+      "class": ".is-selected",
+      "applies": ".s-tag",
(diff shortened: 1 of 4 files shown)


benjamin-hodgson:
Looks like these styles don't match what we ended up choosing for the tags in job ads (`ba bc-blue-3`). gonna ping @taiwei426 to see what she thinks

taiwei426:
I'm cool with the new design @benjamin-hodgson @aaronshekey

dancormier (approved):
lgtm 🤷‍♂️ 🧨🚀

PR #665, opened 2021-05-28
Adapt tab focus to changes in modals

bnickel:
I have a reusable modal where a couple things can happen between shows:

- The body of the modal gets replaced with new contents, which may or may not contain a link.
- One of two elements that should receive initial focus is hidden.

This surfaces two problems with the current implementation of tab traps:

1. We only capture the list of tabbable elements once, the first time we bind document events.  After that, we just keep reusing that list no matter how the modal's content changes.
2. When using `initialFocusTarget`, Stimulus grabs the first matching element, but doesn't know or care if the element is visible.

To resolve 1, I'm no longer using a stored version of the tabbable list, instead fetching it each time a tab keydown is observed.  Given the nature of modals, I don't expect us to reach a scale where this really impacts users.

To resolve 2, I'm switching from `initialFocusTarget` to `initialFocusTargets` and finding the first visible one using a check from https://stackoverflow.com/a/21696585/860000.  If none is found, it falls back to the first tabbable as before.

### DIFF
--- a/lib/ts/controllers/s-modal.ts
+++ b/lib/ts/controllers/s-modal.ts
@@ -173,27 +173,16 @@ namespace Stacks {
          * Binds tab presses on tabbable items such that tabbing only works within the modal
          */
         private focusInsideModal() {
-
-            // get all tabbable items
-            var allTabbables = this.getAllTabbables();
-
-            if (!allTabbables.length) {
-                return;
-            }
-
-            // focus on the first focusable item within the modal, preferring targets explicitly provided by the DOM.
-
-            var initialFocus = this.firstVisible(this.initialFocusTargets) ?? this.firstVisible(allTabbables);
-            if (initialFocus) {
-                this.modalTarget.addEventListener("s-modal:shown", () => {
-                    // double check the element still exists when the event is called
-                    if (initialFocus && document.body.contains(initialFocus)) {
-                        initialFocus.focus()
-                    }
-                }, {once: true });
-            }
+            this.modalTarget.addEventListener("s-modal:shown", () => {
+                var initialFocus = this.firstVisible(this.initialFocusTargets) ?? this.firstVisible(this.getAllTabbables());
+                initialFocus?.focus();
+            }, {once: true });
         }
 
+
+        /**
+         * Returns keyboard focus to the modal if it has left or is about to leave.
+         */
         private keepFocusWithinModal(e: KeyboardEvent) {
 
             // If somehow the user has tabbed out of the modal or if focus started outside the modal, push them to the first item.
--- a/lib/ts/controllers/s-modal.ts
+++ b/lib/ts/controllers/s-modal.ts
@@ -196,7 +196,7 @@ namespace Stacks {
             }
 
             // If we observe a tab keydown and we're on an edge, cycle the focus to the other side.
-            if (e.keyCode === 9) {
+            if (e.key === "Tab") {
                 var tabbables = this.getAllTabbables();
 
                 var firstTabbable = this.firstVisible(tabbables);
--- a/lib/ts/controllers/s-modal.ts
+++ b/lib/ts/controllers/s-modal.ts
@@ -153,24 +153,32 @@ namespace Stacks {
         }
 
         /**
-         * Gets all elements that could be tabbed to in 
+         * Gets all elements within the modal that could receive keyboard focus.
          */
         private getAllTabbables() {
-            return <HTMLElement[]>Array.from(this.modalTarget.querySelectorAll("[href], input, select, textarea, button, [tabindex]"))
+            return Array.from(this.modalTarget.querySelectorAll<HTMLElement>("[href], input, select, textarea, button, [tabindex]"))
                 .filter((el: Element) => el.matches(":not([disabled]):not([tabindex='-1'])"));
         }
 
+        /**
+         * Returns the first visible element in an array or `undefined` if no elements are visible.
+         */
         private firstVisible(elements: HTMLElement[]) {
             // https://stackoverflow.com/a/21696585
             return elements.find(el => el.offsetParent !== null);
         }
 
+        /**
+         * Returns the last visible element in an array or `undefined` if no elements are visible.
+         */
         private lastVisible(elements: HTMLElement[]) {
-            return [...elements].reverse().find(el => el.offsetParent !== null);
+            return this.firstVisible([...elements].reverse());
         }
 
         /**
-         * Binds tab presses on tabbable items such that tabbing only works within the modal
+         * Attempts to shift keyboard focus into the modal.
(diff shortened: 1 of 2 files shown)


lib/ts/controllers/s-modal.ts
```diff
@@ -153,58 +153,69 @@ namespace Stacks {
...
+        private firstVisible(elements: HTMLElement[]) {
+            // https://stackoverflow.com/a/21696585
+            return elements.find(el => el.offsetParent !== null);
+        }
 
-            if (this.hasInitialFocusTarget) {
-                initialFocus = this.initialFocusTarget;
-            }
+        private lastVisible(elements: HTMLElement[]) {
+            return [...elements].reverse().find(el => el.offsetParent !== null);
```
b-kelly: nit: Good opportunity to reuse existing code

```suggestion
            return this.firstVisible([...elements].reverse());
```

It sucks that `.reverse` mutates the existing array; it'd be much neater without having to copy the array.

lib/ts/controllers/s-modal.ts
```diff
@@ -153,58 +153,69 @@ namespace Stacks {
...
 
         /**
-         * Binds tab presses on tabbable items such that tabbing only works within the modal
+         * Gets all elements that could be tabbed to in 
          */
-        private handleFocusableElements() {
-            // get all tabbable items
-            var allTabbables = Array.from(this.modalTarget.querySelectorAll("[href], input, select, textarea, button, [tabindex]"))
+        private getAllTabbables() {
+            return <HTMLElement[]>Array.from(this.modalTarget.querySelectorAll("[href], input, select, textarea, button, [tabindex]"))
```
b-kelly: optional: Didn't test this locally, but `querySelectorAll` has a generic version to set the return type:

```suggestion
            return Array.from(this.modalTarget.querySelectorAll<HTMLElement>("[href], input, select, textarea, button, [tabindex]"))
```

lib/ts/controllers/s-modal.ts
```diff
@@ -153,58 +153,69 @@ namespace Stacks {
...
                 .filter((el: Element) => el.matches(":not([disabled]):not([tabindex='-1'])"));
+        }
 
-            if (!allTabbables.length) {
-                return;
-            }
-
-            var initialFocus = <HTMLElement>allTabbables[0];
+        private firstVisible(elements: HTMLElement[]) {
```
b-kelly, line 166: nit: Add `/** */` JSDocs to all new functions

lib/ts/controllers/s-modal.ts
```diff
@@ -153,58 +153,69 @@ namespace Stacks {
...
                 }
-            });
 
-            document.addEventListener("keydown", this._boundTabTrap);
+                return;
+            }
 
-            return initialFocus;
+            // If we observe a tab keydown and we're on an edge, cycle the focus to the other side.
+            if (e.keyCode === 9) {
```
b-kelly: nit: Don't use `KeyboardEvent.keyCode`, it is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode). We can use the friendlier [`.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) instead (IE9+):

```suggestion
            if (e.key === "Tab") {
```

I just realized that the code I wrote originally was using `.keyCode`, so... I dunno what I was thinking. Let's go ahead and fix it up here and brush my terrible code under the rug...

lib/ts/controllers/s-modal.ts
```diff
@@ -153,58 +153,69 @@ namespace Stacks {
...
+        private focusInsideModal() {
             this.modalTarget.addEventListener("s-modal:shown", () => {
-                // double check the element still exists when the event is called
-                if (initialFocus && document.body.contains(initialFocus)) {
-                    initialFocus.focus()
-                }
+                var initialFocus = this.firstVisible(this.initialFocusTargets) ?? this.firstVisible(this.getAllTabbables());
+                initialFocus?.focus();
             }, {once: true });
+        }
```
b-kelly, line 189: nit: double newline

```suggestion
```

lib/ts/controllers/s-modal.ts
```diff
@@ -97,6 +96,7 @@ namespace Stacks {
 
             if (toShow) {
                 this.bindDocumentEvents();
+                this.focusInsideModal();
```
b-kelly, line 99: This should be safe to do on each show because the event has `once: true` set.

No changes necessary.

lib/ts/controllers/s-modal.ts
```diff
@@ -4,8 +4,7 @@ namespace Stacks {
         static targets = ["modal", "initialFocus"];
 
         private modalTarget!: HTMLElement;
-        private initialFocusTarget!: HTMLElement;
-        private hasInitialFocusTarget!: boolean;
+        private initialFocusTargets!: HTMLElement[];
```
b-kelly, line 7: This change should be fine because Stimulus maps the return targets as an array even if there's only one, making this change api compatible with what we currently have.

No changes necessary.

lib/ts/controllers/s-modal.ts
```diff
@@ -153,58 +153,69 @@ namespace Stacks {
...
+        }
 
-            if (!allTabbables.length) {
-                return;
-            }
-
-            var initialFocus = <HTMLElement>allTabbables[0];
+        private firstVisible(elements: HTMLElement[]) {
+            // https://stackoverflow.com/a/21696585
+            return elements.find(el => el.offsetParent !== null);
```
b-kelly, line 168: optional: This solution is fine for me, but if we wanted to be more thorough, jQuery's hidden selector is actually _really_ small: https://github.com/jquery/jquery/blob/a684e6ba836f7c553968d7d026ed7941e1a612d8/src/css/hiddenVisibleSelectors.js#L9

b-kelly (approved):
lgtm! Tested locally and everything works as expected. I added a few nits, but nothing blocking merge.

Originally, I had computed the tabbables once to avoid what could be (in theory) an expensive dom search. I agree with you that computing this list on each TAB press should be fine, since user interaction could cause e.g. form elements to show/hide at any point in time. It also follows that we'd want to check the visibility of the tabbables as well.

lib/ts/controllers/s-modal.ts
bnickel: I'm inclined to leave things alone here.  The edge that jQuery _may_ be avoiding is `position: fixed` elements, which shouldn't exist within or be focusable within our modals.

b-kelly: Sounds good to me! Your solution is clean and works _now_, so I'm totally ok with merging as-is. If we run into an edge case in the theoretical future, we can address it at that time.

b-kelly (approved):
lgtm! Thanks for the tweaks

PR #674, opened 2021-06-01
Use opacity only for disabled states of checkbox, radio

dancormier:
https://github.com/StackExchange/Stacks/issues/672

Issue #672, opened 2021-05-31
Disabled states for checkboxes 


CGuindon:
To make all our disabled states consistent, we'd like to update the disabled state of the checked checkbox (and radio button?) to use 50% opacity instead of changing to a grey colour.

_current disabled checked checkbox_
<img width="241" alt="Screen Shot 2021-05-28 at 10 00 47 AM" src="https://user-images.githubusercontent.com/5436760/120192416-278c9f00-c1e9-11eb-91bb-f6a2b6b44e46.png">

_current active checked checkbox_
<img width="268" alt="Screen Shot 2021-05-28 at 10 01 13 AM" src="https://user-images.githubusercontent.com/5436760/120192444-2c515300-c1e9-11eb-899d-84d46765c4b4.png">

_proposed disabled checked checkbox_
<img width="373" alt="Screen Shot 2021-05-28 at 10 01 29 AM" src="https://user-images.githubusercontent.com/5436760/120192541-4be87b80-c1e9-11eb-8ada-440588363b48.png">

PR #675, opened 2021-06-01
Use opacity only for disabled states of checkbox, radio (take 2 🎬)

dancormier:
This PR addresses https://github.com/StackExchange/Stacks/issues/672 by removing color overrides for checkbox and radio inputs to solely use 50% opacity when disabled.

Take 1: https://github.com/StackExchange/Stacks/pull/674


aaronshekey:
Looking good! Didn't realize we already rendered these at `50%` and _then_ recolored. Nice!

I think it'll be worth it for future testing to add the checked (but disabled) state in this example: 
image

Maybe we just double this row and have the second row both `checked`, leaving the copy & paste markup alone.

Like so:
image

aaronshekey (approved):

Issue #620, opened 2021-03-22
Inputs should have explicit `readonly` styling


aaronshekey:
We're currently ignoring the `readonly` attribute and any styling. At the very least, we should probably turn off selected states here?

PR #678, opened 2021-06-04
Style readonly attribute (not read-only)

dancormier:
Addresses: https://github.com/StackExchange/Stacks/issues/620

The attribute to target is `readonly`. It can also be targeted as a pseudo-class in the form of `:read-only`, which is what I'd guess caused the confusion here.

### The `<select />` element

This targets a `readonly` attribute on `<select />` elements, though that element does not support it.
See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#attributes


aaronshekey:
I think this actually needs some more thought than just looking like a `disabled` input. I'll add this to a list of items needing designs.

PR #680, opened 2021-06-06
Better flex documentation

aaronshekey:
Prior to this PR, our flex documentation was mostly just a table of classes without any examples. This expands each of those available classes into an obvious example. It also gets rid of documentation of a few shorthand classes that I think we'll be better off deprecating.

There will be a follow-up there to save us some bytes.


dancormier (approved):
I love good documentation and found nothing out of place.

Two minor tangents that shouldn't hold up a merge here:

1. What's the likelihood https://github.com/StackExchange/Stacks/pull/423 will be merged anytime in the near future? I figure it'll impact this documentation and  necessitate a split of the Grid and Flex sections. Which would be nice because that Grid page has a lot.
2. Is there a convention around indicating deprecation in the docs (seems like maybe sticking `deprecated` in a badge on the offending class)?

aaronshekey:
> 1. What's the likelihood #423 will be merged anytime in the near future? I figure it'll impact this documentation and  necessitate a split of the Grid and Flex sections. Which would be nice because that Grid page has a lot.

I made the call to not wait on these bigger proposed changes and opted to improve documentation in the meantime. This'll definitely thrash those PRs. Oh well :) 

> 2\. Is there a convention around indicating deprecation in the docs (seems like maybe sticking `deprecated` in a badge on the offending class)?

Not yet. A lot of our "deprecations" are mostly based on classes really only being used a time or two in production. Once linting is in, I think we'll handle deprecations there, and simply remove documentation.

PR #683, opened 2021-06-07
Deprecate a few flex atomic classes

aaronshekey:
Most of these `ff-*` classes just map to shorthands that can be achieved with `fd-column` or `fw-wrap`. No need to ship a second meta class there, in my opinion.

Additionally, the `fl-grow` and `fl-shrink` classes are covered by `fl1` and `fl0` very well.


benjamin-hodgson:
`fl0` doesn't really work as advertised

aaronshekey:
> `fl0` doesn't really work as advertised

Do you have a good example of that? I'm not surprised, sadly. Is `.fl0` not working as a replacement for `.fl-shrink0`?

benjamin-hodgson:
It sets grow to 0 and shrink to 1. https://github.com/StackExchange/Stacks/blob/develop/lib/css/atomic/_stacks-flex.less#L104

dancormier:
I _don't believe_ `.fl0` would set shrink to 0.
And just to lob a bomb, I'd suggest that `.fl0` might be more accurately named `.fl-initial`.

From https://css-tricks.com/almanac/properties/f/flex/#flex-0-auto
> `flex: 0 auto;`
This is the same as `flex: initial`; and the shorthand for the default value: `flex: 0 1 auto`. It sizes the item based on its width/height properties (or its content if not set).

benjamin-hodgson:
We do have `fl-none`.

benjamin-hodgson:
my proposal would actually be to deprecate all of the `fl0`/`fl1` classes instead and recommend using `fl-grow`/`fl-shrink` instead

aaronshekey:
Glad I didn't just YOLO this then. Perhaps we take the Tailwind route:

https://tailwindcss.com/docs/flex-grow
https://tailwindcss.com/docs/flex-shrink

Good and clean and still gets rid of all those weird short hands and fl-grow5's.

aaronshekey:
How about this latest approach?

dancormier (approved):

aaronshekey:
Ok, in this case our "deprecation" is just a deletion since these aren't the widest-used classes in production and we can handle the rename in a single commit. When we reach `1.0` we can be way more diligent about deprecations, but this is firmly alpha software.

PR #684, opened 2021-06-07
Add Prettier and see what happens 😬 

dancormier:
Don't worry, this is just a dry run.
I figure it's wise to document how adding [Prettier](https://prettier.io/) goes so when I inevitably think to try this again, I can look at this and have a good think.

You may look at all the changes and think _"shouldn't there be some configuration that prevents most of these changes?_
No, Prettier is opinionated and [almost entirely devoid of options](https://prettier.io/docs/en/option-philosophy.html). Consider it a feature. The point of Prettier is to squash all the debate around code formatting (and configurations of formatters) and just pick formatting rules for you. Some will hate this. I like it.

Note: I didn't bother adding git hooks in here because this isn't getting merged, but I'd recommend it if we ever add Prettier. https://prettier.io/docs/en/install.html#git-hooks


aaronshekey:
Yeah, don't know about this one 
image

dancormier:
I like everything there except that `sm:pl16 sm:pr16`. That bugs me.

PR #689, opened 2021-06-07
Transition `grid` to `d-flex`

aaronshekey:
`css-grid-final__final-v3.psd`

This PR figures out our transition to a proper CSS grid. It duplicates our existing `grid` classes to transition to new flex names. e.g. `.grid` > `.d-flex` and `.grid--cell` to `.flex--item`.

We'll leave both the old `.grid` class names in place for a while, changing only the documentation for now. This PR communicates these changes.

This approach avoids any `.grid--cell` name clashes when we drop the new `.d-grid` [hotness](https://github.com/StackExchange/Stacks/pull/423). It'll just be `.d-grid` and `.grid--item`.

Then, step 100 is removing `.grid` and `.grid--cell` from the Stacks library and documentation. 🎉 

image


dancormier (approved):

aaronshekey:
Let's talk about this during today's sync, but I'm digging this plan :v:

PR #690, opened 2021-06-08
Add Uploader component

### LINKED ISSUE #435: Component: Input type file
We need a way to reliably upload files in a Stacks-styled manner.

dancormier:
Addresses #435 
👋 I've requested @aaronshekey as a check on design and maybe to address some of my questions below.
👋 I've requested @b-kelly to check on my utterly newb stimulus.js. Suggestions welcome!

# Reference
[Figma mockup](https://www.figma.com/file/N3AcooHblPRENtNGBpFXvZ/Uploader?node-id=0%3A1)
[Request for critique on Slack](https://stackexchange.slack.com/archives/CMWHQ4U5V/p1623708153019800)

# Q&A
- Should this include a hover style?
    - [no, but...](https://github.com/StackExchange/Stacks/pull/690#issuecomment-857931243)
- Should we build to assume one or multiple files can be uploaded at once?
    - [It should support both](https://github.com/StackExchange/Stacks/pull/690#issuecomment-857931243)
- Should we assume this accepts images only? images and video? any file type?
    - [Should be file type agnostic, but assume images](https://github.com/StackExchange/Stacks/pull/690#issuecomment-857931243)
- Any touch-specific considerations?
    - [Just copy considerations](https://github.com/StackExchange/Stacks/pull/690#issuecomment-857931243)

# To do
- [x] Fix focus styling (my `.o0` approach is hiding `.f:outline-ring`)
- [x] Add label
- [x] Add support for multiple files
- [x] Add `x` (remove) button to previews
- [x] Add drag-over styling
- [x] Add sane file preview styling
- [x] Compensate for multi-image previews with odd initial sizes (short images primarily)
- [x] Make file preview type-agnostic
- [x] Add sensible documentation
- [x] Add better [placeholder icon](https://github.com/StackEng/StackOverflow/pull/8943#issuecomment-887909317)
- [x] Update copy to consider touch devices


aaronshekey:
> * Should this include a hover style?

I don't think we need a hover style, but it would be nice if we darken the background a bit if I'm dragging a file over this area.

> * Should we build to assume one or multiple files can be uploaded at once?

I think this should be configurable. When multiple images are dropped, hopefully we could stack them a bit. Speaking of previews, I think we should lock the image previews to a height so the uploader area stays the size it is.

> * Should we assume this accepts images only? images and video? any file type?

This should probably also be configurable, but I think we can assume images for most use-cases. I tossed a `.mov` in on my mac and it built a low-framerate preview. In my original explorations, I kept the placeholders a bit agnostic 
image

Note to self: We should get a spot illustration whipped up for this.

> * Any touch-specific considerations?

I don't _think_ so. I think the only consideration I have there might be the copy on "Drag". That may work on an iPad but is less likely to make sense on the phone...

aaronshekey:
While we have the `Cancel` button, perhaps we should also have an `x` on the thumbnails that get added to the control? 🤔

aaronshekey:
Added a few more bits to this work in progress spec. Enough to chew on here? https://www.figma.com/file/N3AcooHblPRENtNGBpFXvZ/Uploader?node-id=0%3A1

aaronshekey:
Some observations:

1. We add the focus state when dragging _to_, but not _from_. Drag something over the control. Without letting go of your mouse, drag it back out. It should head back to the original state.
2. We don't want the control to grow. I think with the right padding, we can just set those child image previews to `.h100`.
3. I think the preview image shouldn't take up the full space. It should fill it proportionately?
4. Got some feedback that my treatment of multiple images isn't super obvious. Maybe we do something with shadows and a wider spread?

dancormier:
> Some observations:
> 
> 1. We add the focus state when dragging _to_, but not _from_. Drag something over the control. Without letting go of your mouse, drag it back out. It should head back to the original state.

Ah, you caught a Safari bug. Nice catch! Looking.

> 2. We don't want the control to grow. I think with the right padding, we can just set those child image previews to `.h100`.

That's totally doable. The problem I'm running into is with our predefined heights.
I'd love to stick to a height that we have in the design system, but not finding anything between `.hs1` and `hs2`. It's a Goldilocks situation. Is there an defined height in Stacks I should go with or should I slap a 143px on the uploader and be done with it?

> 3. I think the preview image shouldn't take up the full space. It should fill it proportionately?

~~Hmm, not sure I follow. Can you give me an example?~~

Nevermind, this is Safari again! I really should've checked it before asking you to take a look. Sorry!

> 4. Got some feedback that my treatment of multiple images isn't super obvious. Maybe we do something with shadows and a wider spread?

I think a subtle shadow and a more generous size for the n+1 images could help it read as _multiple_.


Another idea is to indicate more of a "stack" of files (breaking convention too much?).



A more _hit-you-over-the-head_ approach could be to stick a number on the previews to indicate how many files are selected.

aaronshekey:
> Another idea is to indicate more of a "stack" of files (breaking convention too much?).
> 

I like this one while spreading out the distance a bit.

Should we try .h128 on the control? Too small?

dancormier:
> I like this one while spreading out the distance a bit.

By "_spreading out the distance_" do you mean:

1) Retain the white-space between previews and make it wider.
2) No white-space between previews. Just make each them wider.

> Should we try .h128 on the control? Too small?

`.h128` is just on that edge imo. Might be pushing it, especially in wide parent containers.
Originally, I had the consumer specify the height because I figured it could wind up in all sorts of containers (and at different viewport widths) and figured making it feel proportioned might be tricky to prescribe.

We might be able to tie height to viewport, but I'm not sure if modal usage throws a wrench in that approach. Container queries or aspect-ratio ([we're soooo close](https://caniuse.com/mdn-css_properties_aspect-ratio)) would make this simpler 😢

My gut says we'd need a little more height than `.h128` in most cases, but I'm glad to defer to you on the best approach here.

## Examples of `.h128` sizing:

### 375px viewport width (309px uploader)




### 768px viewport width (540px uploader)


### 1280px viewport width (750px uploader)

aaronshekey:
This is prrreettty elegant so far! Those are now absolutely reading as multiple files 👌 . Docs are even coming together. Nice work.

Some more observations:

1. The height feels ok, but I definitely share your concern at super wide displays. What if we just add something like `.wmx3` and call it a day?
2. There's one tiny tiny tiny nit. In Safari, if you're up in the top left where the input's shadow DOM is, the cursor goes `default`. Is there any way to force `cursor: pointer` even in the shadow DOM?

dancormier:
>     1. The height feels ok, but I definitely share your concern at super wide displays. What if we just add something like `.wmx3` and call it a day?

I've probably spent too long thinking about this problem, so I'm inclined to say yes and move on. Adding it as a class to the HTML seem good? (as opposed to adding to the `.s-uploader` css)

>     2. There's one tiny tiny tiny nit. In Safari, if you're up in the top left where the input's shadow DOM is, the cursor goes `default`. Is there any way to force `cursor: pointer` even in the shadow DOM?

[Yep, good call!](https://github.com/StackExchange/Stacks/pull/690/commits/3ff222fbec7eee0b87bf360e84e702b6acd55fd8)

lib/css/components/_stacks-uploader.less
```diff
@@ -0,0 +1,95 @@
...
+        border-radius: @br-sm;
+        box-shadow: var(--bs-md);
+        box-sizing: border-box;
+        display: flex;
+        height: 100%;
+        overflow: hidden;
+        pointer-events: none;
+        transform-origin: center left;
+
+        each(range(5), {
```
dancormier: 👋 hi, me. You used `range` here. Ya gotta use something else or upgrade some (non-Stacks) deps.

https://lesscss.org/functions/#list-functions-range

aaronshekey:
Switching to SVG to render this dashed border. There are subtle cases where things render super weird with dashed borders.

**CSS border**
image

**SVG border**
image

dancormier:
Ah damn that's a really good catch.

aaronshekey:
Okie dokie. We still need the JS events documented. I'll leave that to you, @dancormier. I want @b-kelly's review on the Stimulus but this looks good to me!

dancormier:
@aaronshekey I came up with a few questions re: the new multiple file layout:

- The previous design capped out at 5 previews thumbnails. How should we handle a cap with this one? 
    - We could append `... and 4 others`, use the heading like `9 items (4 not shown)`, or just cap it and say nothing. I think we need *some* cap so we're not adding 500 DOM elements when someone fat fingers a *select all* and adds 100s of items to the field (though I think it's on the consumer to limit what's potentially uploaded).
- How should we handle long file names? Is a single-line `text-overflow: ellipsis;` sufficient?
- I was using the filename as the thumbnail content when no image could be extracted. Should I continue doing that for single files or use something like the document icon we're using when displaying multiple files?

aaronshekey:
>   * We could append `... and 4 others`, use the heading like `9 items (4 not shown)`, or just cap it and say nothing. I think we need _some_ cap so we're not adding 500 DOM elements when someone fat fingers a _select all_ and adds 100s of items to the field (though I think it's on the consumer to limit what's potentially uploaded).

How about "Showing 10 of 500 items"? Capping at 10 seems reasonable, though this is mostly a hypothetical. We could always adjust later.

> * How should we handle long file names? Is a single-line `text-overflow: ellipsis;` sufficient?

Your instincts are right there, let's just crop with the approach we're using in `.truncate` with an ellipsis.

> * I was using the filename as the thumbnail content when no image could be extracted. Should I continue doing that for single files or use something like the document icon we're using when displaying multiple files?

We can keep doing this for single files, I think. I think our 90% use case will be a single, previewable image.

aaronshekey:
Now that I'm done monkeying with designs, this looks super super close.

- [x] Finish documenting all the JS attributes and events
- [x] Get final Stimulus review from @b-kelly

dancormier:
I added a limit of 10 MB on previews in 79f2bd0 to prevent a memory monster if someone accidentally selects a Blu-ray rip or something.

I think a 10 MB cap on previews seems reasonable. Enjoy these examples:

## Screencaps

<details>
<summary>No limit</summary>

https://user-images.githubusercontent.com/647177/128558062-2e8fb528-65dc-4a70-94dd-29300ac65b53.mp4

</details>

<details>
<summary>10 MB limit</summary>

https://user-images.githubusercontent.com/647177/128558202-ce438a8c-8ba5-4129-a7fa-9fd7e713bc14.mp4

</details>

<details>
<summary>1 MB limit</summary>

https://user-images.githubusercontent.com/647177/128558253-69c55574-a0d9-45ec-9c82-cedd84b5c3a1.mp4

</details>

lib/ts/controllers/s-uploader.ts
```diff
@@ -0,0 +1,171 @@
...
+         * Resets the Uploader to initial state
+         */
+        reset() {
+            this.inputTarget.value = '';
+            this.previewsTarget.innerHTML = "";
+            this.handleVisible(false);
+        }
+
+        /**
+         * @param  {boolean} shouldPreview - Uploader is entering a preview state
```
b-kelly, line 77: nit: To me, the most important part of a function's docs is the description, followed by the param descriptions (followed by types if not using TS). I recommend adding docs for all functions (especially the public ones)

lib/ts/controllers/s-uploader.ts
```diff
@@ -0,0 +1,171 @@
...
+        disconnect() {
+            this.inputTarget.removeEventListener("dragenter", this.boundDragEnter);
+            this.inputTarget.removeEventListener("dragleave", this.boundDragLeave);
+            super.disconnect();
+        }
+
+        /**
+         * Handles rendering the file preview state on input change
+         */
+        handleInput() {
```
b-kelly, line 38: I'm sorta back and forth on "expose this for use in `data-action`" and "just wire it up always, don't make the user do it". In pure stimulus, we'd prefer the former, but I'm not really super into "pure stimulus". This is fine as-is, but I did want to bring it up to prompt the discussion.

b-kelly (commented):
Code looks good, nice work! I made some slight "philosophical" tweaks, but nothing drastic that changes your approach.

lib/ts/controllers/s-uploader.ts
```diff
@@ -1,15 +1,15 @@
 namespace Stacks {
     interface FilePreview {
-        data?: Blob;
+        data?: string | ArrayBuffer;
```
dancormier: @b-kelly `ArrayBuffer` makes sense here. Thanks! I'm wondering why you included `string` as well?

lib/ts/controllers/s-uploader.ts
```diff
@@ -1,15 +1,15 @@
...
         name: string;
-        type: MimeType;
+        type: string;
```
dancormier: @b-kelly Should this not remain `MimeType`? The `File` interface provides `type` as a `MimeType`. https://developer.mozilla.org/en-US/docs/Web/API/File/type

lib/ts/controllers/s-uploader.ts
```diff
@@ -0,0 +1,174 @@
...
+        handleInput() {
+            this.previewsTarget.innerHTML = "";
+
+            if (!this.inputTarget.files) {
+                return;
+            }
+
+            const count = this.inputTarget.files.length;
+                this.getDataURLs(this.inputTarget.files, UploaderController.FILE_DISPLAY_LIMIT)
+                    .then((res) => {
```
dancormier: Thanks for killing the `any` I plopped in there @b-kelly. Would this be reasonable?

```suggestion
                    .then((res: FilePreview[]) => {
```

lib/ts/controllers/s-uploader.ts
b-kelly: I was using the types TS reported in my editor. In this case, `FileReader.result` has [the signature](https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L5259):

```ts
readonly result: string | ArrayBuffer | null;
```

I was a tad concerned about the `string` part, but TS didn't throw any additional type errors where we were using it, so I guess we're good ¯\\\_(ツ)_/¯

lib/ts/controllers/s-uploader.ts
b-kelly: Similar to above, I was going off the types TS provided. The signature for [Blob.type](https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L2401) is 

```
readonly type: string;
```

lib/ts/controllers/s-uploader.ts
b-kelly: Yes, absolutely. In this case (for me, locally anyways) TS was able to infer the type, so I dropped it. You're welcome to add it back according to your preferences - I don't have a strong opinion on leaving it in or out here.

lib/ts/controllers/s-uploader.ts
dancormier: Admittedly, I was leaning heavily on prior art (I can't recall which component I referenced most), but I think stepping away from "pure stimulus" is generally a good call.

It probably does make more sense to just attach this function directly to the target input but I'm good to leave it as-is for now.

PR #696, opened 2021-06-10
Fix whitesp ace issues in post summary

### LINKED ISSUE #686: Icons in post summary are dependent on HTML whitespace
screen_shot_2021-06-07_at_11 58 04_am
screen_shot_2021-06-07_at_11 58 14_am

Removing the space before the 5 in Stacks docs reproduces the issue. Let's explicitly add some margins to those icons so we aren't relying on HTML whitespace.

aaronshekey:
Switch to flex on .s-post-summary--stats-item to control for whitespace or the lack thereof. Also shifts some icon positioning to CSS instead of atomic classes.

Closes #686 

cc: @bnickel

### DIFF
--- a/docs/product/components/post-summary.html
+++ b/docs/product/components/post-summary.html
@@ -46,15 +46,15 @@ description: The post summary component summarizes various content and associate
 <div class="s-post-summary">
     <div class="s-post-summary--stats">
         <div class="s-post-summary--stats-item has-answers has-accepted-answer">
-            @Svg.CheckmarkSm.With("va-text-bottom")
-            … answers
+            @Svg.CheckmarkSm
+            …answers
         </div>
         <div class="s-post-summary--stats-item">
-            … votes
+            …votes
         </div>
         <div class="s-post-summary--stats-item is-supernova">
-            @Svg.FireSm.With("va-text-bottom mrn2")
-            … views
+            @Svg.FireSm
+            …views
         </div>
     </div>
     <div class="s-post-summary--content">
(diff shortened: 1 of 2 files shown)


bnickel:
Casual reminder that `display:flex` is not a panacea.

```
<div class="s-post-summary--stats-item">
    @_s("**#__count#** vote", new { __count })
</div>
```

Screen Shot 2021-06-10 at 11 36 27 AM

Flex treats text nodes next to elements as nodes so you end up with "<b>1</b>" and "vote" as two flex items.

The fix here is the following but I feel that it's less intuitive than the existing behavior:

```
<div class="s-post-summary--stats-item">
    <span>@_s("**#__count#** vote", new { __count })</span>
</div>
```

paintedbicycle:
Nice job on the issue title

aaronshekey:
> Not a panacea

You know what is though? `gap: 3px`. When `flex` supports this across the board, we might be able to control white space in a bulletproof manner here 🤔 

So what's the alternative? Close both these and tell @paintedbicycle he's gotta control his whitespace?

bnickel:
> You know what is though? `gap: 3px`.

Nah, you forget Japanese, where zero space between to items might be preferable. We're in a fun area.

> So what's the alternative? Close both these and tell @paintedbicycle he's gotta control his whitespace?

That's what I'd advocate for.  As I mentioned in #686, this isn't new to Stacks or Teams and we live with this in `.s-btn` today.  `.s-post-summary--stats-item` should just be a plain simple container with as few surprises as possible and the consumer is responsible for making sure their stuff looks right.

aaronshekey:
> and the consumer is responsible for making sure their stuff looks right

Yeah, @paintedbicycle 😝

PR #698, opened 2021-06-11
Fix js selector in stylelint

bnickel:
`"/^[.#]js-/"` looks at the whole selector and fails to catch things like `div .js-foo`, `div.js-foo`, `.s-good, .js-bad`.


aaronshekey:
Beautiful, thanks!

Issue #385, opened 2019-09-06
Can we support disabled fieldsets?


mstum:
**Is your feature request related to a problem? Please describe.**
Assume this kind of layout (I removed some stuff for brevity, like the lock icon in the textbox, or the validation error message):

```html
<fieldset disabled="disabled">

    <div class="grid gs4 gsy fd-column mb16 js-form-field">
        <div class="grid--cell">
            <label class="d-block s-label" for="AuthorizationBearerToken">
                SCIM authorization bearer token
            </label>
        </div>
        <div class="grid ps-relative">
            <input class="s-input" type="text" id="AuthorizationBearerToken" name="AuthorizationBearerToken" value="..............">
        </div>
    </div>
    <fieldset class="js-form-field">
        <div class="grid gs4 mb16">
            <div class="grid--cell">
                <input class="s-checkbox" type="checkbox" name="EnableModeratorPromotion" id="EnableModeratorPromotion" checked="checked">
            </div>
            <label class="grid--cell s-label" for="EnableModeratorPromotion">
                Allow Moderator Promotion via a userType property
            </label>
        </div>
    </fieldset>

</fieldset>
```

This displays like this:

![image](https://user-images.githubusercontent.com/104127/64452205-def78e80-d0b3-11e9-8190-7f4f0ed83255.png)

Two questions:

* The Label of the Edit element is still rather prominent - should it be grayed out?
* While the input inherits the disabled from the fieldset, the checkbox does not. According [to the docs](https://stackoverflow.design/product/components/checkbox), Checkboxes support a separate `is-disabled` class, but I wonder if we can inherit it from the fieldset being disabled?

**Describe the solution you'd like**
Have the disabled property from `<fieldset disabled="disabled">` trickle down to any nested form elements (also buttons, radios, etc.), and make sure it's obvious that these elements are deactivated.

**Describe alternatives you've considered**
JavaScript and a separate `<style>` tag to define some styles - basically using a regular `<div>` instead of a fieldset and targeting that via JavaScript to set disabled on every form element.

**Additional context**
N/A

aaronshekey:
I believe this was addressed in #702

PR #702, opened 2021-06-17
Style disabled fieldset elements as disabled

dancormier:
Addresses https://github.com/StackExchange/Stacks/issues/385

When contained within `<fieldset disabled />`, apply the equivalent styles as adding `.is-disabled`/`[disabled]`  to form elements.


lib/css/components/_stacks-inputs.less
```diff
@@ -108,6 +108,35 @@ fieldset {
...
+            opacity: 0.5;
+        }
+
+        .s-input,
+        .s-textarea,
+        .s-select > select {
+            border-color: var(--black-075);
+            background-color: var(--black-050);
+            color: var(--black-200);
+            cursor: not-allowed;
```
aaronshekey: Should we consider normalizing this around `opacity: 0.5` like we've done with other disabled states?

aaronshekey (approved):

lib/css/components/_stacks-inputs.less
dancormier: Looks like I lifted this from https://github.com/StackExchange/Stacks/blob/develop/lib/css/components/_stacks-inputs.less#L399-L414

I tested out using opacity and it seems to work just fine. Can you think of any reason we would've used colors instead of opacity (just generally; not wrt fieldset)?

aaronshekey: Opacity felt like a shortcut at the time, but it's since been established as a common-practice.

PR #707, opened 2021-06-29
Migrate to Webpack

b-kelly:
This PR changes over to using Webpack for bundling. The package is now shipped as a UMD file and can be imported as a module (along with types in TS) or added via script tag (making itself available via `window.Stacks`). When adding to a page via a script tag, the global object remains backwards compatible.

## Bundle stats

Old bundle taken from develop branch, new bundle taken from this branch. Gzipped files done via `gzip` cli tool. All sizes are in _bytes_. 

| bundle | js | js (gzip) | css  | css (gzip) |
| ------ | -- | --- | --------- | ---------- |
| old | 97,162 | 24,298  | 404,248 | 54,591 |
| new | 98,176 | 23,417 | 387,246 | 54,133 | 

## Testing

Make sure you run `git clean -xfd` first to clean your local install, then `npm i`.

### Testing package as if installed from npm locally

See the README in [this gist](https://gist.github.com/b-kelly/017de797c71c2ddbcaafe9694308cfd0).

## TODO
- [x] docs initial bundle size / imports
  - [x] stop setting imports as global to be used in other scripts
  - [ ] ~~remove global jQuery ($) and migrate away from using it~~ Follow up PR
- [ ] ~~webpack and eleventy watchers stomp on each other a bit~~ Good enough for now
- [x] remove old grunt stuff
- [x] add non-minified versions of bundled files
  - [x] ~~add non-minified css files if able~~ Turns out, we don't care about this


docs/_includes/head.html
```diff
@@ -23,5 +23,5 @@
 
     <!-- CSS -->
     <link rel="stylesheet" href="//cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />
-    <link rel="stylesheet" type="text/css" href="{{ "/assets/css/stacks-documentation.min.css" | relative_url }}?{{page.date | date: '%s%N'}}">
+    <link rel="stylesheet" type="text/css" href="{{ "/assets/dist/stacks.min.css" | relative_url }}?{{page.date | date: '%s%N'}}">
```
aaronshekey: Just a heads up, the way the docs work now is the stacks bundle + whatever's in `stacks-documentation.less`. If I'm not mistaken, this loses any custom docs CSS.

b-kelly: Nope! The file is the same, just the name changed. By default, the style bundles take the same name as the JS bundles, but this can be easily changed. Since the JS file name changed to `stacks.js`, the css file name changed as well. You can see [here](https://github.com/StackExchange/Stacks/pull/707/files#diff-6760708bea0e9d8fe9c272531d931fcf8fd7424db4de10dcba92031a508bcc13R3) that I'm including the correct css file.

dancormier (commented):
I added [8b6451b](https://github.com/StackExchange/Stacks/pull/707/commits/8b6451b6cef574ca9ff3a77171be432e3927bdcf) to this branch to include some missing deps required to build the docs.

Once those deps were in place, everything ran smooth from a DX perspective. I ran through each npm script in the repo and they all performed reasonably well, including hot reloading where applicable.

aaronshekey:
I've taken this for a spin locally, and things seem to be behaving as expected in terms of building and serving the docs.

It seems that when I run `npm pack`, I'm missing the `dist` directory. Should `npm start` also run `npm build` so we always have that `dist` directory? It doesn't get generated otherwise. I can see a world in which I clone the latest `develop`, then run `npm start`, and then `npm publish` and I'm not actually publishing the latest and greatest. Our Grunt hacking was a little more bulletproof in this regard.

Unpkg is referencing `dist/css/stacks.min.css` which I think doesn't exist. I would still very much love to include a `stacks.css` and `stacks.min.css`. If we can't ship an unminified version, the file itself should be called `stacks.min.css` in my opinion. This would be absolutely the most clear and wouldn't break Unpkg.

b-kelly:
@aaronshekey I've addressed both of your comments. For publishing, I've added a [prepublishOnly](https://docs.npmjs.com/cli/v8/using-npm/scripts) script that runs `npm run build` first whenever you run `npm publish`. I've also "doubled" the webpack config so that it runs twice - once to spit out a minified build (with `.min.*` extension), once to spit out an unminified build. This does not happen for local dev, so no worries on dev slowdown there.

docs/brand/colors.html
```diff
@@ -4,7 +4,7 @@
 description: See the <a href="/product/base/colors/">Product</a> section for colors when building with Stacks. While there is some cross-over with the Product colors, using Stacks color classes will mean they adjust depending on context (e.g., dark mode)
 ---
 
-<script src="{{ "/assets/js/feature.notices.js" | url }}" defer></script>
+<script src="{{ "/assets/js/entry.notices.js" | url }}" defer></script>
```
allejo:
```suggestion
<script src="{{ "/assets/dist/entry.notices.js" | url }}" defer></script>
```

Is copying over `assets/js` necessary for the compiled site? Might be worth ignoring that so that there's no question as to where to include the JS from.

aaronshekey (approved):
Yeah! 🦐

b-kelly:
Good call on copying over the assets @allejo! I've updated the copy to only do the dist and image folders. In a followup PR I think I'll change where the image and dist folders live so we don't have this issue later on.

@aaronshekey If you could do a `git clean -xfd` to clear out old compiles, then run `npm i` and test one last time, I'd be forever grateful.

docs/brand/colors.html
b-kelly: Nice catch! I think I straight up forgot to update this import and it just kept working so... See my comment below - I stopped copying over the assets folder in its entirety.

aaronshekey:
Ok, I've noticed _one_ thing just now. In the Editor demo page, if I write some text into the empty editor and flip to markdown, it appears as a code block. This doesn't happen on `.design`.

b-kelly:
@aaronshekey Nice catch. This was due to the external styles being treated as a normal css import (good) and hoisted to the top. This means that the docs css was taking precedence. In this case, I'd prefer the _editor_ styles to take precedence, so I instructed Less to inline them directly. Is this a hack? Yes, kinda. I'll think about making those overridden styles more specific upstream (or the overridding styles less specific here).

aaronshekey:
Seems chill now!

PR #713, opened 2021-07-14
Add a diamond to `.s-badge__moderator`

aaronshekey:
We can't lose the moderator diamond, so let's figure out a way to deliver via CSS-only, so we aren't messing with unicode symbols or SVG or whatever.


paintedbicycle (approved):
Nice solution @aaronshekey. Thank you for keeping the spirit of it

PR #716, opened 2021-07-19
Convert hex, rgb(a) color values to hsl(a)

dancormier:
## Description
This PR changes HEX and RGB color values to HSL.

## Why?

I find HSL to be more readable than RGB and HEX (and [many](https://www.sarasoueidan.com/blog/hex-rgb-to-hsl/) [others](https://elad.medium.com/why-css-hsl-colors-are-better-83b1e0b6eead) [agree](https://stackoverflow.com/a/43795837/652353)).

I think changing from HEX/RGB to HSL will encourage consistency, remove friction, and prevent slip-ups when altering colors.

## Notes

This PR alters color values *ever-so-slightly*. This mostly comes in the form of unifying hue values for a given color. There’s also rounding to the nearest whole number/percent with H/S/L values (alpha values were not rounded). I expect them to be imperceptible (YMMV on a calibrated Apple Pro Display HDR).

## To do
- [x] Tweak the blue values. There’s a percievable difference in those colors from the HEX values

## Screenshots
<details>
<summary>Before</summary>

Screenshot 2021-08-23 at 11-53-23 Stacks

</details>

<details>
<summary>After</summary>

Screenshot 2021-08-23 at 11-50-41 Stacks

</details>


aaronshekey (approved):
Does what it says on the tin without any regressions that I've been able to find. Took screenshots and popped this into photoshop with difference layers and it was barely perceivable. Also ran visual regression testing and nothing popped up.

PR #717, opened 2021-07-20
Tag theming refinements

aaronshekey:
Instead of having a modifier for theming tags, let's theme them by default. This PR explores just that, replacing the powder for `theme-secondary` and removing `__themed`, which was a bit of an anti-pattern.


lib/css/base/_stacks-configuration-dynamic.less
```diff
@@ -107,16 +107,6 @@
...
-    @tags-border:                                    transparent;
-    @tags-background:                                var(--powder-100);
-    @tags-color:                                     var(--powder-700);
-    @tags-hover-border:                              transparent;
-    @tags-hover-background:                          var(--powder-200);
-    @tags-hover-color:                               var(--powder-800);
-    @tags-selected-border:                           transparent;
-    @tags-selected-background:                       var(--powder-400);
-    @tags-selected-color:                            var(--powder-900);
-
```
aaronshekey: This may be a premature optimization—we may need this for current Less-based theming. We'll have to test.

aaronshekey:
This should now be in play again...

b-kelly (commented):
Have you considered adding theming override variables for these like we do buttons? That'd give us free network support and would unify the styles with the legacy `.post-tag` class on Prod. Here are the variables used for network theming:

```
--theme-tag-color
--theme-tag-background-color
--theme-tag-border-color
--theme-tag-color-hover
--theme-tag-background-color-hover
--theme-tag-border-color-hover
```

I also checked Prod to see if the dynamic config variables are used outside of Stacks and... no they are not.

PR #718, opened 2021-07-21
Use the same `:hover` color on `:focus` in Notices

allejo:
**Currently**

wrong-color-for-notices

**Fixed**

right-color-for-notices


aaronshekey:
Thank you!

PR #719, opened 2021-07-22
Improve theme dark mode colors algorithm

aaronshekey:
This increases contrast across the generated colors in dark mode. This is especially evident in our `notices` in dark mode. This should improve things in the spirit of a hotfix.

### DIFF
--- a/lib/css/exports/_stacks-constants-colors.less
+++ b/lib/css/exports/_stacks-constants-colors.less
@@ -532,36 +532,36 @@
     --theme-secondary-color: .assemble-color(theme-secondary-color)[];
 
     // Primary Theming
-    --theme-primary-025: .native-darken-desaturate(theme-primary-color, 15, 56)[];
-    --theme-primary-050: .native-darken-desaturate(theme-primary-color, 12, 49)[];
-    --theme-primary-075: .native-darken-desaturate(theme-primary-color, 10, 42)[];
-    --theme-primary-100: .native-darken-desaturate(theme-primary-color, 8, 35)[];
-    --theme-primary-150: .native-darken-desaturate(theme-primary-color, 6, 28)[];
-    --theme-primary-200: .native-darken-desaturate(theme-primary-color, 4, 21)[];
-    --theme-primary-300: .native-darken-desaturate(theme-primary-color, 2, 14)[];
-    --theme-primary-350: .native-darken-desaturate(theme-primary-color, 0, 7)[];
-    --theme-primary-400: .native-tint(theme-primary-color, 11)[];
-    --theme-primary-500: .native-tint(theme-primary-color, 26)[];
-    --theme-primary-600: .native-tint(theme-primary-color, 41)[];
-    --theme-primary-700: .native-tint(theme-primary-color, 56)[];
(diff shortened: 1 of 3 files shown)


dancormier (commented):
On the whole this is a great improvement, but the contrast on certain button states (particularly in default filled) seems to have dropped. I found myself putting in just a little extra effort to read them.

**Before**




**After**




I included the WCAG ratios because, while they're imperfect, they're quick to grab and I do think they're relevant here.

---

While I'm here, I'll point out a couple of spots I found with rough contrast ratios that might warrant consideration while we're here.

- [User Cards > Highlighted](http://localhost:8080/product/components/user-cards/#highlighted)
- [User Cards > Full](http://localhost:8080/product/components/user-cards/#full)
- [Link Previews > Code Blocks](http://localhost:8080/product/components/link-previews/#code-blocks)

The contrast ratios on all of these are _better_, but still are on the edge of legibility imo.

aaronshekey:
Ok, there should be better contrasts levels in these situations. It's difficult to nail a single algo across all colors, but I think this is completely serviceable and there's always higher contrast level modes we can build in as well.

lib/css/exports/_stacks-constants-colors.less
```diff
@@ -532,37 +532,36 @@
...
-    --theme-secondary-200: .native-darken-desaturate(theme-secondary-color, 15, 25)[];
-    --theme-secondary-300: .native-darken-desaturate(theme-secondary-color, 8, 15)[];
-    --theme-secondary-350: .native-darken-desaturate(theme-secondary-color, 3, 10)[];
-    --theme-secondary-400: var(--theme-secondary-color);
-    --theme-secondary-500: .native-tint(theme-secondary-color, 11)[];
-    --theme-secondary-600: .native-tint(theme-secondary-color, 32)[];
-    --theme-secondary-700: .native-tint(theme-secondary-color, 45)[];
-    --theme-secondary-800: .native-tint(theme-secondary-color, 70)[];
-    --theme-secondary-900: .native-tint(theme-secondary-color, 80)[];
+    --theme-secondary-025: .native-darken-desaturate(theme-secondary-color, 17, 82)[];
```
aaronshekey, line 551: I don't love the big visual jump between `025` and `050`, but no one is ever practically using those colors together like we display them on the colors page. Contrast is king, not a smooth gradient.

dancormier: For _some_ colors, `025` winds up gray though.






Looks much better with our custom theme but I think the default orange might point to an issue.

aaronshekey: Yeah, I pretty massively desaturated those since `025` is only used in the context of the highlighted user card, which was just way too little contrast and saturated at `050`. Happy to revisit this gradient / curve but want to improve things in dark mode in production as soon as possible.

dancormier: That's fair

dancormier (approved):
Approving in the spirit not letting the perfect be the enemy of the good.

Seriously though, this is a really nice improvement to dark mode color and contrast.

aaronshekey:
Color is magic. This keeps it usable while we refine 🎉

PR #720, opened 2021-07-23
Respect `prefers-reduced-motion` for toast animations

allejo:
**With animations (untouched)**
with-animations

**Prefers reduced motion (no animation)**
reduced-motion

### DIFF
--- a/lib/css/components/_stacks-notices.less
+++ b/lib/css/components/_stacks-notices.less
@@ -189,4 +189,12 @@
         transform: translate3d(0, 0, 0);
         transition: visibility 0s 0s, opacity 100ms @te-smooth 0s, transform 100ms @te-smooth 0s;
     }
+
+    @media (prefers-reduced-motion) {
+        transition: none;
+
+        &[aria-hidden="false"] {
+            transition: none;
+        }
+    }
 }


dancormier (approved):
Woo nice one!

bnickel:
@dancormier I have a concern on this approach.  Our code that renders these on SE uses the `transitionend` to know that the element has faded away and can be removed from the DOM.  Without an actual transition, the finalizing code will fail to run.  We'd run into the same issue if we did this with `s-modal`.

One solution, which mirrors how Apple handles `prefers-reduced-motion`, would be to remove the translation and keep the fade.

dancormier:
Oh interesting (and thanks for catching it) @bnickel. I hadn't considered that the JS would depend on CSS transition properties. I'll ~~revert this merge and~~ open a new PR that uses your suggested approach.

dancormier:
@bnickel to clarify, the approach in this PR does retain a transition on `opacity`. Is that sufficient?

bnickel:
Yeah, there just needs to be some transition, so something like:

```
     @media (prefers-reduced-motion) {
         transition: opacity 60ms @te-smooth-slow 0ms, visibility 0s 150ms;

         &[aria-hidden="false"] {
             transition: visibility 0s 0s, opacity 100ms @te-smooth 0s;
         }
     }
```

PR #721, opened 2021-07-23
Retain `.s-toast` transition while respecting prefers-reduced-motion

dancormier:
@bnickel pointed out that our `.s-toast` component has JS that relies on the a CSS transition.

This PR builds on top of @allejo's work in https://github.com/StackExchange/Stacks/pull/720


bnickel (approved):
Tested and works as expected

PR #730, opened 2021-08-04
added atomics for object-fit and object-position

benjamin-hodgson:
As discussed in chat

### DIFF
--- a/docs/_data/atomics.json
+++ b/docs/_data/atomics.json
@@ -1349,17 +1349,17 @@
     {
       "class": ".of-fill",
       "output": "object-fit: fill",
-      "define": "<p class='mb0'>Stretch or squash the image, so that the whole container is filled. The image's aspect ratio may change as a result.</p>"
+      "define": "<p class='mb0'>Stretch or squash the image, so that the whole container is filled. The image’s aspect ratio may change as a result.</p>"
     },
     {
       "class": ".of-none",
       "output": "object-fit: none",
-      "define": "<p class='mb0'>Don't resize the image at all. This may result in the image getting cropped, or there being some empty space in the container.</p>"
+      "define": "<p class='mb0'>Don’t resize the image at all. This may result in the image getting cropped, or there being some empty space in the container.</p>"
     },
     {
       "class": ".of-scale-down",
       "output": "object-fit: scale-down",
-      "define": "<p class='mb0'>If the image fits inside the container, it's not resized. If it's too large, it will be scaled down just like <code>.of-contain</code>.</p>"
+      "define": "<p class='mb0'>If the image fits inside the container, it’s not resized. If it’s too large, it will be scaled down just like <code class='stacks-code'>.of-contain</code>.</p>"
     },
     {
       "class": ".op-center",
--- a/docs/_data/atomics.json
+++ b/docs/_data/atomics.json
@@ -1364,7 +1364,7 @@
     {
       "class": ".op-center",
       "output": "object-position: 50% 50%",
-      "define": "<p class='mb0'>Positions the image in the centre of the container.</p>"
+      "define": "<p class='mb0'>Positions the image in the center of the container.</p>"
     }
   ]
 }
--- a/docs/_data/atomics.json
+++ b/docs/_data/atomics.json
@@ -1339,32 +1339,32 @@
     {
       "class": ".of-contain",
       "output": "object-fit: contain",
-      "define": "<p class='mb0'>Resize the image, while maintaining its aspect ratio, so that the whole image fits inside the container. This may result in some empty space in the container.</p>"
+      "define": "<p class='mb0'>Scale the content to fit its content box. This maintains its aspect ratio and may result in empty space in the content box.</p>"
     },
     {
       "class": ".of-cover",
       "output": "object-fit: cover",
-      "define": "<p class='mb0'>Resize the image, while maintaining its aspect ratio, so that the whole container is filled. The image may get cropped as a result.</p>"
+      "define": "<p class='mb0'>Scale the content to fill its content box. This maintains its aspect ratio and may result in the content being cropped.</p>"
     },
     {
       "class": ".of-fill",
       "output": "object-fit: fill",
-      "define": "<p class='mb0'>Stretch or squash the image, so that the whole container is filled. The image’s aspect ratio may change as a result.</p>"
+      "define": "<p class='mb0'>Resize the content’s dimensions to match its content box. This may result in the content’s aspect ratio being altered.</p>"
     },
     {
       "class": ".of-none",
       "output": "object-fit: none",
-      "define": "<p class='mb0'>Don’t resize the image at all. This may result in the image getting cropped, or there being some empty space in the container.</p>"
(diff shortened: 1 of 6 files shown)


dancormier (approved):

docs/_data/atomics.json
```diff
@@ -1334,5 +1334,37 @@
...
+    },
+    {
+      "class": ".of-scale-down",
+      "output": "object-fit: scale-down",
+      "define": "<p class='mb0'>If the image fits inside the container, it's not resized. If it's too large, it will be scaled down just like <code>.of-contain</code>.</p>"
+    },
+    {
+      "class": ".op-center",
+      "output": "object-position: 50% 50%",
+      "define": "<p class='mb0'>Positions the image in the centre of the container.</p>"
```
dancormier: > centre

@benjamin-hodgson are you British?

benjamin-hodgson: Yes

aaronshekey: Sadly yes. :trollface: 

But! We have [opinions](https://stackoverflow.design/content/guidelines/grammar-and-mechanics/#use-american-spelling-and-standards) on this. 🇺🇸

lib/css/atomic/_stacks-misc.less
```diff
@@ -80,6 +80,18 @@
...
+//  $  OBJECT-FIT / OBJECT-POSITION
+//  ----------------------------------------------------------------------------
+
+.of-contain { object-fit: contain !important; }
+.of-cover { object-fit: cover !important; }
+.of-fill { object-fit: fill !important; }
+.of-none { object-fit: none !important; }
+.of-scale-down { object-fit: scale-down !important; }
+
+.op-center { object-position: 50% 50% !important; }
```
dancormier:
```suggestion
.op-center { object-position: center !important; }
```

This suggestion is just a _very_ semantic one, but I tend to prefer the value to match the class name when possible.

docs/_data/atomics.json
dancormier:
```suggestion
      "define": "<p class='mb0'>Positions the image in the center of the container.</p>"
```

lib/css/atomic/_stacks-misc.less
benjamin-hodgson: I have a vague memory of `object-position: center` not actually working last time I tried it

dancormier: `center` should work just the same. Tested locally and seemed fine.

Which makes me realize we should also add an example for `.op-center`. I'll pop one in here.

aaronshekey (approved):

benjamin-hodgson:
@dancormier I went with "cropping" as the title of the doc page because "object fit" is not very easy to find if you don't already know what it means. I suspect most of the full-stack devs on the team would not know what "object fit" meant if they were browsing around the docs

benjamin-hodgson:
Also, forgive me but I really don't think c369593a85fb9e243a2a147646416ce33c62609b is an improvement

aaronshekey:
What about those who know CSS?

benjamin-hodgson:
I would bet you good money that a large majority of devs on the team don’t know that `object-fit` is a thing. At least three people I’ve talked to *today* didn’t know

dancormier:
I have my reservations about going with `object fit`, but I feel that `cropping` doesn't quite encapsulate what `object-fit` does. To me, cropping specifically implies cutting off edges of an element while `object-fit` may scale, stretch, or resize objects and not necessarily crop them. `Object fit` is a bit more accurate (if obscure to many devs).

I'd be open to other names and realize that the most accurate name is not always the best at communicating.

kylejrp:
👋 I knew about `object-fit` but I don't think that I'm in the majority since I helped maintain a CSS framework before this 😅

I personally like `object fit` while recommending against `cropping`, but also agree with the idea that there's potentially a better name out there. Since we use `crop` in the description of one of the properties, it should come up in the search once we push live, yeah?

Would be nice if the difference between `of-contain` and `of-scale-down` was clear when the image is smaller than the box containing it - `of-contain` will unintuitively scale-up:
image

dancormier:
> Also, forgive me but I really don't think [c369593](https://github.com/StackExchange/Stacks/commit/c369593a85fb9e243a2a147646416ce33c62609b) is an improvement

I was being a bit pedantic and I'm open to changing whatever seems unclear.

- I view `scale` as meaning maintaining aspect ratio while resizing (and considered dropping the line about aspect ratio altogether for this reason). If the content could be stretched, I used `resize`.
- I went for `content` vs `image` because `object-fit` can be used on `<video>`s as well (though this might be an uncommon use case for Stack Overflow).
- I chose `content box` instead of `container` because  `container` implies the parent DOM element. I'm not sure if this is the common interpretation.
- Overall, I tended toward consistency and conciseness to make the descriptions more digestible, figuring that we're basically describing the CSS properties in that table and there are other sources that go into greater detail if needed.

If you have suggestions I'll look it over and we can refine.

Or yell "JUST REVERT IT" and I'll revert that commit 😛

benjamin-hodgson:
There's already tons of precise-but-hard-to-understand CSS documentation out there. If someone wants the precise version they can go to MDN. Remember that your audience is Stack Overflow developers, who mostly know enough CSS to get by but don't particularly like it or know all of the terminology. You probably have the most CSS knowledge out of any developer in the department! So what makes sense to you won't necessarily make sense to the rest of us.

"Scale the object to fit its content box" is more economical and precise than "Resize the image, while maintaining its aspect ratio, so that the whole image fits inside the container", but it's quite a lot harder to understand. It's also way too similar to "Scale the object to fill its content box".

I'd rather be understood than right!

dancormier:
@benjamin-hodgson I hear you and I'm reconsidering the language. At the very least, it'll give me an opportunity to proofread (I noticed a couple typos I made in my haste yesterday).

This is one of those precision vs clarity things that is so tricky to navigate. I have thoughts after looking into this with fresh eyes.

- I can see _too much_ consistency in places (your `fit` vs `fill` example is a good illustration). I'm reconsider overlapping descriptions.
- You're right that there's some exactness that isn't really necessary in the descriptions. I'm looking into the descriptions to make them generally easier to comprehend and more practical.
- I'm still questioning `content` vs `image`. I _generally_ dislike the term `content` for its ambiguity but chose it to not exclude usages outside of `img`. That being said, I recognize that 95% of the time, the class will be used on images. I'm kinda flip-floppy on this.
- I want to hold firm on `content box` vs `container` for the reason I stated and because there's already lots of precedent in the docs (see [Box sizing](https://stackoverflow.design/product/base/box-sizing/#classes) and [Overflow](https://stackoverflow.design/product/base/overflow/#classes)). With the intended audience being consumers of Stacks, I want someone that's already poked around elsewhere to see consistent language here.

I [made a PR](https://github.com/StackExchange/Stacks/pull/733) with some changes that I think makes the documentation a bit more digestible. Feel free to push a commit or two there!

dancormier:
> Would be nice if the difference between `of-contain` and `of-scale-down` was clear when the image is smaller than the box containing it - `of-contain` will unintuitively scale-up:
> image

@kylejrp on it and adding it to #733

PR #735, opened 2021-08-10
Improvements to mod badge

aaronshekey:
This calms down the badges a bit and switches our approach to coloring them.

cc @yellis


yellis:
LGTM @aaronshekey.

Diamond tag for any badge coming in a different PR?

When is this due to hit prod?

aaronshekey:
Got some feedback from the design crew that we'd rather not mix the staff and mod badge. Going to leave them separate for now and see if the aesthetic change is enough.

PR #736, opened 2021-08-12
Add a navigation dropdown

aaronshekey:
Look, sometimes you just have so many links you've gotta drop a bunch of them into a `More` button and call it a day. I don't make the rules.

This can be done responsively, wrapping any of those last elements in an `s-menu`.

image

### DIFF
--- a/lib/css/components/_stacks-navigation.less
+++ b/lib/css/components/_stacks-navigation.less
@@ -26,6 +26,7 @@
         display: flex;
         align-items: center;
         padding: @su6 @su12;
+        position: relative;
 
         // Reset some things for when the navigation item is also a button
         border: none;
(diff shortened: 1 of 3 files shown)


lib/css/components/_stacks-navigation.less
```diff
@@ -50,6 +51,24 @@
...
+
+        &.s-navigation--item__dropdown {
+            padding-right: 2em;
+
+            &:after {
+                content: "";
+                position: absolute;
+                z-index: @zi-active;
+                top: calc(50% - 2px);
+                right: 0.9em;
```
dancormier, line 63: I notice button takes the same approach to positioning a caret. Do you know of any reason we're going with absolute? We _could_ do something like this:

```suggestion
            &:after {
                content: "";
                margin-left: 0.5em;
```

This is small potatoes and I consider "consistency" and "it works fine" a valid reason, so feel free to ignore this suggestion if that's the case.

dancormier (approved):
LGTM! (I made a note but it's not a blocker).

lib/css/components/_stacks-navigation.less
aaronshekey: I think this was to control the vertical alignment to the pixel level. Happy to revisit this!

PR #739, opened 2021-08-17
Generate responsive/non-responsive stylesheets

dancormier:
This PR generates a CSS file that is responsive-only CSS file (and eventually another that is non-responsive-only).

This is so consumers can greatly reduce payload size of Stacks CSS by [using media queries](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-media) when including stylesheets.

Update Tue Aug 17, 2021: Turns out `<link>` media queries don't block loading of an asset but do prevent render blocking CSS. Source: [https://developers.google.com/web/fundamentals/performance/critical-rendering-path/render-blocking-css#tldr](https://developers.google.com/web/fundamentals/performance/critical-rendering-path/render-blocking-css#tldr)


dancormier:
I did some discovery and have some findings:

# Shorthand vocab

I'm going to use the following terms for brevity. Pipe down, pedants.

- `Desktop`: Clients with viewport widths _outside_ of our largest mobile style -`@media(min-width: 1265px)`
- `Mobile`: Clients with viewport widths _within_ of our largest mobile style - `@media(max-width: 1264px)`
- `Non-responsive styles`: Styles that are applied to any viewport width.
- `Responsive styles`: Styles that are conditional on a media query tied to width.


# tl;dr

We can split our CSS files into responsive/non-responsive (and roll them together into stacks.css).

For Stacks, the biggest potential drawback is consumers continuing to only use stacks.min.css (a YAGNI situation).

## Consuming as split responsive/non-responsive

For desktop clients, this would result in less render-blocking CSS on desktop (but an additional non-blocking HTTP request).

For mobile clients, this would only result in an additional HTTP request up front.

# `<link>` and its `media` attribute

I discovered that `<link>`s can accept a `media` attribute. For example:

```
<link href="mobile.css" rel="stylesheet" media="screen and (max-width: 600px)">
```
*The above example is from [MDN `<link>` documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-media)*

I initially assumed that the appropriate `media` attribute on a `<link>` could prevent the client from downloading the asset. I later found that [the `media` attribute only prevents the asset from blocking rendering](https://developers.google.com/web/fundamentals/performance/critical-rendering-path/render-blocking-css#tldr). The client will request the asset, even if the `media` condition is unmet.

# Questions

1. Can we reduce client load times by only serving them a non-responsive CSS if responsive styles are unneeded?
2. Can this change empower us to include responsive conditional classes in Stacks more liberally?
3. Can we do the above without a significant impact server-side?
4. Can we improve Stacks code structure?

## Client load times

> Can we reduce desktop client load times by only serving them a non-responsive CSS if responsive styles are unneeded?

**Answer**: Yes, mostly.

Since unmet conditions on `link[media]` prevent the asset from blocking render. This has the potential to speed up page load for desktop clients. The desktop client will still request the responsive styles, but the render of the page will not be block by the request.

### Mobile impact

This would most likely _increase_ the load time for mobile clients, since it would need to request two files.

### Existing non-responsive views

We _could_ serve only non-responsive styles to non-responsive views. This would reduce request size on these pages. This sounds cool, but considering we have a goal to [make the entirety of Stack Overflow responsive](https://meta.stackexchange.com/questions/367609/deprecating-our-mobile-views), this would only be a short term win.

## Empowering more responsive classes

> Can this change empower us to include responsive conditional classes in Stacks more liberally?

**Answer**: It depends.

Here are some pros and cons:

### Pros

- We could serve any size stylesheet to desktop clients without further delaying render.

### Cons
- A responsive stylesheet will be requested on all clients.
  - It's request will be render-blocking for mobile clients and will result in more data transfer.
  - On desktop, more responsive classes will mean larger total transfer size. 

## Server impact

> Can we do the above without a significant impact server-side?

**Answer**: No… well, it depends on your definition of "_significant_".

The proposed approach would have consumers making an additional CDN hit, regardless of desktop/mobile. I'm not sure how to weigh the cost of that extra request.

## Stacks code structure

> Can we improve Stacks code structure?

**Answer**: I think so, but this is subjective.

The approach I took extracts all responsive style generation into separate files from the non-responsive styles. I personally like this structure. 

# Metrics

|                               | Production | This PR  | Prod GZipped | PR GZipped | 
|-------------------------------|------------|----------|------------|----------|
| stacks.min.css                | 354.4 KB   | 354.1 KB | 48.7 KB | 48.4 KB |
| stacks.non-responsive.min.css | -        | 312.6 KB | - | 42.9 KB |
| stacks.responsive.min.css     | -        | 41.5 KB  | - | 5.8 KB |

Using the split CSS proposed in this PR would:
- Save us ~12% (41.8 KB raw, 5.9 KB GZipped) on render-blocking CSS on Desktop.
- Save us ~nothing on total transfer.
- Add an additional HTTP request.

# Conclusion and recommendations

Coming soon.

## Bonus findings

We have _tons_ of instances (I counted 228) of `@media (prefers-color-mode: dark) {` in `stacks.min.css` We could save ~8 KB if we found a way to combine them (did someone say postcss?).

Speaking of dark mode, this approach could be worth considering for splitting out a `prefers-color-mode: dark` stylesheet. If you're not using dark mode, why have it block the render? (well, it would add another request, that's why. Still worth considering IMO).

aaronshekey:
Apologies for YOLOing responsive flex and inline flex.

dancormier:
> Apologies for YOLOing responsive flex and inline flex.

It's all good. I though I missed it the first go 'round.

aaronshekey:
If putting `links` behind a media query acted as you’d originally expected... when you loaded the page on a desktop at around 1200px, you’d be adding another request at the medium breakpoint. This would likely cause a bit of jank as the styles are downloaded and applied.

I think at our scale, we should pay the closest attention to requests. I'm not sure splitting the bundle would improve things because of that scale.

We should probably collect all our dark mode styles in the same fashion as our responsive ones, but I haven't quit figured it out.

I think the biggest impact would be shifting all styling to Stacks, and deleting the entirety of `primary.css` and `secondary.css`, leaving a single, mostly atomic CSS bundle for all users.

dancormier:
I initially hoped we could make one request for <= 1264px with _everything_ and a single, smaller request for > 1264px with only non-responsive styles. It would provide little benefit for viewports below our largest breakpoint but would give a little boost for those > lg.

And (if it functioned the way I initially hoped) I thought we might want to break it down per media query like so:
```
<link href="stacks.min.css" rel="stylesheet" media="(max-width: 640px)">
<link href="md-lg-and-nonresponsive.min.css" rel="stylesheet" media="(min-width: 641px) and (max-width: 980px)">
<link href="lg-and-nonresponsive.min.css" rel="stylesheet" media="(min-width: 981px) and (max-width: 1264px)">
<link href="nonresponsive.min.css" rel="stylesheet" media="(min-width: 1265px)">
```

With what we know now, it would obviously be a terrible idea to do the above. And the benefits of a single responsive/non-responsive stylesheet split is also questionable.

@aaronshekey forgetting about splitting up CSS files, do you see any benefit to splitting up the responsive (or, thinking bigger picture, conditional) styles from the base styles structurally?

PR #740, opened 2021-08-17
Explain the static width classes

aaronshekey:
Some clarity here

### DIFF
--- a/docs/product/base/width-height.html
+++ b/docs/product/base/width-height.html
@@ -6,7 +6,7 @@ description: Stacks provides atomic sizing classes for percentage-based widths &
 <section class="stacks-section">
     {% header "h2", "Widths" %}
     {% header "h3", "Width classes" %}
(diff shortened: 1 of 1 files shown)


docs/product/base/width-height.html
```diff
@@ -188,8 +188,10 @@
 
 <section class="stacks-section">
     {% header "h2", "Static widths" %}
+    <p class="stacks-copy">Static widths are based on the full width of our 12-column grid. The grid is defined in <code class="stacks-code">rems</code>, which are tied to the base font size. At the default font size, the full width is <code class="stacks-code">1264px</code>. A single column is approximately <code class="stacks-code">105px</code>.</p>
+    <p class="stacks-copy">At the smallest breakpoint, our font size is reduced, and these static widths are therefore also reduced.</p>
```
bnickel, line 192: "Static" widths 🤨

PR #742, opened 2021-08-18
Responsify .bg-inherit

aaronshekey:
This came up today in a profiles refactor as a way to remove a background color at a smaller breakpoint.


aaronshekey:
Meh, found a different direction.

PR #744, opened 2021-08-20
Component color variables

b-kelly:
Allows us to set custom colors for certain components on an opt-in basis. Useful for themes where you want your (for example) buttons to deviate from using secondary colors to something else.

Additionally, this allows us to completely remove the need for our hacky `_stacks-configuration-*` files and `#stacks-internals #load-config()` calls.


lib/css/base/_stacks-configuration-dynamic.less
```diff
@@ -21,38 +21,9 @@
...
-    @button-filled-selected-color:                   var(--theme-secondary-900);
-    @button-filled-selected-background-color:        var(--theme-secondary-300);
-    @button-filled-selected-border-color:            var(--theme-secondary-500);
-
-    //  Button Outline
-    @button-outlined-border-color:                   var(--theme-secondary-300);
-    @button-outlined-selected-border-color:          var(--theme-secondary-400);
+    @body-site-width:                                1060px; // TODO not overridden in prod
 
+    // TODO not overridden in prod
```
aaronshekey: Glad all these TODOs are identified. I wonder if these should actually be `// Deprecated` comments? Or, honestly, I'd be happy to just yeet these straight away.

lib/css/components/_stacks-buttons.less
```diff
@@ -16,13 +16,11 @@
 //  $   BASE STYLE
 //  ----------------------------------------------------------------------------
 & {
-    #stacks-internals #load-config();
```
aaronshekey, line 19: Absolutely thrilled to see these going.

lib/css/base/_stacks-configuration-dynamic.less
```diff
@@ -96,16 +72,7 @@
...
-    @button-primary-hover-color:                     @white;
-    @button-primary-hover-background-color:          var(--theme-secondary-500);
-    @button-primary-active-background-color:         var(--theme-secondary-700);
-    @button-primary-selected-color:                  var(--white);
-    @button-primary-selected-background-color:       var(--theme-secondary-700);
-    @button-primary-number-color:                    var(--theme-secondary-900);
-
+    // TODO not overridden in prod
     //  Tags
     @tags-border:                                    transparent;
```
aaronshekey, line 77: When exploring better theming on tags, I ended up deleting these. #717. I could see this entire file going away.

aaronshekey (approved):
I think this approach looks great. I ran this through Backstop and it didn't catch anything unexpected. I made a note or two about things we could do _now_ instead of later.

lib/css/base/_stacks-configuration-dynamic.less
b-kelly: Changed to `// DEPRECATED` comments. I'm going to come back in a follow up PR to yeet these entirely once I make triple sure these are not used anywhere (I'm 90% sure they aren't though).

lib/css/base/_stacks-configuration-dynamic.less
b-kelly: Nice! I'm going to leave these in at the moment so you don't have to solve a merge conflict in your PR. Happy to see these are going away early.

aaronshekey (approved):

PR #745, opened 2021-08-26
Add a skip to content link

aaronshekey:
This PR adds the ability to hit `tab` and show an otherwise-hidden `Skip to content` button. It allows keyboard users to skip all the nonsense and get straight to the content.


dancormier (approved):
A sensible implementation of a skip link. I like it.

🦐

PR #748, opened 2021-08-31
Staff badge contrast tweaks

b-kelly:
Darkens the staff badge borders to increase contrast. The current colors are a tad hard to see against the background and gives the visual illusion that the Staff badge is smaller than the Mod badge at small sizes. Also reported/suggested on [MSE](https://meta.stackexchange.com/questions/369245/can-the-staff-label-resemble-the-mod-label).

## Old

image
image
image

## New

image
image
image

### DIFF
--- a/lib/css/components/_stacks-badges.less
+++ b/lib/css/components/_stacks-badges.less
@@ -159,7 +159,7 @@
 }
 .s-badge__staff {
     // Staff badges are always "Stack Overflow Orange"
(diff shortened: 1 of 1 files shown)


aaronshekey:
This is a nice refinement. For whatever reason, the 'old' version isn't as pronounced in the docs as when it's displayed on SO. Likely because I was looking at `Admin` vs. `Mod` and not `Staff` vs. `Mod`.

aaronshekey (approved):

PR #750, opened 2021-09-03
Allow for text inside activity indicators

### LINKED ISSUE #749: Activity indicators should allow for numbers
We ought to be able to include a number within `s-activity-indicator` to show _how much_ activity there is.

Like so:

image

aaronshekey:
Closes #749 :v:

### DIFF
--- a/lib/css/components/_stacks-activity-indicator.less
+++ b/lib/css/components/_stacks-activity-indicator.less
@@ -23,6 +23,7 @@
     box-shadow: 0 0 0 @su4 var(--focus-ring);
     border-radius: 1000px;
     font-size: @fs-fine;
+    font-weight: 600;
     color: @white;
     text-transform: uppercase;
 }
--- a/lib/css/components/_stacks-activity-indicator.less
+++ b/lib/css/components/_stacks-activity-indicator.less
@@ -13,7 +13,6 @@
 
 .s-activity-indicator {
     display: inline-block;
-    text-align: center;
     min-width: @su12;
     height: @su12;
     padding-left: @su4;
--- a/lib/css/components/_stacks-activity-indicator.less
+++ b/lib/css/components/_stacks-activity-indicator.less
@@ -26,3 +26,18 @@
     color: @white;
     text-transform: uppercase;
 }
+
+.s-activity-indicator__success {
+    background-color: var(--green-500); // TODO .bg-success
+    box-shadow: 0 0 0 @su4 var(--focus-ring-success);
+}
+
+.s-activity-indicator__warning {
+    background-color: var(--yellow-600); //TODO .bg-warning
+    box-shadow: 0 0 0 @su4 var(--focus-ring-warning);
+}
+
+.s-activity-indicator__danger {
+    background-color: var(--red-500); //TODO .bg-danger
+    box-shadow: 0 0 0 @su4 var(--focus-ring-error);
+}
--- a/lib/css/components/_stacks-activity-indicator.less
+++ b/lib/css/components/_stacks-activity-indicator.less
@@ -17,7 +17,7 @@
     height: @su12;
     padding-left: @su4;
     padding-right: @su4;
-    line-height: 1;
+    line-height: 1.1; // Custom line-height to satisfy 1x screens
     background-color: var(--theme-secondary-400);
     box-shadow: 0 0 0 @su4 var(--focus-ring);
     border-radius: 1000px;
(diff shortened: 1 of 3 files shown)


lib/css/components/_stacks-activity-indicator.less
```diff
@@ -13,9 +13,17 @@
...
     height: @su12;
+    padding-left: @su4;
+    padding-right: @su4;
+    line-height: 1;
     background-color: var(--theme-secondary-400);
     box-shadow: 0 0 0 @su4 var(--focus-ring);
-    border-radius: 100%;
+    border-radius: 1000px;
+    font-size: @fs-fine;
+    font-weight: 600;
```
dancormier, line 25: ~~not~~ **good.**

lib/css/components/_stacks-activity-indicator.less
```diff
@@ -13,9 +13,17 @@
 
 .s-activity-indicator {
     display: inline-block;
-    width: @su12;
+    text-align: center;
+    min-width: @su12;
     height: @su12;
+    padding-left: @su4;
+    padding-right: @su4;
+    line-height: 1;
```
dancormier: At viewport widths < `lg`, line height gets kinda jacked up.



At other widths, the vertical negative space is minimal to the point of impacting legibility IMO.



I'd changing the `height` to a `min-height` and setting an increased `line-height` for this result:



```suggestion
    min-width: @su12;
    min-height: @su12;
    padding-left: @su4;
    padding-right: @su4;
    line-height: @lh-md;
```

lib/css/components/_stacks-activity-indicator.less
```diff
@@ -13,9 +13,17 @@
 
 .s-activity-indicator {
     display: inline-block;
-    width: @su12;
+    text-align: center;
```
dancormier: I don't think this `text-align` is needed. We could save ~20 bytes if you never see this being multi-line.

```suggestion
```

lib/css/components/_stacks-activity-indicator.less
```diff
@@ -13,9 +13,17 @@
...
+    padding-left: @su4;
+    padding-right: @su4;
+    line-height: 1;
     background-color: var(--theme-secondary-400);
     box-shadow: 0 0 0 @su4 var(--focus-ring);
-    border-radius: 100%;
+    border-radius: 1000px;
+    font-size: @fs-fine;
+    font-weight: 600;
+    color: @white;
```
dancormier, line 26: This should be `var(--white)` to play with nice with dark mode, correct?

as is:


using `var(--white)`:


```suggestion
    color: var(--white);
```

dancormier (changes requested):
A few minor suggested tweaks but looking (**SHARP**).

lib/css/components/_stacks-activity-indicator.less
dancormier: BTW, this will increase the height of the activity indicator from 12px to about 14px when it has text inside. I imagine that should be fine but I figured I should call that out.

aaronshekey: I can't repro this, is it a `1x` screen issue? 🤔

lib/css/components/_stacks-activity-indicator.less
aaronshekey: I chose white to try to keep contrast :v:

lib/css/components/_stacks-activity-indicator.less
dancormier: Ah, yep. Shoved my browser onto my high DPR screen and looks better there.

### At 1x (despite what Firefox says)



### At 2x

dancormier: Upon further investigation, this may be a Gecko-only `1x` issue. I checked Chrome/Safari and no weird line-height found there.

b-kelly:
I added new `success, warning, danger` variants for use with the new topbar component:

image

lib/css/components/_stacks-activity-indicator.less
dancormier: **tl;dr**:  I'd suggest we use `line-height: 1.1;` and move on with our lives (after checking different OSes/browsers/fonts in our stack for issues, of course).

---

This issue happens specifically with `line-height: 1;`. It seems that just about anything between `1` and `@lh-sm` could work but introducing a new value for this edge case feels just a little icky. I really wanted `@lh-sm` to work, but it shifts the text down a pixel.

## Examples

### `line-height: 1;`



### `line-height: 1.1;`

aaronshekey: How about now, @dancormier?

dancormier: 👍 LGTM

dancormier (approved):
👍 🚀 🦐

PR #754, opened 2021-09-13
Topbar component

b-kelly:
Stacks topbar component. Aims to achieve visual fidelity with what we have on the Stack Exchange network, while relying on Stacks for providing the bulk of the styling.


lib/css/components/_stacks-topbar.less
```diff
@@ -0,0 +1,366 @@
...
+    z-index: @zi-navigation-fixed;
+    background-color: var(--theme-topbar-background-color);
+    height: var(--topbar-height);
+    display: flex;
+    border-top: var(--theme-topbar-accent-border);
+    position: relative;
+
+    // TODO necessary? used to wrap content so the topbar stretches 100% w/ content at some value below that
+    .s-topbar--container {
+        width: 100%; // TODO note - consumers should use atomic classes to override this
```
aaronshekey: I think by default, we can set this to be `@site-width` or `wmx12`. The non-default behavior we may see is someone adding `wmx-initial` to `.s-topbar--container`.

lib/css/components/_stacks-topbar.less
```diff
@@ -0,0 +1,366 @@
...
+            &:before,
+            &:after {
+                position: absolute;
+                content: '';
+                width: 18px;
+                height: 2px;
+                left: 0;
+                background: var(--theme-topbar-item-color);
+                top: -5px;
+                transition: all ease-in-out 0.1s;
```
aaronshekey: Nit: I try to avoid using `all` since that'll add a transition to color changes and other unpredictable things. Better to just define `transform` and `top` or whatever individual thing.

lib/css/components/_stacks-topbar.less
```diff
@@ -0,0 +1,366 @@
...
+        color: var(--theme-topbar-item-color);
+        display: inline-flex;
+        align-items: center;
+        padding: 0 10px; // TODO @su8 or @su12?
+        text-decoration: none;
+        white-space: nowrap;
+        position: relative;
+        white-space: nowrap;
+
+        &:hover, &:focus,
```
aaronshekey: All separate lines, please. If you've got `stylelint` installed in your editor, it should underline these as linting errors :v:

lib/css/components/_stacks-topbar.less
```diff
@@ -0,0 +1,366 @@
...
+
+// CTAs: Sign in, Sign up, etc. on the far right
+.s-topbar .s-topbar--ctas {
+    padding-left: @su4;
+    padding-right: @su8;
+    // push this section as far to the right as possible
+    margin-left: auto;
+
+    li + li {
+        margin-left: @su4;
```
aaronshekey, line 308: `gap` can not come soon enough.

lib/css/components/_stacks-topbar.less
```diff
@@ -0,0 +1,366 @@
...
+.s-topbar__light {
+    --theme-topbar-background-color:           @black-025;
+
+    // Search input
+    --theme-topbar-search-color:               @black-700;
+    --theme-topbar-search-background:          @white;
+    --theme-topbar-search-placeholder:         @black-200;
+    --theme-topbar-search-border:              @black-200;
+    --theme-topbar-search-border-focus:        @blue-300;
+    --theme-topbar-search-shadow-focus:        0 0 0 @su4 fade(@blue-500, 15%);
```
aaronshekey: This likely still needs to be aware of `var(--theme-primary)`. Borrow the approach we used on buttons? Will `0 0 0 @su4 var(--focus-ring);` cut it?

lib/css/components/_stacks-topbar.less
```diff
@@ -0,0 +1,366 @@
...
+    }
+}
+
+.s-topbar .s-topbar--navigation,
+.s-topbar .s-topbar--content {
+    .s-topbar--item {
+        color: var(--theme-topbar-item-color);
+        display: inline-flex;
+        align-items: center;
+        padding: 0 10px; // TODO @su8 or @su12?
```
aaronshekey: 12, or... if you really wanna be fancy, you could `@su12 - @su2`. We've done worse.

lib/css/components/_stacks-topbar.less
```diff
@@ -0,0 +1,366 @@
...
+            width: 18px;
+            height: 2px;
+            background-color: var(--theme-topbar-item-color);
+            position: relative;
+
+            &:before,
+            &:after {
+                position: absolute;
+                content: '';
+                width: 18px;
```
aaronshekey: We should put this down to `@su16` for more consistent rendering with `@Svg.Hamburger` but we'll need to double check all the rotation math.

lib/css/components/_stacks-topbar.less
```diff
@@ -0,0 +1,366 @@
...
+
+    a.s-topbar--logo.is-selected {
+        background-color: var(--theme-topbar-item-background-hover);
+    }
+
+    .s-topbar--menu {
+        display: flex;// none;
+        padding: 0;
+        height: 100%;
+        width: 44px;
```
aaronshekey: Should we make this a perfect square and just go `@su48`?

docs/product/components/partials/_topbar.html
```diff
@@ -0,0 +1,151 @@
...
+                <li>
+                    <a href="#" class="s-topbar--item" title="Achievements">
+                        {% icon "Achievements" %}
+                        <span class="s-activity-indicator s-activity-indicator__success">+0</span>
+                    </a>
+                </li>
+                <li>
+                    <a href="#" class="s-topbar--item" title="Review queues">
+                        {% icon "ReviewQueue" %}
+                        <div class="s-activity-indicator">
```
aaronshekey: I think we want `s-activity-indicator__danger` here.
```suggestion
                        <div class="s-activity-indicator s-activity-indicator__danger">
```

docs/product/components/partials/_topbar.html
```diff
@@ -0,0 +1,151 @@
...
+                <li>
+                    <a href="#" class="s-topbar--item s-user-card s-user-card__small">
+                        <span class="s-avatar s-avatar__24 s-user-card--avatar">
+                            <img class="s-avatar--image" src="{{ "/assets/img/anonymous-user.svg" | relative_url }}" />
+                            <span class="v-visible-sr">John Doe</span>
+                        </span>
+                        <div class="s-user-card--info">
+                            <ul class="s-user-card--awards">
+                                <li class="s-user-card--rep">3,145</li>
+                                <li class="s-award-bling s-award-bling__gold">3</li>
```
aaronshekey: Speaking of interdependencies, we won't be able to go live with this until we add custom CSS for bling. Otherwise this'll be a regression across the network. Not news to you of course, but noting it here for when we both forget lol.

docs/product/components/partials/_topbar.html
```diff
@@ -0,0 +1,151 @@
...
+                                <li class="s-award-bling s-award-bling__silver">9</li>
+                                <li class="s-award-bling s-award-bling__bronze">20</li>
+                            </ul>
+                        </div>
+                    </a>
+                </li>
+                <li>
+                    <a href="#" class="s-topbar--item" title="Inbox">
+                        {% icon "Inbox" %}
+                        <span class="s-activity-indicator s-activity-indicator__danger">0</span>
```
aaronshekey:
```suggestion
                        <span class="s-activity-indicator">3</span>
```

docs/product/components/topbar.html
```diff
@@ -0,0 +1,42 @@
+---
+layout: page
+title: Topbar
+description: A topbar component
+---
+
+<section class="stacks-section">
+    {% header "h2", "Logged in" %}
+
+    {% assign variations = "s-topbar__auto,s-topbar__dark,s-topbar__light" | split: "," %}
```
aaronshekey: Is this here 'cause you don't wanna pass an empty class or something? `s-topbar__auto` is the same as just a plain ol' `s-topbar` yeah? Think you can just kill it.

lib/css/components/_stacks-topbar.less
```diff
@@ -0,0 +1,366 @@
...
+                background-color: var(--theme-primary-600);
+                color: var(--white);
+            }
+        }
+    }
+
+    // TODO Stacks component? Badge?
+    .s-topbar--item.s-topbar--notice {
+        border-radius: @br-sm;
+        font-weight: bold;
```
aaronshekey: We have more control these days. `700` or `800` ought to do it.

lib/css/components/_stacks-topbar.less
```diff
@@ -0,0 +1,366 @@
...
+        padding-top: @su8;
+        padding-bottom: @su8;
+    }
+}
+
+// Searchbar
+.s-topbar .s-topbar--searchbar {
+    @inputLineHeights: @lh-sm; // ensure the line heights between the two elements match up
+
+    -webkit-transform: translate3d(0, 0, 0); // Solves an issue with Safari clipping things strangely when a popover is present in the topbar
```
aaronshekey: Bust out that mac and see if this is still an issue?

docs/product/components/partials/_topbar.html
b-kelly: On Core I have this working / looking ok with the current markup that we're using for the bling. I want the Stacks docs to be 100% Stacks, but the topbar component is flexible enough in this aspect that we have quite a bit of wiggle room as to what we put inside the `s-topbar--item`s.

lib/css/components/_stacks-topbar.less
b-kelly: Set it to `width: @s-full; max-width: 100%;`. This looks pretty good in the general case as well as on our squished examples in the docs. I'd prefer not to use `@site-width` since I'm actively trying to remove that and doing so would require me to move this component out of `static` and into `dynamic`, which kills my ability to use it in non-unified areas (SE.com)

lib/css/components/_stacks-topbar.less
b-kelly: Good call! I just installed the stylelint extension locally. An option to help enforce this would be to add it to our recommended extensions for VS Code (right click the extension in the pane, then `Add to Workspace Recommendations`).

lib/css/components/_stacks-topbar.less
b-kelly: No math changes needed, thankfully!

docs/product/components/partials/_topbar.html
b-kelly: Kinda. On production that indicator can be either blue or red, depending on if the queue "is highlighted", which is based on a couple of factors (queue length, any item is "in danger"). Since I already had one red item in the examples, I decided to keep this one blue.

docs/product/components/topbar.html
b-kelly: More or less. I show the class above the generated example right now, but that might change once the docs are actually written. In that case, we can make this an empty class or whatever really.

lib/css/components/_stacks-topbar.less
b-kelly: Do you have any specific tips on when to use `bold` vs the numeric values? I don't personally have any specific sense for when one looks better than another or on what fonts these additional values are supported. Fwiw, it seems like Less is compiling this into `700`. I didn't realize it did that!

lib/css/components/_stacks-topbar.less
b-kelly: Nice catch! I'll do one better and not hardcode the width at all. Changing the items inside to 16px means that I can do `padding: 0 16px` and get the same width out, but with some flexibility if the interior size changes up any. This was a holdover from the legacy css I ~~copy/pasted~~ ported over.

lib/css/components/_stacks-topbar.less
b-kelly: Imo, 12 was too much and 8 was too little. I added `(@su12 - @su2)` as suggested so we're at least not _entirely_ hardcoding the value...

lib/css/components/_stacks-topbar.less
b-kelly: Good call on this as well. I was using the value on Prod, but I like yours better.

docs/product/components/partials/_topbar.html
b-kelly: I went back and looked at the code again and for the life of me I can't find any trace of the blue (vs red) indicator. Maybe I dreamt it? I'll look over it again, but I might be describing some legacy topbar styles that are no longer in use... I'll update the indicator to be danger to match.

lib/css/components/_stacks-badges.less
```diff
@@ -20,10 +20,6 @@
     border-color: @border;
     background-color: @background;
     color: @color;
-
-    .s-tag--dismiss {
```
b-kelly, line 24: This was not used anywhere in the product, nor in Core. Furthermore, it was bloating the compiled css file, so I deleted it.

aaronshekey (changes requested):
So very close now!

- [x] Grab the high contrast overrides we have in production and add them here
- [x] Double check the forced dark mode scrollbar colors. Transparency approach doesn't work there
- [x] Should we also apply scrolling to the CTA's section? We could probably kill it.
- [x] Navigation button stays in `x` state when navigating to a new page in docs.

docs/_data/topbar.json
```diff
@@ -0,0 +1,66 @@
...
+        "applies": "Child of `.s-topbar`"
+    },
+    {
+        "class": ".s-topbar--content",
+        "description": "Contains `li > .s-topbar--item` elements that don't look like anything in particular",
+        "applies": "Child of `.s-topbar`"
+    },
+    {
+        "class": ".s-topbar--ctas",
+        "description": "Basically --content above... why do I have this again? Hover classes maybe?",
```
aaronshekey: Right, heh. This was my thought as well.

b-kelly: I'll update this section. The reasoning is that we want to maintain sizing, spacing, etc on these items, but we don't want to style them any further. That way, we can add in buttons or whatever without worrying about stepping on _their_ hover classes. This could probably be renamed to something like `.s-topbar--content.s-topbar--content__unstyled`. Thoughts?

lib/css/components/_stacks-badges.less
aaronshekey: Are we absolutely sure? We use something that looks just like it in the context of the tag editor. I bet if we switch to `s-tag` there, we'll miss this.

lib/css/components/_stacks-topbar.less
b-kelly: Tested it and it didn't seem to be an issue any longer

aaronshekey:
Small regression with the navigation text color:
image

We just gotta do more states here, if I'm not mistaken. Active, and is-selected:hover states.

aaronshekey:
image

This `new` button needs an active state. Clicking it in Safari makes it lose its background.

b-kelly:
> This `new` button needs an active state

Can you expand on that? I have JS added to the docs to toggle the state of the notices on click. I did this so users can easily preview what it looks like with and without the `is-unread` styling.

aaronshekey (approved):
🦐

PR #755, opened 2021-09-15
Kill some Less color vars

dancormier:
This PR moves us a tiny step closer to independence from Less.

- I got rid of any `@focus-ring` Less variables.
    - We only reference them in `_stacks-constants-colors.less`
- I updated the dark theme `--focus-ring-muted` value to be visible (thanks @allejo [for pointing out this issue](https://stackexchange.slack.com/archives/C27RWNQN9/p1631740546149600))
- I replaced `--highlight-*` references to Less color variables with CSS variables.

### DIFF
--- a/lib/css/exports/_stacks-constants-colors.less
+++ b/lib/css/exports/_stacks-constants-colors.less
@@ -368,9 +368,9 @@
 
     // Focus rings
     --focus-ring-success: hsla(@green-h, 40%, 75%, 40%);
-    --focus-ring-warning: hsla(@yellow-h, 79%, 58%);
-    --focus-ring-error: hsla(@red-h, 62%, 47%);
-    --focus-ring-muted: hsla(@black-h, @black-s, 15%);
+    --focus-ring-warning: hsla(@yellow-h, 79%, 58%, 40%);
+    --focus-ring-error: hsla(@red-h, 62%, 47%, 15%);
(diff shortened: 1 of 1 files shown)


lib/css/exports/_stacks-constants-colors.less
```diff
@@ -368,9 +368,9 @@
 
     // Focus rings
     --focus-ring-success: hsla(@green-h, 40%, 75%, 40%);
-    --focus-ring-warning: hsla(@yellow-h, 79%, 58%);
-    --focus-ring-error: hsla(@red-h, 62%, 47%);
-    --focus-ring-muted: hsla(@black-h, @black-s, 15%);
+    --focus-ring-warning: hsla(@yellow-h, 79%, 58%, 40%);
+    --focus-ring-error: hsla(@red-h, 62%, 47%, 15%);
+    --focus-ring-muted: hsla(@black-h, @black-s, 15%, 10%);
```
dancormier: Oh wow, good catch @b-kelly!

aaronshekey (approved):
Ran visual regression testing against this one. Also checked Core and Careers for use of `@focus-ring` and came up with no results. Safe to merge!

b-kelly (approved):
lgtm! There was a bug in light mode where focus rings didn't have transparency that I fixed, but other than that, everything checks out.

PR #758, opened 2021-09-16
Fix print conditional classes

### LINKED ISSUE #757: `.print:d-block` isn't working as expected
We have a conditional class for print that isn't working currently. Also, there's a bug in the documentation itself.

aaronshekey:
Responsive styles were clobbering our print conditionals. Since there are only two of them, this puts those classes _after_ all the responsive classes.

Also fixes a broken example in the conditionals page.

To test: Head to the [conditional classes](https://deploy-preview-758--stacks.netlify.app/product/guidelines/conditional-classes/#print) demo, press print, and check that the page layout is different and the conditional example changes when printing.

Closes #757

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -10,6 +10,7 @@
     "selector-pseudo-element-colon-notation": "single",
     "no-descending-specificity": null,
     "function-whitespace-after": null,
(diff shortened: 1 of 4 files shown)


dancormier (approved):
Well, this looks fine.

🖨️ it!

PR #760, opened 2021-09-21
Add focus/blur events to tooltips

dancormier:
See also: https://github.com/StackEng/StackOverflow/pull/9371


aaronshekey:
Document these events as well?

b-kelly (approved):
lgtm! If we wanted to be fancy with it, we'd remove the hover delay for the focus event, but I'm 100% fine with this as-is.

dancormier:
@aaronshekey I don't _think_ these need documentation. We don't really document the click events.

aaronshekey:
> We don't really document the click events

Yep, confirmed. 🦐

PR #763, opened 2021-09-27
Get our package.json back into sync

aaronshekey:
Dependabot only updates the package-lock.json file by default. This has been reconfigured and now the `package.json` reflects the current versions of everything.


b-kelly (approved):
lgtm! I pulled locally to test. I updated stacks-icons by another minor point (23 to 24) and ran `npm audit fix` while I was at it, since we're already updating the package-lock file. Everything runs fine on my machine.

PR #767, opened 2021-10-01
Topbar light/dark experimental CSS variables

dancormier:
I'm made this branch to see if we could get rid of the light/dark CSS variable generation in the Topbar scope and just define the variables alongside other color variables.

_I think_ what I did here would still be themable by consumers to the extent it is in #754 currently, but it relies on predefined values from the `.theme-dark__forced` and `.theme-light__forced` classes.

@b-kelly let me know if this is a good jumping off point and, if so, I'll dig a little further into getting the topbar color values just right (specifically the dark mode since I think I took some liberties here).


aaronshekey:
'sup with this one? Are we closing this?

dancormier:
Closing time!

PR #769, opened 2021-10-06
Add anchor underline to high contrast

dancormier:
Upstream changes to support https://github.com/StackEng/StackOverflow/pull/9535

Only merge once https://github.com/StackEng/StackOverflow/pull/9535 and https://github.com/StackEng/StackOverflow/pull/9526 are merged.


lib/css/base/_stacks-reset-normalize.less
```diff
@@ -99,11 +99,14 @@ pre {
...
  * 1. Remove the gray background on active links in IE 10.
  * 2. Remove gaps in links underline in iOS 8+ and Safari 8+.
+ * 3. Adds underline to links in high contrast
  */
 
 a {
   background-color: transparent; /* 1 */
   -webkit-text-decoration-skip: objects; /* 2 */
+
+  .highcontrast-mode({ text-decoration: underline; });
```
aaronshekey: Let's do this in `stacks-links.less` instead: https://github.com/StackExchange/Stacks/blob/develop/lib/css/components/_stacks-links.less#L17

There's a future where we kill these resets entirely #609

dancormier: Good call. I knew it was wrong when I was doing it and I did it anyway. 🤷‍♂️

lib/css/base/_stacks-reset-normalize.less
```diff
@@ -99,6 +99,7 @@ pre {
 /**
  * 1. Remove the gray background on active links in IE 10.
  * 2. Remove gaps in links underline in iOS 8+ and Safari 8+.
+ * 3. Adds underline to links in high contrast
```
aaronshekey:
```suggestion
```

aaronshekey (changes requested):
Just 2 teeny tiny bits. Merge whenever you'd like into production SO. That work can definitely ship before the next release of Stacks.

We've got a blue that's a little leaky in the case of `s-card`. 
image

aaronshekey (approved):

PR #770, opened 2021-10-08
Add table layout atomic classes

aaronshekey:
This PR adds 2 table layout atomic classes while documenting the `display: table` on the tables page as well. They already appeared on the Display page, but the redundancy can't hurt.


b-kelly (commented):
I'm generally not a fan of redundant documentation. Rather than re-list the classes in your new paragraph, could you perhaps link to the display page instead?

PR #773, opened 2021-10-14
Data visualization guidelines

abovedave:


CGuindon:
Thinking we can remove the first line (or merge it into the other) and just use the first paragraph below 'Data Viz at Stack Overflow' as the subtitle line below the page title.
Screen Shot 2021-10-18 at 8 56 58 AM

CGuindon:
first word "The"
Screen Shot 2021-10-18 at 3 27 55 PM

CGuindon:
Maybe we switch the title for these to 'Stacked bar graph (100%)' — instead of having the percentage at the start? Would read better I think

Screen Shot 2021-10-18 at 3 30 28 PM

aaronshekey:
Ok but we need that `package-lock.json` so you'll have to sort that one out :v:

abovedave:
what's the deal with the `tscommand-XXX.tmp.txt` files - do we want to add to .gitignore or are they required?

aaronshekey:
> what's the deal with the `tscommand-XXX.tmp.txt` files - do we want to add to .gitignore or are they required?

If the build completes those should get thrown away. Should be fine to add them to the .gitignore or just removing it entirely.

CGuindon:
@CGuindon To do next time — Should swap the red and blue color order (we want them to use blue more often than red — positive stuff first)

CGuindon:
@abovedave Any chance you can tweak the `stacked bar 100%` chart to just be a `stacked bar chart `icon? Like you did for `stacked area` ones (one goes to the top the other doesn't)?
Screen Shot 2021-11-01 at 9 10 08 PM

abovedave:
@CGuindon good spot, sorted!

CGuindon (approved):

CGuindon:
Ship it 🚢

aaronshekey (approved):

PR #798, opened 2021-11-29
Add ARIA tags for input validation message examples

allejo:
Use [`aria-describedby`](https://www.w3.org/TR/wai-aria/#aria-describedby) in the validation message examples for input fields; we should associated these messages to their respective input fields for screen readers.

See https://www.w3.org/TR/WCAG20-TECHS/ARIA1.html#ARIA1-ex4

Also, add `aria-invalid` for the error state input example.


aaronshekey (approved):
Thanks for the callouts here!

PR #807, opened 2021-12-16
Badge and privilege progress bar refinements

aaronshekey:
There are a lot of atomic classes in this component. This PR cleans up our approach and is a bit more opinionated. This would fix the issue with bling as [reported on Meta](https://codegolf.meta.stackexchange.com/questions/24166/badge-progress-circle-misalinged).

## Before
image

## After
image

### DIFF
--- a/docs/product/components/progress-bars.html
+++ b/docs/product/components/progress-bars.html
@@ -290,25 +290,25 @@ description: A component that visually communicates the completion of a task or 
 {% highlight html %}
 <!-- Gold Badge -->
 <div class="s-progress s-progress__badge s-progress__gold">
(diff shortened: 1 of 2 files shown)


dancormier (approved):
This is a nice improvement. LGTM!

_Out of scope, but here a small note since I saw it_ : These progress components tend to get rough when they wrap. Seems to be partially due to setting `height: 100%` on `.progress--label`. I'll branch off of this one and see if I can make a reasonable tweak. **Update**: Upon further investigations, seems it's a result of the label being positioned absolutely. I think addressing this relatively minor nit would require a bit of a refactor, so I'm punting! Issue: https://github.com/StackExchange/Stacks/issues/809

PR #808, opened 2021-12-16
Guard against parent white space in popovers

aaronshekey:
Without this, the content pops *out* of the popover if the parent has `ws-nowrap`. For tooltips, this can be a tricky interaction.


b-kelly (approved):
lgtm. I don't see any circumstances (currently or in the future) where this will bite us. Actually wanting `nowrap` doesn't seem terribly common in popovers and any that I can think of will have `ws-nowrap` on them deliberately.

PR #810, opened 2021-12-17
Add `.g1` atomic class for `gap: 1px`

dancormier:
When trying to do some hacky border stuff, I found myself reaching for `gap: 1px`. I think this will be useful outside of hacky work and it makes more sense than it's legacy `.gs[x]` counterpart since our `gap` atomic classes don't rely on negative margins.


aaronshekey (approved):
You know what's so silly? We've been doing `@su / 2` for the longest time. Glad to see `@su1` here lol.

We'll want to mark `gs[x]` classes as deprecated sometime in the middle of 2022 :v: Can't wait!

PR #816, opened 2021-12-23
Forced theme component colors

### LINKED ISSUE #814: Forced light / dark conditionals aren’t applying to buttons.
Our `.theme-light__forced` and `.theme-dark__forced` modifiers aren’t applying to buttons. In a use case shared by a designer, we’ve got a banner with several calls to action. We’d like to force light colors on it, but currently these modifiers only affect the `.bg-*` and `.fc-*` classes. The buttons are left untouched.

b-kelly:
Fixes component-colors not properly resetting to their desired colors when the theme is forced.

Fixes #814


aaronshekey:
Forced dark seems to behave correctly, but buttons in forced_light still change when toggling the global dark mode.

dancormier:
I added a `.light-mode-forced` mixin to basically drop the hammer on the cascade. It creates a little output specific to `.theme-light__forced` but I think it's our best option that avoids a big refactor.

aaronshekey:
We've thrashed this branch with all the latest `1.0` stuff. Worth considering including this?

dancormier:
> We've thrashed this branch with all the latest `1.0` stuff. Worth considering including this?

I vote 🦐 .

b-kelly:
This looks ok to me, but I'm still a little hesitant. I feel like with the light forced mixin, we're going to need to add that everywhere we have the dark mixin... This approach needs a bit more testing.

dancormier (approved):

PR #819, opened 2022-01-05
Further refinement of post summary states

aaronshekey:
image


CGuindon (changes requested):
1. In review should be "Review"
2. The vertical alignment on the Closed and Archived icons look good in the screenshot but is off when I preview the link.
Screen Shot 2022-01-05 at 12 57 28 PM

CGuindon (approved):

PR #820, opened 2022-01-06
Switch to native gaps on user cards

aaronshekey:
This reduces some CSS and simplifies our presentation of user cards by switching to native gaps. This will require some scrutiny in production, but should allow us to drop a bunch of margin overrides, and rearrange things more easily.

This also cleans up some documentation.


dancormier (approved):
New team name:  The Gap Band

PR #826, opened 2022-01-10
Justify user-cards to the right in post-summary

b-kelly:
When the `user-card` is wrapped to the next line, but that _itself_ is also wrapped, the card appears to be aligned left. Attached is a realistic use case - mobile view, user has badges.

image


dancormier (approved):
This approach results in an _ever so slight_ misalignment among items in some cases:



That being said, I'm confident https://github.com/StackExchange/Stacks/pull/820 will resolve this.

aaronshekey (approved):

PR #842, opened 2022-01-26
Add focus states to multiple options toggle switch

### LINKED ISSUE #840: Toggle switches need focus styling
As part of our accessibility audit, it's come up that toggle switches may need their focus state revisited.

aaronshekey:
Closes #840

With your keyboard, tab to the multiple toggles, then move between options with your arrow keys like a radio group :v:

image


dancormier (approved):
The same great toggle switch taste, now with fewer accessibility flaws.

```
✨✨✨
✨🦐✨   
✨✨✨
```

PR #844, opened 2022-01-26
Post summary refinements

aaronshekey:
This PR explores some changes based on feedback. More description to come.

1. Increases font size overall on stats
2. Moves watched and ignored to the tags themselves
4. Highlights various states in a way that looks closer to the status quo
5. Aligns post titles better with the view stats
6. Adds an `s-post-summary-stats--item__emphasized` for when we need to add emphasis to voting or views, etc.
7. Removes the fire icon
8. Makes the post title a block level element so it can render mathjax


dancormier (approved):
Does what it says on the tin and does it well.

dancormier (approved):
👍👍👍

PR #845, opened 2022-01-27
Add support for all the different JS packaging methods that exist

allejo:
## DO NOT MERGE

Everything builds, however it's largely untested atm. These are the big things I still need to test,

- [ ] I didn't break the current distribution method
  - [ ] Make sure `windows.Stacks` still exists/works
- [ ] The CJS/MJS organization stuff is properly exported
- [ ] Check to see what I missed exporting

---

## What this PR **DOES NOT** do

~~This PR does not migrate Stacks to Webpack or any fancy packaging system, we can revisit that in PRs like #707 or whatever supersedes that.~~

**Edit:** This PR now uses Webpack to bundle up the browser distributable.

## What this PR hopes to solve

All of the SO repos that uses Stacks currently make a copy of all of its source files and add it to their version control. Since, we have been working on [migrating the JS for these repos to Webpack](https://meta.stackexchange.com/questions/372590/were-migrating-our-javascript), all of our dependencies have started to get locked in the respective `package.json` files except for Stacks; this PR allows us to include our Stacks in our dependencies instead of having to clone a copy of its source.

## What has changed in this PR

- Introduces two new `tsconfig` files, one that will compile everything for CJS and one that will compile for MJS.
  - Node has introduced [MJS as its official module system](https://nodejs.org/docs/latest-v16.x/api/packages.html#introduction) and migration in the ecosystem has started but it's not without compatibility issues.
  - CJS compilation has also been added and will be supported while the ecosystem irons out MJS import compatibility
  - A single JS file with everything included remains untouched to continue supporting unpkg.com; this is still done via Grunt
  - The CJS and MJS compilation is done with `tsc` and different configuration options
  - These improvements allow us to use Stacks the modern way:
    ```typescript
    import { showModal } from '@stackoverflow/stacks';

    showModal();
    ```

    Instead of the old way,

    ```js
    window.Stacks.showModal()
    ```
- The TypeScript definitions for all of the components and functions in Stacks are now automatically exported and listed in `package.json`; this removes the need of having manually maintained definition files in our SO repos because definitions are missing
- TypeScript namespaces have been removed as they're no longer necessary to compartmentalize the code

## What this changes for Stacks development

For the most part, nothing! Continue writing Less and Stimulus controllers as you have. Things should Just Work™️ at build/release time.

However, here are a few things that did change:

- TypeScript namespaces are no longer needed
- If you would like a piece of code to be exported and be available as part of the Stacks API, make sure to export it at `lib/ts/index.ts`


aaronshekey:
@allejo Thank you so much for the pull request. Excited for the team to dig in—also excited to see this deferring the webpack stuff for now.

tsconfig-base.json
```diff
@@ -0,0 +1,11 @@
+{
+  "compilerOptions": {
+    "strict": true,
+    "target": "es5",
+    "lib": ["dom", "es6", "dom.iterable", "scripthost"], // es6 stuff is polyfilled by stimulus
+    "sourceMap": true,
+    "moduleResolution": "node",
+    "skipLibCheck": true
```
benjamin-hodgson, line 8: There are probably a few more flags you wanna include here, eg `isolatedModules`

benjamin-hodgson (commented):
So it looks like `window.Stacks` is out the window? There'll need to be a file by file migration on the consuming end. @aaronshekey @b-kelly Are you OK with breaking compat for this?

If the plan is to include Stacks in the webpack bundle then we'll need to think about Jobs.

allejo:
> So it looks like `window.Stacks` is out the window? There'll need to be a file by file migration on the consuming end. @aaronshekey @b-kelly Are you OK with breaking compat for this?
> 
> If the plan is to include Stacks in the webpack bundle then we'll need to think about Jobs.

The PR as-is currently has killed `window.Stacks` inadvertently but I was planning on adding it back for the old, single-file distribution method that's used for CDNs to keep backward compat. I've added this to my to do list. But if we're okay with dropping that, then I won't add it back in.

benjamin-hodgson:
> the old, single-file distribution method

I think you'll need Webpack for that

lib/ts/browser.ts
```diff
@@ -0,0 +1,16 @@
...
+    interface Window {
+        Stacks: any;
+    }
+}
+
+window.Stacks = {
+    application: application,
+    _initializing: application.isInitialized(),
+    ...Stacks,
+};
```
benjamin-hodgson, line 14: You can [configure webpack to output a UMD module](https://webpack.js.org/configuration/output/#outputlibrary) instead of directly talking to `window` here

dancormier:
Hmm, that merge changed a lotta files. Looking into it (and possibly reverting the merge).

Update: yeah, idk what's going on with this merge. I'll revert now and come back to it later.

b-kelly:
Instead of building on this PR, I've decided to bring #707 up to date and finish that instead. That PR is now ready for prime time, so I'm closing this one.

PR #846, opened 2022-01-28
Add inset atomic classes

dancormier:
In production code, I've seen many instances of positioning like `l0 r0 t0 b0`, `t6 b6 l6`, and `l50 t50` which could be replaced with an [`inset`](https://developer.mozilla.org/en-US/docs/Web/CSS/inset) shorthand. This PR adds an atomic class with the prefix `i` for the `inset` property

I think it'll be useful and [the browser support has been there for a while now](https://developer.mozilla.org/en-US/docs/Web/CSS/inset#browser_compatibility). In fact, we're already using it in Stacks exactly once (in the Uploader component).

**Sidenote**: The `inset` property is not available in the https://browsers.stackoverflow.design/ "can I use …" search. I haven't dug in to it, but seems odd 🤷‍♂️.

## TODO

- [x] Add example (hard to do and have it not stomp on the other examples)


docs/_data/spacing.json
```diff
@@ -115,6 +115,10 @@
     {
       "prefix": "b",
       "values": "0, 1, 2, 4, 6, 8, 12, 16, 24, 32, 48, 64, 96, 128, 50, 100"
+    },
+    {
+      "prefix": "i",
+      "values": "0, 1, 2, 4, 6, 8, 12, 16, 24, 32, 48, 64, 96, 128, 50, 100"
```
aaronshekey: Does inset make sense with 50% and 100%?

docs/_data/spacing.json
```diff
@@ -370,6 +382,10 @@
       "term": "ln",
       "define": "negative left"
     },
+    {
+      "term": "ln",
```
aaronshekey: This should be `in` right? I wonder if we even need negative inset. That seems rare?

docs/product/base/position.html
```diff
@@ -84,7 +84,9 @@
 <div class="ps-absolute t48 r24"> … </div>
 <div class="ps-absolute t64 l64"> … </div>
 <div class="ps-absolute bn8 rn8"> … </div>
+<div class="ps-absolute i48"> … </div>
 {% endhighlight %}
+<!-- TODO: Add inset example -->
```
aaronshekey: If we get the z-index right, this should be just fine to slap back there.

docs/_data/spacing.json
dancormier: Nah, I don't think so. `inset: 50%` would be the same as `left: 50%; right: 50%; top: 50%; bottom: 50%;` which would only work well on an element with no dimensions. Same goes for `100%`.

Removing them!

docs/_data/spacing.json
dancormier: > This should be `in` right?

Yep, you caught a typo. I'll push a fix up*. 

> I wonder if we even need negative inset. That seems rare?

Yeah, it kinda goes against the concept of "inset". I included it just for consistencies sake, but it's probably solidly in YAGNI territory. I'll get rid of it.

*On second thought, see above.

docs/product/base/position.html
dancormier: Sticking it in the back of class is probably the way to go. I'll add this example in so we can get this sucker merged.

aaronshekey (approved):

PR #851, opened 2022-02-09
Add `.bs-xl` XL box-shadow atomic class

dancormier:
@aaronshekey a couple questions:

- The docs "output" for box-shadow look incorrect. The actual styles are combinations of multiple shadows (and they're different in dark mode). Feels a little wordy to stick the entire output of each box-shadow in the docs. Any better alternative in mind or should I just update to have the long rule in there?
- Are we cool with `.bs-xl` not being responsive?

### DIFF
--- a/docs/_data/atomics.json
+++ b/docs/_data/atomics.json
@@ -21,6 +21,11 @@
       "hover": true
     },
     {
+      "class": "bs-xl",
+      "output": "TODO: Add output here. BTW, pretty sure the above outputs are way off.",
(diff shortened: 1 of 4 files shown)


aaronshekey:
Lookin' good!

> * The docs "output" for box-shadow look incorrect.

Oh yeah, way out of date. I think we could probably put the real output in there, and just format it with line breaks as necessary?

> * Are we cool with `.bs-xl` not being responsive?

Sure!

PR #853, opened 2022-02-09
Additional post summary refinements

aaronshekey:
- [x] Upstreams breaking words in excerpts
- [x] Switches to `h3` for post titles
- [x] Guards against various environment differences in post titles
- [x] Removes reliance on `s-link`
- [x] Introduces new theming variables in the post summary
- [x] Adds deprecated "Legacy" layout
- [x] Adds pinned state

To manage in production:

- Add the prop for legacy layout and handle markup moving around based on that state
- Add prop for pinned
- Configure theming variables in production

### DIFF
--- a/docs/product/components/post-summary.html
+++ b/docs/product/components/post-summary.html
@@ -1335,6 +1335,17 @@ Is anyone aware of a fully functional implementation of this API (to generate th
     <p class="stacks-copy">Post summaries change their appearance based on being watched, ignored, or deleted. Articles can also have varying draft states.</p>
     <div class="stacks-preview">
 {% highlight html %}
+<div class="s-post-summary s-post-summary__pinned">
+    <div class="s-post-summary--stats">
+        <div class="s-post-summary--stats-item is-pinned">
+            @Svg.TackSm
+            Pinned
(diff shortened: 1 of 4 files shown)


dancormier (approved):
👍👍👍

b-kelly:
Updated `--theme-question-*` to `--theme-post-*` since post summaries are not just designed for questions any longer. Currently, the shim in Core will keep working once this is deployed there, but we'll want to follow up by removing it and replacing all instances of `--theme-question-*` network theming variables with the `post` versions.

Additionally, I moved these variables to shadow the other existing theme variables for links/body font. This way, you can change e.g. links and it'll work everywhere, including post-summary (this is the behavior I'd expect as a consumer).

aaronshekey:
The only thing that bothers me is we've got font stuff in a file called `constants-colors` 🤔

b-kelly:
@aaronshekey That's fair, but if we don't put them there then they won't inherit properly for child themes.

b-kelly:
Honestly, that file's been needing a rename for a long time now. Some of that file is constants, but most of it is theming variables now.

PR #863, opened 2022-02-21
Adjust Segoe UI's baseline

aaronshekey:
As explored by @b-kelly in https://codepen.io/bk-SO/pen/ExbErYV


aaronshekey:
This works incredibly well for body copy, but when the typeface is italic or bold, the letters change shape and it appears that our headings get tighter letter-spacing. It's as though the browser chooses faux bolding / italics instead of the font's actual letterforms.

aaronshekey:
Barring any showstoppers we haven't found, this looks like a viable solution to Segoe's baseline.


| Before | After |
| -- | -- |
| Before | After |

Pay closest attention to things like inputs and navigation. In the after, these things keep their baseline between macOS and Windows, meaning things like icons are far more likely to look vertically aligned. After adjustments, it seems like our headlines jump a bit, but there are far fewer times we're trying to align very large text with something like an icon, vertically.

We could attempt to adjust the baseline at various font sizes, but I'm not sure it's worth it.

I would personally be comfortable with shipping this.

aaronshekey:
I think I've found the final adjustment at 95%.

final

lib/css/exports/_stacks-constants-type.less
```diff
@@ -13,9 +13,71 @@
...
+
+// Add Segoe UI Adjusted as a font family, and adjust its baseline
+@font-face {
+    font-family: "Segoe UI Adjusted";
+    src: local(Segoe UI);
+    ascent-override: 95%;
+}
+
+// TODO Segoe UI is variable, but only in Windows 11, so add all the variants
+// TODO add Light/Semilight variants: https://docs.microsoft.com/en-us/typography/font-list/segoe-ui
```
b-kelly: Are you interested in adding these variants as well? If not, we can ditch both of these TODO comments

lib/css/exports/_stacks-constants-type.less
```diff
@@ -13,9 +13,71 @@
...
 //  ----------------------------------------------------------------------------
+
+// Add Segoe UI Adjusted as a font family, and adjust its baseline
+@font-face {
+    font-family: "Segoe UI Adjusted";
+    src: local(Segoe UI);
+    ascent-override: 95%;
+}
+
+// TODO Segoe UI is variable, but only in Windows 11, so add all the variants
```
b-kelly:
```suggestion
// Segoe UI is variable, but only in Windows 11, so add all the variants
```

b-kelly:
We'll have to test this on a Windows 11 machine as well to make sure this approach still works there.

b-kelly:
Looks good on Windows 11 with its changes to the Segoe font. Very long image warning incoming:

<details>
<summary>s-prose screenshots</summary>

| Base | Adjusted |
| ----- | ---------- |
| native | adjusted |

</details>

lib/css/exports/_stacks-constants-type.less
```diff
@@ -13,9 +13,70 @@
...
+    src: local(Segoe UI Black Italic);
+    ascent-override: 95%;
+    font-style: italic;
+    font-weight: 800;
+}
+
 @ff-sans:
     -apple-system, BlinkMacSystemFont, // San Francisco on macOS and iOS
-    "Segoe UI", // Windows
+    "Segoe UI Adjusted", // Windows
```
b-kelly: I wonder if we want to be extremely cautious and add Segoe as a fallback. Not required, food for thought.

```suggestion
    "Segoe UI Adjusted", "Segoe UI", // Windows
```

b-kelly (approved):
🦐

PR #866, opened 2022-03-03
Update .s-post-summary__pinned styling

ycantu:
1. Updates the `.s-post-summary--stats-item.is-pinned` background color from `--black-800` to `--black-700`.
2. Removes all `.s-post-summary__pinned` styling (We will not have a background color for pinned posts at this time)

### DIFF
--- a/lib/css/components/_stacks-post-summary.less
+++ b/lib/css/components/_stacks-post-summary.less
@@ -393,15 +393,6 @@
     }
 }
 
-.s-post-summary__pinned {
-    background-color: var(--black-025);
-    border: 1px solid var(--black-075);
-
-    & + & {
-        border-top-width: 0;
-    }
-}
-
 .s-post-summary__ignored,
 .s-post-summary__deleted {
     .s-post-summary--content {
--- a/lib/css/components/_stacks-post-summary.less
+++ b/lib/css/components/_stacks-post-summary.less
@@ -395,6 +395,11 @@
 
 .s-post-summary__pinned {
     background-color: var(--black-025);
+    border: 1px solid var(--black-075);
+
+    & + & {
+        border-top-width: 0px;
+    }
 }
 
 .s-post-summary__ignored,
--- a/lib/css/components/_stacks-post-summary.less
+++ b/lib/css/components/_stacks-post-summary.less
@@ -398,7 +398,7 @@
     border: 1px solid var(--black-075);
 
     & + & {
-        border-top-width: 0px;
+        border-top-width: 0;
     }
 }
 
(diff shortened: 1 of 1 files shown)


lib/css/components/_stacks-post-summary.less
```diff
@@ -234,7 +234,8 @@
 
         &.is-pinned {
             color: var(--white);
-            background-color: var(--black-800);
+            background-color: var(--black-700);
+            border-color: var(--black-075);
```
aaronshekey:
```suggestion
```

It was @CGuindon's intention that the entire post summary get outlined in `var(--black-075)`

aaronshekey (changes requested):
Yesssss! Glad to see this PR. Thank you! Just wanted to double check @CGuindon's intent here.

lib/css/components/_stacks-post-summary.less
ycantu: I misunderstood...we do not want the border color on here.

ycantu:
@aaronshekey I have updated this PR to include a border around the pinned post summary, not sure if I should be using `--black-075` or `--bc-light` for the color. 

We also have a collision with the border around pinned posts and the `.s-post-summary:last-child` rule. Should we handle that bottom border on the code side as it comes up? Or add a rule here in Stacks?

Something like,
```
.s-post-summary:not(.s-post-summary__pinned):last-child {
    border-bottom-width: 0;
}
```

aaronshekey:
Our post summaries, by design, only intend to include a border _between_ them. You'll notice that post summaries don't have a border on the left or right side, since, generally, wrapping or grouping is provided by the parent element. Compare our documentation examples with various list views throughout Stack Overflow. Right now, if we add a border to the left and right side, we'll get double borders. See below:

image

image

2 potential solutions to work out with @CGuindon:

1. Rely exclusively on the badge and the background color. Personally, I think it's enough to differentiate.
2. Pass additional atomic classes to your particular use case. e.g. `extraClasses = "ba bc-black-075"` cc: @b-kelly

CGuindon:
@ycantu Ok! You're going to hate me today but... let's push the pinned post without the border (around the entire post) or the background colour. We're just going to rely on the pinned post having the Pinned label and it being stuck at the top. Adding colour/borders makes it more prominent but it complicates things (other lists we might used pinning in and other components that have a the same background+border). So I'm thinking we leave it plain and I'm going to test it in the next usability session with users (see if the label is enough).

I do want the top and bottom borders though on all the lists but that has nothing to do with pinning. I'll look at what Paul suggested and see if I can open a PR or make a new ticket.

Screen Shot 2022-03-04 at 8 28 53 AM

CGuindon:
@aaronshekey Can I request to remove the light grey background from the pinned post? Leave it clear.

aaronshekey:
Works for me. We had a no background color approach on all our post states but it was too big a change on "watched" and "ignored" for most our meta users.

aaronshekey (approved):

PR #813, opened 2021-12-20
Add high density theme

dancormier:
This PR adds a _high density theme_ to Stacks. It does this by scaling spacing units (`--su-[x]` variables) and line-heights (`--lh-[x] variables`) using their base variable values (`--su-base` and `--lh-base` respectively).
To that end, this PR also:
- Converts `@su[x]` Less variables to CSS variables based off of `--su-base`
- Adds `--sus[x]` variables for static spacing units _not_ impacted by high density mode
- Removes `@lh-[x]` Less variables, replacing them with CSS variables based off of `--lh-base`

<details>
<summary>Screenshots</summary>

</details>

PR #873, opened 2022-03-10
Use CSS variables for spacing

dancormier:
This PR (only) includes the switch to CSS variables to hopefully get this stuff merged.

Variables introduced:
- variable spacing : `--su[x]` ( based off of `--su-base`)
- static spacing: `--sus[x]`
- line-heights: `--lh-[x]`

These changes should result in no visible differences.

### A note on "high density"

#813 had two major changes:

1) Conversion of `@su*` to `--su*`/`--sus*` (what you'll see in this PR)
2) A "high density" theme toggle and some accompanying styles

This PR only includes the first of those two changes. Introducing a high density theme requires debate on approach, product work, ongoing support, communication, etc. When it's time for high (or variable) density, it'll be another PR that benefits from the work here.


lib/css/atomic/_stacks-flex.less
```diff
@@ -196,6 +196,7 @@
 //  $   FLEX SPACING
 //      Insert spaces or gutters in between flex items
 //  ----------------------------------------------------------------------------
+// TODO: Make this play nice with spacing units… or yeet this in the age of gap
```
aaronshekey, line 199: Happy to punt here and just use the oldschool Less. Refactoring away from `gs` classes might take some doing, but needs to be done. I'd like to wait until we're firmly into iOS 16 before truly throwing these classes away. CSS ☹️

lib/css/atomic/_stacks-spacing.less
```diff
@@ -6,163 +6,307 @@
...
+    --sus8: 8px;
+    --sus12: 12px;
+    --sus16: 16px;
+    --sus24: 24px;
+    --sus32: 32px;
+    --sus48: 48px;
+    --sus64: 64px;
+    --sus96: 96px;
+    --sus128: 128px;
+    --su1: clamp(var(--sus1), calc(var(--sus1) * var(--su-base)), calc(var(--sus1) * var(--su-base)));
```
aaronshekey: Can we drop a comment on what this actually does?

lib/css/atomic/_stacks-typography.less
```diff
@@ -77,12 +77,12 @@ p {
...
-.lh-md { line-height: @lh-md !important; }
-.lh-lg { line-height: @lh-lg !important; }
-.lh-xl { line-height: @lh-xl !important; }
-.lh-xxl { line-height: @lh-xxl !important; }
+.lh-xs { line-height: var(--lh-xs) !important; }
+.lh-sm { line-height: var(--lh-sm) !important; }
+.lh-md { line-height: var(--lh-md) !important; }
+.lh-lg { line-height: var(--lh-lg) !important; }
+.lh-xl { line-height: var(--lh-xl) !important; }
+.lh-xxl { line-height: var(--lh-xxl) !important; }
```
aaronshekey: Honestly, these might be used most in density modes.

aaronshekey (commented):
Just a few notes after running regression testing and not finding anything.

lib/css/atomic/_stacks-spacing.less
dancormier: On it!

lib/css/atomic/_stacks-flex.less
dancormier: Yeah, I consider this a long-term TODO. 

That said, if we wind up with `@su` only here (and maybe one other spot), it might be worth it to just hardcode the pixel values here for the sake of simplicity. No rush from me though.

aaronshekey (commented):
I think `sus` could be called `w` for width since we use it with stuff that doesn't change size, borders, box shadows, etc. But that's a whole other can of worms, isn't it?

That way, if you're scanning some CSS you can tell which ones are a fixed width and which ones are variable spacing units 🤔

dancormier:
> I think `sus` could be called `w` for width since we use it with stuff that doesn't change size, borders, box shadows, etc. But that's a whole other can of worms, isn't it?
>
> That way, if you're scanning some CSS you can tell which ones are a fixed width and which ones are variable spacing units 🤔

Huh, I'm picking up what you're putting down. 

It does seem like a can of worms just by the fact that I feel 7 conflicting thoughts come up by the idea. Mostly around the fact that I think all `*-width` CSS properties should be actually be `*-size`. _Width_ implies horizontal size specifically but there's value in abiding by spec conventions set decades ago. I could go on. Kinda like how the `left` and `right` values for `text-align` are so pervasive as opposed to `start` and `end`. But I digress…

All this is to say, I can see a justification for a rename and I can't think of anything way better than `w` for width, but I also don't love `w` for width.

aaronshekey:
After some Slackin', we think `--su-static` might be the clearest.

aaronshekey (approved):

Issue #872, opened 2022-03-10
PopperJS to Floating-UI


aaronshekey:
Popper.JS has been renamed Floating-UI. We'll have to take a look at if there are any breaking changes and possibly update a package.

PR #878, opened 2022-03-15
Swap of popperjs with floating-ui

dancormier:
Addresses https://github.com/StackExchange/Stacks/issues/872

This is very much a WIP.


b-kelly:
The changes floating-ui have made look really cool, so I'm excited to see the progress on this one. Moving to floating-ui is definitely the right path forward for us, eventually. It looks like it is still in the `0.x.x` stages, but if it works and is stable for our uses, I don't see why that should hold us back.

PR #882, opened 2022-03-22
Remove dynamic config

dancormier:
This PR removes `lib/css/base/_stacks-configuration-dynamic.less`, and replaces all references to dynamic config Less variables to CSS variables (with the exception of `@body-site-width`, which has been replaced with the hard-coded value of `1060px`).

I've pulled these changes into Core, tried a few different themes, and have yet to notice any regressions.


lib/css/base/_stacks-body.less
```diff
@@ -33,10 +32,9 @@ body {
 }, @force-selector: true);
 
 body {
-    #stacks-internals #load-config();
     box-sizing: border-box;
     min-height: 100%;
-    background-color: @body-background;
+    background-color: var(--white);
```
b-kelly: I'd recommend making this a `theme` variable. We use `--theme-background-color` in Core, so shadowing this would allow us to drop some legacy code there.

lib/css/base/_stacks-internals.less
```diff
@@ -1,15 +1,5 @@
...
-    //     a {
-    //         #stacks-internals #load-config();
-    //         color: @link-color-regular;
-    //     }
-    #load-config() {
-        #stacks-config();
-    }
-
     #load-static-config() {
         #stacks-config-static();
```
b-kelly, line 4: We're coming for you next `#stacks-config-static`

lib/css/components/_stacks-banners.less
```diff
@@ -73,9 +73,8 @@
 
 //  --  When we want to keep hero content capped
 .s-banner--container {
-    #stacks-internals #load-config();
     position: relative;
     width: 100%;
-    max-width: @body-site-width;
+    max-width: 1060px; // TODO consider creating a CSS variable for this
```
b-kelly: ~~Definitely move this out to a variable. If we don't intend on allowing consumers to change this value, let's at least make it a Less variable so we can do math on it.~~

Wait, we're literally only using this variable in this one spot? Let's see if we can reuse a different variable instead. Perhaps we could use `@s-full`? That's what `.wmx12` and the topbar container uses by default.

lib/css/components/_stacks-buttons.less
```diff
@@ -230,8 +230,7 @@
     //      Creates a gray button style.
     //  ----------------------------------------------------------------------------
     .s-btn__muted {
-        #stacks-internals #load-config();
-        color: @button-muted-color;
+        color: var(--black-500);
```
b-kelly, line 233: Overall, lgtm. I'd like to consider making these theme variables, but I don't think that's necessary at the moment since we don't have any consumers overriding them.

lib/css/components/_stacks-tags.less
```diff
@@ -166,46 +164,40 @@ a.s-tag:not(.is-selected) {
 //  $$  Required Tag
 //  ---------------------------------------------------------------------------
 .s-tag__required {
-    #stacks-internals #load-config();
-    .s-tag-styles(@tags-required-border, @tags-required-background, @tags-required-color);
+    .s-tag-styles(var(--bc-darker), var(--black-075), var(--black-700));
```
b-kelly, line 167: Same comment as above with buttons, lgtm.

b-kelly (changes requested):
Almost lgtm! Some small suggestions left inline.

b-kelly:
Side note: I'm _super duper stoked_ to finally see this happening!

lib/css/components/_stacks-banners.less
dancormier: I'm not sure why this was set to `1060px` to begin with. That value doesn't map to any existing value, so the banner will never be aligned with other content. I kept that value in place just to keep from changing **any_** values in this PR, but I think we'd be safe to tweak it.

How about `@s-step * 10`? That would be equivalent to `.wmx10` which is roughly `1054px`.

b-kelly: I think mimicking `.wmx10` would be an excellent compromise. We can always revisit later if we want to.

lib/css/exports/_stacks-constants-colors.less
```diff
@@ -198,7 +198,7 @@
 
 .component-colors {
     --theme-body-font-color: var(--black-800);
-
+    --theme-background-color: var(--white);
```
b-kelly, line 201: lgtm. This should "Just Work™" in Core, but I'll be sure to test for regressions there.

b-kelly (approved):
Code looks good to me. I haven't had a chance to really dive into a local/visual test yet, but I doubt there will be any issues there. I'll give that a go later and report back.

PR #883, opened 2022-03-22
Add code copy buttons to docs

### LINKED ISSUE #829: Add copy to clipboard button to code blocks
A suggestion from @MaxHorstmann that I'd love to include at some point. Code blocks could use their own copy to clipboard button.

There have been other requests for a proper stimulus copy to clipboard button, so perhaps we could break this down a bit and combine the components where necessary?

dancormier:
Addresses https://github.com/StackExchange/Stacks/issues/829

For now, I've only applied it to `product/components/cards` to keep myself from having to change it in 1000 places. It's out of draft to solicit some 👍/👎 on the approach, but consider this in "do-not-merge" territory.

### TODO

- [ ] Add copy button to all code snippets


docs/assets/js/controllers/clipboard.ts
```diff
@@ -0,0 +1,32 @@
+(function(){
+    var application = Stimulus.Application.start();
```
b-kelly: Once #707 merges we can do the following:

```
import { application } from "stacks";
application.register(...);
```

docs/product/components/cards.html
```diff
@@ -9,13 +9,18 @@
     <p class="stacks-copy">The base card styling applies a border and padding to the card.</p>
     <p class="stacks-copy">Cards can be any size and it’s ok to increase the body text size for larger cards.</p>
     <div class="stacks-preview">
+        <div class="stacks-clipboard-content ps-relative" data-controller="clipboard">
```
b-kelly: We could/should probably extend/replace the `highlight` shortcode to include the copy/paste button as well. That way we'd get it "for free" and wouldn't have to add boilerplate all over the docs.

b-kelly (commented):
I like this approach! I'd recommend implementing this on the [checkbox page](https://deploy-preview-883--stacks.netlify.app/product/components/checkbox/) as well to see how the button looks in a codeblock that has horizontal scrolling (partially obscuring the code).

docs/assets/js/controllers/clipboard.ts
```diff
@@ -0,0 +1,32 @@
...
+            const hideElements = scope.findAllElements('[data-hide-on-copy]');
+            const showElements = scope.findAllElements('[data-show-on-copy]');
+
+            hideElements.map(el => el.classList.add("d-none"));
+            showElements.map(el => el.classList.remove("d-none"));
+
+            setTimeout(function () {
+                hideElements.map(el => el.classList.remove("d-none"));
+                showElements.map(el => el.classList.add("d-none"));
+            }, 3000);
```
double-beep: [Is performing a mapping operation without using returned value an antipattern?](https://stackoverflow.com/questions/56903693)

docs/.eleventy.js
```diff
@@ -134,6 +134,35 @@ module.exports = function(eleventyConfig) {
...
+    output +=  '<span data-hide-on-copy>' + copyIcon + '</span>';
+    output += '</button>'
+    output += '<div class="s-popover s-popover__tooltip"';
+    output += ' id="' + tooltipId + '"';
+    output += ' role="tooltip"';
+    output += ' aria-hidden="true">';
+    output +=   '<div class="s-popover--arrow"></div>';
+    output +=   '<span class="d-none" data-show-on-copy>Copied</span>';
+    output +=   '<span data-hide-on-copy>Copy</span>';
+    output += '</div>';
```
double-beep: No comment...

double-beep (changes requested):

docs/.eleventy.js
```diff
@@ -134,6 +134,35 @@ module.exports = function(eleventyConfig) {
     return output;
   });
 
+  // Copy button shortcode
+  eleventyConfig.addLiquidShortcode("copybutton", function(name) {
+    var checkmarkIcon = Icons["Checkmark"];
+    var copyIcon = Icons["Copy"];
+    var tooltipId = "tooltip-" + (name || Math.floor(Math.random() * 1000));
+
+    var output = '';
```
allejo: Is there something stopping you from using a template string (` `` `) for this block?

dancormier: Nothing's stopping me from using a template string. Originally, I lifted the implementation from other similar functions within this file out of ~~laziness~~ consistency, but using template strings here does make it much more readable/maintainable. Commit incoming.

docs/assets/less/stacks-documentation.less
```diff
@@ -231,6 +231,20 @@
...
+//  ----------------------------------------------------------------------------
+.stacks-copy-btn {
+    .svg-icon,
+    .svg-icon * {
+        pointer-events: none;
+    }
+
+    .iconCheckmark {
+        color: var(--green-500);
+    }
```
aaronshekey, line 245: My normal trick of applying `pointer-events: none` to SVG isn't quite working. When moving the mouse within the button there's weird behavior with the tooltip hiding and showing.

docs/assets/less/stacks-documentation.less
```diff
@@ -255,10 +269,12 @@
         box-shadow: none;
     });
 
-    > pre.s-code-block {
+    > pre.s-code-block,
+    > .stacks-clipboard-content pre.s-code-block {
         border-radius: var(--br-md) var(--br-md) 0 0;
         border: 1px solid var(--bc-medium);
         max-height: 24rem;
+        padding-right: var(--su32);
```
aaronshekey, line 283: When you scroll all the way to the right, we want to be able to see and select everything.

docs/plugins/copy-button.js
```diff
@@ -0,0 +1,29 @@
+const { default: Icons } = require("@stackoverflow/stacks-icons");
+
+module.exports = {
+    configFunction(eleventyConfig) {
+      eleventyConfig.addLiquidShortcode("copybutton", function(name) {
+        var tooltipId = "tooltip-" + (name || Math.floor(Math.random() * 1000));
+
+        var output = `<button
+          class="stacks-copy-btn s-btn s-btn__muted s-btn__outlined s-btn__icon ps-absolute p4 bg-black-050 bc-black-100 t8 r8"
```
aaronshekey: Some button customization here.

docs/product/components/cards.html
dancormier: I'd like to add the copy button to the `highlight` shortcode, but also have it available standalone. This will allow us to add a copy button to any element, like so:

image

It's a contrived example, but I think it illustrates the point.

MaxHorstmann:
Love this! Any chance we'll see it on SO one day? :)

dancormier:
Hey @MaxHorstmann! Nice to hear from you!

These specific copy buttons are intended for the Stacks docs only, _but_ I personally love the idea of introducing them to Stack Overflow, writ large. We have the `s-codeblock` component and we could enhance that component with some copy JS. I'll spread the idea around and see if it takes.

MaxHorstmann:
@dancormier hey Dan, good to hear from you as well! Would be great to have this on code snippets on SO questions & answers, like on GitHub and many docs pages ([random example](https://posthog.com/docs/apps/build/tutorial)) these days.

PR #890, opened 2022-03-29
Add linters

b-kelly:
Adds prettier and eslint with configs based on those in StackExchange/Stacks-Editor. This only adds the linters, but does not clean up any linting errors (follow up PR).

### Future work

- clean up errors
- enable eslint errors for JS files (or migrate 11ty files to TS)
- add linting checks to GH actions [like in Stacks-Editor](https://github.com/StackExchange/Stacks-Editor/blob/main/.github/workflows/main.yml#L13-L33)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,6 +7,7 @@ end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
+max_line_length = 80
 
 [*.{json,yml,yaml}]
 indent_size = 2
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,3 @@
+node_modules
+dist
+**/*.js
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,7 +17,5 @@ module.exports = {
         "no-console": "error",
         "no-alert": "error",
         "no-process-env": "error",
-        // TODO this one is a pain to fix w/ ProseMirror's Class<T extends Schema = any> types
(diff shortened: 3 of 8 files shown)


dancormier (approved):
Nice!

Looking forward to the mammoth cleanup PR that comes after this one. Taking bets on lines +/-

.editorconfig
```diff
@@ -7,6 +7,7 @@ end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
+max_line_length = 80
```
aaronshekey, line 10: Boooooooooooooo but I get it.

dancormier: I tend to be more of 120 fella myself, but 🤷 .

b-kelly: I could be swayed here to change to a different value (or remove entirely). This is what we're using on Stacks-Editor and has worked out well enough there, so I moved it over for consistency's sake.

aaronshekey: I actually really love soft wrapping on my editor, especially when writing documentation and the like so ¯\_(ツ)_/¯ . I would _not_ complain if we left that bit of unconfigured :v:

dancormier: I'm down with the consistency

b-kelly: @aaronshekey The formatter only applies the max line length to code _I think_. Comments and html text are both excluded iirc.

b-kelly: Merging for now, can trivially revisit if we hate it

PR #892, opened 2022-03-30
Remove the static config

dancormier:
This PR removes `lib/css/base/_stacks-configuration-static.less`, and moves necessary Less variables (`@css-reset`, `@breakpoints`) to `lib/css/exports/_stacks-constants-helpers.less`.

~~I've yet to test this against core so standby.~~

I've tested this against core and it LGTM but I'd love some other eyes on it.

---

See also:

PR comment saying "we're coming for you next #stacks-config-static"


lib/css/stacks-static.less
```diff
@@ -83,4 +83,3 @@
 
 //  --  CONFIG
 @import "base/configuration-static.less";
```
aaronshekey: I think this one's gotta go yet.

dancormier: Ah, damn funky merge. Removed!

Also, thanks for catching this!

aaronshekey (approved):

b-kelly (changes requested):
Looks like we've lost the ability to set css resets and multiple breakpoints - both of which we're using in prod. I'm going to see if I can come up with a good solution here.

dancormier:
> Looks like we've lost the ability to set css resets and multiple breakpoints - both of which we're using in prod. I'm going to see if I can come up with a good solution here.

Is the reset not handled through `lib/css/base/reset.less`?

```
// the default CSS reset is normalize.css; other possibilities are the Eric Meyer reset, and no reset at all.

#stacks-internals() {
    #include-css-reset(meyer) {
        @import "_stacks-reset-meyer.less";
    }
    #include-css-reset(normalize) {
        @import "_stacks-reset-normalize.less";
    }
    /* stylelint-disable block-no-empty */
    #include-css-reset(none) {
    }
    /* stylelint-enable */
}

& {
    #stacks-internals #load-static-config();
    #stacks-internals #include-css-reset(@css-reset);
}
```

b-kelly:
@dancormier Unfortunately, no. I just double checked and downstream consumers aren't setting this value with the current syntax. Working on a downstream fix now to see if we need to make any changes here.

b-kelly:
I'm starting to think we can't merge this without an overhaul on how we deliver breakpoints and resets. Any concerns on leaving this for 2.0?

dancormier:
> I'm starting to think we can't merge this without an overhaul on how we deliver breakpoints and resets. Any concerns on leaving this for 2.0?

Yep, sounds good.

PR #893, opened 2022-03-30
Replace Less border-radius (`@br-*`) variables with CSS variables

dancormier:


aaronshekey (approved):
This looks painless and well-executed on the Stacks side. The hardest part will be managing the changes on the production side which we'll need to sync on.

PR #894, opened 2022-03-30
Replace Less box-shadow (`@bs-*`) variables with CSS variables

dancormier:
**Note:** Breaking change for the following locations in Core:
https://github.com/StackEng/StackOverflow/blob/master/StackOverflow/Content/Shared/Channels/_channels-admin.less#L179
https://github.com/StackEng/StackOverflow/blob/master/StackOverflow/Content/Shared/Channels/channels.less#L252


lib/css/exports/_stacks-constants-helpers.less
```diff
@@ -44,10 +44,12 @@
...
-@br-sm:                 3px;
-@br-md:                 5px;
-@br-lg:                 7px;
-@br-circle:             50%;
+body {
+    --br-sm:                 3px;
+    --br-md:                 5px;
+    --br-lg:                 7px;
+    --br-circle:             50%;
+}
```
aaronshekey:
```suggestion
@br-sm:                 3px;
@br-md:                 5px;
@br-lg:                 7px;
@br-circle:             50%;
```

We don't need this change in _this_ PR, do we?

aaronshekey (changes requested):

lib/css/exports/_stacks-constants-helpers.less
dancormier: Whoops!

aaronshekey (approved):

PR #895, opened 2022-03-30
Replace Less z-index (`@zi-*`) variables with CSS variables

dancormier:

### DIFF
--- a/lib/css/exports/_stacks-constants-helpers.less
+++ b/lib/css/exports/_stacks-constants-helpers.less
@@ -32,8 +32,8 @@ body {
     --zi-banners:            4000;   //  Banners
     --zi-navigation:         5000;   //  Navigation Bars
     --zi-navigation-fixed:   5050;   //  Fixed navigation bars
+    --zi-modals-background:  8050;   //  Modals background
     --zi-modals:             9000;   //  Modals
-    --zi-modals-background:  calc(var(--zi-modals) - 50);
 }
 
 //  ============================================================================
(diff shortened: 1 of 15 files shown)


lib/css/exports/_stacks-constants-helpers.less
```diff
@@ -21,18 +21,20 @@
...
+    --zi-selected:           25;     //  Pop above siblings
+    --zi-active:             30;     //  Pop above selected siblings
+    --zi-dropdown:           1000;   //  Dropdowns
+    --zi-popovers:           2000;   //  Popovers, Popups, Autocompletes
+    --zi-tooltips:           3000;   //  Tooltips
+    --zi-banners:            4000;   //  Banners
+    --zi-navigation:         5000;   //  Navigation Bars
+    --zi-navigation-fixed:   5050;   //  Fixed navigation bars
+    --zi-modals:             9000;   //  Modals
+    --zi-modals-background:  calc(var(--zi-modals) - 50);
```
aaronshekey: Maybe we just hard code this. It's not like `zi-modals` varies.

lib/css/exports/_stacks-constants-helpers.less
```diff
@@ -21,18 +21,20 @@
...
-@zi-active:             30;     //  Pop above selected siblings
-@zi-dropdown:           1000;   //  Dropdowns
-@zi-popovers:           2000;   //  Popovers, Popups, Autocompletes
-@zi-tooltips:           3000;   //  Tooltips
-@zi-banners:            4000;   //  Banners
-@zi-navigation:         5000;   //  Navigation Bars
-@zi-navigation-fixed:   5050;   //  Fixed navigation bars
-@zi-modals:             9000;   //  Modals
-@zi-modals-background:  @zi-modals - 50;
+body {
```
aaronshekey, line 24: Some discussion might need to be had about where we actually put some of these. I don't imagine a theme or a user modifying these specific variables, so they could be tied to `:root`? 🤔 

This works for now, of course.

aaronshekey (approved):

PR #896, opened 2022-03-30
Replace Less font-family (`@ff-*`), font-size (`@fs-*`), line-height (`@lh-*`) variables with CSS variables

dancormier:
This PR places these variables within `html, body`. All other variables are in the scope of `body` only, but using that context would break the changes in `_stacks-body.less`. I ran backstop and this change resulted in no visual regressions (3 errors popped up but they were because different randomized avatar images popped up).

IMO we should consider adding CSS variables in the scope of `html` instead of `body` as standard.

### DIFF
--- a/lib/css/exports/_stacks-constants-type.less
+++ b/lib/css/exports/_stacks-constants-type.less
@@ -74,6 +74,7 @@
     font-weight: 800;
 }
 
+// TODO remove font-family Less variables. Keeping them here for now for comments.
 @ff-sans:
     -apple-system, BlinkMacSystemFont, // San Francisco on macOS and iOS
     "Segoe UI Adjusted", "Segoe UI", // Windows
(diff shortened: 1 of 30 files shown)


aaronshekey:
> IMO we should consider adding CSS variables in the scope of `html` instead of `body` as standard.

Ah, yes. I'm happy to use `html` and `:root` for things that make sense. The decision to use `body` was really just because it was much easier to modify the `body` class in production at the time.

lib/css/exports/_stacks-constants-type.less
```diff
@@ -87,66 +88,66 @@
...
 // Now the various line-height variables
-@lh-xs:    1;
-@lh-sm:    ((@_stacks-internals-lh-unit + 2) / @_stacks-internals-lh-unit);
-@lh-md:    ((@_stacks-internals-lh-unit + 4) / @_stacks-internals-lh-unit);
-@lh-lg:    ((@_stacks-internals-lh-unit + 8) / @_stacks-internals-lh-unit);
-@lh-xl:    ((@_stacks-internals-lh-unit + 12) / @_stacks-internals-lh-unit);
-@lh-xxl:   2;
+    --lh-xs:    1;
+    --lh-md:    calc((var(--stacks-internals-lh-unit) + 4) / var(--stacks-internals-lh-unit));
+    --lh-sm:    calc((var(--stacks-internals-lh-unit) + 2) / var(--stacks-internals-lh-unit));
```
aaronshekey:
```suggestion
    --lh-sm:    calc((var(--stacks-internals-lh-unit) + 2) / var(--stacks-internals-lh-unit));
    --lh-md:    calc((var(--stacks-internals-lh-unit) + 4) / var(--stacks-internals-lh-unit));
```

aaronshekey:
```suggestion
    --lh-sm:    calc((var(--stacks-internals-lh-unit) + 2) / var(--stacks-internals-lh-unit));
    --lh-md:    calc((var(--stacks-internals-lh-unit) + 4) / var(--stacks-internals-lh-unit));
```

aaronshekey (changes requested):
Just one little flip here, but this looks good otherwise.

aaronshekey (approved):

PR #897, opened 2022-03-30
Replace Less transition (`@te-*`) variables with CSS variables

dancormier:

### DIFF
--- a/lib/css/atomic/_stacks-misc.less
+++ b/lib/css/atomic/_stacks-misc.less
@@ -316,9 +316,9 @@
 
 // -- Provide a sensible default
 .t {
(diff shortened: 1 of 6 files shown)


aaronshekey (approved):
Some of this is pretty hypothetical at this point. Pretty sure we only use `te-smooth` and `te-smooth-quick` in production. Looks good.

dancormier:
> Some of this is pretty hypothetical at this point. Pretty sure we only use `te-smooth` and `te-smooth-quick` in production. Looks good.

Yeah, I considered adding a note that we don't use a few of these but figured I'd just change them to CSS vars and leave it for now.

PR #898, opened 2022-03-31
Migrate  s-topbar--ctas to s-topbar--item__unset

b-kelly:
Scrolling is a bit funny when both a `--content` and a `--ctas` exist together. The solution is to combine the two and just not apply `--content` styling to the ctas. This approach gives us more flexibility in what a user can add to a topbar's content.

Addresses [this concern on MSE](https://meta.stackexchange.com/a/377605/395497).

### DIFF
--- a/docs/_includes/topbar.html
+++ b/docs/_includes/topbar.html
@@ -141,7 +141,9 @@
             {% else %}
                 <li>
                     <a href="#" class="s-topbar--item s-topbar--item__unset s-btn s-btn__filled ws-nowrap">Log in</a>
-                    <a href="#" class="s-topbar--item s-topbar--item__unset s-btn s-btn__primary ws-nowrap ml4">Sign up</a>
+                </li>
+                <li>
+                    <a href="#" class="s-topbar--item s-topbar--item__unset ml4 s-btn s-btn__primary ws-nowrap">Sign up</a>
                 </li>
             {% endunless %}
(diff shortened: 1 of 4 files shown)


docs/_includes/topbar.html
```diff
@@ -117,37 +138,12 @@
...
-                        <span class="s-activity-indicator s-activity-indicator__danger">3</span>
-                    </a>
-                </li>
-                <li>
-                    <a href="#" class="s-topbar--item" title="Achievements">
-                        {% icon "Achievements" %}
-                        <span class="s-activity-indicator s-activity-indicator__success">+10</span>
-                    </a>
+                    <a href="#" class="s-topbar--item s-topbar--item__unset s-btn s-btn__filled ws-nowrap">Log in</a>
+                    <a href="#" class="s-topbar--item s-topbar--item__unset s-btn s-btn__primary ws-nowrap ml4">Sign up</a>
```
dancormier: These should probably be two separate `<li>`s, right?

dancormier (approved):
This approach seems reasonable to me!

docs/_includes/topbar.html
b-kelly: Done. That's something I had considered, but then decided not to do (for arbitrary reasons). It works either way and (in hindsight) your way is "more correct", so I've reverted that back.

PR #903, opened 2022-04-04
Use CSS variables for s-step and s-full

aaronshekey:
What it says in the title. This allows us to change widths in the client by overriding `--s-full`


dancormier (approved):
I love being able to mess with a single variable in the browser and see everything resize so cleanly. 🦐

PR #904, opened 2022-04-04
Remove `@su*` Less vars and references

dancormier:
So, we don't _have_ to do this. These `@su*` variables aren't increasing bundle size or anything.

That being said, I figured I'd make a branch where we remove `@su*` and hardcode values that reference them just in case we decide it's worth it.

See also https://github.com/StackEng/StackOverflow/pull/11388


aaronshekey:
Latest commits remove any reliance on hardcoded pixel values. Mostly just interpolation and rethinking internal properties.

lib/css/atomic/flex.less
```diff
@@ -103,7 +103,7 @@
             @itemWidth: (@count * 100% / @num);
 
             > .flex--item@{count} {
-                margin: (@spacing / 2);
+                margin: calc(~"@{spacing}" / 2);
```
dancormier, line 106: > [No one has ever done anything like this.](https://youtu.be/j0Pa6-rqGuI?t=22)

PR #905, opened 2022-04-04
Explore replacing jQuery with Cash

dancormier:
jQuery works but I figured Stacks would be a good, isolated environment to explore replacing it with [Cash](https://github.com/fabiospampinato/cash) and saving a few KBs (see [the comparison chart](https://github.com/fabiospampinato/cash#comparison) and [cash-dom@8.1.1](https://bundlephobia.com/package/cash-dom@8.1.1)/[jQuery@3.6.0](https://bundlephobia.com/package/jquery@3.6.0)) sizes.

## Worth noting

- This PR ditches much of the navigation functionality (see https://github.com/StackExchange/Stacks/pull/905/commits/fc04f40c64881d3fc0d918e922986528e86816cb). It didn't seem necessary, but there's totally a chance I'm completely overlooking the need. We can add it back but we'll need to extend Cash or add some more vanilla JS.
- Cash doesn't include `queue`/`dequeue` methods, so we'll need to find a replacement for toast docs functionality.
- We could import Cash on a per-file basis instead of adding it globally, but I opted for a light-touch in this PR.


docs/assets/js/global.navigation.js
```diff
@@ -1,92 +1,6 @@
...
-        })
-
-        window.onpopstate = history.onpushstate = function(e) {
-            if(e.state){
-                // Update the page
-                $('title').text(e.state.title)
-                $('#nav').html(e.state.nav)
-                $('#content').html(e.state.content)
-            }
-        }
```
aaronshekey, line 90: This is what makes the navigation between sections seamless.

In our production docs, head all the way down to the user cards section. Then click the uploader section. You'll see that the navigation stays put, since we're replacing the main content area via AJAX instead of doing a hard page load.

Without this functionality, the UX suffers since you have to re-scroll the entire way to choose the uploader section.

aaronshekey (changes requested):

docs/assets/js/global.navigation.js
dancormier: Ah, I see now. We can probably replace `$.ajax` with `fetch` pretty easily. I think there might be a couple more methods in there that don't exist in Cash.

aaronshekey: This is all a roundabout way of saying this is basically turbolinks or pjax or whatever it's called.

b-kelly:
Hear me out on this one - what if we just dropped jquery entirely? We don't have _that much_ JS using it.

aaronshekey:
> what if we just dropped jquery entirely

Better get crackin' 🥰😇

dancormier:
> Hear me out on this one - what if we just dropped jquery entirely? We don't have _that much_ JS using it.

The thought crossed my mind. I figured this was a less risky stopgap for now that would help ~~kick the can down the road~~ save a few KBs. That said, if we dropped jQuery, I wouldn't shed a tear.

PR #906, opened 2022-04-05
Drop `_stacks-` from filenames

aaronshekey:
Stacks is now lives in `node_modules` in production, meaning the files are no longer accessible in the project. We can now drop `_stacks-` from each filename.

This PR explores this breaking change prior to `1.0.0`


b-kelly (approved):
🎉 Works on my machine™

As you mention, this will have breaking changes downstream, but they should be fairly trivial to spot and then clean up.

dancormier (approved):
What Ben said!

PR #914, opened 2022-04-12
Add state badge styling and documentation

### LINKED ISSUE #913: Abstract out post summary state badges
There's been [a](https://stackexchange.slack.com/archives/C27RWNQN9/p1649793368848369) [few](https://stackexchange.slack.com/archives/C27RWNQN9/p1648572388785989?thread_ts=1648572283.445809&cid=C27RWNQN9) [requests](https://stackexchange.slack.com/archives/C27RWNQN9/p1649793105455299) to make the post summary badges portable and general-purpose.

dancormier:
This PR addresses #913 by abstracting several post summary states to badge styles. This will make state badges portable and ready for more generalized use. 

Once this PR is merged and in Stacks is updated in core, we can replace the changes made in https://github.com/StackEng/StackOverflow/pull/11575

To see the state badges, visit https://deploy-preview-914--stacks.netlify.app/product/components/badges/#badge-states

### A note on badges within post summary stats

I avoided any breaking changes in this PR but we should deprecate a few post summary state classes. We'll need to make minor syntactical changes to the post summary component used in core. Once that's complete, we should probably rip out some styling on `.is-*` classes and cut a new version.


b-kelly:
I like this approach. I'd abstract out a tad more with a `s-badge__icon` class, so we could add icons to any badge we want. I'd also like to make these badge names a bit more generic - warning, info, muted, etc but that's not a huge priority.

I'd also like to implement these into post-summary as well while we're at it.

gkoscky (approved):
Looks great. This will greatly simplify our tag helper.

My only nit is that while on one hand I agree with the idea of making these color schemes generic, the "sentiment-adjacent" (for lack of a better word) names (warn, info, etc) might get confusing if we ever decide to change the color-scheme for the post-state badges.

That's one big if, but just something that caught my eye. If "Review" goes from the current yellow to a nice purple (this is why I don't design things) the `__warn` qualifier would make little sense.

But it's just a tiny nit. The badges and code itself looks great

lib/css/components/post-summary.less
```diff
@@ -158,7 +160,14 @@
...
         }
 
+        &.s-badge {
+            font-size: var(--fs-body1);
+            line-height: var(--lh-md);
+            padding: var(--su2) var(--su4);
+        }
+
         &.has-answers,
+        // TODO: Remove .has-bounty styling
```
b-kelly: nit: Adding a specific label here will allow us to quickly and trivially find all deprecated functionality when we eventually do a new major/breaking release.

```suggestion
        // TODO DEPRECATED: Remove .has-bounty styling
```

lib/css/components/post-summary.less
```diff
@@ -186,6 +195,7 @@
             }
         }
 
+        // TODO: Remove and replace with references with `s-badge s-badge__bounty`
```
b-kelly: nit

```suggestion
        // TODO DEPRECATED: Remove and replace with references with `s-badge s-badge__bounty`
```

b-kelly: Do the same with all below - I'm being lazy and not marking all with a code suggestion.

lib/css/components/badges.less
```diff
@@ -59,6 +59,15 @@
     margin-left: calc((var(--su4) + var(--su1)) * -1);
 }
 
+//  $   BADGE ICON
+//      When we have a badge icon
+//  ---------------------------------------------------------------------------
+.s-badge__icon {
+    .svg-icon {
+        margin-right: var(--su4);
+    }
```
b-kelly: I think we could reuse the code from s-post-summary here so we don't have to re-override it there (and it gives a bit of extra consistency as well)

```suggestion
    gap: 0.3em;
```

lib/css/components/badges.less
```diff
@@ -174,6 +183,44 @@
...
 }
 
+//  $$  States
+//  ---------------------------------------------------------------------------
+&.s-badge__danger {
+    color: var(--red-900);
+    background-color: var(--red-100);
+    border-color: var(--red-600);
+
+    &.s-badge__filled {
```
b-kelly, line 191: Do we want to add filled versions to all badges? As of right now, the api is a tad confusing since the filled modifier doesn't work on all types. Alternatively, this suggestion may fall into YAGNI territory.

lib/css/components/badges.less
```diff
@@ -174,6 +183,44 @@
...
+    }
+}
+
+&.s-badge__info {
+    color: var(--blue-900);
+    background-color: var(--blue-100);
+    border-color: var(--blue-600);
+}
+
+&.s-badge__warn {
```
b-kelly: Let's match our other alias class names and use `warning` here

lib/css/components/post-summary.less
```diff
@@ -143,12 +143,14 @@
     color: var(--fc-light);
 
     .s-post-summary--stats-item {
-        display: inline-flex;
-        gap: 0.3em;
-        align-items: center;
-        justify-content: center;
-        white-space: nowrap;
-        border: 1px solid transparent;
+        &:not(.s-badge) {
```
b-kelly, line 146: optional: Other than `border`, we can get away with not excluding badge from any of these. Up to you on whether we want to do so or not.

lib/css/components/badges.less
b-kelly: Also, since these are all already "filled' (unlike s-btn), should we perhaps follow the convention set by s-notice and use "important" instead?

b-kelly (changes requested):
Looking good! Left some comments on a few small changes, but lgtm otherwise

lib/css/components/badges.less
dancormier: D'oh! Nice catch!

lib/css/components/badges.less
dancormier: I feel you on this. `s-badge__important` is already taken though (see [badges/#number-counts](https://deploy-preview-914--stacks.netlify.app/product/components/badges/#number-counts)). `__filled` was the best option I could come up with but I'm not totally happy with it. I'll see if I can come up with anything better.

b-kelly: Ah shoot, I forgot about that one. Good call. Let me think on this one a bit as well.

lib/css/components/post-summary.less
dancormier: My preference is to apply as few styles to a nested component as possible.

lib/css/components/badges.less
b-kelly: Let's stick with "filled" for now. We can switch over to "important" as a breaking change once we have a few more under our belt.

b-kelly (approved):
lgtm! Made a few slight tweaks to the docs wording (and fixed a warn -> warning), but everything else looks great. Nice work

PR #915, opened 2022-04-14
Restore 0 margin and padding

aaronshekey:
This PR restores `mt0`, etc. by calling it separately outside of our CSS variables. It also responsifies each.


aaronshekey:
Another approach would be to introduce `--su0` but... why?

dancormier (approved):
This works!

`--su0` wouldn't make sense except for the sake of consistency. I have a hard time imagining that we'd need to alter pervasive `0` padding/margin values.

Thanks for tackling this bug! 🦐

PR #926, opened 2022-04-27
Remove grunt-concurrent dependency

dancormier:
I saw https://github.com/StackExchange/Stacks/pull/925 and was :confused:

I'm _pretty sure_ we can ditch this dependency, but I wanna ask the subject matter expert to be certain. We don't need `grunt-concurrent` for any special reason that I'm just completely blind to, correct @b-kelly?


b-kelly (approved):
lgtm, nice catch - we _definitely_ don't need this. It may have been accidentally added back during a merge conflict or similar.

PR #932, opened 2022-05-06
add accessibility attributes to s-popover

mukunku:
# Summary

In order for popovers to be accessible to screen readers and other assistive technologies there are a few requirements that need to be met as laid out here: https://techservicesillinois.github.io/accessibility/examples/popover.html

> A popover must indicate if it is expanded (visible) or collapsed (hidden) and the trigger for the popover must be associated with the popover region to be displayed.
> 
> * button (role) – set on the popover trigger.
> * aria-expanded (state) – Set to true if the popover is visible; false if the popover is hidden.
> * aria-controls (property) (optional) – Set to the HTML id of the region toggled by popover trigger.

I think it makes the most sense to take care of this in the Stacks component level so all popovers are automatically accessible.

# Implementation

It looks like we're already handling the optional `aria-controls` attribute 🎉 . So this PR adds the `role` attributes automatically to all elements that have a popover, as well as add the `aria-expanded` state as appropriate to the elements.

Based on this post the `role` attribute is supported on most, if not all, html elements so hopefully setting the `role` automatically is not an issue: https://stackoverflow.com/a/18664038/1458738

### DIFF
--- a/lib/ts/controllers/s-popover.ts
+++ b/lib/ts/controllers/s-popover.ts
@@ -302,8 +302,6 @@ export class PopoverController extends BasePopoverController {
     public override connect(): void {
         super.connect();
 
-        //Make sure the popover is screen reader friendly
-        this.referenceElement.setAttribute("role", "button");
         this.toggleAccessibilityAttributes();
     }
 
--- a/lib/ts/controllers/s-popover.ts
+++ b/lib/ts/controllers/s-popover.ts
@@ -278,19 +278,33 @@ export class PopoverController extends BasePopoverController {
     private boundHideOnEscapePress!: any;
 
     /**
-     * Toggles optional classes in addition to BasePopoverController.shown
+     * Toggles optional classes and accessibility attributes in addition to BasePopoverController.shown
      */
-    protected shown(dispatcher: Element|null = null) {
+    protected override shown(dispatcher: Element|null = null) {
         this.toggleOptionalClasses(true);
+        this.toggleAccessibilityAttributes(true);
         super.shown(dispatcher);
     }
 
     /**
-     * Toggles optional classes in addition to BasePopoverController.hidden
+     * Toggles optional classes and accessibility attributes in addition to BasePopoverController.hidden
      */
-    protected hidden(dispatcher: Element|null = null) {
+    protected override hidden(dispatcher: Element|null = null) {
         this.toggleOptionalClasses(false);
+        this.toggleAccessibilityAttributes(false);
         super.hidden(dispatcher);
+    }
+
+
+    /**
+     * Initializes accessibility attributes in addition to BasePopoverController.connect
+     */
+    public override connect(): void {
+        super.connect();
+
+        //Make sure the popover is screen reader friendly
+        this.referenceElement.setAttribute("role", "button");
+        this.toggleAccessibilityAttributes();
     }
 
     /**
@@ -307,7 +321,7 @@ export class PopoverController extends BasePopoverController {
     /**
      * Unbinds global events to the document for hiding popovers on user interaction
      */
-    protected  unbindDocumentEvents() {
+    protected unbindDocumentEvents() {
         document.removeEventListener("mousedown", this.boundHideOnOutsideClick);
         document.removeEventListener("keyup", this.boundHideOnEscapePress);
     }
@@ -356,6 +370,14 @@ export class PopoverController extends BasePopoverController {
         this.data.get("toggle-class")!.split(/\s+/).forEach(function (cls: string) {
             cl.toggle(cls, show);
         });
+    }
+
+    /**
+     * Toggles accessibility attributes based on whether the popover is shown or not
+     * @param {boolean=} show - A boolean indicating whether this is being triggered by a show or hide.
+     */
+    private toggleAccessibilityAttributes(show?: boolean) {
+        this.referenceElement.ariaExpanded = show?.toString() || this.referenceElement.ariaExpanded || 'false';
     }
 }
 


lib/ts/controllers/s-popover.ts
```diff
@@ -278,21 +278,35 @@ export class PopoverController extends BasePopoverController {
...
 
+
+    /**
+     * Initializes accessibility attributes in addition to BasePopoverController.connect
+     */
+    public override connect(): void {
+        super.connect();
+
+        //Make sure the popover is screen reader friendly
+        this.referenceElement.setAttribute("role", "button");
```
dancormier: I'd be curious to get @b-kelly's take, but I might prefer that including this attribute be the responsibility of the consumer of Stacks rather than forcing it.

I can't think of an edge case that this will cause issues with but could their be a case where adding the role is unwanted and the consumer would have to write additional JS to remove it?

lib/ts/controllers/s-popover.ts
```diff
@@ -357,6 +371,14 @@ export class PopoverController extends BasePopoverController {
...
         });
     }
+
+    /**
+     * Toggles accessibility attributes based on whether the popover is shown or not
+     * @param {boolean=} show - A boolean indicating whether this is being triggered by a show or hide.
+     */
+    private toggleAccessibilityAttributes(show?: boolean) {
+        this.referenceElement.ariaExpanded = show?.toString() || this.referenceElement.ariaExpanded || 'false';
+    }
```
dancormier, line 379: Nice! This is all upside IMO.

dancormier (commented):
Thanks so much @mukunku. Making accessibility changes like this have the potential to make a big impact.

Only real concern here is with the addition of the `role` attribute via JS philosophically. Curious to get some thoughts there before I make a solid judgement.

mukunku:
The PR as it is satisfies the _bare minimum_ requirements for accessibility of popovers. There are a few other things we can do to enhance this further. If anyone has the time could you kindly review the following proposal/ideas and see if they make sense to implement as well?

## A) `aria-haspopup="dialog"` vs `role="region"`

There are two conflicting pieces of information regarding popovers that I'd like to discuss.

1. The link I shared above suggests adding the `role="region"` attribute to the popover element. This is reinforced by the w3c spec here: https://www.w3.org/TR/wai-aria-practices-1.1/#wai-aria-roles-states-and-properties
image
2. However according to another w3c standard: https://www.w3.org/TR/wai-aria-1.2/#aria-haspopup (and a few other sources) the popover element should have its role as `dialog` or `menu` and the element that triggers the popover should have the `aria-haspopup` attribute that matches the same role as the popover's.

With option 1, we can easily attach a role to the popover element and mark it as "region" but not sure if that makes sense.

With option 2, there's unfortunately no way for us to know the role of the popover ahead of time. However if the popover element itself has a role already defined such as `role="menu"` then we could dynamically add the `aria-haspopup="menu"` to the element that triggers the popover. This would increase accessibility somewhat. ([Here's one example](https://github.com/StackEng/StackOverflow/blob/e8e73facb98b587e9e3d6db350a273d1702cfea1/StackOverflow/Views/Shared/LeftSidebar.cshtml#L325) where we could automatically populate the `aria-haspopup` role because the target popover element has its role defined)

With all that said, its worth mentioning that the `aria-haspopup="dialog"` type isn't supported very well. See: https://github.com/w3c/aria-practices/issues/1383 

So not sure if we should go for Option 1, Option 2 or skip this altogether for now.

## B) Automatically remove `aria-hidden` ?

This is a problem I've seen in some places but not sure if it makes sense to apply this generically to everything.

I noticed that some people are adding `aria-hidden="true"` to popover elements marked with `s-popover`. However because `s-popover` already applies the `display: none` style it doesn't need an additional `aria-hidden` attribute if I'm understanding things correctly.

So if the user put `aria-hidden="true"` on the popover element and it gets displayed it won't get picked up by screen readers, even though it's visible. So my idea is that when we're initializing our popover component on an html element we should simply remove any `aria-hidden` attributes on the popover itself. I couldn't think of a scenario where it would need to remain there but of course just because I can't think of one doesn't mean it doesn't exist. So any thoughts on this would be great as well.

The [same example above](https://github.com/StackEng/StackOverflow/blob/e8e73facb98b587e9e3d6db350a273d1702cfea1/StackOverflow/Views/Shared/LeftSidebar.cshtml#L325) is an example of this as well.

lib/ts/controllers/s-popover.ts
mukunku: I definitely understand the concern. 

From the literature I've read there are a few conditions that need to be met for a popover to be considered accessible. Not sure how many of them we can 'force' like this on the Stacks side: https://www.w3.org/TR/wai-aria-practices-1.1/#wai-aria-roles-states-and-properties-14

For example it also states that we need to add the `aria-haspopup` attribute but I also shared some concerns around that [below](https://github.com/StackExchange/Stacks/pull/932#issuecomment-1119979072). So maybe these parts of the markup can be left to the author like you said. Happy to go either way. It _feels_ like it would be okay to do because we know its a popover and we're initializing our component on it. But like you said, libraries dictating how you write your markup can be very annoying to deal with.

b-kelly: Regarding `role`, I think it would be better to have the consumer set it. The reference element is not always a button and in fact, is not always even interactive. Additionally, you can have multiple triggers for a single popover (less apparent for popover, but this concept is commonly used for modals).

b-kelly:
@mukunku Thanks so much for the contribution! I'll take a peek at the links you suggest in regards to auto-adding attributes. My gut feeling is that we don't want to auto-add any attribute that a user may want to set differently or on multiple elements (`aria-role=button` for instance), as it would increase the amount complexity and surface area that we would have to support.

That being said, there's absolutely no reason why we couldn't add these attributes to the documentation and examples instead. This would encourage consumers to use them and increase friction to removing them when they / copy-paste an example.

As a side note, a11y is something we want to essentially _force_ once we release our component framework (starting with server-side w/ Razor, then moving on to client-side). This will take care of the problem at render time, so we any code we add here to force them will end up getting ripped out anyways.

lib/ts/controllers/s-popover.ts
mukunku: Thanks for the feedback Ben/Dan. I went ahead and removed the role assignment.

Just some more thoughts I wanted to share here:
From what I've read if the popover triggering element is clickable it should have the `role="button"` attribute regardless of which type of html element it is. But in cases where it's not clickable and only triggerable elsewhere via JS then I agree the role shouldn't be there although I've also read that that isn't very good design in terms of accessibility (Deque's documentation mentions how UI's need to be "predictable"). But we can't predict all use cases like you said.

b-kelly: Definitely agreed on both cases. Feel free to edit the docs as you please to showcase the role attribute. If that's a bit of scope creep for you or if you simply don't have the time to get to it, then we can add the content in a follow-up PR.

mukunku:
> @mukunku Thanks so much for the contribution! I'll take a peek at the links you suggest in regards to auto-adding attributes. My gut feeling is that we don't want to auto-add any attribute that a user may want to set differently or on multiple elements (`aria-role=button` for instance), as it would increase the amount complexity and surface area that we would have to support.
> 
> That being said, there's absolutely no reason why we couldn't add these attributes to the documentation and examples instead. This would encourage consumers to use them and increase friction to removing them when they / copy-paste an example.
> 
> As a side note, a11y is something we want to essentially _force_ once we release our component framework (starting with server-side w/ Razor, then moving on to client-side). This will take care of the problem at render time, so we any code we add here to force them will end up getting ripped out anyways.

That's really good context, thanks Ben. I was thinking maybe we could add some custom analyzers to our IDE to help guide developers into writing accessible html but having examples with accessibility details in https://stackoverflow.design/ + having another automated way to auto fix these issues sounds like a good plan to me 👍🏼

lib/ts/controllers/s-popover.ts
mukunku: I think updating the documentation would be better served as a follow up PR. I won't be able to give it the care it needs right now and don't want to push for a half-baked solution.

mukunku: Just had a random thought as well. Have we considered adding a "Accessibility Considerations" section to our Stacks components pages? We could have one on the [Popover component page](https://stackoverflow.design/product/components/popovers/) for example.

We could lay out details on accessibility there to inform users on the how and why. The reason I think that may be important is because there's no cookie cutter solution for everything so each UI element needs to be considered in its own context and the solution may vary accordingly. 

Although having a section like this would be more maintenance work for us as the standards around accessibility change over time.

b-kelly: I would absolutely consider having an "Accessibility" section on the component pages. We have one on the [modals page](https://stackoverflow.design/product/components/modals/#accessibility) already, so there's some prior art there. That's essentially what I was envisioning with the popovers section.

mukunku: Ooh I hadn't noticed that. Probably because I never had to add a modal component, lol. That's awesome. I'll try and look into adding a section like that for popovers when I can 👍🏼

b-kelly: No prob if you don't get to it - it isn't required for this PR to be merged.

lib/ts/controllers/s-popover.ts
b-kelly: Would this apply to [tooltips](https://stackoverflow.design/product/components/popovers/#hover-tooltips) as well? If so, we'll want to move it into the `BasePopoverController` class

b-kelly (changes requested):
Overall lgtm. I think the a11y toggling could also apply to tooltips as well, so we may want to move the code into the base controller. Side note: I'm ok with the `override` additions going in, but typically we'd prefer that any code changes that are completely unrelated to the main goal of the PR are added separately.

lib/ts/controllers/s-popover.ts
dancormier: ~~Yes, I believe this should apply to tooltips too. Good thinking @b-kelly. Should be pretty trivial to move to `BasePopoverController`.~~

dancormier: Oh, I take it back. [From MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role):

> Though a tooltip may appear and disappear, as its appearance is automatic and not intentionally controlled by the user, the [aria-expanded](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded) role is not supported.

b-kelly: Ok, in that case, let me change this to an approval. Thanks for the check on that!

b-kelly (approved):
I take back my comment regarding this being useful for tooltips. @dancormier points out [here](https://github.com/StackExchange/Stacks/pull/932#discussion_r868461913) that:

> [From MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role)
> > Though a tooltip may appear and disappear, as its appearance is automatic and not intentionally controlled by the user, the [aria-expanded](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded) role is not supported.

mukunku:
> I take back my comment regarding this being useful for tooltips. @dancormier points out [here](https://github.com/StackExchange/Stacks/pull/932#discussion_r868461913) that:
> 
> > [From MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role)
> > > Though a tooltip may appear and disappear, as its appearance is automatic and not intentionally controlled by the user, the [aria-expanded](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded) role is not supported.

I like the accessibility concerns they've laid out there, lol:
image

Thanks for the approval. Merging now. I'll be more careful not to change anything extra next time.

PR #939, opened 2022-05-11
fix: add back x/y margin negative classes

dancormier:
Thanks for reporting @CGuindon!

In the switch over to CSS custom properties, we neglected to account for `.mxn` and `.myn` atomic classes. Luckily/unluckily, these are used so sparingly that their omission went unnoticed for about a month.

Tangent: Should we consider adding a page/state where the popovers are exposed so we can use visual regression testing to detect issues like these?

### DIFF
--- a/lib/css/atomic/spacing.less
+++ b/lib/css/atomic/spacing.less
@@ -163,7 +163,17 @@ body {
 );
 
 #stacks-internals #build-classes(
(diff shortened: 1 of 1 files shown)


b-kelly (approved):
lgtm! Looks like these accidentally got stripped out [here](https://github.com/StackExchange/Stacks/pull/873/files#diff-709446d6a67158d13436f9987de123956ee2e87d566e798ad575066d4f8cec95L78). Due to the original implementation, it's easy to miss and subtle to detect among the other changes.

Thanks for the fix!

PR #956, opened 2022-05-26
Fix accessibility issues found by Mabl

dancormier:
I've addressed _most_ accessibility issues found by Mabl.

## Local Mabl run results

 Issue type | Before fixes | After fixes
-|-|-
critical | 15 | 6
serious | 7 | 1
moderate | 25 | 5
minor | 7 | 3

## Relevant links

- [MDN: Multiple \<h1\> elements on one page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#multiple_h1_elements_on_one_page)

- Context on radio elements using `aria-expanded`: https://github.com/w3c/aria/issues/1404

### DIFF
--- a/docs/product/components/avatars.html
+++ b/docs/product/components/avatars.html
@@ -154,39 +154,45 @@ description: Avatars are used to quickly identify users or teams.
 {% highlight html %}
 <a href="…" class="s-link s-link__muted">
     <div class="s-avatar">
-        <div class="s-avatar--letter">H</div>
+        <div class="s-avatar--letter" aria-hidden="true">H</div>
         @Svg.ShieldXSm.With("native s-avatar--badge")
     </div>
     <span class="pl4">Team name</span>
 </a>
 
 <a href="…" class="s-avatar s-avatar__24">
-    <div class="s-avatar--letter">H</div>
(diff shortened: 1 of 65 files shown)


docs/product/components/avatars.html
```diff
@@ -154,39 +154,45 @@
 {% highlight html %}
 <a href="…" class="s-link s-link__muted">
     <div class="s-avatar">
-        <div class="s-avatar--letter">H</div>
+        <div class="s-avatar--letter" aria-hidden="true">H</div>
```
b-kelly, line 157: I'm guessing these are hidden from screenreaders because they're "images" and therefore should not be read aloud?

b-kelly (approved):
lgtm, thank you for taking this on. There's some interesting stuff in here, but overall:

image

docs/product/components/avatars.html
dancormier: Yep, these letter avatars should include the team name wrapped in a node including `.v-visible-sr`.

dancormier: Thanks for taking a look btw!

PR #967, opened 2022-05-31
feat: add post summary micro variant

### LINKED ISSUE #966: Add "micro" variant of post summary
We currently have very minimal post summaries that don't use the Stacks post summary component. 

### Question page sidebar

image

### Profile Question/Answer sections

image

dancormier:
resolves #966

See https://deploy-preview-967--stacks.netlify.app/product/components/post-summary/#micro

## Questions

- **Should we truncate titles? If so, how many lines?** (This _could_ be handled in core as needed. Curious if we should prescribe truncation though. There's implications when it comes to vertically aligning elements within the post summary.)
- **What's the recommended title font size?** (Profile currently uses `body2`, sidebar currently uses `body1`)
- **Should items include borders?** (I'm assuming they can be added as needed by consumers but I'm not sure if we want to unify this.)
- **Should we include support for timestamps? Or any other optional data, for that matter?** (Profile currently includes timestamp)

## TODOs

- [ ] Update documentation

### Second draft

Below are screenshots with a couple variations (unanswered/short title/no date/no menu, longer title/short date/longest title/long date). There's more that _could_ exist in these post summaries.

image

image

<details>
<summary>First draft</summary>

image

</details>

### DIFF
--- a/docs/product/components/post-summary.html
+++ b/docs/product/components/post-summary.html
@@ -607,9 +607,6 @@ description: The post summary component summarizes various content and associate
                         <a href="#">Clicking through a viewpager2?</a>
                     </h3>
                 </div>
-                <a href="#" class="s-btn s-btn__muted s-post-summary--content-menu-button">
-                    {% icon "EllipsisVertical" %}
-                </a>
             </div>
             <div class="s-post-summary s-post-summary__micro">
                 <div class="s-post-summary--stats">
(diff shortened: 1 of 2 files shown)


paintedbicycle:
- I think body1 seems fine
- Yes, the component should handle truncation
- Do we need the `Q` and `A` icon back? And the Article and Collection icons?

paintedbicycle:
Looking at the profile page I think we need the icons, otherwise everything would look exactly the same. This could then also be used in places like the sidebar and in search like below

Screen Shot 2022-06-02 at 8 33 43 AM

dancormier:
I'm closing this PR since it will have many, _many_ conflicts once https://github.com/StackExchange/Stacks/pull/1091 is merged. We shall revisit post-PPCP refactor.

PR #975, opened 2022-06-06
chore: remove unnecessary Less var escaping

dancormier:
This is cruft as of like 2020. See https://lesscss.org/#operations-calc-exception

Requesting a review out of an abundance of caution, but really if everything looks good in staging, I'm feeling this is pretty safe.


b-kelly:
There's a slight possibility that this is a breaking change for downstream clients using an older version of Less. Not sure if we care so much there, but I'd double check that these changes work in Core before merging.

PR #981, opened 2022-06-13
fix: revert non-js banner changes from #707

dancormier:
This PR reverts some (probably unintentional) changes made in #707 from the banners docs page.

- add back the figma link
- revert class name changes to `grid` back to `flex`

---

From Stacks feedback received Mon Jun 13, 2022 9:26 AM:

> **New submission in stacks's feedback-banners**
> The example is broken: the items in the banner are in a column instead of a row. Also the style changers (with the "Important?" and "Pinned?" checkboxes) are in a column too. (Chrome 102.0.5005.63, on Windows 10 Pro 19044.1741).

### Before

image

### After

image

### DIFF
--- a/docs/product/components/banners.html
+++ b/docs/product/components/banners.html
@@ -1,6 +1,7 @@
 ---
 layout: page
 title: Banners
+figma: https://www.figma.com/file/9MG6BQ5X0FlHV0rQ3c9ugA/Banners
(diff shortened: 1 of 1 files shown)


b-kelly (approved):
Ouch! Yeah, definitely unintentional. Might have been a bad merge? Either way, lgtm and thanks for the fix!

PR #999, opened 2022-06-27
fix(a11y): apply focus style to button unset, link

### LINKED ISSUE #843: Revisit focus styling on `s-btn s-btn__link`
If `:focus-visible` were bulletproof on Safari, we could keep link-like behavior while enabling keyboard focus 🤔 

Right now, we strip focus styling from buttons that need to appear as links.

### LINKED ISSUE #996: Focus indicator missing on link style button
**Describe the bug**
There is no focus state indicator for buttons with the class `.s-btn__link`. Per WCAG guidelines, all focusable elements must have a visual focus indicator when in focus.

**To Reproduce**
1. Go to [documentation > Buttons > Reset](https://stackoverflow.design/product/components/buttons/#resets)
2. Using keyboard navigation only, tab to the "Link button" example.
3. See that there is no indication that the button is in focus

**Expected behavior**
The focus state of a button with `.s-btn__link` is the same as a default link (using an anchor tag).

**Screenshots**
The link's focus state:
Screen Shot 2022-06-27 at 5 11 46 PM

**Additional context**
In addition to in-line text use cases for a button that has been styled this way,`.s-btn__link` is also used in the header in Teams under the help menu for "Getting Started." All other items in that menu are anchor tags so behave as expected.

dancormier:
fixes #996
See it in action here: https://deploy-preview-999--stacks.netlify.app/product/components/buttons/#resets

### DIFF
--- a/lib/css/components/buttons.less
+++ b/lib/css/components/buttons.less
@@ -560,7 +560,7 @@
     .s-btn__unset:focus {
         padding: 0;
         border: none;
-        outline: none;
+        outline: initial;
         font: unset;
         border-radius: 0;
         color: unset;
@@ -575,7 +575,7 @@
         padding: 0;
         border: none;
         border-radius: 0;
-        outline: none;
+        outline: initial;
         font: inherit;
         background: none;
         box-shadow: none;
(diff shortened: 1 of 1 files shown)


dancormier:
FWIW, I tested this in Safari 15.5 and it works great. Although `:focus-visible` is supported only in Safari 15.4 and above, I think this is better than the nothing that we have prior to this PR.

PR #1001, opened 2022-06-30
docs(a11y): add accessibility scan button

dancormier:
This PR adds an "accessibility scan" button to the Stacks docs. We might want to instead just opt for an extension like the [axe DevTools extension](https://www.deque.com/axe/devtools/).

Btw, this is real rough with just a quick-once over to kinda have this as a proof of concept.

### DIFF
--- a/docs/_includes/layouts/page.html
+++ b/docs/_includes/layouts/page.html
@@ -100,14 +100,19 @@
             </div>
         </aside>
     </div>
-    <button class="s-btn s-btn__filled z-nav-fixed ps-fixed b12 r12 js-a11y-btn" role="button"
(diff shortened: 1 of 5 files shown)


allejo:
Just a drive-by thought here, if this PoC goes anywhere, why not bundle it as a separate package within the `@StackExchange` npm namespace and ship it? That way we're not shipping with axe-tools with Stacks but it's easy enough to pull in and replace the current implementation that exists in some places.

b-kelly:
@allejo I agree. This PoC was quickly thrown together at the request of an employee internally. I personally think that this is something that would be better suited for a Tampermonkey script

PR #1011, opened 2022-07-13
feat(components): add indeterminate checkbox state

KyleMit:
Allows for [.indeterminate](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes) checkbox property to be set to true by using [`:indeterminate`](https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate) selector

* Here's the [current indeterminate handling (doesn't exist)](https://codepen.io/KyleMit/pen/YzaGdWr)
* Here's an [example of the fix inlined into a pen](https://codepen.io/KyleMit/pen/XWEjoMX)
* Here's some [decoded svgs that I used to get the background image right](https://codepen.io/KyleMit/pen/yLKaGzL)
  * I started with using `<rect>` svg element, but that didn't inline nicely for some reason, so stuck to path, which is what we're using now

Also added new section to the docs:

image

### DIFF
--- a/docs/product/components/checkbox.html
+++ b/docs/product/components/checkbox.html
@@ -341,9 +341,6 @@ description: A checkable input that visually shows if an option is true or false
                     </div>
                 </div>
             </fieldset>
-            <script>
-                document.getElementById("indeterminate-checkbox-1").indeterminate = true;
-            </script>
         </div>
     </div>
 </section>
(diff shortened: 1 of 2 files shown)


dancormier (approved):
I spent like 15 minutes messing with that SVG to reduce aliasing before I realized it was different from the one in the PR 🤦

Anyways, this looks good! I made some tweaks to the documentation but nothing major.

Thanks for this!

PR #1012, opened 2022-07-13
Fix typescript errors when using Stacks package

mukunku:
# Summary

We are using the Stacks package in our solution in Visual Studio but there are these typescript compilation related errors that show up within the IDE:
image

I'm not sure if my proposed changes are the best way to address this but having those errors in my Visual Studio can be quite annoying and I haven't been able to figure out a way to hide/ignore them. This PR is my naïve attempt at trying to get rid of those errors.

If I make the proposed changes on my machine the errors go away: 
image
But this doesn't solve it for others and the issue would most likely reoccur if we were to update the package and the files are overwritten.

### DIFF
--- a/lib/ts/controllers/s-table.ts
+++ b/lib/ts/controllers/s-table.ts
@@ -142,7 +142,7 @@ function getCellSlot(cell: HTMLTableCellElement): number {
     if (!(cell.parentElement && cell.parentElement.parentElement instanceof HTMLTableSectionElement)) {
         throw "invalid table"
     }
-    const result = buildIndexOrGetCellSlot(cell.parentElement.parentElement, cell);
+    const result = buildIndexOrGetCellSlot(cell.parentElement.parentElement as any as HTMLTableSectionElement, cell);
     if (typeof result !== "number") {
         throw "shouldn't happen"
     }
--- a/lib/ts/controllers/s-tooltip.ts
+++ b/lib/ts/controllers/s-tooltip.ts
@@ -101,7 +101,7 @@ export class TooltipController extends BasePopoverController {
 
         var popover = document.getElementById(popoverId);
         if (!popover) {
-            popover = document.createElement("div");
+            popover = document.createElement("div") as any as HTMLElement;
             popover.id = popoverId;
             popover.className = "s-popover s-popover__tooltip pe-none";
             popover.setAttribute("aria-hidden", "true");


dancormier:
Thanks @mukunku! FYI, I opened a [PR with a similar goal](https://github.com/StackExchange/Stacks/pull/1009) recently. I'll take a look at your changes here soon.

lib/ts/controllers/s-table.ts
```diff
@@ -142,7 +142,7 @@ function getCellSlot(cell: HTMLTableCellElement): number {
     if (!(cell.parentElement && cell.parentElement.parentElement instanceof HTMLTableSectionElement)) {
         throw "invalid table"
     }
-    const result = buildIndexOrGetCellSlot(cell.parentElement.parentElement, cell);
+    const result = buildIndexOrGetCellSlot(cell.parentElement.parentElement as any as HTMLTableSectionElement, cell);
```
KyleMit, line 145: `any` is a bit of an early escape hatch here.  You can use a [type predicate/guard](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) to safely convince TS you are looking at the right type.  

We're supposedly already doing that above with this check:

```ts
if (!(cell.parentElement && cell.parentElement.parentElement instanceof HTMLTableSectionElement)) {
    throw "invalid table"
}
```

But perhaps TS remains unconvinced?

We can add this type guard:

```cs
function isHTMLTableSectionElement(el: HTMLElement | null | undefined): el is HTMLTableSectionElement {
    return el instanceof HTMLTableSectionElement;
}
```

You then rewrite like this:

```ts
function getCellSlot(cell: HTMLTableCellElement): number {
    const grandParent = cell.parentElement?.parentElement
    if (!isHTMLTableSectionElement(grandParent)) {
        throw "invalid table"
    }
    const result = buildIndexOrGetCellSlot(grandParent, cell);
    if (typeof result !== "number") {
        throw "shouldn't happen"
    }
    return result
}
```

allejo: > But perhaps TS remains unconvinced?

Yup, this is known TS behavior, see https://github.com/microsoft/TypeScript/issues/10421. If these objects/values are coming from built-in browser functions or we're positive there's no way the wrong type could exist here. Then we can create a utility (`assumeType<T>()`) to make it easy for us to fix this in the future,

```typescript
function assumeType<T>(x: unknown): asserts x is T {
    return; // ¯\_(ツ)_/¯
}

function processGeometry(obj: Geometry): void {
    if ("width" in obj) {
        assumeType<AARect>(obj);
        let width = obj.width;
        // ...
    }
    if ("halfDimX" in obj) {
        assumeType<AABox>(obj);
        let halfDimX = obj.halfDimX;
        // ...
    }
    else if ("radius" in obj) {
        assumeType<Circle>(obj);
        let radius = obj.radius;
        // ...
    }
    // And much more...
}
```

Source: https://github.com/microsoft/TypeScript/issues/10421#issuecomment-518806979

mukunku:
> Thanks @mukunku! FYI, I opened a [PR with a similar goal](https://github.com/StackExchange/Stacks/pull/1009) recently. I'll take a look at your changes here soon.

Thanks @dancormier.  Out of curiosity I tried using the versions of those two files you have in your PR to see if the error was already fixed however the errors still persisted on my end. I'm happy to close this PR out if you'd rather add it to your changes. Let me know!

lib/ts/controllers/s-table.ts
mukunku: The double cast to any then to the element is definitely an ugly band-aid. 

I tried the type guard approach but I'm still getting some errors from the typescript compiler:
image

The assume type approach also didn't work for me. Not sure if I'm doing something incorrectly:
image

dancormier:
> Out of curiosity I tried using the versions of those two files you have in your PR to see if the error was already fixed however the errors still persisted on my end. 

Ah, I'm a fool! `s-table.ts` in my PR has no type-related fixes. 🤦 I was operating under the (incorrect) assumption that all errors would surface in Stacks standalone. That being said, I don't get any errors on `.s-tooltip.ts` in Stacks _or_ Core 🤔

> I'm happy to close this PR out if you'd rather add it to your changes. Let me know!

If the errors are resolved here, this PR should be good to merge in and #1009 can be merged separately.

I'm gonna poke around at that [`s-table.ts`](https://github.com/StackExchange/Stacks/pull/1012/files/4440de59cbc8cea93669cacb8aa1c5121b354d03#diff-be89c5194c8d7ac58409a18dd9aaa6b184f17354f91936a1a67754409213d552) fix in ~20 minutes and report back.

lib/ts/controllers/s-table.ts
KyleMit: Huh, I can't repro your error @mukunku, either in my editor or in this [TS Playground Demo](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAcwKZQMKoDbYMrZxQAUEO2AXIgBIAqAsgDK0CGARtqlrgKKcC2qMFACUVMCH5tUAJ0QBvALAAoRGsQQEAZygoZLMABMACixlDdAXg3kAdAAczFvqkHCA-A6fCXbqCvVEGGBEYgBCGC06JlYOVDxUaHgwXwtiZH0jU3NhEREFAMD1KAALGTgAd0QAIhgwADcWbBhDRCh2TmrCtQBfbo1tXXMtEGwrRDYQGGxDAEkjVAAPAHkZAHF0bnxCEgyDE28oABobXBEAbn7g0KgAT3tUOBDh0d0wy2tqiSlZavylVRFNSlcpVapaEpwUaGMAAcl0JRY9geYC6gN6-XMUBAMiQLzGKj6ylAkFgCCCUQYzA68USZJSAjSOCo0UYqWEiAAPogJLguYhwIZUMA6qhDGJEDgKTQqbFOAkkgh2boAYEsTikFK6joDGQnjKYjSFfTlZdlESVCTFUhJtM5gsVutNuQCERiFo6ckWbKjZ6lYzhCcRUYtu5vYa4ltlf9+powFo4JxbIRkMR5B7rUdg4Ytj0LoSgA)

If I control click on `HTMLTableSectionElement ` it provides the following source code:

```ts
/** Provides special properties and methods (beyond the HTMLElement interface 
  it also has available to it by inheritance) for manipulating the layout and 
  presentation of sections, that is headers, footers and bodies, in an HTML table. 
*/
interface HTMLTableSectionElement extends HTMLElement {}
```

So it does seem fair that it should be assignable in a predicate.

But I believe the squiggles in your screenshot; I just wonder where they're coming from.

dancormier:
Closing in favor of https://github.com/StackEng/StackOverflow/pull/12361

PR #1024, opened 2022-07-20
fix(button): contrast, focus state fixes

dancormier:
This PR fixes two accessibility issues with `.s-btn` in dark mode:

<details>
<summary>Increases color contrast for primary button</summary>

### Before
Contrast: 3.39:1
image

### After
Contrast: 6.15:1
image

</details>

- Adds focus outline (as box-shadow) to any buttons missing it


shainanigans (approved):
LGTM :) I was a little unsure about whether our focus ring is sufficient, but the docs say if it is at least 2px thick it's fine, so seems like we're good!

PR #1025, opened 2022-07-20
modify ignored post summary selectors to be more specific

mukunku:
# Summary

This is to address a visual glitch due to some changes I'm making within the StackOverflow project.

Right now the opacity and grey filters are applied to all children of ignored post summaries. Which means if I add a popover within the post summary via javascript it too gets its opacity and filter modified. It also throws off the coordinates of the absolute positioning:

My PR that isn't working correctly right now: https://github.com/StackEng/StackOverflow/pull/12396
Codepen of suggested fix: https://codepen.io/mukunku/pen/RwMVorL

# Testing

I compared the Stacks product pages for the post summary component before and after my change and it looks like it is working okay:

After this change my popovers are rendering correctly:

### DIFF
--- a/lib/css/components/post-summary.less
+++ b/lib/css/components/post-summary.less
@@ -403,10 +403,9 @@
 
 .s-post-summary__ignored,
 .s-post-summary__deleted {
-    .s-post-summary--content > *:not(.s-post-summary--content-menu-button) {
+    .s-post-summary--content > *:not(.s-post-summary--content-menu-button):not(.s-post-summary--meta) {
         opacity: 0.6;
     }
-
     // TODO DEPRECATED remove `:not(.is-deleted)`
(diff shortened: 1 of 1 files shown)


dancormier (approved):
LGTM! Thanks @mukunku!

lib/css/components/post-summary.less
```diff
@@ -403,10 +403,9 @@
 
 .s-post-summary__ignored,
 .s-post-summary__deleted {
-    .s-post-summary--content > *:not(.s-post-summary--content-menu-button) {
+    .s-post-summary--content > *:not(.s-post-summary--content-menu-button):not(.s-post-summary--meta) {
```
dancormier, line 406: I see a future where this has like 10 `:not()`s. That's:not(good).

I've captured this issue in my audit doc.

PR #1026, opened 2022-07-20
chore(post-summary): remove legacy variant

dancormier:
AFAICT, we no longer use the legacy post summary format in production. We _could_ deprecate legacy, but I think we should just delete it and save a few bytes. What do ya say?


b-kelly (approved):
Agreed. I was thinking about this literally just yesterday.

PR #1038, opened 2022-07-25
refactor(button): style with pseudo-private custom properties

dancormier:
This PR refactors the CSS for `.s-btn` and `.s-btn-group` to use pseudo-private custom properties.

> **Note** see also: https://github.com/StackExchange/Stacks/pull/1135

### `.s-btn-group` using `input[type="radio"]`
This PR includes [changes to `.s-btn-group` to support implementation with `input[type="radio"]`](https://github.com/StackExchange/Stacks/pull/1135). Normally, I'd separate these into unique PRs, but this and `.s-btn` are pretty tightly coupled, so I've included the change here.

## Bugs fixes

I noticed a few issues that I resolved by preventing CSS rule conflicts.

### High contrast primary button active state

In `develop`, the high contrast primary button style doesn't change on `:active`. This PR resolves that.

### High contrast borders

Some buttons variants had inconsistent borders in high contrast ([Primary style selected state](https://stackoverflow.design/product/components/buttons/#primary), for example). This PR resolves those issues. 

## TODO

- [x] Rename PPCP to match format (`--_btn` should become `--_bt`)
- [x] Fix hover styling on selected button within `.s-btn-group` using `input[type="radio"]`


docs/_data/button-groups.json
```diff
@@ -6,9 +6,9 @@
       "description": "Base button group style."
     },
     {
-      "class": ".s-btn-group--container",
```
b-kelly, line 9: Looks like this is a breaking change. How should we handle this?

lib/css/components/buttons.less
```diff
@@ -1,634 +1,464 @@
...
-        //      This means we need to off set the margin y-axis so we don't add
-        //      additional height to the button.
-        //  ------------------------------------------------------------------------
-        .svg-icon {
-            vertical-align: baseline;
-            margin-top: -0.3em;         // [1]
-            margin-bottom: -0.3em;      // [1]
-            transition: opacity 200ms var(--te-smooth); // Animate the transition to .is-loading
-        }
+    // .s-btn-group &--radio:checked + &,
```
b-kelly: Was this left in intentionally? If so, add a TODO comment w/ a GH issue number

lib/css/components/buttons.less
```diff
@@ -1,634 +1,464 @@
...
+    --_bu-p: 0.8em;
+    // STATE AND INTERACTION CUSTOM PROPERTIES
+    --_bu-bc-selected: transparent;
+    --_bu-bg-active: var(--theme-button-active-background-color);
+    --_bu-bg-hover: var(--theme-button-hover-background-color);
+    --_bu-bg-selected: var(--theme-button-selected-background-color);
+    --_bu-fc-active: var(--theme-button-hover-color); // Note: hover color used here intentionally
+    --_bu-fc-hover: var(--theme-button-hover-color);
+    --_bu-fc-selected: var(--theme-button-selected-color);
+    // Filled
```
b-kelly, line 20: This is fine for now, but sometime after merge I'd like to see if we can't deduplicate these variables (i.e. remove the "filled/outlined/etc" and override the base variables instead) in a follow up PR. This approach will result in some slight selector duplication under e.g. `--danger`, but I think the overall readability will improve

b-kelly (changes requested):
Excellent work! Visually, this PR doesn't appear to introduce any regressions (tested all combos between light/dark, standard/custom theme, highcontrast on/off). Code-wise, the organization of this file has been greatly improved (dropping ~200 lines of duplicated selectors/properties in the process).

I do have a few easily addressed concerns:

1. Looks like the removal of the documented `s-btn-group--container` class is a breaking change. I'd prefer to avoid this if at all possible. Any chance we can put a shim in here with a deprecation note in the changelog?
2. Radio based buttons don't have focus rings. Since this syntax is new to this PR, we could do one of the following:
   * just add the checked selectors to the focused selectors, much like you did for `.is-selected`
   * change the markup to have the input _inside_ of the label, then drop the special selector casing for `:focus-within` (in addition to `:focus`)
3. I also spotted an existing bug you can fix while you're in here: Primary buttons don't change their BG color when hovering over them in _dark mode_

lib/css/components/buttons.less
dancormier: Not intentional! I'll remove it

lib/css/components/buttons.less
dancormier: Totally agree. In fact, I attempted doing away with the "filled/outlined/etc" variables when revisiting this PR recently, but I found it too complex to tackle here. I'd love to revisit in the future.

dancormier:
> Excellent work! Visually, this PR doesn't appear to introduce any regressions (tested all combos between light/dark, standard/custom theme, highcontrast on/off). Code-wise, the organization of this file has been greatly improved (dropping ~200 lines of duplicated selectors/properties in the process).

🎉 

> I do have a few easily addressed concerns:
> 
>     1. Looks like the removal of the documented `s-btn-group--container` class is a breaking change. I'd prefer to avoid this if at all possible. Any chance we can put a shim in here with a deprecation note in the changelog?

Yeah, I thought that it could be deprecated and any relevant CSS could be removed, but I guess I didn't compensate for this properly (bottom row includes `form.s-btn-group--container`):

image

I'll have to reconsider removing this class before merging this PR.

>     2. Radio based buttons don't have focus rings. Since this syntax is new to this PR, we could do one of the following:
>        
>        * just add the checked selectors to the focused selectors, much like you did for `.is-selected`
>        * change the markup to have the input _inside_ of the label, then drop the special selector casing for `:focus-within` (in addition to `:focus`)

Fixed in https://github.com/StackExchange/Stacks/pull/1038/commits/e946105efad6abb90de5368385556c9ffc415a01

>     3. I also spotted an existing bug you can fix while you're in here: Primary buttons don't change their BG color when hovering over them in _dark mode_

Ah, nice catch! I'll fix that. Update: fixed in [e519ab4](https://github.com/StackExchange/Stacks/pull/1038/commits/e519ab470d499d74b7f712c4cc3df8e11cd9bf11)

docs/_data/button-groups.json
dancormier: So, I originally intended this to be a non-breaking change, but a deprecation. In 77c425c9c1f68cb8d06015f89a58ba180b22ee6d, I've targeted the `form` element and once again made this a deprecation. Basically with the PR as it is now, the changes should be backwards compatible. 



image

> **Note**
> Image for demo purposes only. Don't worry. I didn't change the docs to this permanently.

Should we reintroduce `.s-btn-group--container`? I know targeting a HTML tag feels a little icky but it doesn't feel totally wrong here. That said, I'm fine with targeting that class and punting on any of this and FWIW it's pretty low stakes since I only found it used in [one place in Core](https://github.com/StackEng/StackOverflow/blob/master/StackOverflow/Views/Channels/Admin/UserManagement/_UserSyncSettings.cshtml#L79-L88).

---

PS: `.s-btn-group--container` previously just set any nested button border radius to 0. That means if the button was at the start or end of the button group, it would get squared off. I've addressed this in 77c425c9c1f68cb8d06015f89a58ba180b22ee6d.

b-kelly (approved):
lgtm, thanks for the updates

PR #1039, opened 2022-07-26
refactor(color): adjust high contrast colors

dancormier:
@allejo mentioned that `red-050` is indistinct from black in dark mode (and from white in light mode). This PR adjust lower color stops to provide adequate contrast on similar monochromatic colors.

<details>
<summary>Before</summary>
image
</details>

<details>
<summary>After</summary>
image
</details>


allejo (approved):
I only tested the `-050` stops and am assuming the higher stops are okay. But this fixed the colors for me, they're more visible now in HC 😄

PR #1040, opened 2022-07-26
Pull in classes from Stacks-Editor

b-kelly:
This PR adds in a new `is-readonly` input class and abstracts out the button `is-loading` class to be used anywhere. I've added examples / docs for both items, as well as tested in both dark and highcontrast modes.

Ideally, this would be multiple PRs since they're individual features, but we can merge w/o squashing if you'd like each commit to appear in the changelog.

### DIFF
--- a/docs/_data/spinner.json
+++ b/docs/_data/spinner.json
@@ -27,8 +27,8 @@
     },
     {
       "class": ".is-loading",
-      "applies": "Any element",
-      "description": "Prepends a spinner pseudo-element to the element"
+      "applies": "Any text-based elements",
+      "description": "Prepends a spinner pseudo-element to the element. Prefer using the spinner component when possible."
     }
   ]
 }
--- a/docs/_data/spinner.json
+++ b/docs/_data/spinner.json
@@ -24,6 +24,11 @@
       "class": ".s-spinner__lg",
       "applies": ".s-spinner",
       "description": "A large style for the largest layouts"
+    },
+    {
+      "class": ".is-loading",
+      "applies": "Any element",
+      "description": "Prepends a spinner pseudo-element to the element"
     }
   ]
 }
--- a/lib/css/components/inputs.less
+++ b/lib/css/components/inputs.less
@@ -469,6 +469,9 @@ fieldset {
     }
 
     //  Readonly
+    //  Note: The readonly attribute is not supported on select elements
+    //  See https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly
+    //  and https://github.com/StackExchange/Stacks/pull/1040#discussion_r931144086
     &[readonly],
     .is-readonly & {
         border-color: var(--bc-light);
--- a/lib/css/components/inputs.less
+++ b/lib/css/components/inputs.less
@@ -462,21 +462,23 @@ fieldset {
 .s-input,
 .s-textarea,
 .s-select > select {
-    //  Disabled, Read-only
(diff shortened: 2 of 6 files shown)


lib/css/components/inputs.less
```diff
@@ -462,21 +462,23 @@ fieldset {
 .s-input,
 .s-textarea,
 .s-select > select {
```
dancormier, line 464: > **Note**
> The `readonly` attribute does not apply to `select` elements. This style was pre-existing and won't hurt anything to include, but I was a little confused, so I'm making a note.

From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly):
> The attribute is not supported or relevant to [`<select>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select) 

The [section detailing the differences between `readonly` and `disabled`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly#attribute_interactions) is worth a read too:
> The difference between [disabled](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled) and readonly is that read-only controls can still function and are still focusable, whereas disabled controls can not receive focus and are not submitted with the form and generally do not function as controls until they are enabled.

docs/_data/spinner.json
```diff
@@ -24,6 +24,11 @@
       "class": ".s-spinner__lg",
       "applies": ".s-spinner",
       "description": "A large style for the largest layouts"
+    },
+    {
+      "class": ".is-loading",
+      "applies": "Any element",
```
dancormier: I tried this on a bunch of miscellaneous components and it works _to varying degrees_. If an element already has pseudo-elements, this class is likely to overwrite (or at least append) them. I found this with `.s-select`. In many other cases, the loading spinner isn't quite aligned the way you'd expect.

Anyone applying this class will need to use it thoughtfully and I can see a future where someone comes to the Stacks team in need of a component-specific application. That being said, I think this is progress.

dancormier (approved):
I left a couple of notes that require no action. This is 🦐 status. Thanks!

> Ideally, this would be multiple PRs since they're individual features, but we can merge w/o squashing if you'd like each commit to appear in the changelog.

Merge w/o squash would be nice but I'm [not sure GitHub will allow that](https://github.com/community/community/discussions/10809), so a single commit is fine in this case if it's trouble.

lib/css/components/inputs.less
b-kelly: I can also drop a note in a comment in the file. I did test the readonly styles on selects (because why not), but they looked very "meh", so I dropped them from the `.is-readonly` rule below since they aren't technically supported anyways. Good callout

docs/_data/spinner.json
b-kelly: Hmm... that's a good call. I'll add some notes to the docs. Instead of "any element" maybe something like "any text display element"? That's pretty confusing though. I'll also drop a note to prefer using the spinner markup if possible.

Alternatively, we could leave this relatively undocumented and just officially support the class on `s-btn`, but I'd like to avoid that scenario if we can.

dancormier: I updated the documentation in https://github.com/StackExchange/Stacks/pull/1040/commits/ac44000f502cdd247d342c7d86bde920fad81528. I'm not 100% satisfied with the wording, but I don't think we need to spend too much more time on it.

PR #752, opened 2021-09-09
High Contrast Theme: Part II

dancormier:
I've opened a new PR to merge into w3c/wcag#556 to 1) make sure what I'm doing makes sense before I completely blow up w3c/wcag#556 and 2) be able to easily compare and _contrast_ with deploy previews.

PR #1047, opened 2022-08-02
docs(a11y): add accessibility guidelines page

dancormier:
I think our documentation could benefit from a standalone accessibility-focused page. Consider this a big ol' work in progress atm.

### TODO
- [ ] Complete "Overview" section
- [ ] Complete "Accessibility testing" section
- [ ] Complete "Structural markup" section
- [ ] Complete "Using WAI-ARIA" section
    - [ ] Change this heading to be more friendly to those not familiar with the world of a11y
- [ ] Complete "Color contrast" section
- [ ] Complete "Additonal resources" section
- [ ] Consider cross-linking opportunities
- [ ] Consider level of detail per section necessary

cc @CGuindon @shainanigans

### DIFF
--- a/docs/product/guidelines/accessibility.html
+++ b/docs/product/guidelines/accessibility.html
@@ -1,0 +1,38 @@
+---
+layout: page
+title: Accessibility
+description: A non-comprehensive guide to accessibility best practices when using Stacks.
+---
+
+<section class="stacks-section">
+    {% header "h2", "Overview" %}
+    <!-- TODO: Mention dry stuff - trying to meet WCAG 2.1 -->
+    <!-- TODO: State Stacks' goals/aspirations wrt a11y broadly. Think inclusive design principles https://inclusivedesignprinciples.org/ -->
+    <!-- TODO: Mention constant "work in progress" nature, "no such thing as 100% accessible" -->
+</section>
+
+<section class="stacks-section">
+    {% header "h2", "Accessibility testing" %}
+    <!-- TODO: Comment on how we test for accessibility. Mention mabl, dev tools, extensions, assistive technologies. -->
+</section>
+
+<section class="stacks-section">
+    {% header "h2", "Structural markup" %}
+    <!-- TODO: Comment on the benefits of using semantically correct structural markup -->
+</section>
+
+<section class="stacks-section">
+    {% header "h2", "Using WAI-ARIA" %}
+    <!-- TODO: Document best practices for using WAI-ARIA, mention visually hidden content (link to https://stackoverflow.design/product/base/visibility/#classes). Consider recommending an approach for aria-label vs visually hidden content. -->
+</section>
+
+<section class="stacks-section">
+    {% header "h2", "Color contrast" %}
+    <!-- TODO: Mention (and link to) high contrast and dark modes. Mention the benefits wrt accessibility. -->
+    <!-- TODO: Touch on limitations of WCAG 2.1 in this context. Maybe give a nod to WCAG 3.0 standards. -->
+</section>
+
+<section class="stacks-section">
+    {% header "h2", "Additonal resources" %}
+    <!-- TODO: Link to informative resources. Bootstrap had a pretty good list https://getbootstrap.com/docs/5.2/getting-started/accessibility/#additional-resources -->
+</section>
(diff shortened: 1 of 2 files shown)


docs/product/guidelines/accessibility.html
```diff
@@ -0,0 +1,38 @@
...
+</section>
+
+<section class="stacks-section">
+    {% header "h2", "Accessibility testing" %}
+    <!-- TODO: Comment on how we test for accessibility. Mention mabl, dev tools, extensions, assistive technologies. -->
+</section>
+
+<section class="stacks-section">
+    {% header "h2", "Structural markup" %}
+    <!-- TODO: Comment on the benefits of using semantically correct structural markup -->
```
dancormier: Avoid value judgements in favor of an implied benefit.

```suggestion
    <!-- TODO: Comment on the benefits of using semantically precise structural markup -->
```

docs/product/guidelines/accessibility.html
```diff
@@ -0,0 +1,38 @@
...
+
+<section class="stacks-section">
+    {% header "h2", "Using WAI-ARIA" %}
+    <!-- TODO: Document best practices for using WAI-ARIA, mention visually hidden content (link to https://stackoverflow.design/product/base/visibility/#classes). Consider recommending an approach for aria-label vs visually hidden content. -->
+</section>
+
+<section class="stacks-section">
+    {% header "h2", "Color contrast" %}
+    <!-- TODO: Mention (and link to) high contrast and dark modes. Mention the benefits wrt accessibility. -->
+    <!-- TODO: Touch on limitations of WCAG 2.1 in this context. Maybe give a nod to WCAG 3.0 standards. -->
```
dancormier, line 69: See https://github.com/StackExchange/Stacks/pull/752#issuecomment-918652311 and https://github.com/StackExchange/Stacks/pull/752#issuecomment-923601203

shainanigans:
Looks good @dancormier ! 

If you're having a section on color contrast, would it also make sense to cover some other broad accessibility categories of things to look out for? For example: 

- Screen reader accessibility
- Focus and keyboard navigation / interactive components
- Language (don't use directional instructions because screen readers don't behave that way, don't use action-descriptive language like "click" or "hover" because it excludes keyboard users, etc)

Some of these ideas I've stolen from [Boostrap's](https://getbootstrap.com/docs/4.0/getting-started/accessibility/) and [Material's](https://material.io/archive/guidelines/usability/accessibility.html) guides if you want to check them out and haven't already :)

Will structural markup also include semantic elements, or is that a different category?

dancormier:
> Looks good @dancormier !

Thank you! And thanks for giving it a look!

> 
> If you're having a section on color contrast, would it also make sense to cover some other broad accessibility categories of things to look out for? For example:
> 
>     * Screen reader accessibility

Can you give me a rough idea of what we'd mention in this section?

> 
>     * Focus and keyboard navigation / interactive components

Yes, this section should be added. Good call!

> 
>     * Language (don't use directional instructions because screen readers don't behave that way, don't use action-descriptive language like "click" or "hover" because it excludes keyboard users, etc)
> 

I can go either way on language guidelines being included in this page. I'm trying to keep this page scoped to the technical approaches to accessibility but I see it worth while to mention language _somewhere_. Maybe as a "[Content](https://stackoverflow.design/content/guidelines/principles/)" section?

> 
> Some of these ideas I've stolen from [Boostrap's](https://getbootstrap.com/docs/4.0/getting-started/accessibility/) and [Material's](https://material.io/archive/guidelines/usability/accessibility.html) guides if you want to check them out and haven't already :)

I've poked into a few other design systems' accessibility-focused pages but only really skimmed them. There's for sure more opportunity to lift some inspiration.

> 
> Will structural markup also include semantic elements, or is that a different category?

Yeah, "structural markup" is probably the wrong heading label. I imagine it'll be a section broadly outlining how markup should be written for the greatest accessibility.

shainanigans:
> Screen reader accessibility

Maybe some of this doesn't make sense for Stacks documentation, but I'm not sure where else we might want to put it. 
- Aria roles like `aria-hidden`, `aria-expanded`, `aria-live`... possibly more that aren't coming to mind :) Unless this is already covered by your WAI-ARIA section?
- How the order of elements in DOM is how a screen reader reads content (regardless of CSS styling to make it different)
- Link to the screen reader visibility class
- Writing good aria labels or screen reader text (this could be the language part I mentioned)
- Hiding visually hidden elements from screen readers too
- Heading order, with links to the classes to visually change the style of headings as needed

dancormier:
Closing in favor of https://github.com/StackExchange/Stacks/pull/1670

PR #1086, opened 2022-08-29
Post summary accessibility updates

ellenchanhy:
Updating the component to meet color contrast requirements.

Updated colors as per [Figma](https://www.figma.com/file/XEKb1HjXVQOgLbqtLvW7i7/Post-summary-updates?node-id=37%3A8312)

- Updated green color for answers from var(--green-500) to var(--green-700)
- Updated hotness colors for post summary stats 
-  Added 4 new colors for var(--red-050) for light, dark, high contrast light, high contrast dark modes (for deleted style)
- Updated deleted post summary styles
- Updated ignored post summary styles 
- Updated watched post summary styles


lib/css/components/post-summary.less
```diff
@@ -309,38 +305,94 @@
...
 
-    .is-deleted {
-        color: @white;
-        background-color: var(--red-500);
+        &:hover,
+        &:active,
+        &:visited {
+            color: var(--blue-700);
+        }
+    }
```
dancormier: Did you intend for the watched post summary title anchor to have the same color regardless of state?

lib/css/components/post-summary.less
```diff
@@ -309,38 +305,94 @@
...
+    .is-deleted {
+        color: var(--white);
+        background-color: var(--red-600);
     }
 
-    .s-post-summary--content-title a {
-        color: var(--black-600);
+    .s-post-summary--meta-tags > a {
+        background-color: var(--black-075);
+    }
```
dancormier: The interactive states of tags within deleted post summaries are still blue. Is this as expected?

Hovered tag screenshot:
image

lib/css/components/post-summary.less
```diff
@@ -309,38 +305,94 @@
...
+        }
+
+        .s-post-summary--stats-item__emphasized {
             color: var(--black-500);
         }
+    }
+
+    .s-post-summary--meta-tags > a {
+        background-color: var(--black-050);
+    }
```
dancormier: The interactive states of tags within ignored post summaries are still blue. Is this as expected?

Hovered tag screenshot:
image

dancormier (commented):
Thanks for popping into the Less and making these changes! They look good overall, I just have a few details I wanted to mention and make sure they're as expected. 

BTW, I made some commits that are cleanup and documentation based. They haven't affected the styles you've set.

lib/css/components/post-summary.less
ellenchanhy: It looks like active & hover are the same, and can be the same color as the title link color which is blue-700. I checked to make sure that's ok for accessibility, as long as the hover state has the icon. 

For visited though, that can probably go one darker to blue-800 to follow what visited states look like on other post summary styles.

lib/css/components/post-summary.less
ellenchanhy: Thanks for catching this! It should not be blue, it should be one darker at background black-075.

lib/css/components/post-summary.less
ellenchanhy: Thanks for catching this! It should not be blue, it should be one darker at background black-075.

philBrown:
Not tested in StackOverflow dark or high-contrast modes and ruins theme link colours for watched tags.

This is not a good change at all.

PR #1087, opened 2022-08-29
fix(topbar): add required `nav` element around navigational `<ol>`s

dancormier:
Navigational elements _must_ be wrapped in `nav` elements (or include `aria` roles to indicate they're navigational).

This PR wraps topbar nav element groups (`<ol>`) in `<nav>`s, adds `.s-topbar--navigation` for styling, and updates the docs/examples as necessary.

### DIFF
--- a/docs/_data/topbar.json
+++ b/docs/_data/topbar.json
@@ -46,7 +46,7 @@
     {
         "class": ".s-topbar--navigation",
         "description": "Apply to `nav` to ensure proper layout",
-        "applies": "Child of `.s-topbar`"
+        "applies": "Parent of `.s-topbar--container`"
     },
     {
         "class": ".s-topbar--notice",
--- a/docs/_data/topbar.json
+++ b/docs/_data/topbar.json
@@ -44,6 +44,11 @@
         "applies": "`.s-topbar--item`"
     },
     {
+        "class": ".s-topbar--navigation",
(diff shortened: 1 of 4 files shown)


shanefoster:
@dancormier Should we also include [role="navigation"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role) or does the aria-label essentially handle that? I am not suggesting we do. This is a legitimate question for my own education :)

dancormier:
> @dancormier Should we also include [role="navigation"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role) or does the aria-label essentially handle that?

Close! The _HTML element `<nav>`_ handles that. All `<nav>` elements have an implicit role of `navigation` which cannot be overridden. From [MDN's page on `<nav>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav): 

> **Implicit ARIA role**: [navigation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role)
> **Permitted ARIA roles**: No `role` permitted

shanefoster:
Ahh, that makes sense. That would be rather redundant.

shanefoster (approved):
👍🏻

PR #1088, opened 2022-08-29
fix(table): ensure disabled rows meet contrast minimums

dancormier:
> Inactive rows do not pass. Should they? Let’s discuss and potentially solve.

`.s-table` rows that have `.is-disabled` need to meet WCAG contrast minimums*. 

The closest exception would be [**incidental** UI elements](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum#dfn-user-interface-component):

> Text or images of text that are part of an inactive [user interface component](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum#dfn-user-interface-component), that are [pure decoration](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum#dfn-pure-decoration), that are not visible to anyone, or that are part of a picture that contains significant other visual content, have no contrast requirement.

<details>
<summary>Screenshot (before)</summary>

</details>

<details>
<summary>Screenshot (after)</summary>

</details>

### TODO
- [ ] Include dark mode background color
- [ ] Add a `aria` attribute to examples to indicate state

> **Note**
> *I'm still finding supporting evidence that disabled table rows aren't exempt

### DIFF
--- a/lib/css/components/tables.less
+++ b/lib/css/components/tables.less
@@ -307,11 +307,24 @@
     //  $$  Disabled rows
     //  ------------------------------------------------------------------------
     tr.is-disabled {
-        background-color: var(--black-025);
+        position: relative;
+
+        &:after {
+            content: '';
+            inset: 0;
+            position: absolute;
+            z-index: -1;
+        }
 
         th:not(.is-enabled),
         td:not(.is-enabled) {
-            opacity: calc(var(--_o-disabled) * 0.6); // 0.5 * 0.6 = 0.3
+            color: var(--fc-light);
+            text-decoration: line-through;
+        }
+    }
+    &:not(.s-table__stripes) {
+        tr.is-disabled:after {
+            background-color: var(--black-025);
         }
     }
 }


ondrejkonec (changes requested):
Hi there,

I'm Ondřej and I'm designer and developer from Prague, Czech Republic 👋. 

I really love **CSS** and **accessible HTML** and systems and I would like to try to help as a contributor, if you will find it helpful.

My option on this is, that disabled element is only decorative and having lower contrast is desirable (as you mentioned).

According to [W3C](https://www.w3.org/TR/WCAG21/#contrast-minimum), this is exception. 

From a design perspective, if the element is disabled and the user cannot interact with this element, it does not need to be accessible (because element is inaccessible for user).

There is another thing too: Strikethrough text is primarily used to mark text that is **mistaken**, to be **removed**, **no longer relevant** or **no longer accurate**. Therefore, it should not replace the disabled state.

My request is to keep only background color for disabled tables. What do you think?

dancormier:
Hi @ondrejkonec and thanks for the feedback!

This PR was something I slapped together to demonstrate an idea but I think my idea is a little half-baked. I'm going to delegate figuring out this issue to our designers and make sure your feedback is communicated.

PR #1090, opened 2022-09-02
docs: button group docs mention the wrong aria attribute in one place

balpha:
The rest of this section correctly says `aria-current` already.


dancormier (approved):
Nice catch! Thanks!

PR #1091, opened 2022-09-02
refactor(post-summary): use pseudo-private custom properties

dancormier:


lib/css/exports/constants-colors.less
```diff
@@ -197,6 +197,10 @@
     --theme-link-color: var(--theme-secondary-400);
     --theme-link-color-hover: var(--theme-secondary-350);
     --theme-link-color-visited: var(--theme-secondary-500);
+    // Below variables set as theme override in consumers
+    // --theme-link-color-dark: undefined
+    // --theme-link-color-dark-hover: undefined
+    // --theme-link-color-dark-visited: undefined
```
b-kelly: I'd rather not set a precedence for this. If consumers want to override specific variables for specific modes, then they can do so with our mode mixins.

For example:

```less
.dark-mode({
--theme-link-color: red;
})
```

dancormier: Yeah, that sounds reasonable. I'm trying (and failing) to recall why I went about it how I did. If I can remember and it's worth considering, I'll mention it here. Otherwise, I'm good to remove these dark-specific custom properties.

dancormier: Ah, my memory came through!

 I think we have a naming issue here. These variables have nothing to do with the _dark_ theme specifically. They're intended to define a color to use when links are rendered on a non-standard background (think watched or deleted post summaries).

I'd still be open to reconsidering this approach altogether, but the surface issue here is my use of the term `dark`. Maybe a better alternative would be:

```
    // Below variables set as theme override in consumers
    // --theme-link-color-emphasised: undefined
    // --theme-link-color-emphasised-hover: undefined
    // --theme-link-color-emphasised-visited: undefined
```

giamir:
Hello @dancormier and @b-kelly,
I spent some time doing a high level review of this PR (with the limited context I have).

Here are some comments/questions/ideas:
- What do we hope to achieve by introducing "pseudo-private custom properties"? My guess here is that we hope they will make the css codebase more readable/maintainable. Are there any other reasons I miss (e.g. do we expect consumers to leverage the dynamic nature of CSS custom properties - changing values at runtime)?
- If the main goal is to make things more readable, maintainable and organised I have 2 suggestions:
   - If we don't need any dynamic behavior from these properties, what about converting them to preprocessor static variables instead (as recommended [here](https://www.smashingmagazine.com/2018/05/css-custom-properties-strategy-guide/#local-static-variables-are-ok-sometimes))? Such approach would also prevent consumers to change the value of the properties at runtime reinforcing their "private" nature  
   - What would you think about declaring all the variables at the top of the CSS file (similarly to what the author describes here as [the logic fold](https://www.smashingmagazine.com/2018/05/css-custom-properties-strategy-guide/#the-logic-fold))? This might establish a stronger pattern to recognise across the codebase (all the files starting with variable declarations).
 - About the naming "pseudo-private custom properties" - what "pseudo" represents? If my thinking above is correct and there are no other reasons to introduce those variables apart from maintainability what about calling them just "local variables"?
 - Is there already a way in Stacks to keep track about why certain architecture decisions were taken? If not, it could be beneficial to start [lightweight architecture decision records](https://github.com/peter-evans/lightweight-architecture-decision-records#lightweight-architecture-decision-records). I have used similar tools in the past: they are relatively inexpensive and they hold important context for new contributors but also for established ones. I'd be happy to give an introduction to ADRs if you don't know them and are interested. 

As you can see I have more questions than answers but I hope this input is useful anyway. 😊
Happy to talk more sync about this topic.

ondrejkonec:
> Hello @dancormier and @b-kelly, I spent some time doing a high level review of this PR (with the limited context I have).
> 
> Here are some comments/questions/ideas:
> 
> * What do we hope to achieve by introducing "pseudo-private custom properties"? My guess here is that we hope they will make the css codebase more readable/maintainable. Are there any other reasons I miss (e.g. do we expect consumers to leverage the dynamic nature of CSS custom properties - changing values at runtime)?
> * If the main goal is to make things more readable, maintainable and organised I have 2 suggestions:
>   
>   * If we don't need any dynamic behavior from these properties, what about converting them to preprocessor static variables instead (as recommended [here](https://www.smashingmagazine.com/2018/05/css-custom-properties-strategy-guide/#local-static-variables-are-ok-sometimes))? Such approach would also prevent consumers to change the value of the properties at runtime reinforcing their "private" nature
>   * What would you think about declaring all the variables at the top of the CSS file (similarly to what the author describes here as [the logic fold](https://www.smashingmagazine.com/2018/05/css-custom-properties-strategy-guide/#the-logic-fold))? This might establish a stronger pattern to recognise across the codebase (all the files starting with variable declarations).
> * About the naming "pseudo-private custom properties" - what "pseudo" represents? If my thinking above is correct and there are no other reasons to introduce those variables apart from maintainability what about calling them just "local variables"?
> * Is there already a way in Stacks to keep track about why certain architecture decisions were taken? If not, it could be beneficial to start [lightweight architecture decision records](https://github.com/peter-evans/lightweight-architecture-decision-records#lightweight-architecture-decision-records). I have used similar tools in the past: they are relatively inexpensive and they hold important context for new contributors but also for established ones. I'd be happy to give an introduction to ADRs if you don't know them and are interested.
> 
> As you can see I have more questions than answers but I hope this input is useful anyway. 😊 Happy to talk more sync about this topic.

Hi guys,

I was trying to fix few bugs here as I wanted to help you, but I'm often fighting with CSS architectural problems and poor component structure. I would like to support idea with making this design manipulation more simple, clear and straightforward.

Do you think I could try to suggest some improvements? I'm happy to get involved, but right now I find working with CSS in this code little bit inflexible and unscalable.

dancormier:
> **Note**
> For brevity, I'll refer to "pseudo-private custom properties" as PPCP

---

> What do we hope to achieve by introducing "pseudo-private custom properties"? My guess here is that we hope they will make the css codebase more readable/maintainable. Are there any other reasons I miss (e.g. do we expect consumers to leverage the dynamic nature of CSS custom properties - changing values at runtime)?

There's a few potential benefits to introducing PPCP, IMO:

- Flattening the cascade. With PPCP, we (mostly) only change custom property value at the base component level, so we can change values without having to untangle complex specificity. Long term, I think this could be useful to alter styles for nested components based on context (though it would kinda undermine the whole "private" aspect to PPCP).
- Improved readability. With PPCP, we explicit call out of properties changed by variants/modifiers/themes/etc and, since a given variant is only changing those properties, it's arguably easier to skim and see what's changing without needing to parse through a cascade of rules.
- Lighter-weight CSS. Because we're changing custom properties at the top level, we don't need to repeat the entire structure of a rule inside a given variant/modifier. We create the structure for the CSS once on the component, which makes it more maintainable and lighten the CSS output for more complex components.
- Reducing our dependency on Less. With [CSS nesting](https://www.w3.org/TR/css-nesting-1/) and [better approaches to colors](https://www.smashingmagazine.com/2021/11/guide-modern-css-colors/#lab) in native CSS on the horizon, I'm anticipating a day we no longer need a preprocessor. I figure this would speed up development and eliminate the need for contributors to understand CSS _and_ preprocessor syntax. This one is far off in the future, though.

Some of these benefits are personal preference and up for debate, and some benefits don't materialize in every context, but I've found it to be a useful approach on the whole. For further information on PPCP, I recommend [Lea Verou's great CSS Variable Secrets talk](https://www.youtube.com/watch?v=ZuZizqDF4q8), which has been the inspiration for this approach.

> If we don't need any dynamic behavior from these properties, what about converting them to preprocessor static variables instead (as recommended [here](https://www.smashingmagazine.com/2018/05/css-custom-properties-strategy-guide/#local-static-variables-are-ok-sometimes))? Such approach would also prevent consumers to change the value of the properties at runtime reinforcing their "private" nature

I'd like to tend toward native CSS and reduce Stacks' dependency on a preprocessor . I think the "pseudo" part of PPCP stands out, but I don't personally see a problem with the properties in question not truly being private.

> What would you think about declaring all the variables at the top of the CSS file (similarly to what the author describes here as [the logic fold](https://www.smashingmagazine.com/2018/05/css-custom-properties-strategy-guide/#the-logic-fold))? This might establish a stronger pattern to recognise across the codebase (all the files starting with variable declarations).

I might be misunderstanding, but I think that's the approach I've taken. Each component's PPCP are declaired at the top within the context of the given component. See [_styles-template.less](https://github.com/StackExchange/Stacks/blob/develop/lib/css/components/_styles-template.less#L4-L14) for more details. And let me know if I'm misunderstanding!

> About the naming "pseudo-private custom properties" - what "pseudo" represents? If my thinking above is correct and there are no other reasons to introduce those variables apart from maintainability what about calling them just "local variables"?

The naming is just pedantry and I'm not too tied to a specific name, honestly. I adopted the name from [Lea Verou](https://lea.verou.me/2021/10/custom-properties-with-defaults/). For some details, the "pseudo" refers to the fact that they're not _truly_ private. In CSS, all custom properties are exposed in the output but we treat them as if they're private. And I call them "custom properties" because that's [_technically_ their name](https://developer.mozilla.org/en-US/docs/Web/CSS/--*), but I have not problem with using the term "variables".

> Is there already a way in Stacks to keep track about why certain architecture decisions were taken? If not, it could be beneficial to start [lightweight architecture decision records](https://github.com/peter-evans/lightweight-architecture-decision-records#lightweight-architecture-decision-records). I have used similar tools in the past: they are relatively inexpensive and they hold important context for new contributors but also for established ones. I'd be happy to give an introduction to ADRs if you don't know them and are interested.

I'm a fan of ADRs (in fact, I was just talking to Alex about them last week) and I would love to adopt them. It would probably be a useful exercise to create some while the team grows. We should sync up on this and make some choices about where they should live and what we should document.

> As you can see I have more questions than answers but I hope this input is useful anyway. 😊 Happy to talk more sync about this topic.

Keep the questions coming! It's been useful to think through answers and I like having a little documentation to point to when others have the same questions! 💛

giamir:
> I'd like to tend toward native CSS and reduce Stacks' dependency on a preprocessor . I think the "pseudo" part of PPCP stands out, but I don't personally see a problem with the properties in question not truly being private.

I did not know that we are aiming to stick to native CSS as much as possible. I can certainly get behind that principle though - I like to leverage the platform as much as possible. Can I assume then that we are not supporting IE?

> I might be misunderstanding, but I think that's the approach I've taken. Each component's PPCP are declaired at the top within the context of the given component. See [_styles-template.less](https://github.com/StackExchange/Stacks/blob/develop/lib/css/components/_styles-template.less#L4-L14) for more details. And let me know if I'm misunderstanding!

In the `post-summary` file it looks like we are [continuing to declare css custom properties](https://github.com/StackExchange/Stacks/blob/67a1643ec9aa4e61bf5dc7dd65ceb3d558b3059b/lib/css/components/post-summary.less#L56) (`L56`) also after some [CSS properties have been already defined](https://github.com/StackExchange/Stacks/blob/67a1643ec9aa4e61bf5dc7dd65ceb3d558b3059b/lib/css/components/post-summary.less#L31) (`L31`). I was thinking to take things a step further and take care of **all** custom properties declarations at the very beginning of the file. In the [file you shared](https://github.com/StackExchange/Stacks/blob/develop/lib/css/components/_styles-template.less) that is the case so can I assume that for the `post-summary` file we are targeting the same but the refactoring is still ongoing?

Thanks a lot for the detailed answer(s) @dancormier. 🙏
I really appreciate it and I think the approach makes a lot of sense (so much sense that it might deserve a lightweight ADR). Starting with ADRs doesn't have to be expensive. In one of my previous team we just started by creating an ADRs.md file in the repo and we followed this simple template:

---

### [ADR Title]

**Date:** [DD.MM.YYYY]

**Issue:** [describe in few sentences the problem we need to solve]

**Status:** [ 🚧 Proposed | ✅ Accepted | 🚫 Rejected ]

**Decision:** [describe in few sentences the proposed solution to the issue] 

---

I personally like when ADRs are close to the codebase. 😊

dancormier:
> > I'd like to tend toward native CSS and reduce Stacks' dependency on a preprocessor . I think the "pseudo" part of PPCP stands out, but I don't personally see a problem with the properties in question not truly being private.
> 
> I did not know that we are aiming to stick to native CSS as much as possible. I can certainly get behind that principle though - I like to leverage the platform as much as possible. Can I assume then that we are not supporting IE?

Yes! We are not supporting IE. For a list of browsers we support, see https://browsers.stackoverflow.design/

> > I might be misunderstanding, but I think that's the approach I've taken. Each component's PPCP are declaired at the top within the context of the given component. See [_styles-template.less](https://github.com/StackExchange/Stacks/blob/develop/lib/css/components/_styles-template.less#L4-L14) for more details. And let me know if I'm misunderstanding!
> 
> In the `post-summary` file it looks like we are [continuing to declare css custom properties](https://github.com/StackExchange/Stacks/blob/67a1643ec9aa4e61bf5dc7dd65ceb3d558b3059b/lib/css/components/post-summary.less#L56) (`L56`) also after some [CSS properties have been already defined](https://github.com/StackExchange/Stacks/blob/67a1643ec9aa4e61bf5dc7dd65ceb3d558b3059b/lib/css/components/post-summary.less#L31) (`L31`). I was thinking to take things a step further and take care of **all** custom properties declarations at the very beginning of the file.  In the [file you shared](https://github.com/StackExchange/Stacks/blob/develop/lib/css/components/_styles-template.less) that is the case so can I assume that for the `post-summary` file we are targeting the same but the refactoring is still ongoing?

Ah, I see! The properties declared on `L56` should be defined _at the end_ of the file. Good catch! I've come to define the structure of these component styles more rigidly since creating this PR. I'll make some updates soon to better adhere to the correct structure.

> Thanks a lot for the detailed answer(s) @dancormier. 🙏 I really appreciate it and I think the approach makes a lot of sense (so much sense that it might deserve a lightweight ADR). Starting with ADRs doesn't have to be expensive. In one of my previous team we just started by creating an ADRs.md file in the repo and we followed this simple template:
> ### [ADR Title]
> 
> **Date:** [DD.MM.YYYY]
> 
> **Issue:** [describe in few sentences the problem we need to solve]
> 
> **Status:** [ 🚧 Proposed | ✅ Accepted | 🚫 Rejected ]
> 
> **Decision:** [describe in few sentences the proposed solution to the issue]
> 
> I personally like when ADRs are close to the codebase. 😊

Thanks for the template! I'll aim to make a PR for a simple ADR for PPCP this week. 

P.S. OMG tech uses a lot of acronyms 😝

lib/css/exports/constants-colors.less
dancormier: I've removed those variables for now. We can punt and try to get this one across the finish line.

giamir (changes requested):
Hi @dancormier, 

I have rebased the branch and have done a preliminary review. I have found a couple of visual issues but I did not find the time to look further into details.

The answers preview looks off:
Screenshot 2023-01-04 at 16 33 19

And there are some layout issues for the stats:

Screenshot 2023-01-04 at 16 35 52
Screenshot 2023-01-04 at 16 36 21

Apart from that the rest looks good to me.
Thanks 🙏

dancormier:
@giamir odd! Thanks for looking at this branch. Were there any merge conflicts when merging in the latest `develop`?

giamir:
@dancormier No conflict, but the branch looked pretty outdated so it might be that some other refactor like the gap/gutter one have affected the visuals.

dancormier:
Sorry about the mess @giamir! There were a few rough edges that I've smoothed out. Honestly, it had nothing to do with the age of this branch but was just a few issues I missed in the course of refactoring. It should be good for a review now 🤞

lib/css/components/post-summary.less
```diff
@@ -1,436 +1,407 @@
...
     }
 
-    #stacks-internals #screen-md({
-        flex-direction: row;
-        align-items: center;
-        width: auto;
-    });
-}
+    // Child Elements
+    & & {
```
giamir, line 143: I had no clue that using 2 ampersands like that generate a selector targeting the children. I have to admit sometimes all these ampersands give me a headache. I learned something though. 🙂

lib/css/components/post-summary.less
```diff
@@ -1,436 +1,407 @@
...
-        .v-truncate2;
-        .break-word;
+                &__lg {
+                    .v-truncate4;
+                }
+            }
 
-        &.s-post-summary--content-excerpt__sm {
-            .v-truncate1;
+            .break-word;
```
giamir, line 196: Just thinking out loud: in this file we make extensive usage of "atomic classes as mixins". I don't have strong opinions about that but I was wondering if we could do something similar also when we will write the styles of our new WCs; it is certainly handy to extend/refer the rules like that instead of rewriting them over and over, at the same time though we create "implicit" coupling across 2 less files. I think it would be more maintainable if there would be an "explicit" import at the beginning of the file showing that this file depends on typography.less (something like `@use typography.less as t` in SCSS I believe). No action required here just food for thoughts in terms of maintainability. 🙂

giamir (approved):
Hello @dancormier,

I have gone through the screens again and everything looks good to me now.
I have also checked the code to the best of my abilities. I left a couple of comments (one about maintainability specifically) but there are no actions to be taken in this PR.

Great work! You can go ahead and merge from my end! 🙏

lib/css/components/post-summary.less
dancormier: The way I think of it is that an ampersand just pops the parent selector in place to make a new selector. A double ampersand (be it `&&` or `& &`) can be pretty useful and, at this point, my mental model is pretty locked in so it's intuitive to me.

I like the noise it removes but I could see it being tough to read in some cases, so if you see any egregious uses, call them out and we can consider long-hand or other tactics to make our Less more approachable.

lib/css/components/post-summary.less
dancormier: I like the explicit import idea. Seeing as we're considering splitting up components/atomic classes/etc into standalone repos, it seems like a natural progression.

PR #1100, opened 2022-09-09
fix(input): increase placeholder text contrasts

dancormier:
From WCAG SC 1.4.3:

> The minimum contrast success criterion (1.4.3) applies to text in the page, including placeholder text and text that is shown when a pointer is hovering over an object or when an object has keyboard focus. If any of these are used in a page, the text needs to provide sufficient contrast.

## Before

image

## After

image

### DIFF
--- a/lib/css/components/inputs.less
+++ b/lib/css/components/inputs.less
@@ -73,21 +73,13 @@
 
     //  --  PLACEHOLDER
     //  ------------------------------------------------------------------------
+    &::placeholder,
     &::-webkit-input-placeholder {
-        color: var(--black-200);
-
-        .highcontrast-mode({
-            color: var(--black-400);
-        });
+        color: var(--black-500);
     }
 
     &::placeholder {
-        color: var(--black-200);
         opacity: 1;
-
-        .highcontrast-mode({
-            color: var(--black-400);
-        });
     }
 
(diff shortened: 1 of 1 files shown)


dancormier:
👋 @kplsn I added you as a reviewer to this PR so you could checkout this design change. You can see some inputs with placeholder text in the [deploy preview](https://deploy-preview-1100--stacks.netlify.app/product/components/inputs/#sizes).

I'm a little on the fence about this since, with the placeholder text so dark, it can give the user the impression an input has been filled out when it hasn't and I'm fine to hold off on this if it's not a viable change at the moment.

kplsn:
The consensus amongst Deque and the US/UK governments is that placeholder text should be avoided. 

- https://www.deque.com/blog/accessible-forms-the-problem-with-placeholders/ 
- https://designsystem.digital.gov/components/text-input/
- https://design-system.service.gov.uk/components/text-input/

I think we should first do an audit of existing placeholder text usage and decide where it can be removed, and where (if anywhere) it might serve a purpose. Also, #767676 on a white background passes, but is not currently a Stacks colour.

Screenshot 2022-09-15 at 15 57 55

dancormier:
Closing to first perform an audit and reassess our approach

PR #1104, opened 2022-09-15
Darkened user card time for highlighted

ellenchanhy:
For `s-user-card__highlighted`, the `s-user-card--time` text colour was darkened from `--black-500` to `--black-600`.

Before:
image

After:
image


dancormier (approved):
👍

PR #1107, opened 2022-09-16
feature(btn): extend disabled styling to include `aria-disabled` attribute

dancormier:
This PR adds disabled styling to `.s-btn` elements with `[aria-disabled="true"]` (and a little docs update to mention it).

cc @allejo


allejo (approved):

b-kelly (commented):
lgtm, nice change (late review)

PR #1116, opened 2022-09-20
docs(icon): update data-icon usage instructions

KyleMit:
Originally discussed when we updated codepen links to use the new template in this commit: https://github.com/StackExchange/Stacks/commit/e3b1d4512c3d25447afb01075da0e296fafd3d10

Icon data attributes now need to be prefixed with "Icon":

```diff
- <svg data-icon="Logo"></svg>
+ <svg data-icon="IconLogo"></svg>
```

Updates this section of documetation - https://deploy-preview-1116--stacks.netlify.app/product/resources/icons/#prototypes

image

### DIFF
--- a/docs/product/resources/icons.html
+++ b/docs/product/resources/icons.html
@@ -36,12 +36,12 @@ description: Stacks provides a complete icon set, managed separately in the <a h
     </div>
 
     {% header "h3", "Prototypes" %}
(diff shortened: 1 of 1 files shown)


dancormier (approved):
Lovely! You even caught a typo 💛

b-kelly (approved):
lgtm. Technically, `spots.html` should be updated as well (e.g. `SpotWaveXl`), but it doesn't actually have the same docs you're editing here. 🤷 No need to do it in this PR as I'd consider adding those docs to be out of scope.

PR #1121, opened 2022-09-21
refactor(toggle-switch): simplify markup

dancormier:
This PR simplifies the markup for the `.s-toggle-switch` component, updates the Less to use pseudo-private custom properties, and includes minor docs updates.

[Deploy preview: toggle switch component](https://deploy-preview-1121--stacks.netlify.app/product/components/toggle-switch/)

---

> **Warning**  - DEPRECATION
> This PR contains changes to the `.s-toggle-switch` markup. While changes to styling are currently backwards compatible, consumers should update their markup to match the new markup.

> **Note**
> The only UI change in this PR is the application of `cursor: default` to `disabled` toggles switches.

---

### Old markup
```
<label class="s-label" for="toggle-example">…</label>
<div class="s-toggle-switch">
    <input id="toggle-example" type="checkbox">
    <div class="s-toggle-switch--indicator"></div>
</div>
```

### New markup
```
<label class="s-label" for="toggle-example">…</label>
<input class="s-toggle-switch" id="toggle-example" type="checkbox">
```


b-kelly (changes requested):
I like the changes so far, but I think we could (and should) go a step further. I'd personally love to see the toggle switch class placed on the parent element, much like we do for multiple. Currently, the control doesn't work _at all_ without the right atomic classes on the parent along with the label, so let's enforce that. Additionally, it'd align the structure with the multi-toggle, making our future Razor/React/Svelte/whatever components much easier to design.

Something like this:

```html
<div class="s-toggle-switch">
    <label class="s-label" for="toggle-example-checked">…</label>
    <input type="checkbox" id="toggle-example-checked" checked>
</div>
```

dancormier:
I'm punting on https://github.com/StackExchange/Stacks/pull/1121#pullrequestreview-1131921845 for now for the sake of backwards compatibility and merging this PR. We can (and should) revisit at a later date.

Issue #957, opened 2022-05-27
When a post is both "watched" and "deleted", the background color styling for deleted should take precendence


johnmwright:
When a post is both "watched" and "deleted", the watched styling takes precedence instead of the deleted styling.  In my opinion, it should be the other way around (the fact it's deleted is more important than the fact it's watched).

Specifically, a post that has `.s-post-summary__deleted` and `.s-post-summary__watched` will have it's background color set to the watched value.

Here is a screenshot showing four posts:
- a deleted post with red background
- a watched & delete post with the yellow background (this is the one I think should be red)
- a watched post with yellow background
- a normal post with neither watched or deleted

![image](https://user-images.githubusercontent.com/3373249/170733222-be40502b-0ff6-4117-b8fb-d6b9361d8650.png)

Note that the style appears to come from `primary.css` instead of `stacks.css`:

![image](https://user-images.githubusercontent.com/3373249/170733314-1851d429-e817-4a3a-bdc9-bf90ffe0f6ed.png)

dancormier:
Seems the issue is an [additional post summary style in `_base-primary.less`](https://github.com/StackEng/StackOverflow/blob/master/StackOverflow/Content/less/_base-primary.less#L1619-L1621).

```less
// TODO upstream?
.s-post-summary__watched {
    .tagged-interesting;
}
```

ondrejkonec:
Hi there!

I take a look and tried to fix this bug in `post-primary.less` file in Stacks here #1134 

Because the component structure does does not have strict methodology, I chose a way to just raise the priority of the selector a bit. 

What do you think @dancormier ?

Hope it helps!

Ondřej

PR #1134, opened 2022-10-05
Increased specificity to get a higher priority for rewriting styles from primary.css

ondrejkonec:
Hi!

I took care about this bug #957 

Variants does not work without `s-post-summary` selector, so it is possible to increase the specificity of the selector with certainty and achieve that it will have a higher priority and override `s-post-summary__watched` if both selectors are there. We can remove this selector override when it's not in `primary.css` file.

Tested in Stacks and https://stackoverflow.com/

Shot from Stackoverflow.com

I have one question about that. What is `primary.css` used for or why are the same elements there?

Hope this helps!

### DIFF
--- a/lib/css/components/post-summary.less
+++ b/lib/css/components/post-summary.less
@@ -318,7 +318,7 @@
     }
 }
 
-.s-post-summary__deleted {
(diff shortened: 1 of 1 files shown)


ondrejkonec:
:(

PR #1157, opened 2022-10-24
refactor(flex): deprecate gutter classes

### LINKED ISSUE #1010: Deprecate gutter classes; shuffle related docs sections
We're generally using native gap more and more often to control layout and it's time make that explicit. We have been waiting on being deeper into the iOS 15 release cycle since, gap was only supported as of [iOS 14.5](https://caniuse.com/?search=gap).

It's probably been long enough! Here's what we need to do:

- [x] Remove usage of [gutter classes (`.gs*`)](https://stackoverflow.design/product/base/flex/#gutters) from Stacks components
- [x] Add a deprecation message to gutter classes documentation
- [x] Consider moving gap class documentation to it's own page or [Spacing](https://stackoverflow.design/product/base/spacing/) (since it doesn't _only_ apply to grid)

~~This will be a major breaking change so it'll take some coordination.~~

giamir:
This PR:
- replaces all the usages of gutter classes across the codebase (`gs[number]`, `gsx` and `gsy`) with gap classes (`g[number]`, `gx[number]` and `gy[number]`). Some examples:
  - `gs4` becomes `g4`
  - `gs4 gsx` becomes `gx4`
  - `gs4 gsy` becomes `gy4` 
- introduces a [deprecation warning](https://github.com/StackExchange/Stacks/blob/gbuoncristiani/deprecate-gutter-classes/docs/product/base/flex.html#L267) in the flex documentation under the gutters section
- introduces a couple of [less mixins](https://github.com/StackExchange/Stacks/pull/1157/files#diff-d3945b34790d45853feb2f00f6238921de62df644765fca9994b4914c1baf8f7R233) to adjust flex items witdth to account for gap spacing

Closes #1010


giamir:
@dancormier shall we create a section in the flex documentation page to illustrate how to use gap in a flex layout?
We could create something similar to what we have now for gutters but for gap: https://deploy-preview-1157--stacks.netlify.app/product/base/flex/#gutters
We could also push the current gutters section at the very end of the page (since it is deprecated and it will eventually be removed in the next major release).

lib/css/atomic/flex.less
```diff
@@ -151,6 +153,38 @@
...
+
+            #stacks-internals #flex-builder-helpers .flex-fixed-item-spacing-gap(@spacing, @num, (@count + 1));
+        }
+
+
+        .gap-spacing(@spacing) {
+            #stacks-internals #flex-builder-settings();
+
+            #stacks-internals #flex-builder-helpers .flex-fixed-item-spacing-gap(@spacing, @flex-cols);
+        }
```
dancormier, line 187: ## No action required

### Eventual `.flex--item` deprecation

Long term, I'd love to consider deprecating the `.flex--item` class in favor of assuming all direct children should gain whatever styling currently is applied by `.flex--item`. Deprecating `.flex--item` is out of scope for this PR and would be a pretty substantial lift, but I figured I'd plant the seed while we're here.

docs/product/base/flex.html
```diff
@@ -264,6 +264,9 @@
 
 <section class="stacks-section">
     {% header "h2", "Gutters" %}
+    {% tip, "warning" %}
+        <strong>Deprecation:</strong> gutters will be removed in a future release.<br /> Please use <a href="/product/base/gap">gap classes</a> to set spacing on flex items.
+    {% endtip %}
```
dancormier, line 269: ## No action required

I'm not a huge fan of our naming convention here. The terms `gap`, `gutter`, and `alley` can all mean the same thing. 

From [MDN docs on gap](https://developer.mozilla.org/en-US/docs/Web/CSS/gap):

> The gap [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS) property sets the gaps ([gutters](https://developer.mozilla.org/en-US/docs/Glossary/Gutters)) between rows and columns.

And [MDN's glossary entry on "gutters"](https://developer.mozilla.org/en-US/docs/Glossary/Gutters):

> Gutters or alleys are spacing between content tracks.

I find this hilarious and have always felt a little hesitancy in labeling one approach `gutters` and another `gap` because of these definitions. [Two hard things](https://martinfowler.com/bliki/TwoHardThings.html), am I right?

docs/product/guidelines/conditional-classes.html
```diff
@@ -59,9 +59,9 @@
...
+<pre class="language-html s-code-block"><code class="language-html s-code-block"><span class="hljs-tag">&lt;<span class="hljs-name">div</span> <span class="hljs-attr">class</span>=<span class="hljs-string">"d-flex flex__center <span class="md:highlight">md:fd-column</span> g16"</span>&gt;</span><br>    <span class="hljs-tag">&lt;<span class="hljs-name">div</span> <span class="hljs-attr">class</span>=<span class="hljs-string">"flex--item wmx3 fs-body3 <span class="lg:highlight">lg:ta-center</span>"</span>&gt;</span><br>        …<br>    <span class="hljs-tag">&lt;/<span class="hljs-name">div</span>&gt;</span><br>    <span class="hljs-tag">&lt;<span class="hljs-name">div</span> <span class="hljs-attr">class</span>=<span class="hljs-string">"flex--item <span class="md:highlight">md:order-first</span> <span class="sm:highlight">sm:d-none</span>"</span>&gt;</span><br>        <span class="hljs-tag">&lt;<span class="hljs-name">svg</span>&gt;</span>…<span class="hljs-tag">&lt;/<span class="hljs-name">svg</span>&gt;</span><br>    <span class="hljs-tag">&lt;/<span class="hljs-name">div</span>&gt;</span><br><span class="hljs-tag">&lt;/<span class="hljs-name">div</span>&gt;</span></code></pre>
```
dancormier, line 62: ## No action required

It took me a second to notice the specific change here. There's a `gs16` buried in there that has been changed into `g16`.

lib/css/atomic/flex.less
```diff
@@ -194,9 +228,24 @@
...
+.g12, .gx12, .gy12 { #stacks-internals #flex-builder-helpers .gap-spacing(var(--su12)); }
+.g16, .gx16, .gy16 { #stacks-internals #flex-builder-helpers .gap-spacing(var(--su16)); }
+.g24, .gx24, .gy24 { #stacks-internals #flex-builder-helpers .gap-spacing(var(--su24)); }
+.g32, .gx32, .gy32 { #stacks-internals #flex-builder-helpers .gap-spacing(var(--su32)); }
+.g48, .gx48, .gy48 { #stacks-internals #flex-builder-helpers .gap-spacing(var(--su48)); }
+.g64, .gx64, .gy64 { #stacks-internals #flex-builder-helpers .gap-spacing(var(--su64)); }
+
+
+// DEPRECATED - START
+//
```
dancormier, line 248: Nice stop-gap!

I have no suggestion here… I just wanted to say that pun 😝

dancormier (commented):
> **Note**
> This is looking good so far, but I'm holding off on a approval until after the rebase of recent gap class changes.

Great work @giamir! ~~I don't see any need for changes (outside of what comes from the rebase)~~ I found a single issue I'd like to see addressed (relating to `<label>` elements being `display: table` and therefore unaffected by gap classes), but that's all so far. I left a few notes in the course of my review, but they're all purely minor mentions of fine details or out of scope items I noticed.

### Speaking of out of scope…

Since `justify content`, `align items`, and `order` CSS rules can apply to both flex and grid children, I propose we move these items from their current flex context into separate doc pages and restructure the Less files to reflect this. I plan on doing this in an upcoming PR.

docs/product/base/colors.html
```diff
@@ -599,7 +599,7 @@
 <p>…</p>
 <p class="fc-dark">…</p>
 {% endhighlight %}
-        <div class="stacks-preview--example d-flex fd-column gs16 gsy my0 fs-body2">
+        <div class="stacks-preview--example d-flex fd-column gx16 my0 fs-body2">
```
dancormier: ## No action necessary

This renders slightly differently than what's in `develop`.

<details>
<summary>develop</summary>

image

</details>

<details>
<summary>This PR</summary>

image

</details>

<details>
<summary>Backstop diff</summary>

image

</details>

This is a result of the newly applied `.gx16` class no longer overriding margin on the child `<p>` elements. Previously, the margin hack from `.gs16.gsy` replaced any default margins set on `<p>`. IMO this is fine in the context of docs, but we'll want to be conscious of how this might affect similar code in production.

docs/product/components/checkbox.html
```diff
@@ -10,7 +10,7 @@
 {% highlight html %}
 <!-- Base -->
 <fieldset>
-    <div class="d-flex gs8">
+    <div class="d-flex g8">
```
dancormier, line 13: ## Change requested

Since ~~`<label>`~~ `<legend>` elements are rendered as `display: table`, the gap property doesn't apply. Any ~~`<label>`~~ `<legend>` with parents that include a gap property need either to have a margin added or have their `display` property altered to honor `gap`.

docs/product/components/expandable.html
```diff
@@ -153,10 +153,10 @@
                     </div>
                 </div>
                 <div class="flex--item">
-                    <div class="d-flex gs8 gsx">
+                    <div class="d-flex gy8">
```
dancormier: ## Change requested

Since ~~`<label>`~~ `<legend>` elements are rendered as `display: table`, the gap property doesn't apply. Any ~~`<label>`~~ `<legend>` with parents that include a gap property need either to have a margin added or have their `display` property altered to honor `gap`.

docs/product/components/inputs.html
```diff
@@ -10,7 +10,7 @@
 
     <div class="stacks-preview">
 {% highlight html %}
-<div class="d-flex gs4 gsy fd-column">
+<div class="d-flex gx4 fd-column">
```
dancormier: ## Change requested

Since ~~`<label>`~~ `<legend>` elements are rendered as `display: table`, the gap property doesn't apply. Any ~~`<label>`~~ `<legend>` with parents that include a gap property need either to have a margin added or have their `display` property altered to honor `gap`.

docs/product/components/labels.html
```diff
@@ -14,15 +14,15 @@
     {% header "h2", "Base style" %}
     <div class="stacks-preview">
 {% highlight html %}
-<form class="d-flex gs4 gsy fd-column">
+<form class="d-flex gx4 fd-column">
```
dancormier: ## Change requested

Since ~~`<label>`~~ `<legend>` elements are rendered as `display: table`, the gap property doesn't apply. Any ~~`<label>`~~ `<legend>` with parents that include a gap property need either to have a margin added or have their `display` property altered to honor `gap`.

docs/product/components/links.html
```diff
@@ -206,11 +206,11 @@
 <a class="s-block-link s-block-link__left is-selected" href="#">…</a>
 {% endhighlight %}
         <div class="stacks-preview--example">
-            <div class="d-flex gs16 fd-column ws3">
+            <div class="d-flex g16 fd-column ws3">
                 <a class="flex--item s-block-link s-block-link__left" href="#">Block link</a>
             </div>
 
-            <div class="d-flex gs16 fd-column ws3">
+            <div class="d-flex g16 fd-column ws3">
```
dancormier, line 213: ## No action necessary

This renders slightly differently than what's in `develop`. Same as https://github.com/StackExchange/Stacks/pull/1157#discussion_r1003817900

<details>
<summary>Diff</summary>

image

</details>

docs/product/components/post-summary.html
```diff
@@ -1105,7 +1105,7 @@ <h3 class="s-post-summary--content-title">
 </div>
 {% endhighlight %}
         <div class="stacks-preview--example">
-            <div class="d-flex ai-center gs24 fw-wrap">
+            <div class="d-flex ai-center g24 fw-wrap">
```
dancormier, line 1108: ## No action necessary

This renders differently than `develop`.

<details>
<summary>develop</summary>

image

</details>

<details>
<summary>This PR</summary>

image

</details>

I haven't figured out why, but I suspect it's another negative margin difference.

docs/product/components/radio.html
```diff
@@ -11,7 +11,7 @@
 <!-- Base -->
 <fieldset>
     <legend class="flex--item s-label">Which radio is best?</legend>
-    <div class="d-flex gs8">
+    <div class="d-flex g8">
```
dancormier, line 14: ## Change requested

Since ~~`<label>`~~ `<legend>` elements are rendered as `display: table`, the gap property doesn't apply. Any ~~`<label>`~~ `<legend>` with parents that include a gap property need either to have a margin added or have their `display` property altered to honor `gap`.

docs/product/components/textarea.html
```diff
@@ -8,13 +8,13 @@
     {% header "h2", "Base style" %}
     <div class="stacks-preview">
 {% highlight html %}
-<div class="d-flex fd-column gs4 gsy">
+<div class="d-flex fd-column gx4">
```
dancormier: ## Change requested

Since ~~`<label>`~~ `<legend>` elements are rendered as `display: table`, the gap property doesn't apply. Any ~~`<label>`~~ `<legend>` with parents that include a gap property need either to have a margin added or have their `display` property altered to honor `gap`.

lib/css/atomic/flex.less
giamir: 👍 Might be worth creating an issue detailing the deprecation with pros and cons.

docs/product/guidelines/conditional-classes.html
giamir: Yeah, the diffing viz could be better

lib/css/atomic/flex.less
giamir: 😂

docs/product/base/colors.html
giamir: Good catch! I am not entirely sure why we use Flex for these paragraphs to be honest. I just left the paragraphs and removed flex all together in my local machine and it looks pretty much the same to the original version. I will update the PR.

docs/product/base/flex.html
giamir: > [Two hard things](https://martinfowler.com/bliki/TwoHardThings.html), am I right?

If Martin says so. 😂

docs/product/components/links.html
giamir: 👍

docs/product/components/inputs.html
giamir: I cannot see any visual difference here. Can you have another look in the updated PR and tell me if this is still relevant? Thanks

docs/product/components/labels.html
giamir: Also here cannot see visual differences.

docs/product/components/textarea.html
giamir: Also for `textarea` cannot see visual differences.

giamir:
Thanks @dancormier for the detailed review.

I have rebased now the branch with the new `gap` properties you released yesterday.
I guess rewriting history of PR branches is ok. If you prefer an approach where the changes from main (or develop in this case) are merged back let me know. I don't have a particular preference. Rebasing can be a bit annoying if more than one person are working on a branch since the history is rewritten and can cause conflicts to resolve on pull.

>  I found a single issue I'd like to see addressed (relating to <label> elements being display: table and therefore unaffected by gap classes)

Good catch! The actual elements affected though are not the `<labels>` but the `<legend>` ones. I ended up fixing the issue by giving an appropriate margin bottom ([see this commit for context](https://github.com/StackExchange/Stacks/pull/1157/commits/fa45b384ccdd83fc79b6b8425c941adfc0a359b2)).
I tried to make it work by making the `legend` a `d-block` too but I gave up because there is some very hard to debug behavior where `legend` elements ignore the `gap`properties. This might be related on how browsers treat [rendered legend](https://html.spec.whatwg.org/multipage/rendering.html#rendered-legend) as described in the html specs. I spent 1h or so investigating without success. Feel free to give it a shot too if you deem that important.

> Since justify content, align items, and order CSS rules can apply to both flex and grid children, I propose we move these items from their current flex context into separate doc pages and restructure the Less files to reflect this. I plan on doing this in an upcoming PR.

I think that's a good idea and it follows what the folks at Tailwind are doing. I guess the more we make our atomic classes similar to Tailwind ones the easier should be for new developers to ramp up and use Stacks.
https://tailwindcss.com/docs/justify-content

docs/product/components/inputs.html
dancormier: Ah, I was a little hasty in assuming the `<legend>` issue was here. Backstop show some significant differences on this page and it looks to be due to `.s-input-message` being rendered with the `<p>` tag, which comes with margins that were previously overwritten with the `.gs4` class.

docs/product/components/labels.html
dancormier: Yep, not sure what issue I saw here yesterday! Sorry for the noise!

docs/product/components/textarea.html
dancormier: Ah, this is the same issue as reference in https://github.com/StackExchange/Stacks/pull/1157#discussion_r1004616938

> Ah, I was a little hasty in assuming the `<legend>` issue was here. Backstop show some significant differences on this page and it looks to be due to `.s-input-message` being rendered with the `<p>` tag, which comes with margins that were previously overwritten with the `.gs4` class.

dancormier:
> I guess rewriting history of PR branches is ok. If you prefer an approach where the changes from main (or develop in this case) are merged back let me know. I don't have a particular preference. Rebasing can be a bit annoying if more than one person are working on a branch since the history is rewritten and can cause conflicts to resolve on pull.

Since we squash PRs down to a single commit, we should be good with a merge in these cases to prevent any conflicts on pull.

> Good catch! The actual elements affected though are not the <labels> but the <legend> ones. I ended up fixing the issue by giving an appropriate margin bottom ([see this commit for context](https://github.com/StackExchange/Stacks/pull/1157/commits/fa45b384ccdd83fc79b6b8425c941adfc0a359b2)).

That's what I get for copying and pasting. Instead of misspeaking once, I've done it _a bunch_. Yay! I'm gonna go ahead and update those comments for clarity/posterity. Sorry for the confusion!

> I tried to make it work by making the `legend` a `d-block` too but I gave up because there is some very hard to debug behavior where `legend` elements ignore the `gap` properties. This might be related on how browsers treat [rendered legend](https://html.spec.whatwg.org/multipage/rendering.html#rendered-legend) as described in the html specs. I spent 1h or so investigating without success. Feel free to give it a shot too if you deem that important.

Thanks for investigating. If you spent an hour on this without success, I say you're good to move on. The juice isn't worth the squeeze, as they say!

docs/product/base/typography.html
```diff
@@ -50,7 +50,7 @@
 <a class="td-none">…</a>
 {% endhighlight %}
         <div class="stacks-preview--example">
-            <div class="d-flex fd-column gs16 gsy">
+            <div class="d-flex fd-column gy16">
```
dancormier, line 53: ### Nitpick

This page contains many flexbox elements comprised solely of `<p>` children. For this page specifically, we could probably remove the `d-flex` and gap classes altogether and have a more reasonable layout. It's not 100% necessary, but I'll leave it up to you @giamir.

docs/product/components/inputs.html
```diff
@@ -131,7 +131,7 @@
     {% header "h4", "Warning" %}
     <div class="stacks-preview">
 {% highlight html %}
-<div class="d-flex gs4 gsy fd-column has-warning">
+<div class="d-flex gy4 fd-column has-warning">
```
dancormier, line 134: ### Changes requested

We should determine how we want to handle `p.s-input-message` both in the example code blocks as well as the rendered example. I think adding `.mb0` to the paragraph element will suffice.

docs/product/components/textarea.html
```diff
@@ -8,13 +8,13 @@
     {% header "h2", "Base style" %}
     <div class="stacks-preview">
 {% highlight html %}
-<div class="d-flex fd-column gs4 gsy">
+<div class="d-flex fd-column gy4">
```
dancormier, line 11: ### Changes requested

We should determine how we want to handle `p.s-input-message` both in the example code blocks as well as the rendered example. I think adding `.mb0` to the paragraph element will suffice.

dancormier (changes requested):
**tl;dr**: I ran Backstopjs. Outside of a few minor margin issues, this PR is good! The issues are called out in review comments.

@giamir feel free to move it out of _Draft_ if you think this has what it needs to merge (outside of requested changes).

---

## Notes on the margin issue

The margin issue arises from flexbox elements with `<p>` children. Previously, the gutter classes set would _override_ the default margin on `<p>` elements. Now by using gap classes, the `<p>` element's default margin is no longer overridden and the gap adds to the space between elements. This is an improvement IMO (i.e. it's less hacky and unpredictable), but it results in some funky layout issues when transitioning to these gap classes in practice.

I'm less concerned with how things render in the docs, to be honest. I'm more concerned with what we recommend for our main consumer when they transition to gap classes and run into the same issue. Overall, not a big issue but I figured it's worth calling out so it's at least in the back of our heads as we make the transition.

### Here's an example of the difference in our docs:

<details>
<summary>Old gutter method</summary>

image

</details>

<details>
<summary>New gap method with excess spacing</summary>

image

</details>

## Backstop results

This latest review is based off of differences found when running Backstopjs (`npm run test`). Before running the test, I increased the `misMatchThreshold` value from `0.1` to `0.2` to cut down on diffs reporting minor text rendering differences.

image

As the image above shows, Backstop reports 58 failures across our docs pages which are tested a three different viewport sizes. Those failures occur across 28 different pages (if I counted correctly). _Most_ of them are counted as failures either because a) we intentionally made visual changes to the page, b) the page includes dynamic content which will always count as a failure, c) backstop messed up the rendering, or d) the transition to gap classes materially affected the layout in an undesirable way. Here are the results separated into groups for the sake of posterity:

### Intentional/minor visual changes or dynamic content
#### Product
##### Base
- backgrounds
- colors
- flex
- overflow
##### Components
- avatars
- checkbox
- expandable
- labels
- links
- popovers
- post summary
- progress bars
- radio
- select
- toggle switch
### Content
- Alt text
### Brand
- Logos
- Colors

### Backstop screwed up (and I've checked there's no impactful diff)
#### Email
- Getting started
- Short transactional
- Long transactional
- Promotional
- Major announcements

### Differences worth addressing
#### Product
##### Base
- typography
##### Components
- inputs
- textarea

docs/product/base/typography.html
giamir: 👍

docs/product/components/textarea.html
giamir: 👍

docs/product/components/inputs.html
giamir: 👍

docs/product/components/inputs.html
giamir: 👍

docs/product/components/textarea.html
giamir: 👍

giamir:
Thanks @dancormier for the detailed breakdown. 
I took care about the `p.s-input-message` in `textarea` and `input`.
I think we should good to go for merging this PR. By the way, since the PR touches many files we should consider merging sooner rather than later. I already had to fix some conflicts with `develop` and they are never fun (on top of being error-prone). 🙂

dancormier (approved):
Great work @giamir!

PR #1158, opened 2022-10-24
fix(gap): swap gap axis names

dancormier:
This PR addresses a mistake introduced in https://github.com/StackExchange/Stacks/pull/1143. This shouldn't impact any existing production code. 

This was a really silly mistake on my part. I assumed (for who knows what reason) that gap values were `gap: x-axis y-axis`. While reviewing https://github.com/StackExchange/Stacks/pull/1157, I realized that I implemented gap axis values backwards. [MDN says](https://developer.mozilla.org/en-US/docs/Web/CSS/gap#formal_syntax) `gap = 
  <'row-gap'> <'column-gap'>`. Really embarrassing!

In any case, this PR will fix the issue before it gets to prod and any meaningful adoption of `.gx`/`.gy` classes occurs.

### DIFF
--- a/docs/_data/gap.json
+++ b/docs/_data/gap.json
@@ -75,124 +75,124 @@
   ],
   "row": [
     {
-      "class": "gx0",
+      "class": "gy0",
       "define": "Add no space between rows",
       "responsive": true
     },
     {
-      "class": "gx1",
+      "class": "gy1",
       "define": "Space out rows by 1px",
       "responsive": true
     },
     {
-      "class": "gx2",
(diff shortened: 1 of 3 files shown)


giamir (approved):
@dancormier I had a look and it makes sense. Good catch!
I also double check with what they are doing here: https://tailwindcss.com/docs/gap
 Tomorrow I will rebase my work.

dancormier:
> Good catch!

Would've been better if I didn't make the mistake to begin with! 😖

In any case, thanks for the review!

PR #1160, opened 2022-10-27
docs(adr): add a template for adrs

giamir:
We have been talking in several occasions within our Stacks Core Team about documenting our architecture decisions in a format that is easily consumable by all contributors.

This is my attempt to get us kick started with that.
I would like to keep the process as lean as possible in the beginning and explore more complex avenues (e.g. RFCs) only when we feel the need for it.
 
 In my "backlog" I have already 2 items which will eventually benefit from written ADRs:
- Deprecation Strategy
- Testing Strategy

I know that Dan have something about:
- Pseudo Private Custom Properties

 Looking forward to get your feedback. 🙂

### DIFF
--- a/adrs/0000-template.md
+++ b/adrs/0000-template.md
@@ -1,0 +1,16 @@
+# [ADR Title]
+
+**Date:** [YYYY-MM-DD]
+
+**Status:** [ 🚧 Proposed | ✅ Accepted | 🚫 Rejected ]
+
+**Related PR:** [#xxxx (when applicable)]
+
+### Issue
+[describe in few sentences the problem we need to solve]
+
+### Decision
+[describe in few sentences the proposed solution to the issue]
+
+##### Disclaimer
+This is just a template: while we expect the above items to appear in every ADRs, feel free to add any additional points as you see fit based on your ADR context (e.g. Consequences, etc...).
--- a/adrs/0000-template.md
+++ b/adrs/0000-template.md
@@ -2,8 +2,6 @@
 
 **Date:** [YYYY-MM-DD]
 
-**Status:** [ 🚧 Proposed | ✅ Accepted | 🚫 Rejected ]
-
 **Related PR:** [#xxxx (when applicable)]
 
 ### Issue
--- a/adrs/0000-template.md
+++ b/adrs/0000-template.md
@@ -12,5 +12,5 @@
 ### Decision
 [describe in few sentences the proposed solution to the issue]
 
-##### Disclaimer
-This is just a template: while we expect the above items to appear in every ADRs, feel free to add any additional points as you see fit based on your ADR context (e.g. Consequences, etc...).
+### Additional info
+[anything else you'd like to add; remove this section if it is empty]
--- a/adrs/README.md
+++ b/adrs/README.md
@@ -1,0 +1,21 @@
+# Architecture Decision Records (ADRs)
+An architecture decision record (ADR) is a document that captures an important architecture decision made along with its context and consequences. We strive to keep these records small and lean.
+
+### Goal
+We want to record important design decisions to help current and future contributors of the library to understand the reasons for doing things a certain way.
+
+
+### Lightweight Process
+
+- Create a new branch on this repository and copy `adrs/0000-template.md` to a new file with the next free number.
+- Modify the template to add details about your proposal and add any supporting picture under the `adrs/images` folder.
+- Make a Pull Request (PR) for your branch.
+- Once consensus is reached and approvals given using the Github review system, the PR can be merged.
+
+In case the team has already agreed on a decision and no further async discussion is necessary just go ahead an record it directly without opening any PR.
+
+
+### Disclaimer
+
+As of today, the primary audience for these ADRs artefacts is the Stacks Core Team. If circumnstances will change and we feel a broader audience should get involved in the decision making process we could look into evolving our process in something a bit more formal like the internal [Stack Overflow RFC process](https://github.com/StackEng/StackOverflow/blob/master/docs/rfcs/0001-lightweight-rfc-process.md).
+
--- a/adrs/README.md
+++ b/adrs/README.md
@@ -4,7 +4,6 @@ An architecture decision record (ADR) is a document that captures an important a
 ### Goal
 We want to record important design decisions to help current and future contributors of the library to understand the reasons for doing things a certain way.
 
-
 ### Lightweight Process
 
 - Create a new branch on this repository and copy `adrs/0000-template.md` to a new file with the next free number.
@@ -13,7 +12,6 @@ We want to record important design decisions to help current and future contribu
 - Once consensus is reached and approvals given using the Github review system, the PR can be merged.
 
 In case the team has already agreed on a decision and no further async discussion is necessary just go ahead an record it directly without opening any PR.
-
 
 ### Disclaimer
 
--- a/adrs/README.md
+++ b/adrs/README.md
@@ -7,7 +7,7 @@ We want to record important design decisions to help current and future contribu
 ### Lightweight Process
 
 - Create a new branch on this repository and copy `adrs/0000-template.md` to a new file with the next free number.
-- Modify the template to add details about your proposal and add any supporting picture under the `adrs/images` folder.
+- Modify the template to add details about your proposal and add any supporting diagram preferably in [ASCII](https://textik.com/).
 - Make a Pull Request (PR) for your branch.
 - Once consensus is reached and approvals given using the Github review system, the PR can be merged.
 
--- a/adrs/README.md
+++ b/adrs/README.md
@@ -17,5 +17,5 @@ In case the team has already agreed on a decision and no further async discussio
 
 ### Disclaimer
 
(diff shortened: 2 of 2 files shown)


adrs/README.md
```diff
@@ -0,0 +1,21 @@
...
+- Modify the template to add details about your proposal and add any supporting picture under the `adrs/images` folder.
+- Make a Pull Request (PR) for your branch.
+- Once consensus is reached and approvals given using the Github review system, the PR can be merged.
+
+In case the team has already agreed on a decision and no further async discussion is necessary just go ahead an record it directly without opening any PR.
+
+
+### Disclaimer
+
+As of today, the primary audience for these ADRs artefacts is the Stacks Core Team. If circumnstances will change and we feel a broader audience should get involved in the decision making process we could look into evolving our process in something a bit more formal like the internal [Stack Overflow RFC process](https://github.com/StackEng/StackOverflow/blob/master/docs/rfcs/0001-lightweight-rfc-process.md).
```
dancormier: Caught a couple misspellings

```suggestion
As of today, the primary audience for these ADRs artifacts is the Stacks Core Team. If circumstances will change and we feel a broader audience should get involved in the decision making process we could look into evolving our process in something a bit more formal like the internal [Stack Overflow RFC process](https://github.com/StackEng/StackOverflow/blob/master/docs/rfcs/0001-lightweight-rfc-process.md).
```

dancormier (approved):
IMO this LGTM and is GTG! 🚢 

Once this is merged in, I'll add an ADR for Pseudo Private Custom Properties

giamir:
Thanks @dancormier for the review.  🙏
Let's give Ben a chance to share his opinion next week and then we merge it. 🙂

adrs/README.md
```diff
@@ -0,0 +1,21 @@
+# Architecture Decision Records (ADRs)
+An architecture decision record (ADR) is a document that captures an important architecture decision made along with its context and consequences. We strive to keep these records small and lean.
+
+### Goal
+We want to record important design decisions to help current and future contributors of the library to understand the reasons for doing things a certain way.
+
+
```
b-kelly: nit

```suggestion
```

adrs/README.md
```diff
@@ -0,0 +1,21 @@
...
+### Lightweight Process
+
+- Create a new branch on this repository and copy `adrs/0000-template.md` to a new file with the next free number.
+- Modify the template to add details about your proposal and add any supporting picture under the `adrs/images` folder.
+- Make a Pull Request (PR) for your branch.
+- Once consensus is reached and approvals given using the Github review system, the PR can be merged.
+
+In case the team has already agreed on a decision and no further async discussion is necessary just go ahead an record it directly without opening any PR.
+
+
```
b-kelly: nit

```suggestion
```

adrs/README.md
```diff
@@ -0,0 +1,21 @@
...
+An architecture decision record (ADR) is a document that captures an important architecture decision made along with its context and consequences. We strive to keep these records small and lean.
+
+### Goal
+We want to record important design decisions to help current and future contributors of the library to understand the reasons for doing things a certain way.
+
+
+### Lightweight Process
+
+- Create a new branch on this repository and copy `adrs/0000-template.md` to a new file with the next free number.
+- Modify the template to add details about your proposal and add any supporting picture under the `adrs/images` folder.
```
b-kelly: Given this repo's history with images, I don't know if I want to encourage committing images to the repo as it's another thing we have to keep track of size-wise. Maybe we can punt this down the line and encourage images to be added the the GH PR? When the ADR is merged, the PR number should be in the commit message, so there should still be a canonical link to it.

adrs/README.md
```diff
@@ -0,0 +1,21 @@
...
+- Make a Pull Request (PR) for your branch.
+- Once consensus is reached and approvals given using the Github review system, the PR can be merged.
+
+In case the team has already agreed on a decision and no further async discussion is necessary just go ahead an record it directly without opening any PR.
+
+
+### Disclaimer
+
+As of today, the primary audience for these ADRs artifacts is the Stacks Core Team. If circumstances will change and we feel a broader audience should get involved in the decision making process we could look into evolving our process in something a bit more formal like the internal [Stack Overflow RFC process](https://github.com/StackEng/StackOverflow/blob/master/docs/rfcs/0001-lightweight-rfc-process.md).
```
b-kelly: nit: We use the term "Core" (capital C) to refer to the Stack Overflow "Core" codebase. I've further tweaked the wording here to match my personal tastes. Let me know if you disagree and we can change things back.

```suggestion
As of today, the primary audience for these ADRs artifacts is the core Stacks Team. If circumstances change and we feel a broader audience should get involved in the decision making process we could look into evolving our process in something a bit more formal like the internal Stack Overflow RFC process.
```

Side note: I removed the link to the private/internal SO repo. I could go either way on this, but something about linking to a private repo in an OSS project doesn't sit right with me.

adrs/0000-template.md
```diff
@@ -0,0 +1,16 @@
+# [ADR Title]
+
+**Date:** [YYYY-MM-DD]
+
+**Status:** [ 🚧 Proposed | ✅ Accepted | 🚫 Rejected ]
```
b-kelly: Do we need the status field here? Will we ever be committing anything _but_ accepted ADRs? Proposed is a pre-merge concept and rejected never gets merged, which only leaves "accepted" as the sole valid value.

adrs/0000-template.md
```diff
@@ -0,0 +1,16 @@
...
+
+**Related PR:** [#xxxx (when applicable)]
```
b-kelly, line 5: I don't mind having this here in general, but we should also ensure that the PR number is in the commit message (following the standard already laid out in Stacks-Editor w/ conventional commits)

adrs/0000-template.md
```diff
@@ -0,0 +1,16 @@
...
+
+### Issue
+[describe in few sentences the problem we need to solve]
+
+### Decision
+[describe in few sentences the proposed solution to the issue]
+
+##### Disclaimer
+This is just a template: while we expect the above items to appear in every ADRs, feel free to add any additional points as you see fit based on your ADR context (e.g. Consequences, etc...).
```
b-kelly: I think we can more fluidly integrate this section into the template:

```suggestion
### Additional info
[anything else you'd like to add; remove this section if it is empty]
```

b-kelly (commented):
Overall, I like this approach. Something we _may_ consider is simply letting these ADRs live in GH - no files committed at all.

Pros:
- rejected ADRs will still have a place to live
- no additional non-code files littering the git repo
- can use labels, PR/issue templates, etc as necessary
- Can automate steps with GH actions (e.g. adding/removing labels)
- Can use GH Projects to keep track of everything, including status
  - GH projects also supports cross-repo tracking, so this would work well for our other repos too

Cons:
- not easily discoverable outside of GH

In the interest of keeping this simple, I don't mind just continuing forward with this approach, but it's something I'd still like to think about.

adrs/0000-template.md
giamir: I think it's important to have this here for people browsing the markdown directly. It might not be immediate for people that additional information is available in the PR linked in the commit message.

adrs/0000-template.md
giamir: Agree. Will remove the status section and the redundancy.

adrs/README.md
giamir: Thanks for clarifying. Happy to remove the direct link.

adrs/0000-template.md
b-kelly: Sounds good to me! Agreed that having it here is not terribly redundant and greatly improves visibility. I think having it in both places is totally fine.

adrs/README.md
giamir: I understand. Maybe we could incentivise the use of ASCII diagrams for the time being and decide how we want to host images when there is a first use case.

b-kelly: Ooh excellent idea, especially since GH has native support for [mermaid](https://mermaid-js.github.io/mermaid/#/)

giamir:
> Something we may consider is simply letting these ADRs live in GH - no files committed at all.
> - no additional non-code files littering the git repo

The idea of [lightweight ADRs](https://www.thoughtworks.com/en-us/radar/techniques/lightweight-architecture-decision-records) is to create simple artefacts which lives as close as possible to the codebase to decrease the chances for them to go out of sync with the code. I don't see ADRs as litter in this context.

> - rejected ADRs will still have a place to live

We will close rejected ADRs but we won't delete the related PRs so they will still be browsable under the label `adr`

> - can use labels, PR/issue templates, etc as necessary

I think it's important to keep ADRs as lean as possible. I don't think we will need GH extra tools.

> - Can automate steps with GH actions (e.g. adding/removing labels)

If we need automation, I believe the ADR process have become more complicated than it should be. 🙂

> - Can use GH Projects to keep track of everything, including status

I personally would prefer to keep ADRs in the repo and treat them in a similar way we do with code. If there is a strong preference in the team to use GH though I am fine to explore that avenue.

> - GH projects also supports cross-repo tracking, so this would work well for our other repos too

This is a valid point. There might be some ~~ADRs~~ guidelines which affect all our codebases and this is **not** the place for them. We should store here only ADRs that are relevant for the Stacks codebase and they should be fairly detailed.

For high level architecture guidelines which span across all our codebases we need a different solution.
From the top of my ahead I can only think about a separate repo in the same organisation. 
I see this as a separate problem though: how do we capture high level guidelines shared across all our projects.

b-kelly (approved):
@giamir All excellent points, thank you for gut checking my assumptions. This is not a topic that I feel strongly about, so I'm 1000% ok with moving forward with our current approach. Thank you for putting all this together!

PR #1165, opened 2022-10-29
Highlight the current column that a table is being sorted by

allejo:
Instead of squinting and trying to make sense of the little arrow icons in the table headers, can't there be a more visible way of showing which column is currently being used for sorting?

Say no more! Sorting columns now highlights them!!

image
image
image
image

Extracted from #1144


dancormier:
Committed to the wrong branch 🤦 Ignore me!

PR #1169, opened 2022-10-31
refactor(prose): use pseudo-private custom properties

dancormier:
This PR refactors `.s-prose` component styling to use pseudo-private custom properties.

> **Note**
> This PR should result in no visual changes for the `.s-prose` component.


dancormier:
@StackExchange/stacks this component is complex enough that I'd like some eyes on it to ensure I didn't introduce regressions. I _think_ this shouldn't cause visual changes to prose, but I want some verification on that. Thanks!

giamir (approved):
@dancormier it looks good to me. I have confronted the old version and the refactored one and there are no visual differences I could spot. I even ran backstop test suite and it looked everything ok also there. Feel free to merge.

PR #1172, opened 2022-11-03
docs: add ADR for pseudo-private-custom-properties

dancormier:
This PR adds an ADR for our _pseudo-private-custom-properties_ approach to component styling.

**Note to reviewers**: I didn't catch this minor issue in my [ADR PR](https://github.com/StackExchange/Stacks/pull/1160) review, but the ADR template jumps from heading level 1 to 3. This is a _very minor_ issue, but worth discussing if we care sooner rather than later.


adrs/0001-pseudo-private-custom-properties.md
```diff
@@ -0,0 +1,28 @@
+# Refactor component styling using pseudo-private custom properties
+
+**Date:** [2022-11-03]
+
+### Issue
+Component styles are difficult to maintain. This is due to:
+
+1. Disjointed styles .less files amongst components. Within components, we currently sometime will output contextual, variant, modifier, interaction styles with the base component class and sometimes we don’t. These styles will might be in alphabetical order, by some logical order, in the order added, or seemingly randomly. Many component style .less files include a large preamble and table of contents that are seldom updated to reflect the current state of the component.
```
giamir: > Many component style .less files include a large preamble and table of contents that are seldom updated to reflect the current state of the component.

For my understanding: do PPCPs have self documenting characteristics as well?

giamir: > Disjointed styles .less files amongst components. Within components, we currently sometime will output contextual, variant, modifier, interaction styles with the base component class and sometimes we don’t.

How complicated would be to add a simple snippet example from one of our files? Sometimes few lines of code are easier to digest than words. 🙂

adrs/0001-pseudo-private-custom-properties.md
```diff
@@ -0,0 +1,28 @@
...
+**Date:** [2022-11-03]
+
+### Issue
+Component styles are difficult to maintain. This is due to:
+
+1. Disjointed styles .less files amongst components. Within components, we currently sometime will output contextual, variant, modifier, interaction styles with the base component class and sometimes we don’t. These styles will might be in alphabetical order, by some logical order, in the order added, or seemingly randomly. Many component style .less files include a large preamble and table of contents that are seldom updated to reflect the current state of the component. 
+2. Complex cascade for component styles within variants, modifiers and theme contexts. This makes scanning .less files cumbersome, maintaining styles difficult, and changes liable to cause regressions.
+
+### Decision
+Defining a structure for component styles using pseudo-private custom properties (PPCP) addresses the issues and includes a few other benefits:
```
giamir, line 19: > addresses the issues

So if we add a snippet highlighting the problem in the issue section can we have here the same snippet with PPCPs applied so that the reader can see at a glance the improvement?

giamir (approved):
Looks good to me Dan. Thanks for writing this up! 🙏
I have only added some suggestions. Once addressed, feel free to merge directly. 🙂

giamir:
> the ADR template jumps from heading level 1 to 3

Feel free to adjust the template (and this ADR to heading level 2). Personally I don't mind as long as we stay consistent. 🙂

adrs/0001-pseudo-private-custom-properties.md
```diff
@@ -0,0 +1,28 @@
+# Refactor component styling using pseudo-private custom properties
+
+**Date:** [2022-11-03]
```
giamir: Maybe before merging the PR we should add a link here to this PR for future reference (see ADR template). 🙂

ondrejkonec:
Hi guys!

I would like to make a small comment about the structure of LESS files and the structure of components.

Stacks uses `foo--child` syntax for child elements in a component (eg `s-btn__primary`) and `foo__variant` syntax for variants (`foo--item`).

Is it some standard syntax? It reminds me of BEM, but it's exactly the opposite of how it should be (where the child element is like `foo__child` and the variant is `foo--variant`)

Am I overlooking something or is it your specific notation?

Thanks!

adrs/0001-pseudo-private-custom-properties.md
dancormier: The way they're implemented in Stacks, yes. Specifically, the custom properties adhere to a naming convention to make them as clear as possible. They follow this format (all groups after the prefix separated by `-`):

- `--_` prefix to indicate it's a PPCP
- two-letter abbreviation of the component (if the component name is comprised of multiple words, I take the first letter of the first two words, otherwise user the first two letters of the component name)
- the child component name if applicable
- atomic class equivalent abbreviation or name if it exists, otherwise the logical name of the property
- interactive state if applicable

Examples:

- `--_ba-bc`: `border-color` on `.s-badge`
- `--_lp-footer-fd`: `flex-direction` on `.s-link-preview .s-link-preview--footer` 
- `--_ta-bg-hover`: `background-color` on `.s-tag` when hovering

I took this approach to make discerning the property's purpose obvious and to prevent collisions. I generally prefer self-documenting code with some support from short, inline comments over verbose code comments and find that in practice, large comments tend to get outdated often and can become a liability. Like most things, it depends, but that's my general feeling.

Should I include these details in the ADR or is that overkill?

dancormier: Oh, I should have proofread this! It's tough to understand. I'll give this section an edit. A snippet might be tough, since the issue could be quite verbose, but I'll give it a shot.

Thanks for pointing this out!

adrs/0001-pseudo-private-custom-properties.md
dancormier: Good idea! I'll try to work that in.

adrs/0001-pseudo-private-custom-properties.md
giamir: Thanks for explaining Dan. 🙂

You read my mind. These details/conventions we started using should be part of the "Decision" section of the ADR in my opinion.

Agree with you about self documenting code (or even using tests as documentation) before resorting to long comments. 👍

dancormier: It just occurred to me that linking to [`lib/css/components/_styles-template.less`](https://github.com/StackExchange/Stacks/blob/develop/lib/css/components/_styles-template.less) would probably make sense here _**or**_ if we're ok with these ADRs being verbose, I could add the contents of that file here and remove the file itself. Thoughts?

b-kelly: I vote to add it into the ADR in a language tagged code block. Imo, we should aim to keep unshippable code out of the core library folders as best we can (sometimes it's unavoidable, but it's my personal philosophy).

Something like the following:

~~~markdown
```less
.s-component-name {
    //...
}
```
~~~

dancormier:
> Is it some standard syntax? It reminds me of BEM, but it's exactly the opposite of how it should be (where the child element is like `foo__child` and the variant is `foo--variant`)

@ondrejkonec You're correct! It's inverted BEM syntax. The choice to do this predates my tenure at Stack Overflow and, when I've searched/asked for a reason for this, I haven't found a satisfying justification for it.

I'd prefer to conform to BEM syntax, _but_ it would be a huge breaking change at this point to update something that is a minor annoyance/stumbling block. The juice isn't worth the squeeze, IMO.

adrs/0001-pseudo-private-custom-properties.md
```diff
@@ -0,0 +1,28 @@
...
+Defining a structure for component styles using pseudo-private custom properties (PPCP) addresses the issues and includes a few other benefits:
+
+- **Increase consistency within our codebase**. This is not specific to PPCP as much as it applies because we’ve chosen an unified approach towards structuring our component styles.
+- **Increased maintainability and readability**. With PPCP, we change custom property values as close to the base component level as possible and nearly flattening the cascade. This ensures that we keep the cascade as simple as possible, increasing readability, making maintenance easier, and decreasing the potential for styling conflicts in the process.
+
+#### Secondary benefits
+- **Prepare us for native CSS component styles**. Adopting a PPCP approach gives us the opportunity to reduce our dependency on a preprocessor and anticipate a future where a preprocessor is entirely unnecessary.
+
+### Additional info
+Inspiration for this approach initially came from [Lea Verou’s presentation titled “CSS Variable Secrets at CSS Day 2022](https://www.youtube.com/watch?v=ZuZizqDF4q8&amp;ab_channel=WebConferencesAmsterdam)
```
b-kelly: ultra nit: we can simplify the link a bit for a slightly cleaner source code (source: visiting this link and using YouTube's "Share" button)

```suggestion
Inspiration for this approach initially came from [Lea Verou’s presentation titled “CSS Variable Secrets at CSS Day 2022](https://youtu.be/ZuZizqDF4q8)
```

adrs/0001-pseudo-private-custom-properties.md
```diff
@@ -0,0 +1,28 @@
...
+
+#### See also
+[Stacks components _style-template.less](https://github.com/StackExchange/Stacks/blob/develop/lib/css/components/_styles-template.less)
```
b-kelly, line 122: Personally, I'm in favor of adding code samples at the bottom here. GH assigns ids to all of its headers, so we could even hand wire up a little table of contents here if the snippets make the ADR too long. Alternatively, we could take advantage of some commonmark/html rendering edge cases with `<details>` to shorten things up a bit (note the carefully placed whitespace to terminate the html_block tokens):

~~~markdown
<details><summary>Component template</summary>

```less
.s-component-name {
    foo: bar;
}
```

</details>
~~~

adrs/0001-pseudo-private-custom-properties.md
b-kelly:
```suggestion
**Date:** 2022-11-03

**Related PR:** #1172
```

b-kelly (changes requested):
I'm in favor of this approach. Independently, I rediscovered this technique in a side project and really enjoyed it, so I'm happy this is an actual movement in the front-end scene.

I'm marking as "Request changes" because I'd love to see a few very short examples laid out that we can canonically point to when someone asks "how _should_ I write PPCPs in the Stacks codebase".

ondrejkonec:
> > Is it some standard syntax? It reminds me of BEM, but it's exactly the opposite of how it should be (where the child element is like `foo__child` and the variant is `foo--variant`)
> 
> @ondrejkonec You're correct! It's inverted BEM syntax. The choice to do this predates my tenure at Stack Overflow and, when I've searched/asked for a reason for this, I haven't found a satisfying justification for it.
> 
> I'd prefer to conform to BEM syntax, _but_ it would be a huge breaking change at this point to update something that is a minor annoyance/stumbling block. The juice isn't worth the squeeze, IMO.

Thanks @dancormier for explaining this, now I get it. I also think that it is better to use the established syntax and not to invent something new that has no added value (it can just be confusing). I understand that it is something minor. Thanks again!

adrs/0001-pseudo-private-custom-properties.md
giamir: I agree with Ben. I wonder if there is an opportunity to remove the template file from that lib folder now that we have dedicated section for architecture decisions. 🤔

giamir:
@dancormier Thanks for following up on this.
I just wanted to give you a heads up about how the current markdown is rendered by Github:
https://github.com/StackExchange/Stacks/blob/bb29ca23ee48f0632f4d44ffc2b25d14f2da0d64/adrs/0001-pseudo-private-custom-properties.md
You can see that there is a bullet section not behaving as you would expect.
Also the "Related PR" value is not an hyperlink to this PR somehow. 
I really like the detailed code snippets you have included. 👍

dancormier:
@giamir good catch!

I plugged that markdown into a comment input on GitHub, checked out the "Preview" tab, and it rendered as expected… I guess I can't trust those to render the same 😞  I'll resolve this issue shortly.

dancormier:
Ok! I resolved the rendering issues so I believe this is good to merge now. Gone mergin'!

PR #1180, opened 2022-11-10
docs(banner): improve example interaction

dancormier:
I got tired of clicking that "Update example" button every time I changed the example settings, so I made it so that any change to the settings just goes ahead and updates the example.

Here's an example of the example update example:


dancormier (commented):
I'm giving myself a "LGTM" on this one 🚀

PR #1184, opened 2022-11-15
refactor(inputs): use pseudo-private custom properties

dancormier:
This PR refactors component styling to use pseudo-private custom properties for the following components (completed refactors checked):

- [x] `.s-check-control`
- [x] `.s-check-group`
- [x] `.s-checkbox`
- [x] `.s-input`
- [x] `.s-input-fill`
- [x] `.s-input-icon`
- [x] `.s-input-message`
- [x] `.s-select`
- [x] `.s-radio`
- [x] `.s-textarea`
- [x] Validation state classes
    - [x] `.is-disabled`
    - [x] `.is-readonly`
    - [x] `.has-success`
    - [x] `.has-error`
    - [x] `.has-warning`

> **Note**
> This PR should result in no visual changes to the above components.

All of these components are being refactored at once because their styling is currently tightly-coupled. This PR will decouple styling where ever it's logical.

### Deprecations and removals

This PR gets rid of the one and only reference to `.s-input.has-focus` in Stacks. I only found it used once in Core and that was in our developer theme testing page.


lib/css/base/reset-normalize.less
```diff
@@ -311,7 +311,9 @@ button:-moz-focusring,
  */
 
 fieldset {
-  padding: 0.35em 0.75em 0.625em;
```
b-kelly: I'd like to be cautious tweaking either of the resets, as consumers may be depending on those for _other_ styles.

I'd recommend making this change in your component css instead if it is required.

lib/css/components/description.less
```diff
@@ -0,0 +1,9 @@
+.s-description {
```
b-kelly, line 1: Good call pulling this one out. We use this "component" in a few places, but never really document it.

lib/css/exports/global-modifiers.less
```diff
@@ -0,0 +1,7 @@
+.is-disabled,
```
b-kelly: Imo, we should inline this one style anywhere it is used. This is something that could bite us in the rear later on, especially if we're including it in `exports` (essentially making it part of our public LESS api)

b-kelly (commented):
I haven't reviewed visually yet, but I love that you're splitting these components off into their own files.

I have a few small nits that I think could be easily rectified before merging.

lib/css/exports/global-modifiers.less
dancormier: Ah, I didn't really consider the implications of including this in our `/exports/` directory. Yeah, this really isn't the best home for these classes.

The issue is that these classes have this rule applied without any parent/sibling classes applied, so I'm not sure where the best permanent home would be (besides a file like this one). My first instinct is to move this file to `lib/css/atomic` since they're applied _kinda_ like atomic classes, but really they're modifiers 🤷‍♂️ Any thoughts?

ondrejkonec: Hi guys! Quick note: as you know, this State rules comes from SMACSS approach (http://smacss.com/book/type-state) and they are mostly tied to changing behavior using JS. I think they are good for naming component states (you can see very quickly what is going on in CSS and what variant it is) and IMHO they don't need to be in a separate file because they have no general meaning.

ondrejkonec: If you plan to use for e.g. "is-hidden" or "is-invisible" that can be used everywhere, I think some /atomic class folder should work here.

lib/css/base/reset-normalize.less
dancormier: Ah, good catch! I [restored this style within the the reset and created fieldset.less](https://github.com/StackExchange/Stacks/pull/1184/commits/c364106cbbc6c591917756f0da7ed5bfc71796bb) which includes the styles previously in `inputs.less`. I chose the standalone `lib/css/base/fieldset.less` file since the style is not specific to any specific component.

lib/css/exports/global-modifiers.less
dancormier: @giamir @b-kelly any ideas wrt [my comment](https://github.com/StackExchange/Stacks/pull/1184#discussion_r1038536330)?

giamir: I haven't reviewed in details this PR but at a first glance I also find odd to see these classes globally available across the whole document.  As a rule of thumbs we should always think twice before exposing more API surface to consumers because we are implicitly making the system harder for us to change/evolve.
I would prefer if we could keep those classes scoped to the component that needs them (similar to what Ben suggested). This will create a bit of duplication but I would not mind to prioritize encapsulation over performance in this context.
If we have instances where these classes are just used in random contexts then we could put the rules in a place where it's clear we will deprecate them in the next major release (maybe a `pending-deprecation.less` file?) How much work do you foresee in Core for replacing these classes usage?
By the way the deprecation strategy ADR and the css dev bundle experiment is still in my todo list. I should probably create an issue for that but won't be able to tackle it before next year.
I hope this rambling is somewhat useful. 🙂

giamir: >  If we have instances where these classes are just used in random contexts then we could put the rules in a place where it's clear we will deprecate them in the next major release (maybe a pending-deprecation.less file?) How much work do you foresee in Core for replacing these classes usage?

@dancormier @b-kelly Do you think is possible to deprecate those classes in our next major release? How painful would that be?

dancormier: > How much work do you foresee in Core for replacing these classes usage?

I think it'd be a pretty heavy lift to replace those classes in Core. There's lots of usage of `.is-disabled` and `.has-error` that would need to be replaced. Often, these classes are set with JS conditionally, so time would be needed to understand context and prevent regressions.

> How painful would that be?

I think it would be pretty painful and honestly I feel like this is something we should punt on. In the context of this PR, nothing would materially change for consumers wrt these `.is-` and `.has-` classes. I'm good with being overruled if I'm being too pessimistic, but felt I should voice my opinion.

dancormier: Outside of the deprecation consideration, where could be the best place for these styles @giamir @b-kelly? `lib/css/atomic`?

lib/css/components/link.less
```diff
@@ -68,6 +68,13 @@ a,
         }
     }
 
+    // MODIFIERS
+    fieldset[disabled] & {
```
b-kelly, line 87: Move fieldset stuff all into `fieldset.less`. Two ways to think about it:

1. Styles are targeting the child, put them in the child file
2. Styles are in the context of the parent, put them in the parent file

This is not a hill I'm willing to die on, just food for thought.

lib/css/stacks-static.less
```diff
@@ -1,22 +1,11 @@
...
-//  visit https://stackoverflow.design/
-//
-//  This is where all the magic happens.
-//
-//  ============================================================================
-//  $   STATIC ELEMENTS
-//      The following items are elements which we DO NOT allow communities
-//      to modify via variables.
-//  ----------------------------------------------------------------------------
+// stacks-static.less contains styles which we DO NOT allow communities to modify via variables
```
b-kelly, line 1: One note to make here is that when we share Stacks from SO to other projects/sites (e.g. StackExchange.com) _only_ the static styles are included (for legacy theming reasons that don't matter any longer now that we're using css variables after we remove `stacks-dynamic.less`).

No action necessary.

lib/css/components/checkboxes-radios.less
```diff
@@ -0,0 +1,181 @@
...
+    .s-check-group & {
+        margin-top: calc(var(--su2) + var(--su1)); // 3px
+    }
+
+    input& {
+        flex-shrink: 0;
+    }
+
+    // STATE
+    &::-ms-check { // TODO do we still need this? I don't think so, but help me Windows peeps!
```
b-kelly: Double checked this - this was to support Edge Legacy and IE (which we do not support). This is safe to remove.

lib/css/components/checkboxes-radios.less
```diff
@@ -0,0 +1,181 @@
...
+    border: var(--_ch-baw) solid var(--_ch-bc);
+
+    appearance: none;
+    cursor: pointer;
+    font-size: inherit;
+    height: 1em;
+    margin: 0; // A guard against Core's default margins
+    outline: 0;
+    vertical-align: middle;
+    width: 1em;
```
b-kelly, line 46: Looks like these styles were formally gated behind a `supports` media query? Still necessary with our supported browser list?

https://github.com/StackExchange/Stacks/blob/develop/lib/css/components/inputs.less#L289

b-kelly: Double checking the support list, we're widely in the clear here (Chrome 1, Safari 3, Firefox 54)

lib/css/components/checkboxes-radios.less
```diff
@@ -0,0 +1,181 @@
...
+
+    // CONTEXTUAL STYLES
+    .highcontrast-dark-mode({
+        &:checked {
+            --_ch-bc: var(--blue-300);
+            outline: var(--su-static1) solid var(--black);
+        }
+    });
+
+    .dark-mode({
```
b-kelly, line 101: PR NOTE: Review visually, added style

lib/css/components/checkboxes-radios.less
```diff
@@ -0,0 +1,181 @@
...
+        &:checked {
+            --_ch-bg: var(--black);
+        }
+    });
+
+    // STATES
+    &:checked {
+        --_ch-baw: 0.30769231em;
+        --_ch-bc: var(--theme-secondary-400);
+        --_ch-bg: var(--white); // TODO seems to not change anything. Investigate
```
b-kelly: Can we clean up this TODO?

lib/css/components/checkboxes-radios.less
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    border-radius: var(--br-circle);
+}
+
+.s-checkbox,
+.s-radio:not(:checked) {
+    .has-error &,
```
b-kelly: Note to self - this could likely be a mixin that is shared amongst all inputs. Doing so would help to deduplicate values

lib/css/components/checkboxes-radios.less
```diff
@@ -0,0 +1,181 @@
...
+.s-check-group {
+    --_cg-fd: column;
+
+    // MODIFIERS
+    &&__horizontal {
+        --_cg-fd: row;
+    }
+
+    // CHILD ELEMENTS
+    legend.s-label { // TODO HACK? <legend> isn't respecting gap...
```
b-kelly: optional: Personal preference to keep the width low (and sometimes putting on the same line as an opening bracket screws with formatting).

Ignore if you like.

```suggestion
    // TODO HACK? <legend> isn't respecting gap...
    legend.s-label {
```

lib/css/components/link.less
b-kelly: This was originally applied to `a,.s-btn,.s-link`. We're covering `a, .s-link`, but not `.s-btn` I believe.

https://github.com/StackExchange/Stacks/blob/develop/lib/css/components/inputs.less#L91-L93

lib/css/components/select.less
```diff
@@ -0,0 +1,194 @@
+.s-select {
+    --_se-arrow-bc: currentColor transparent;
+    --_se-arrow-size: var(--su-static4); // Constant // TODO create name/styling convension for constants
```
b-kelly: Cleanup TODO?

lib/css/components/select.less
```diff
@@ -0,0 +1,194 @@
...
+    .has-error &,
+    .has-warning & {
+        position: relative;
+    }
+
+    .is-disabled & {
+        --_se-arrow-bc: var(--bc-darker) transparent;
+    }
+
+    &&__sm {
```
b-kelly, line 32: Sizing could also likely benefit greatly from a mixin that is shared between all inputs

lib/css/components/select.less
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    .has-warning & {
+        --_se-select-bc: var(--yellow-600);
+        --_se-select-bc-focus: var(--yellow-600);
+        --_se-select-bs-focus: 0 0 0 var(--su-static4) var(--focus-ring-warning);
+    }
+
+    // CHILD ELEMENTS
+    &,
```
b-kelly: Added - double check visually. To make this more safe we should probably do:

```suggestion
    select&,
```

lib/css/components/select.less
```diff
@@ -0,0 +1,194 @@
...
+    & > select {
+        &::-webkit-contacts-auto-fill-button {
```
b-kelly: These may need to be pulled out into a mixin since it's fairly complicated and is shared amongst a few components

lib/css/components/select.less
```diff
@@ -0,0 +1,194 @@
...
+        &::-moz-focus-inner {
+            outline: none !important;
+        }
+
+        &:-moz-focusring {
+            color: transparent;
+            text-shadow: 0 0 0 #000;
+        }
+
+        &::-ms-expand {
```
b-kelly: Double check - can almost assuredly remove (Edge Legacy)

lib/css/components/select.less
```diff
@@ -0,0 +1,194 @@
...
+
+            cursor: not-allowed;
+        }
+
+        // CHILD ELEMENTS
+        &::-moz-focus-inner {
+            outline: none !important;
+        }
+
+        &:-moz-focusring {
```
b-kelly: PR Note - double check, necessary or is this on a standards track now?

lib/css/components/select.less
```diff
@@ -0,0 +1,194 @@
...
+            color: transparent;
+            text-shadow: 0 0 0 #000;
+        }
+
+        &::-ms-expand {
+            display: none;
+        }
+
+        // INTERACTION
+        &:focus {
```
b-kelly, line 109: Potentially abstract back out into a mixin?

lib/css/components/toggle-switches.less
```diff
@@ -5,6 +5,14 @@
     --_ts-multiple-bg: unset;
     --_ts-multiple-fc: var(--black-500);
 
+    fieldset[disabled] & {
+        &,
+        & label {
```
b-kelly, line 10: Optional: simplification? removes a level of indentation, but don't know if it is cleaner

```
    fieldset[disabled] &, fieldset[disabled] & label {
```

lib/css/exports/global-modifiers.less
b-kelly: Looking at some of the other code, we may need to keep around an e.g. `input-utils.less` file for mixins. These could likely go in there as well.

b-kelly (changes requested):
Code only review, will follow up with a visual review

lib/css/components/link.less
dancormier: > Styles are in the context of the parent, put them in the parent file

I opted for option 2 in the spirit of component styles being self-contained. Basically, my goal is that anyone trying to see what could be affecting a given component's styling can look in a single file. Big picture, if we every decided to get rid of or rename a component, all the styles would (optimally) live in the component's `.less` file and hopefully be slightly easier to handle without leaving crust around our codebase.

---

> This was originally applied to a,.s-btn,.s-link. We're covering a, .s-link, but not .s-btn I believe.

I'm not sure how we missed it, but I added this style to [buttons.less](https://github.com/StackExchange/Stacks/blob/43e3ebe364639827fc12b7b0a525c388c56f1d77/lib/css/components/buttons.less#L60) in https://github.com/StackExchange/Stacks/commit/939f1025b0e3a4b108637f3cb9e1adf3c6b2f5ca

lib/css/components/select.less
dancormier: AFAICT, we're good to dump this. I tested and saw no difference in FF with or without this rule.

image

Furthermore, [MDN says](https://developer.mozilla.org/en-US/docs/Web/CSS/:-moz-focusring):
> Non-standard: This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

lib/css/components/select.less
dancormier: I'm removing it. It's hard to _even find_ documentation on it, so I'm pretty we can trash it. Best docs I found https://udn.realityripple.com/docs/Archive/Web/CSS/::-ms-expand

lib/css/components/toggle-switches.less
dancormier: I honestly prefer what we have already, but I think we'd be wise to look out for ugly nested `&`. I've been bit by misinterpretation there and I'll keep in mind where this style might be a little more confusing than it's worth.

lib/css/components/select.less
dancormier: I added a mixin in this commit https://github.com/StackExchange/Stacks/pull/1184/commits/718c52ee7e194d55b04f313349f61ad91613d047. I went pretty programmatic since we're using pseudo-private custom properties. It came out looking a little wacky, though I think the mixin is pretty effective. Let me know if it's too wacky (or not wacky enough).

lib/css/components/select.less
dancormier: This would only partially dedupe in `.s-select` and `.s-input`/`.s-textarea`, so I don't think an abstraction would be worth it here.

lib/css/components/checkboxes-radios.less
dancormier: Added `.validation-states` mixin here: https://github.com/StackExchange/Stacks/pull/1184/commits/2086b63f5460027b9d19bebb6c2283bf2060ff18

lib/css/components/checkboxes-radios.less
dancormier: This resolves a bug where the center dot in checked radio elements would fill with the wrong color in dark mode.

### Before

image

### After
image

PR #1185, opened 2022-11-17
refactor(sidebar-widgets): use pseudo private custom properties

giamir:
This PR refactors component styling to use pseudo-private custom properties for the `.s-sidebarwidget` component.

> **Note**
> This PR should result in no visual changes to the sidebar widgets.


lib/css/components/sidebar-widgets.less
```diff
@@ -0,0 +1,245 @@
...
+}
+
+// [1] We must support things like collapsible (in particular invisble) elements, wrapper elements,
+// etc. Therefore every .-content and .-header must stand on its own; we cannot rely on things
+// like :first-child, because the *first* child may not be the first *visible* child, and it may
+// also be the :first-child of some wrapper. This is why every .-header and .-content has a
+// border-top. But because you shouldn't see the first visible item's or header's top border
+// (the widget itself provides for that border), we shift everything up by one pixel with the following pseudo-element.
+
+// [2] However, because of the above, the top item's divider line now sits above the .s-sidebarwidget's top
```
dancormier: Seeing as these comments no longer live in their original context, they could probably benefit from some slight rewording.

lib/css/components/sidebar-widgets.less
```diff
@@ -0,0 +1,245 @@
...
+        position: absolute;
+        top: -1px;
+        right: -1px;
+        left: -1px;
+        // TODO: this makes no sense. revisit.
+        height: calc(2px + var(--br-sm)); // we need 2px border + 2px border radius to have the correct corner shape
+        border-radius: var(--br-sm);
+        pointer-events: none; // if the top item is clickable, then we don't want to prevent clicking the top 2 pixels
+        clip-path: polygon(0% 0%, 100% 0%, 100% 50%, 0% 50%); // [3]
+        border: 1px solid var(--_sw-bc);
```
dancormier: I'd suggest creating a new custom property named `--_sw-after-bc`. If it's always the same as `--_sw- bc`, then we can define it as `--_sw-after-bc: var(--_sw-bc)`.

Basically, I want to ensure that a contributor can quickly parse what _might_ change from just a glance at the variables defined at the start of the file.

lib/css/components/sidebar-widgets.less
```diff
@@ -0,0 +1,245 @@
+.s-sidebarwidget {
+    --_sw-content-px: calc(var(--su16) - var(--su1)); // subtract 1px for border
+    --_sw-content-spacing-inner: var(--su12); // the spacing between two adjacent simple items
+    --_sw-content-spacing-outer: var(--su16); // the spacing at the start/end of a group of simple items, as well as between a complex item and its separator line
+    --_sw-bc: var(--bc-medium);
+    --_sw-content-bc: var(--bc-light);
```
dancormier, line 10: IMO, a little redundancy is a good thing here. For instance, I see that `--_sw-content-bc` is being used for both the `content` and `header` child elements. I'd prefer to do something like:

```
--_sw-content-bc: var(--bc-light);
--_sw-header-bc: var(--_sw-content-bc);
```

This would be so a contributor can quickly parse that a) the same value us used for the border color in both `content` and `header` and b) modifying `--_sw-content-bc` will also modify the value for `--_sw-header-bc`.

lib/css/components/sidebar-widgets.less
```diff
@@ -0,0 +1,245 @@
...
+            border-top-right-radius: var(--br-sm);
+        }
+
+        &:active {
+            outline: none;
+        }
+
+        border-top: 1px solid var(--_sw-content-bc);
+        padding: var(--_sw-content-spacing-inner) var(--_sw-content-px);
+        background: var(--_sw-header-bg);
```
dancormier: Since it's not redefined anywhere, we can just add the value of `--_sw-header-bg` (`var(--black-025)`) here and get rid of the custom property.

lib/css/components/sidebar-widgets.less
```diff
@@ -0,0 +1,245 @@
+.s-sidebarwidget {
+    --_sw-content-px: calc(var(--su16) - var(--su1)); // subtract 1px for border
+    --_sw-content-spacing-inner: var(--su12); // the spacing between two adjacent simple items
+    --_sw-content-spacing-outer: var(--su16); // the spacing at the start/end of a group of simple items, as well as between a complex item and its separator line
```
dancormier, line 5: These three custom properties seem like constants. It took me a moment to realize that, but I'd want that difference to be clearer. Naming conventions for custom properties that act as scoped constants is an issue I'd like to solve.

@b-kelly you said you've seen the PPCP approach in the wild. Have you noticed a convention for differentiating constant from variable properties in CSS?

lib/css/components/sidebar-widgets.less
```diff
@@ -0,0 +1,245 @@
...
+    --_sw-bc: var(--bc-medium);
+    --_sw-content-bc: var(--bc-light);
+    --_sw-header-bg: var(--black-025);
```
dancormier: I think we can get rid of this custom property. See my comment below where it's referenced.

lib/css/components/sidebar-widgets.less
```diff
@@ -0,0 +1,245 @@
...
+
+    & &--header {
```
dancormier: I've been taking the simple approach of alphabetizing each grouping of elements. It's the simplest way to order and allows for easy jumping to what you might be looking for. 

_That said_, it's something I don't think we need to be dogmatic about and I'm open to whatever ordering makes sense in context. @giamir Did you have a specific order for these child elements?

lib/css/components/sidebar-widgets.less
```diff
@@ -0,0 +1,245 @@
...
+        &.s-sidebarwidget__small-bold-text {
+            font-size: var(--fs-caption);
+            font-weight: bold;
+            .s-sidebarwidget--action {
+                font-weight: normal;
+                line-height: calc(1.3 * var(--fs-caption));  // line-height should be the same as in the outside element, so the header and action baselines line up
+            }
```
dancormier: **Another ordering nitpick**: To match the ordering standard in the ADR, I'd suggest placing the style rules for a given element to come at the end of it's block of styles. It might seem weird at first, but I've been doing it that way to ensure that any redefinition of a given PPCP happens before the style is declared.

lib/css/components/sidebar-widgets.less
```diff
@@ -0,0 +1,245 @@
...
+                padding: 0;
+            }
+        }
+
+        padding: calc(var(--_sw-content-spacing-outer) - var(--_sw-content-spacing-inner)) calc(var(--_sw-content-px) - var(--_sw-content-spacing-inner));
+        border-collapse: separate;
+        border-spacing: var(--_sw-content-spacing-inner);
+    }
+
+    &:not(.s-anchors) a:not(.button):not(.s-btn):not(.post-tag):not(&--action):not(.s-user-card--link) {
```
dancormier, line 173: Oh, that's something I haven't noticed before…`.button`! Oh boy!

dancormier (changes requested):
Awesome work Giamir! Thanks so much for taking this on.

I haven't done a thorough comparison to our `.s-sidebarwidget` component to ensure there's no regressions, but I did take a peek at the code.

The most significant changes I'd like to see is closer adherence to the formatting defined in the [PPCP ADR](https://github.com/StackExchange/Stacks/blob/gbuoncristiani/refactor-sidebar-widgets/adrs/0001-pseudo-private-custom-properties.md). If you have opinions on logical ordering/naming, I'm absolutely open to them and we can modify the ADR.

lib/css/components/sidebar-widgets.less
giamir: Would not the same apply then also for the `--_sw-header-bc` property you suggested to introduce in your previous comment? It's only use in a single place too. https://github.com/StackExchange/Stacks/pull/1185#discussion_r1026795089

lib/css/components/sidebar-widgets.less
giamir: Makes sense. Will address this.

lib/css/components/sidebar-widgets.less
giamir: Will address this.

lib/css/components/sidebar-widgets.less
giamir: You essentially want to keep the relationship between content bc and header bc higher up in the file but still keep them separated for readability. Got it. Will address this.

lib/css/components/sidebar-widgets.less
giamir: I now understand that for `--_sw-header-bc` you want to preserve the fact that the color is inherit by `--_sw-content-bc` which you can only do with custom properties. Will address this.

lib/css/components/sidebar-widgets.less
giamir: What will be the interim alternative here?  Inlining the constants?
I haven't seen PPCP in the wild but if we want to borrow a concept from the JS community I guess we could simply make the word uppercase (e.g. `_SW-CONTENT-PX`)?

lib/css/components/sidebar-widgets.less
giamir: Will address this.

lib/css/components/sidebar-widgets.less
giamir: no, I just left the same order they were in when I started the refactoring. Will address this and order alphabetically.

lib/css/components/sidebar-widgets.less
giamir: I think I got that rule. This was a genuine miss on my end because of the nesting. Will address it.

lib/css/components/sidebar-widgets.less
giamir: yeah, not sure about what we can do about those generic random classes `.button` and `.post-tag`. It's probably too risky to remove them at this point I suppose - it will certainly be a breaking change.
Resolving this conversation for now.

giamir:
Thanks @dancormier for reviewing the PR. I have addressed all your points expect the one about "differentiating constants from variable properties". Perhaps @b-kelly could have a look and we can then take a call about what to do and move on.

From my end I still have to fix the [weird specificity issue](https://github.com/StackExchange/Stacks/pull/1185/files#diff-bf53a5d6f8282a028c66917aa7bb6bb53ce62188ef7a00909205c1ef3452bad7R43) I encountered with expandables inside sidebars.

On top of that I have 2 generic questions that goes beyond the scope of this PR:
- It looks like there is a bit of inconsistency between how we name css classes and the file names (e.g. `sidebar-widgets.less` and the class `s-sidebarwidget` or `buttons.less` and `s-btn`). Is there any rule we follow? Shall we create one? 
- I also observed that we use atomic classes as mixins in the component classes ([see this example](https://github.com/StackExchange/Stacks/pull/1185/files#diff-bf53a5d6f8282a028c66917aa7bb6bb53ce62188ef7a00909205c1ef3452bad7R24)).
Is this something we are doing consciously?

lib/css/components/sidebar-widgets.less
giamir: For the time being I differentiated constants from variables with a comment similar to what has been done already [here](https://github.com/StackExchange/Stacks/blob/develop/lib/css/components/expandable.less#L40). I am happy to leave this conversation open but I believe this should not block the PR from being merged since it goes beyond its initial scope. If we make a decision about it we can document it and add a task to our backlog to rename the constants all at once (it should not be a lot of work if they are already highlighted by a comment).

giamir:
> From my end I still have to fix the [weird specificity issue](https://github.com/StackExchange/Stacks/pull/1185/files#diff-bf53a5d6f8282a028c66917aa7bb6bb53ce62188ef7a00909205c1ef3452bad7R43) I encountered with expandables inside sidebars.

So the reason why this was happening was due to an incorrect usage of the expandable content class in one of our documentation example (see [this commit](https://github.com/StackExchange/Stacks/pull/1185/commits/946a1b1764fb3484d9f393984318a00525195f36) for more context). 

Longer explanation:
The refactoring increased the [specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) weight for the sidebar widgets `content` children (from `.s-sidebarwidget--content` `0-1-0` to `.s-sidebarwidget .s-sidebarwidget--content` `0-2-0`) . This made it exactly the same weight of `.s-expandable .s-expandable--content` `0-2-0`.
`.s-expandable .s-expandable--content` styles were not able to override anymore`.s-sidebarwidget .s-sidebarwidget--content` styles in a situation where an element would have both `s-expandable--content` and `.s-sidebarwidget--content` classes applied like in the incorrect example we had in the documentation page:
```
<div class="s-expandable" id="favorites">
    <div class="s-sidebarwidget--content s-sidebarwidget__items s-expandable--content">
        <div class="s-sidebarwidget--item">
            <i>You have no favorite jobs</i>
        </div>
    </div>
</div>
```
This was fixed by moving the `s-expandable--content` class to a dedicated parent element as it is in fact done for all other examples in the page.
```
<div class="s-expandable" id="favorites">
    <div class="s-expandable--content">
        <div class="s-sidebarwidget--content s-sidebarwidget__items">
            <div class="s-sidebarwidget--item">
                <i>You have no favorite jobs</i>
            </div>
        </div>
    </div>
</div>
```
I run a [quick search](https://github.com/search?q=repo%3AStackEng%2FStackOverflow+s-expandable--content&type=code) in the core codebase to see if we have any occurence where `s-expandable--content` is used in the same element together with `s-sidebarwidget--content`.  There are none which should give us confidence to move ahead without having to take care of regressions in core.

<hr>

@dancormier: This PR should now be good to be merged after an additional review.

b-kelly:
@giamir Regarding the questions you've outlined:

- I don't have a strong preference on file naming, so long as we're consistent. Some files may have more than one component in them, so the plural makes sense for these. Otherwise, my personal opinion is that every component gets its own file and we standardize on singular nouns (e.g. `button.less`)
- I have used the atomic classes as mixins in the past for things like width (no longer necessary), but nowadays we probably shouldn't use them inside our components since they all use `!important`. Imo, any atomic class that is sophisticated enough to want to reuse should be made into a mixin (such as [scrollbar-styles](https://github.com/StackExchange/Stacks/blob/develop/lib/css/exports/constants-helpers.less#L67)) or css variables (such as [--lh-*](https://github.com/StackExchange/Stacks/blob/develop/lib/css/exports/constants-type.less#L140-L145))

lib/css/components/sidebar-widgets.less
b-kelly: Regarding constants (as in, a reused, static value that should not be changed by others), I'd recommend simply using Less variables to bake it in on the server-side. If we want to remain client-side only, then there's no other way to really make our "pseudo-private" into "true private". One thing I've seen before that may of use to us is to use a double underscore (`--__c-constant-var`) to really underscore (heh) the point that you _really_ shouldn't be changing those.

Of the two options, I prefer the former. This may be a nuance we'd like to record in the ADR.

giamir: Thanks @b-kelly for the input. I actually remember vaguely having [this conversation](https://github.com/StackExchange/Stacks/pull/1091#issuecomment-1273040356) about using less variables for "constants" a while back already in an older PR. 
I also do prefer the former approach although I remember it was going against the principle of sticking to native CSS as much as possible.
[This article](https://www.smashingmagazine.com/2018/05/css-custom-properties-strategy-guide/#local-static-variables-are-ok-sometimes) from Smashing Magazine support keeping static variable in preprocessor variables too.
Let's wait for @dancormier to be back, make a decision together, update our ADR and potentially create a tech task to go through the codebase and adjust all constants to whatever approach we have decided.

As I said already in my previous comment. This decision should not block the merging of this PR though.

giamir:
Thanks @b-kelly for the context.

> I have used the atomic classes as mixins in the past for things like width (no longer necessary), but nowadays we probably shouldn't use them inside our components since they all use !important.

I started by removing the `.d-block` usage from this component.

I wonder if we could add a point to [this contribution rules list](https://github.com/StackExchange/Stacks/blob/develop/CONTRIBUTING.md#css--less) saying "Do not use atomic classes as mixins". Thoughts?

lib/css/components/sidebar-widgets.less
dancormier: Using Less variables is a simple way to differentiate constants from variables and it would be easy for us to change in the future if the need arises. I'm 👍 on that approach.

@giamir do you think this calls for a ticket along the lines of "_use Less variables for component constants_"?

---

**A note on moving away from Less variables (and general dependency on Less)**: I have a ticket for an ADR wrt less reliance on Less but, until that's done, I'll just mention I'm not dogmatic about this. There are and will be exceptions where Less just makes the most sense and this is one of them.

dancormier:
> * I don't have a strong preference on file naming, so long as we're consistent. Some files may have more than one component in them, so the plural makes sense for these. Otherwise, my personal opinion is that every component gets its own file and we standardize on singular nouns (e.g. `button.less`)

👍

> * I have used the atomic classes as mixins in the past for things like width (no longer necessary), but nowadays we probably shouldn't use them inside our components since they all use `!important`. Imo, any atomic class that is sophisticated enough to want to reuse should be made into a mixin (such as [scrollbar-styles](https://github.com/StackExchange/Stacks/blob/develop/lib/css/exports/constants-helpers.less#L67)) or css variables (such as [--lh-*](https://github.com/StackExchange/Stacks/blob/develop/lib/css/exports/constants-type.less#L140-L145))

There might be a couple of exceptions to this ([modal.less](https://github.com/StackExchange/Stacks/blob/develop/lib/css/components/modals.less#L28) for example), but I generally agree.

> I wonder if we could add a point to [this contribution rules list](https://github.com/StackExchange/Stacks/blob/develop/CONTRIBUTING.md#css--less) saying "Do not use atomic classes as mixins". Thoughts?

👍

dancormier (approved):
This is looking great! To save some back-and-forth, I added some minor cleanup commits ([diff](https://github.com/StackExchange/Stacks/compare/cefbad0..a5e43da)). Looks like this is good to go!

lib/css/components/sidebar-widgets.less
giamir: > @giamir do you think this calls for a ticket along the lines of "use Less variables for component constants"?

@dancormier Yes, I think that's a good idea. Would you mind adding the ticket? I also believe we could put it under the "Operational Maintenance" epic and prioritize it as any other tech debt/investment. We should also adjust the PPCP ADR to reflect this decision - we can either do it straight away or as part of the ticket. 🙂

PR #1191, opened 2022-11-23
Monorepo ADR

b-kelly:
Adds an ADR detailing the "why" and partial "how" for migrating Stacks to a monorepo with multiple packages.


adrs/0002-monorepo.md
```diff
@@ -0,0 +1,34 @@
...
+
+The Stacks repo is growing in complexity and will soon require many different dependencies that are optional for most consumers.
+
+## Decision
+
+Despite shipping many different concepts (atomic classes, CSS components, JS components), the interdependency chains are relatively linear. This means that we can fairly trivially split the repo into smaller portions.
+
+Our current codebase could be split like so:
+- Atomic classes (no dependencies)
+- CSS Components (depends on Atomic components)
```
giamir: I would prefer to keep CSS components and JS components in the same package (in the spirit of a true component based architecture - see also [folder-by-feature vs folder-by-type](https://softwareengineering.stackexchange.com/questions/338597/folder-by-type-or-folder-by-feature)).  I would even go a step further and create folders for individual components where we keep style (less), behavior (stimulus - when applicable) and component test files (tbd - perhaps wtr + testing-library) together.
In an ideal scenario even the "docs" would be in the same component folder but I understand that it won't be an easy thing to achieve given our current setup. I would be ok to settle for docs in a separate package (revaluate perhaps at a later stage).

adrs/0002-monorepo.md
```diff
@@ -0,0 +1,34 @@
...
+- JS Components (depends on Stimulus, Popper.js, CSS + Atomic components)
+- Documentation site (depends on JS components, 11ty + plugins)
```
giamir: if we keep js and css components in the same package then this will depends on a "stacks-components" package and transitively on the atomic classes package (perhaps "stacks-base")

adrs/0002-monorepo.md
```diff
@@ -0,0 +1,34 @@
...
+
+When the codebase expands in the future, we'll be able to add the following without worrying about consumers needing to import dependencies they don't need:
+- Front-end component library or libraries (e.g. React, Web Components)
+- Large, complex components (e.g. autocomplete)
+- Extra build chain items, such as a [Style Dictionary](https://amzn.github.io/style-dictionary/)
```
giamir: Nice that you mention a design tokens tool here - it might be something worth exploring in the future (especially if we would have to start supporting non-web platforms too). 🙂

adrs/0002-monorepo.md
```diff
@@ -0,0 +1,34 @@
...
+
+We'd also be able to fold the [Stacks-Icons](https://github.com/StackExchange/Stacks-Icons) and [Stacks-Utils](https://github.com/StackExchange/Stacks-Utils) repos into this main monorepo as well.
```
giamir: I like this idea.

adrs/0002-monorepo.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Additional Info
+
+On the technical front, we could manage the monorepo with [npm workspaces](https://docs.npmjs.com/cli/v9/using-npm/workspaces?v=true) or another dedicated tool such as [Rush](https://rushjs.io/pages/intro/welcome/), [Lerna](https://lerna.js.org/) or [Bit](https://bit.dev/).
```
giamir: I have had mixed experiences with npm workspaces in my previous team (e.g. poor support of selective deps resolutions, upgrade deps pain, etc...) but in the end it did its job and it's nice you don't need to install any extra tool. Workspaces in yarn is way more mature though (or at least it was when I evaluated both tools half a year ago or so - maybe npm caught up a bit in the meantime).

adrs/0002-monorepo.md
```diff
@@ -0,0 +1,34 @@
...
+
+In regards to changelog / release notes generation, we'd continue using Conventional Commits, but moving to use [release-please](https://github.com/googleapis/release-please) which [does support monorepos](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md).
```
giamir: I don't have hands on experience with release-please but I am excited to try that out. I believe we should automatize our release process as much as possible (including publishing packages to npm).
I have used conventional commits and I really like it. We might be even able to enforce those conventions with commitlint and nudge people in the right direction with comittizen (in my previous project most teammates they defaulted to run `npx cz` instead of using the git cli to create commits). 🙂

giamir (commented):
Thanks @b-kelly for getting this started. ❤️

I left few comments here and there but overall I like where this is going.
We should think about how to prioritize the work and get started. This is purely technical work so it might fit into that 30% (ballpark) of tech tasks we should pick up every iteration.
Part of this work will also become a blocker when we want to start to build more complex UI components. I really like you brought this up already.

I suggest to finalize the ADR, merge it and then start breaking down what we need to do in bite-size tech tasks. Happy to support there and also excited to pick some of those tasks up. 🙂

adrs/0002-monorepo.md
b-kelly: The reason I had suggested we separate the two is two-fold:
1. If you only need the Stacks classes and not the JS components, you don't have to pull them in
2. We're looking to modernize our JS component story, which means potentially having more than one supported component library (e.g. Stimulus + Web Components / React / Svelte / whatever), at least for a short while

That being said, I'm open to discussion on this point. I'm very much interested in hanging with the team, hashing out the pros / cons and coming to a hard decision. I'll schedule something for us for before the end of the calendar year.

adrs/0002-monorepo.md
b-kelly: I did some exploration on this during my personal "Learn, Share, Grow" day. _Technically_, we already support a "non-web" (non-CSS?) platform - Figma 😉

adrs/0002-monorepo.md
b-kelly: Definitely down to do some discovery on this. Ultimately, I don't really care what tool we use so long as it stays out of the devs' way.

adrs/0002-monorepo.md
b-kelly: I've [played with release-please](https://github.com/StackExchange/stacks-utils/blob/main/.github/workflows/release-please.yml) only a little, but I've been using [commit linting](https://github.com/StackExchange/Stacks-Editor/blob/main/.husky/commit-msg) (plus [a GH action](https://github.com/StackExchange/Stacks-Editor/blob/main/.github/workflows/changelog.yml) to enforce) over in Stacks-Editor for a while now to _great success_. It's something that we've unofficially adopted across _all_ of the Stacks repos recently, but haven't yet had the opportunity to add formal enforcement mechanisms.

adrs/0002-monorepo.md
giamir: Point 1 could be achieved even if we ship a single npm "stacks-components" package. 
We will still have a way for consumers to opt in for css only if they want to I suppose.

The reason why I am suggesting for the component styles to go hand in hand with Stimulus is because I don't see us reusing any of that source code (Stimulus and Less) when we will build web components (or anything other component library). 
My thinking as of now is that new components will depend exclusively on the atomic classes package and work in any context without depending on global styles being applied to a page or not.

I believe styles will have to be rewritten but I might be miscalculating things. Discovery will certainly help.
In case we will reuse the same component styles we have today in Stacks with the new components then I would understand why we would keep components styles in a separate package.

dancormier: It's just my ¢2, but I agree with Giamir that a single package for both CSS components and JS components makes the most sense. 

On top of Giamir's points, it's plausible that we could convert a CSS component to a JS component at some point. Correct me if I'm wrong, but I imagine we'd need to deprecate the component in the CSS package and port it to the JS component package. That seems like it would be messy and a single Stacks components package makes more sense to me intuitively.

adrs/0002-monorepo.md
dancormier: I've used Commitizen in the past and appreciated it (but I'm open). _If_ we used Commitizen, I'd suggest configuring it on the side of `warn` to start, because it can get too annoying to have it yell at you for trying to commit with a message like `cleanup` when you're gonna squash it anyways. I want to make sure our tooling resides in the Goldilocks zone of annoyingness.

dancormier (commented):
Overall, I'm pretty 👍 on this ADR

adrs/0002-monorepo.md
giamir: I think this is the only pending topic before we could merge this ADR too. 🙂

giamir:
@b-kelly heads up: before merging remember to change the sequential number to `0003` since the testing strategy PR took the `0002`.

adrs/0002-monorepo.md
b-kelly: This is actually something I feel strongly about and would be better suited to discuss over a hangout. In the meanwhile, I'm going to go ahead and merge as-is, since that line is just a suggestion and implementation can differ from what I have here.

We can always come back and update this later. The PR for that will serve as its own little mini ADR, since it'll contain the reasoning for the change.

giamir: I have mixed feelings about this PR being merged in without reaching an agreement on this point first. This is primarily because what we merge we consider automatically as approved (we don't have a status flag). 
Can I suggest to add a comment/line directly in the ADR content to remind ourselves that further discussion is required about packages organisation.
Don't get me wrong, I like to move things forward fast but I also think it's ok to slow down a bit when we cannot agree on something as a team. @b-kelly Can I leave this small adjustment with you together with scheduling a meeting for us to talk further about this topic when the time is right? Thanks and sorry if this request sounds a bit pedantic. 🙂

b-kelly: @giamir I'll make an adjustment. I see this section as a "for example" section, not as something I was proposing in this PR. The wording `could be` alludes to that. Would it make more sense if I reworded it like:

> For example, our current codebase could be split like so

giamir: Sounds good to me. Thanks. 
I guess the more important action item is to come together as a team, take a decision and retrospectively justify our reasoning in this ADR too (we are not in a rush though). 🙂

b-kelly: Wording tweaked in 84a917cfa59ca69c789b4e425406de6a3448227f

PR #1194, opened 2022-11-28
Introduce Testing to Stacks

giamir:
This PR builds on top of the [PoC Testing PR](https://github.com/StackExchange/Stacks/pull/1137) and supports the [Testing Strategy ADR](https://github.com/StackExchange/Stacks/pull/1195) with a concrete implementation.

## Context

**Prerequisit**: we are now tracking visual regression tests baseline images with git lfs (github supports that by default) and you need to have the CLI installed in your local machine to pull and upload those images. You can follow [this guide](https://docs.github.com/en/repositories/working-with-files/managing-large-files/installing-git-large-file-storage) to do so.
Apart from that there is nothing else to learn. You can use git as normal.

~~There are several file changes because we had many files violating the existing linting rules. You can review those fixes in isolation by checking the individual commits:~~
~~- fix style rules violations (https://github.com/StackExchange/Stacks/commit/664bef55569cb68d850bd5c638e2d28c74f22e6c)~~
~~- fix eslint rules violations (https://github.com/StackExchange/Stacks/commit/53937b39ee621399b3459205dce391fab8d02927)~~
~~- fix prettier formatting violations (https://github.com/StackExchange/Stacks/commit/b5d1aea0eb4dd6126e99c659017df3c27a6a6905)~~

**Fix existing linting and formatting issues has been now moved into a [separate PR](https://github.com/StackExchange/Stacks/pull/1198).** 

I also updated the [README.md](https://github.com/StackExchange/Stacks/blob/688c07195a8f074403a0c03726566f33b0396961/README.md) extensively and added a basic [GH workflow](https://github.com/StackExchange/Stacks/blob/688c07195a8f074403a0c03726566f33b0396961/.github/workflows/main.yml) to build and run tests on every push. 

### Testing Library Overhead in Browser/ESM Context
Testing Library was initially created to run in a Node.js environment and as a consequence it is relying on several dependencies which supports only CJS and cannot be made ESM-compatible in an easy way by bundler plugins (in our case we use `@rollup/plugin-node-resolve` and `@rollup/plugin-commonjs`).
By using `@web/test-runner` to run our tests all the code is evaluated in a real browser context and everthing is dynamically imported as an ES module. Since not all our dependencies are ESM we need a bundler (in our case `rollup`) which takes care to resolve (plugin-node-resolve) and "esmify" (plugin-commonjs).

I had to use [npm ovverides](https://github.com/StackExchange/Stacks/pull/1194/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R105) to fix a transitive dependency of testing library (`aria-query`) to a version which does not consume `deep-equal` (they introduced that last month in a minor release). `deep-equal` is using a couple of old dependencies (like `call-bind`) which rollup was not able to easily "esmify" and resolve. This could be fixed by instructing rollup to do something special about those deps with a small plugin. I did not want get into that yet since we are not even sure if we will keep testing-library (therefore the temporary npm override).

Testing library comes with a great a11y-friendly API which is familiar to most frontend developers. It also come with a tooling overhead cost to make it [compatible with esm](https://github.com/testing-library/dom-testing-library/issues/995) and, as a heads up, if we will use shadow dom in the future we also will have to work around [some other limitations](https://github.com/testing-library/dom-testing-library/issues/413). 

I personally like very much the structure testing-library bring to our tests and I would be more than willing to pay up some of these costs. I might even foresee a nice opportunity to contribute back to the OS community over time since we are not the only one facing similar challenges ([see this Adobe's RFC](https://github.com/adobe/spectrum-web-components/pull/420)). That said I would also understand if we want to stay away from this complexity and use directly DOM APIs to query elements under test and simulate interactions.  

### Web Dev Server Temporary Patches

 Web Dev Server (the dev server behind our test runner) is currently unable to execute correctly the latest version of the `@rollup/plugin-commonjs`. This is something I am not too worried about since the maintainer are aware of the issues and already working on a fix. For the time being we have a [patch file](https://github.com/StackExchange/Stacks/pull/1194/files#diff-9af47a07f4fc1b314266b307f79267a8dc9d7a5c9f53216462efd5ef586ac1d4) which we could easily discard once the fix will be [released](https://github.com/modernweb-dev/web/pull/2050).

On top of that we have found an [issue where the rollup adapter won't resolve virtual modules in windows environments](https://github.com/modernweb-dev/web/issues/2078). [A PR with a fix](https://github.com/modernweb-dev/web/pull/2079) is waiting to be reviewed and merged by the mantainers. In the meantime we point npm to an [ad-hoc patched tarball](https://github.com/StackExchange/Stacks/pull/1194/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R54).

### Prettier and Style Formatting

~~I have experienced a strange behavior running prettier against our less files: the formatter will break the style syntax.
For the time being I have decided to [exclude less files from prettier](https://github.com/StackExchange/Stacks/pull/1194/files#diff-b640b344ee7f3f03d2a443795a5d0708ef50e2e6e34214109ab2aad13ad6ba98R4) since they already get some format check with stylelint. We might want to revisit this decision though.~~

**Fix existing linting and formatting issues has been now moved into a [separate PR](https://github.com/StackExchange/Stacks/pull/1198).** 

### Typescript/Eslint checking for test files.

At the moment test files are not linted. This is because I did not want to mess around with the [tsconfig.json](https://github.com/StackExchange/Stacks/blob/d03dea1fff80b9e8c34dcdf5bb5e24dc0d0159f0/lib/tsconfig.json) file. Furthermore I have encountered a strange problem where the assertions (e.g. `expect`) we import from the `@open-wc/testing` package seems to don't inherit correctly the related `Chai` types. Further investigation is required although I think this is not very urgent given that we will need to readjust the structure of our repository soon to adhere to the Monorepo ADR. That might be a good time to address this problem.


README.md
```diff
@@ -32,8 +34,70 @@ To contribute to Stacks documentation or its CSS library, you’ll need to build
...
+npm run test:unit
+```
+or if you prefer watch mode run:
+```sh
+npm run test:unit:watch
+```
+
+### Visual Regression Tests
+
+**Prerequisit:**
```
b-kelly: typo:

```suggestion
**Prerequisite:** 
```

README.md
```diff
@@ -32,8 +34,70 @@ To contribute to Stacks documentation or its CSS library, you’ll need to build
...
+```
+or if you prefer watch mode run:
+```sh
+npm run test:unit:watch
+```
+
+### Visual Regression Tests
+
+**Prerequisit:** 
+In order to pull and upload baseline images you need to have `git lfs` installed in your local machine. Please follow [this guide](https://docs.github.com/en/repositories/working-with-files/managing-large-files/installing-git-large-file-storage) to install that.
```
b-kelly, line 82: We could add that command to the the `prepare` script in `package.json`. We do something similar in Stacks-Editor to initialize Husky locally: https://github.com/StackExchange/Stacks-Editor/blob/main/package.json#L28

b-kelly (changes requested):
Preliminary comments while I'm attempting to run locally:

- Can we split out the formatting fixes into it's own commit? I'm totally fine with committing those straight to `develop` without a PR, assuming they're whitespace only changes.
- I can't get the tests to run I keep getting `Failed to fetch dynamically imported module` style errors. I'll debug this locally and see what's up there.

b-kelly:
The issue is that the path of script modules being pulled in are pointing to the `file://` path on my drive, which the browsers are blocking. The issue is definitely with `fromRollupWithFix`. Switching that out to the original `fromRollup` results in a completely different set of files, but they're at least being served via the dev server instead of pulled in from disk.

README.md
giamir: `git-lfs` is something we will be installing **globally** in the user machine. 
Personally I'd rather have the user opt-in explicitly for that one-off installation.
Since the tool is only necessary for people who wants to run (visual) tests we might add a `pretest` warning if `git-lfs` is not in path - we could further optimise it with a console prompt (y/n) style. I will add the warning for now.
Thanks for the idea! 🙂

giamir:
> Can we split out the formatting fixes into it's own commit? I'm totally fine with committing those straight to develop without a PR, assuming they're whitespace only changes.

Formatting and linting are already isolated in their own commits so I could easily cherry pick them to any other branch.
They are not only whitespaces though (especially the fix for the linter errors - although they are still trivial things). Here is an idea I will create another PR just for the formatting/linting fixes and ask @dancormier to quickly double check it before merging it to develop. This branch will be based off that PR then to improve the diffs.

Here are the commits I will cherry pick to the new PR:
- fix style rules violations (https://github.com/StackExchange/Stacks/commit/664bef55569cb68d850bd5c638e2d28c74f22e6c)
- fix eslint rules violations (https://github.com/StackExchange/Stacks/commit/53937b39ee621399b3459205dce391fab8d02927)
- fix prettier formatting violations (https://github.com/StackExchange/Stacks/commit/b5d1aea0eb4dd6126e99c659017df3c27a6a6905)

**Update:** this has now been addressed in [this PR](https://github.com/StackExchange/Stacks/pull/1198)
**Update2:** formatting PR has been merged to develop. Now the diffs should be easier to read. @b-kelly

giamir:
> The issue is that the path of script modules being pulled in are pointing to the file:// path on my drive, which the browsers are blocking. The issue is definitely with fromRollupWithFix. Switching that out to the original fromRollup results in a completely different set of files, but they're at least being served via the dev server instead of pulled in from disk.

This is probably a windows specific issue. I tested both in macos and ubuntu (for the pipeline) and everything worked out fine. I have a windows VM too - will try to make the time to reproduce your error in that environment.

**Update**: @b-kelly I have described the problem in [this issue](https://github.com/modernweb-dev/web/issues/2078) and made a [PR with a fix](https://github.com/modernweb-dev/web/pull/2079). In the meantime we point npm to an [ad-hoc patched tarball](https://github.com/StackExchange/Stacks/pull/1194/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R54). You should be able to continue your review and run the tests successfully in windows now.

README.md
giamir: @b-kelly I'd like to postpone this bit if it's ok with you as a "nice to have" we can do any time. While I was setting up my windows environment to debug the issue you experienced I realized git installation from chocolatey comes with `lfs` by default. I think `lfs` won't cause a lot of friction but if you have something specific in mind feel free to go ahead and update this PR.

web-test-runner.config.mjs
```diff
@@ -0,0 +1,76 @@
...
+    ],
+
+    // settle on a font-family for visual regression tests across operating systems
+    testRunnerHtml: (testFramework) =>
+        `<html>
+            <body>
+                <style>
+                    body {
+                        font-family: Arial;
+                    }
```
dancormier: This doesn't account for components that have an explicit font family definition (`.s-code-block`, for example). I setting the font-related custom properties here instead to avoid false negatives due to differing fonts in different operating systems. 

```suggestion
                    body {
                            --ff-sans: Arial;
                            --ff-mono: "Courier New";
                    }
```

Sources for common fonts included in OSes: [Mac OS](https://support.apple.com/en-us/HT212587), [Windows](https://learn.microsoft.com/en-us/typography/fonts/windows_11_font_list)

> **Note**
> I didn't redefine `--ff-serif` since [we prefer `Georgia`](https://github.com/StackExchange/Stacks/blob/gbuoncristiani/testing/lib/css/exports/constants-type.less#L83), which is common amongst different OSes.

---

All of the above considered, I think we _still_ could wind up with false failures (even considering the 5% threshold) due to OS font-rendering nuances, particularly on tests where slight line height or letter spacing differences could add up. There are heavy-handed solutions to this problem (like running tests in a container), but I think we're best off crossing that bridge when we get there.

dancormier (changes requested):
[I like it](https://www.youtube.com/watch?v=tH3phKBbVLs&t=269)!**\*** I have a minor note on the font-family rule for the visual regression tests, but that's my only note. Great job @giamir!

### A few notes for posterity (no action required):

- I love seeing BackstopJS get removed ❤️ It helped at times, but I'm happy to see WTR/Playwright here now.
- I feel like now/soon would be a good time to switch to a component-based file structure. I went searching for the tooltip test, and that little bit of friction reminded me that we should make the switch. I'm glad to make that change soon, as long as we can time it thoughtfully to avoid a ton of conflicts.

---

### \* This note is for my own entertainment. Please ignore it.

For years, I've [used](https://github.com/StackExchange/Stacks/pull/745#pullrequestreview-740020174) [the](https://github.com/StackExchange/Stacks/pull/745#pullrequestreview-740020174) [phrase](https://github.com/StackExchange/Stacks/pull/1198#pullrequestreview-1201575231) "_I like it_" in place of _LGTM_. When I heard the phrase in my head, it's always been in this yell, like "I LIKE IT!". I couldn't quite place it, but it finally came to me today. _Turns out_, I've been hearing [the very end of the song Indiscipline](https://www.youtube.com/watch?v=tH3phKBbVLs&t=269) from the classic 1981 album Discipline by English progressive rock band King Crimson. Mystery solved!

web-test-runner.config.mjs
giamir: > This doesn't account for components that have an explicit font family definition

Will apply your suggestion and we can take action as we start encountering problems (see comment below). 

> All of the above considered, I think we still could wind up with false failures

Absolutely agree, we should address this as soon as we see visual tests becoming unreliable. If I find some extra time I will already start experimenting with a containerized solution on the side. It will of course require extra software installed locally but flaky/unreliable tests are even worse than no tests at all.
Will keep this out of scope for this PR though.

giamir:
Thanks @dancormier for the review. 

> I feel like now/soon would be a good time to switch to a component-based file structure. I went searching for the tooltip test, and that little bit of friction reminded me that we should make the switch. I'm glad to make that change soon, as long as we can time it thoughtfully to avoid a ton of conflicts.

If @b-kelly agrees as well, we should create a task for our backlog and do this even before starting the work to migrate to the monorepo structure. 
I remember there was an open topic about [keeping component styles in a separate package from the respective TS](https://github.com/StackExchange/Stacks/pull/1191#discussion_r1031720000). Personally I think style, behavior and tests for these components should live in the same package but be exported/published in a way that the consumers can exclusively use styles if they want to. I am making the assumption that any other components we will be creating in future (e.g. web components or framework X) won't attempt to reuse any of those components styles (apart from the atomic classes which will almost certainly leave in a different package).

b-kelly:
It runs locally on my machine now! 🎉 🎉 

The tests pass on Chromium+Webkit, but not on Firefox (looks like text rendering stuff):

s-tooltip-diff
s-tooltip

I'm re-reviewing the code again now.

README.md
b-kelly: I'm ok with pushing this out. It wasn't so much a hard requirement, but a request for thoughts / comments. I'm a little jumpy about committing images to the repo (because of prior mistakes made in this repo), but that's not really based in reality.

.gitattributes
```diff
@@ -0,0 +1 @@
+screenshots/** filter=lfs diff=lfs merge=lfs -text
```
b-kelly, line 1: 🎉 Nice, this should keep us from accidentally committing images to the git history

.gitignore
```diff
@@ -78,11 +78,13 @@ dist/
...
+/screenshots/**/failed/**
 
 # GitHub Actions #
 ###############
 /.github/main.workflow
 .vscode/
+
+# JetBrains Editors #
+###############
+/.idea
```
b-kelly: Not your fault, but making a suggestion since I'm just now seeing it:

```suggestion

# Editors #
###########
.vscode/
/.idea
```

lib/test/s-tooltip.test.ts
```diff
@@ -0,0 +1,62 @@
...
+                role="button"
+                data-controller="s-tooltip"
+                title="tooltip content"
+                data-s-tooltip-placement="bottom-start"
+            >
+                Hover tooltip popover
+            </button>
+        `);
+
+        expect(screen.queryByRole("tooltip")).to.be.null;
```
b-kelly, line 22: Ooooh, I like this! Forcing the tests to use the accessibility tree is also a way to force a11y testing in general.

package.json
```diff
@@ -38,11 +41,20 @@
...
+    "@rollup/plugin-replace": "^5.0.1",
     "@stackoverflow/stacks-editor": "^0.8.2",
     "@stackoverflow/stacks-icons": "^3.1.0",
+    "@testing-library/dom": "^8.19.0",
+    "@testing-library/user-event": "^14.4.3",
     "@typescript-eslint/eslint-plugin": "^5.42.0",
     "@typescript-eslint/parser": "^5.44.0",
-    "backstopjs": "^6.1.4",
+    "@web/dev-server-esbuild": "^0.3.3",
+    "@web/dev-server-rollup": "https://github.com/giamir/web/raw/fix-dev-server-rollup/packages/dev-server-rollup/web-dev-server-rollup-0.3.19.tgz",
```
b-kelly, line 54: I'm ok with this as a short-term fix (until upstream releases a patched version).

No changes necessary.

b-kelly (approved):
Thank you for getting this in place. I'm very excited about the approach you've implemented here. Other than the missing upstream fix, this looks good to merge!

package.json
giamir: Yeah, I think so too. We are also covered from a security perspective since we have the content integrity check thanks to the `package-lock.json`. I will follow up again about my [one-liner PR](https://github.com/modernweb-dev/web/pull/2079) for WTR.

giamir:
> The tests pass on Chromium+Webkit, but not on Firefox (looks like text rendering stuff)

That is not good, flaky tests are as good as no tests (maybe even worse). I will add an issue for us to explore running the visual regression test suite in a container as Dan suggested. Might use tomorrow's LSG to learn a thing or two about that.

giamir:
Thanks @b-kelly for the review. I will go ahead then and merge this and the testing ADR PR. 
I have followed up also with the maintainers of WTR to get that upstream fix merged.
Finally I have a todo item to write an issue for running visual regression tests in a container.
Very excited to have testing tools at our disposal from now on in Stacks!

PR #1195, opened 2022-11-28
Testing Strategy ADR

giamir:
This PR adds an initial testing strategy ADR for Stacks.
It is supported by this Testing PR (implementation): https://github.com/StackExchange/Stacks/pull/1194


adrs/0003-testing-strategy.md
```diff
@@ -0,0 +1,49 @@
...
+- Every component should be accompanied by related component and visual regression tests (atomic classes are excluded)
+
+### Component/Unit Tests
+These tests should provide full coverage of the component being tested. Happy path, edge and error cases should be all covered.
+
+Guidelines:
+- Tests should not reach for internal implementation/methods but rather tests the component through its public APIs (from the point of view of a consumer)
+- Complex DOM-agnostic business logic can be extracted in separate functions which can be unit tested in isolation
+
+### Component Visual Regression Tests
```
b-kelly: What visual regression library do you propose to use here? I'd personally like use to cease use of backstop full-stop. Internally, the product teams use [mabl](https://www.mabl.com/), which does have support for visual regression tests.

giamir: I would prefer to use a lightweight option to start with which allows us to keep the tests checked in the source code. I also believe it is more important to have a visual regression suite for each individual components instead of taking screenshots of entire documentation pages.
I would recommend to start using [this WTR plugin](https://www.npmjs.com/package/@web/test-runner-visual-regression) as a first step and see how it goes. ([we won't be alone using this approach](https://github.com/adobe/spectrum-web-components/blob/899b6eb36206cea9c61a890ba304136616b4cd71/web-test-runner.utils.js#L13)).
I have already setup a working example in [this PR](https://github.com/StackExchange/Stacks/pull/1194). It is also integrated in our new CI workflow.

Perhaps we could give mabl a try if we want to keep regression tests for the documentation pages (exactly what backstop is doing today). Personally I consider those tests a bit too broad though. I would prefer to invest on other categories such as a11y automated testing.

adrs/0003-testing-strategy.md
```diff
@@ -0,0 +1,49 @@
...
+
+Guidelines:
+- Baseline images for these tests should be committed to the repository with `git lfs` to avoid bloating the history.
+- When relevant, screenshots at different viewport sizes should be taken
+
+### Additional Guidelines
+
+- All tests and linters runs in a GH workflow for every commit on any branch.
+
+- BackstopJS tests will be gradually retired in favor of the more targeted component visual regression tests approach.
```
b-kelly: I don't think we actually have any backstop tests (and if we do, I don't think I've ever run them), so I don't see any reason we can't change this to "immediately retire[d]"

b-kelly (commented):
Overall, I don't have any issues with this ADR. I'll review your PR to see how I feel about the technologies you've chosen and come back to approve this PR afterwards

adrs/0003-testing-strategy.md
giamir: I run them manually sometimes while reviewing PRs which should not have any visual changes (usefulness: low - lot of false positives cluttering the output). So I would be happy to get rid of them straight away. 
I also saw @dancormier running them though - @dancormier would you feel comfortable ditching backstop straight away? As I said in my other comment if we want to keep snapshotting entire documentation pages perhaps we could look into mabl (or just use playwright again 😀).

adrs/0003-testing-strategy.md
```diff
@@ -0,0 +1,49 @@
...
+
+- Tests will be gradually added to existing components as developers need to change them. New components should always be accompanied by tests.
+
+- Consider this is a starting point for Stacks Testing Strategy. More categories, such as a11y testing, might be introduced in future ADRs.
```
dancormier: FWIW, we already test for accessibility issues with mabl, though it's kinda cumbersome to run. I anticipate that [Playwright will lend it self well to a11y testing](https://playwright.dev/docs/accessibility-testing).

adrs/0003-testing-strategy.md
dancormier: Honestly, BackstopJS is so noisy that I've recently resorted to doing manual image diffing of docs pages. I'm good with ditching backstop straight away.

dancormier (commented):
ADR looks good to me! I'll take the same approach as Ben and review [the other testing PR](https://github.com/StackExchange/Stacks/pull/1194) before approving this one.

adrs/0003-testing-strategy.md
giamir: Alright I will update both PRs (ADR and implementation) to ditch BackstopJS straight away.

adrs/0003-testing-strategy.md
giamir: Agree, we could do all sort of things really: 
- https://modern-web.dev/docs/test-runner/commands/#accessibility-snapshot
- https://open-wc.org/docs/testing/chai-a11y-axe/

I would recommend to have a dedicated ADR about accessibility testing though to don't bloat this one up. I will document the task as "operational maintenance" in our internal backlog so that we don't forget.

adrs/0003-testing-strategy.md
giamir: Resolving this conversation since I have been given 2 thumbs up. 🙂

giamir:
@b-kelly @dancormier Thanks for the review, the [implementation PR ](https://github.com/StackExchange/Stacks/pull/1194)should now be working also in windows environments. Looking forward to your feedback.

b-kelly (approved):
Reviewed the implementation PR and I like the approach this describes. Approved there, so approving here. 🦐 it!

PR #1196, opened 2022-11-29
refactor(topbar): use pseudo-private custom properties

dancormier:


dancormier:
Before I go too crazy with this one, I'd like to get some thoughts on the worthiness of pseudo-private custom properties (PPCP) for this component. 

Since the topbar is specifically targeted with so many theme variables that _aren't_ intended to be private, I'm hesitating at refactoring this with _more_ custom properties. @b-kelly do you think it's wise to refactor the topbar styles but omit the PPCPs?

b-kelly:
@dancormier I'm biased, but I think the topbar styles can be lowered in priority. When I wrote the topbar component I took a pseudo-PPCP approach (yes, I used pseudo twice there. For fun :trollface: Treat it as a double negative). The result is that I'm essentially using the PPCP-style architecture, but the custom properties are purposefully not private. Now, we _could_ make a few of them private if we wanted to decrease our API surface area, but I think that work has less priority than getting the other components out of the way first.

dancormier:
Closing this for now as it seems like the juice isn't worth the squeeze for this refactor.

PR #1198, opened 2022-12-01
fix(linters): address all existing violations in the codebase

giamir:
This PR fixes existing issues with our linters and formatters.
Running `npm run lint` now executes `stylelint`, `eslint` and `prettier --check` concurrently.
All checks are now passing in preparation for the [testing PR](https://github.com/StackExchange/Stacks/pull/1194) where we will start running linters (as well as tests) in a github workflow.

### Prettier and Style Formatting
I have experienced a strange behavior running prettier against our less files: the formatter will break the style syntax.
For the time being I have decided to exclude less files from prettier since they already get some format check with stylelint. We might want to revisit this decision though.

### DIFF
--- a/lib/ts/controllers/s-popover.ts
+++ b/lib/ts/controllers/s-popover.ts
@@ -1,8 +1,12 @@
-import { createPopper, Placement } from '@popperjs/core';
-import type * as Popper from '@popperjs/core';
+import { createPopper, Placement } from "@popperjs/core";
+import type * as Popper from "@popperjs/core";
 import * as Stacks from "../stacks";
 
-type OutsideClickBehavior = "always" | "never" | "if-in-viewport" | "after-dismissal";
+type OutsideClickBehavior =
+    | "always"
+    | "never"
+    | "if-in-viewport"
+    | "after-dismissal";
 
 export abstract class BasePopoverController extends Stacks.StacksController {
     private popper!: Popper.Instance;
@@ -31,7 +35,9 @@ export abstract class BasePopoverController extends Stacks.StacksController {
      */
     get isVisible() {
         const popoverElement = this.popoverElement;
-        return popoverElement ? popoverElement.classList.contains("is-visible") : false;
+        return popoverElement
+            ? popoverElement.classList.contains("is-visible")
+            : false;
     }
 
     /**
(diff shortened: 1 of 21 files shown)


lib/ts/controllers/s-popover.ts
```diff
@@ -1,8 +1,12 @@
-import { createPopper, Placement } from '@popperjs/core';
-import type * as Popper from '@popperjs/core';
+import { createPopper, Placement } from "@popperjs/core";
+import type * as Popper from "@popperjs/core";
 import * as Stacks from "../stacks";
 
-type OutsideClickBehavior = "always" | "never" | "if-in-viewport" | "after-dismissal";
+type OutsideClickBehavior =
+    | "always"
```
dancormier, line 6: Sweet leading pipe!

lib/ts/controllers/s-table.ts
```diff
@@ -214,9 +253,11 @@ function buildIndexOrGetCellSlot(section: HTMLTableSectionElement, findCell?: HT
             }
             curSlot++;
         }
-        if (curRow) {
-            curRow = curRow.nextElementSibling!;
+        if (curRow?.nextElementSibling) {
+            curRow = curRow.nextElementSibling;
```
dancormier: Isn't the original code saying "if `curRow` has a value, assert that  `curRow.nextElementSibling` is non-null and set `curRow` to it's value", while the update code is saying "if `curRow` and `curRow.nextElementSibling` has a value, set `curRow` equal to `curRow.nextElementSibling`"? Might be a _slight_ difference in functionality unless I'm interpreting it incorrectly.

Was this done by a linter or manually?

dancormier (commented):
I like it! Feels good to see our codebase tend towards consistency. Thanks for picking this up!

I had a question or two, so I'm not explicitly approving it, but if my questions don't require any changes, feel free to merge. I know this will create a handful of conflicts with existing branches, so I understand the urgency here.

Once this is merged, I'll take a minute to update my outstanding branches.

lib/ts/controllers/s-table.ts
giamir: Good catch! This was done manually to address a tslint error and it's my mistake! 
I am reverting this (instead I am going to type the `curRow` variable as `Element | null`).

giamir:
Thanks @dancormier for the review (and for a great catch!). Will merge the PR now into develop and rebase the testing branch accordingly.

PR #1203, opened 2022-12-05
fix(button): fix base box-shadow styling

dancormier:
The recent refactor of button styling introduced a bug where an inner box shadow was set on the base button styling. This was an unintended change and, since these buttons featured a transparent background, the box shadow would create a rather unpleasant style.

image
image

This PR fixes this styling bug. Once merged, I'll cut a new patch release.

---

Thanks for the report @abovedave! I've added you as a reviewer for this. Can you take a peek at the deploy preview and let me know if the box shadows render for you as expected?

### DIFF
--- a/lib/css/components/buttons.less
+++ b/lib/css/components/buttons.less
@@ -4,7 +4,7 @@
     --_bu-bc: transparent;
     --_bu-bg: transparent;
     --_bu-br: var(--br-sm);
-    --_bu-bs: inset 0 var(--su-static1) 0 0 hsla(0, 0, 100%, 0.7);
+    --_bu-bs: none;
     --_bu-fc: var(--theme-button-color);
     --_bu-focus-ring: 0 0 0 var(--su-static4) var(--focus-ring);
     --_bu-fs: var(--fs-body1);
@@ -111,6 +111,8 @@
 
     // MODIFIERS
     &&__filled {
+        --_bu-bs: inset 0 var(--su-static1) 0 0 hsla(0, 0, 100%, 0.7);
+
         border-color: var(--_bu-filled-bc);
         background-color: var(--_bu-filled-bg);
         color: var(--_bu-filled-fc);


abovedave (changes requested):
Fixed the base button, but I'm wondering if it needs moving into the `&.s-btn__filled` modifier for `danger` and `muted` as well @dancormier?

https://github.com/StackExchange/Stacks/blob/c75943aa2d49dc4c2da4941f405d48c1b587d4bf/lib/css/components/buttons.less#L230

https://github.com/StackExchange/Stacks/blob/c75943aa2d49dc4c2da4941f405d48c1b587d4bf/lib/css/components/buttons.less#L271

abovedave:
Should the alpha for those be `0.7` instead of `0.4` as well?

dancormier:
> Should the alpha for those be `0.7` instead of `0.4` as well?

My goal with [the button CSS refactor](https://github.com/StackExchange/Stacks/pull/1038) was to maintain styles as-is while cleaning up the structure and cascade. The box shadow for the [base filled button](https://github.com/StackExchange/Stacks/blob/ead9bcaa16f6859b9b4f77e9cac9a9f0930ac1db/lib/css/components/buttons.less#L198) has `0.7` as the alpha value, while the [`danger`](https://github.com/StackExchange/Stacks/blob/ead9bcaa16f6859b9b4f77e9cac9a9f0930ac1db/lib/css/components/buttons.less#L389) and [`muted`](https://github.com/StackExchange/Stacks/blob/ead9bcaa16f6859b9b4f77e9cac9a9f0930ac1db/lib/css/components/buttons.less#L278-L283) filled variants both use `0.4`, so I've left them in tact. I'm assuming this was intended to unify the look of that box shadow across button variants (since the base style background color is lighter and requires a more opaque shadow to contrast with it) but I'm not 100% sure of the reasoning.

Issue #1125, opened 2022-09-30
`s-table__bx` has a 2px gap between cell borders


kylejrp:
**Describe the bug**
When using `s-table__bx`, it turns the appropriate right/left borders the color `transparent`. However, they still have their original width of `1px`. This means that there is a `2px` gap between borders:
![image](https://user-images.githubusercontent.com/6742479/193363944-f62a648c-b2fe-48c9-a88a-ca48ecb485c1.png)

**To Reproduce**

Zoom in on the example table:
https://stackoverflow.design/product/components/tables/#horizontal-borders

Or in Stack Overflow's post timeline:
https://stackoverflow.com/posts/50534533/timeline

**Expected behavior**
The horizontal lines should not have any breaks in them:
![image](https://user-images.githubusercontent.com/6742479/193364689-82c21475-6d57-4064-aa00-76330d6ead83.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 105.0.5195.127 (Official Build) (64-bit) (cohort: Stable) 

**Additional context**
🤠

ondrejkonec:
Hi @kylejrp,
I've created Pull request #1130 to solve this bugs.

PR #1204, opened 2022-12-05
refactor(tables): use pseudo private custom properties

dancormier:
This PR refactors `.s-table` component styling to use pseudo-private custom properties. With this refactor, I didn't flatten the cascade for variants/modifiers _entirely_ as I have with many other refactors. Tables include some very targeted styles (targeting`first`/`last`/`nth` of nested elements) that would've made a  _complete*_ flattening more confusing and hard to maintain than the alternative.

> **Note**
> This PR should result in no visual changes for the `.s-table` component.

---

\* "_complete_" as _consistent_, not as in _total_ or _done_.


ondrejkonec:
Hi @dancormier 

I see that you are refactoring table component and there is [Issue](https://github.com/StackExchange/Stacks/issues/1125 ) that I tried to fix.

So maybe there is chance now for merging my PR, if you feel it make sense.

lib/css/components/tables.less
```diff
@@ -1,313 +1,296 @@
...
+    --_ta-td-va: middle;
+    --_ta-th-w: unset;
+    --_ta-thead-th-bg: var(--black-025);
+    --_ta-th-bbw: 0;
+    --_ta-th-bc: var(--bc-medium);
+    --_ta-th-fs: unset;
+    --_ta-th-p: var(--su8);
+    --_ta-th-ta: left;
+    --_ta-th-va: middle;
+    --_ta-th-w: unset;
```
giamir: Looks like a duplicate of line 20

lib/css/components/tables.less
```diff
@@ -1,313 +1,296 @@
...
 
-    //  --  Horizontal Only with Table Outline
-    &.s-table__bx {
         tr {
-            > *:not(:first-child) {
-                border-left-color: transparent;
-            }
-            > *:not(:last-child) {
-                border-right-color: transparent;
+            &:nth-child(2n) {
```
giamir: It looks like this selector has higher specificity (`.s-table tbody tr:nth-child(2n)`) now than the `.s-table tr.is-disabled` selector. This prevents the disabled `background-color` from being applied correctly for even rows.

<details>
    <summary>Before</summary>
    
Screenshot 2022-12-13 at 13 38 21
</details>

<details>
    <summary>After</summary>
    
Screenshot 2022-12-13 at 13 39 38
</details>

docs/assets/less/stacks-documentation.less
```diff
@@ -318,12 +318,15 @@
...
-    font-size: var(--fs-caption);
-    color: var(--fc-dark);
+    &,
+    .s-table td& {
+        padding: var(--su8);
+        border: 1px solid var(--bc-medium);
+        background-color: var(--black-075);
+        font-family: var(--ff-mono);
+        font-size: var(--fs-caption);
+        color: var(--fc-dark);
```
giamir: The specificity of this selector has increased significantly with this change. The result is a slightly different color for the td text in some of the examples. It is negligible and I think there is no action here.
Before the example text was using `color: var(--fc-medium)` (see line 226 in this file) instead of `--fc-dark`.

giamir (changes requested):
Great work @dancormier!
There are a couple of small fixes to address which I have highlighted in my comments.
Apart from that this looks good to me. I don't think I need to re-review after you have applied the changes - feel free to go ahead and merge. 🙂

lib/css/components/tables.less
dancormier: Oh, I believe the custom property on line 20 should be `--_ta-td-w`. Good catch!

lib/css/components/tables.less
dancormier: Ah, yep! We need `.is-disabled` to also redefine the custom property setting the even row background color. On it!

dancormier:
Thanks for another great review @giamir! It really helps to have your eyes on these. Much like https://github.com/StackExchange/Stacks/pull/1206, I'm holding on merging for now until Core is updated to 1.6.x.

dancormier:
Core's been updated. Time to merge!

PR #1206, opened 2022-12-08
refactor(link): use pseudo private custom properties

dancormier:
This PR refactors `.s-anchor`, `.s-block-link`, and `.s-link` component styling to use pseudo-private custom properties.

> **Note**
> This PR should result in no visual changes for the `.s-anchor`, `.s-block-link`, and `.s-link` component, with the exception of minor consistency improvements to `:visited` styles.

### No longer targeting `.s-anchor .s-anchor.s-anchor__{variant-name} …`

This PR no longer sets styles on `.s-anchor .s-anchor.s-anchor__{variant-name} …`. I dug through Core and found no applications of such a structure. Also, using this structure would result in the same styles _without_ targeting styles in this way.


lib/css/components/links.less
```diff
@@ -1,214 +1,102 @@
...
-    }
-
-    &.s-block-link__danger {
-        color: var(--red-500);
 
-        &:hover,
-        &:active {
-            color: var(--red-700);
-        }
+        appearance: none; // TODO in the PR, explain the removal of `-moz` and `-webkit` prefixes here
```
dancormier: We previously used vendor prefixes here. That's no longer necessary, since we only support browsers that play nice with plain old `appearance`.

lib/css/components/links.less
```diff
@@ -1,214 +1,102 @@
...
-    }
 
-    &.s-link__muted {
-        color: var(--black-500);
-
-        &.s-link__visited:visited {
-            color: var(--black-700);
+        &__visited { // TODO can we get rid of this? It's only in 3 places in Core and seems to me it doesn't even serve a purpose.
+            --_li-fc-hover: var(--theme-link-color-hover);
+            --_li-fc-visited: var(--theme-link-color-visited);
```
dancormier: I _think_ we could get rid of this. Keeping it around for now just for parity with the old CSS.

docs/.eleventy.js
```diff
@@ -26,7 +26,7 @@ module.exports = function(eleventyConfig) {
   eleventyConfig.addPlugin(markdownPlugin);
 
   // Add submenu generation
-  eleventyConfig.addPlugin(pluginTOC, {tags: ['h2', 'h3'], wrapper: 'nav aria-label="Table of contents"'});
+  eleventyConfig.addPlugin(pluginTOC, {tags: ['h2', 'h3'], wrapper: 'nav aria-label="Table of contents"', wrapperClass: 'toc s-anchors s-anchors__muted'});
```
giamir, line 29: I was wondering if moving the anchors classes here actually creates now 2 places for us to check how the `toc` is styled (this file and the `stacks-documentation.less` file). I would probably prefer to keep everything in one place as much as possible. I don't feel strongly about it though - feel free to ignore the suggestion.

lib/css/components/links.less
```diff
@@ -1,214 +1,102 @@
 a,
```
giamir: What about renaming this file to `link.less` (singular) now that it only contains `s-link` related styles?

giamir (approved):
Great work @dancormier!
Looks good to me. Everything seems in order visually as well. 
I just left a couple of minor suggestions.

### Potentially out of scope
I noticed that in [this anchors section](https://develop.stackoverflow.design/product/components/links/#descendent-anchor-examples) of our documentation the example code looks incorrect to me (see screenshot):
Screenshot 2022-12-13 at 12 07 10
I believe the `s-anchors` classes should be applied to a parent element (e.g. a `div` element) not to the `a` element.
Perhaps we could adjust the documentation as part of this refactoring PR.

docs/.eleventy.js
dancormier: I hear you on this. This is the best solution I could come up with that avoided either a refactor of how our `toc` was rendered or including the equivelent `.s-anchors.s-anchors__muted` styling ported manually to `toc`. It's a trade-off, but I think a reasonable one.

lib/css/components/links.less
dancormier: Good call! Consider it done!

dancormier:
Thanks for the review @giamir! I've addressed your review comments, but I'm holding off on merging until after 1.6.x ships in Core just in case I need to cut another release. I don't want to introduce any new, potentially volatile changes at this point.

dancormier:
Core is updated to 1.6.x. Time to merge!

PR #1211, opened 2022-12-12
docs(readme): add instructions on releasing Stacks

dancormier:
This PR includes instructions for cutting and shipping a new version of Stacks. It _does not_ include instructions on how to update or incorporate Stacks into another codebase.

I chose to include these instructions in `README.md`, but I'm open to moving it to a [`RELEASING.md`](https://github.com/thoughtbot/templates/blob/main/RELEASING.md.template) or similar file. I figure that basic information should default to `README.md` for the sake of visibility, but I can understand wanting to include this in another markdown file since we don't necessarily need this information to be highly visible. Thoughts?


aaronshekey:
Don't forget about https://stackoverflow.design/product/guidelines/releasing/

README.md
```diff
@@ -59,6 +60,36 @@ Stacks has implemented visual regression testing with [Backstop](https://github.
...
+1. Choose your new version from the "Choose a tag" dropdown
+1. Click "Generate release notes"
+1. Cleanup and complete the release notes
+    - Prominently mention any breaking changes, if applicable
+    - Include a "What's Changed" section in the release notes
+    - Mention significant bug fixes
+    - Mention new features
+    - Mention significant under-the-hood changes that could impact consumers
+
+### Ship your newly created version to [npm](https://www.npmjs.com/package/@stackoverflow/stacks)
```
giamir, line 88: I guess this command will require the user to be logged in to be executed. Perhaps it might be worth mentioning that.

giamir (approved):
I like this @dancormier. It is short and it goes straight to the point. 
I would recommend to keep it in the readme for now. Ideally all these manual steps will gradually disappear as we automate everything in scripts running in GH actions.

@aaronshekey made a good point about the releasing page in the documentation website. Personally I don't think our consumers need to know the procedure to release the library (unless they are contributors, but then they will check the readme anyway).

My recommendation would be to remove the releasing page (and avoid having to keep things in sync). If we want to keep it, it should probably contain information which is relevant to consumers and doesn't change often (something like "we are following semver", etc...). That alone won't justify a whole releasing page though so I think we should simply remove that.

dancormier:
> Don't forget about https://stackoverflow.design/product/guidelines/releasing/

@aaronshekey You're right! This page reminds me that I should also add a mention of updating `production` and publishing it on Netlify. Thanks!

> My recommendation would be to remove the releasing page (and avoid having to keep things in sync). If we want to keep it, it should probably contain information which is relevant to consumers and doesn't change often (something like "we are following semver", etc...). That alone won't justify a whole releasing page though so I think we should simply remove that.

I love removing stuff @giamir! Weird question, but should we consider a deprecation strategy for docs pages? I could imagine us pointing to the new "Releasing" section in the README from the existing docs page instead of deleting it altogether. Seems unnecessary here but I can imagine other situations where it could come up (something like the gutter class deprecation, for instance).

README.md
dancormier: The CLI prompts the user to log in if needed (usually, it is because there's a max of 5 minutes for "don't ask me to verify" when publishing), so I feel like that doesn't need to be called out. I'm happy to add a mention though if you have strong feelings here.

giamir:
> should we consider a deprecation strategy for docs pages?

In general yes, in this specific case though I would simply remove the page all together and perhaps add a redirect [here](https://github.com/StackExchange/Stacks/blame/develop/netlify.toml) to our readme to mitigate the potential for broken links. 

Side note: we should be able to monitor in netlify how many hits a certain route gets over time and clean up some of those old redirection rules.

dancormier:
> and perhaps add a redirect [here](https://github.com/StackExchange/Stacks/blame/develop/netlify.toml) to our readme to mitigate the potential for broken links.

Good idea! Added in https://github.com/StackExchange/Stacks/pull/1211/commits/32f956744188ad569a718dbf75907a4d44297dd7.

README.md
giamir: No strong feelings :)

PR #1216, opened 2022-12-19
feat(notice): add show/hide controller

dancormier:
This PR adds a controller for `.s-notice` components to provide simple show/hide functionality. It takes _heavy_ inspiration from the modal controller.

---

### TODO

- [x] Documentation!
- [x] Tests!
- [x] ~~Consider generalizing this controller to broadly show/hide anything~~
- [x] Provide focus capture functionality for toast
- [x] Add auto dismissal after interval (for toast)
- [x] Add dismissal interval data attribute
- [x] Add option to prevent auto-dismissal
- [x] Implement for `.s-banner`


lib/ts/controllers/s-toast.ts
```diff
@@ -0,0 +1,354 @@
...
+            },
+            { once: true }
+        );
+    }
+
+    /**
+     * Hide the element after a delay
+     */
+    private hideAfterTimeout() {
+        if (this.data.get("prevent-auto-hide") === "true" || this.data.get("hide-after-timeout") === "0") {
```
balpha, line 199: Sorry for the drive-by review -- I just stumbled over this, but feel free to ignore me.

The data map API is a remnant from Stimulus 1. It's tricky to update existing usages, but for new stuff we should probably use the value API.

Quoting the [v2 release notes](https://github.com/hotwired/stimulus/releases/tag/v2.0.0),

> The data map API from Stimulus 1.0 will continue to work but is no longer documented and should be considered internal.

giamir:
@dancormier it took me a whole morning of debugging but I found the issue with the test. 😅
This private member definition in the banner controller file
```ts
 private bannerTarget!: HTMLElement;
```
needs to be changed to a readonly declaration
```ts
 declare readonly bannerTarget: HTMLElement;
```
as per [Stimulus documentation](https://stimulus.hotwired.dev/reference/using-typescript#define-target-properties):
```
The declare keyword avoids overriding the existing Stimulus property, and just defines the type for TypeScript.
```

My (unverified) theory here is that WTR esbuild is converting typescript module on the fly and it (correctly) creates an actual [private member](https://www.typescriptlang.org/docs/handbook/2/classes.html#private) when it compiles. The webpack ts-loader somehow behave slightly differently and this is why controllers like modal, etc... still works in the docs ...and production.

### Side note
I believe there are few things we could simplify from an API perspective. For example why do we ask our consumers to assign a target to the banner element `data-s-banner-target="banner"` when we could simply reach out for the element via `this.element`?
I think there is also some room to extract common functionalities from the modal, toast. I did not look at the code extensively but I can see there are some patterns which are repeated across all files (this is an optimisation though).

This was fun, I learned a thing or two about stimulus today. I am still not in love with it though. 😅
Especially I found the separation of js from the markup quite tedious for testing/debugging compared to other component based architectures.

PR #1217, opened 2022-12-21
docs(adr): add draft a11y testing strategy adr

dancormier:
This PR adds an accessibility testing strategy ADR for Stacks.

### DIFF
--- a/adrs/0004-accessibility-testing-strategy.md
+++ b/adrs/0004-accessibility-testing-strategy.md
@@ -5,7 +5,7 @@
 **Related PR:** [#1217](https://github.com/StackExchange/Stacks/pull/1217)
 
 ## Issue
-We currently rely on manual testing, third-party audits, and user reports to identify and fix accessibility issues. The only semi-automated accessibility testing we do is via Mabl and is initiated manually. Furthermore, while we've generally aspired to WCAG 2.1 AA compliance, we haven't codified the accessibility success criteria for our components.
+We currently rely on manual testing, third-party audits, and user reports to identify and fix accessibility issues. The only semi-automated accessibility testing we do is via Mabl, which is initiated manually and tests instead of individual components. Furthermore, while we've generally aspired to WCAG 2.1 AA compliance, we haven't codified the accessibility success criteria for our components.
 
 **A defined success criteria and automated accessibility testing will reduce the effort needed to build and maintain accessibile components.**
 
@@ -18,12 +18,15 @@ Introduce the following success criteria for the accessibility Stacks components
 
 Introduce the following automated accessibility testing for Stacks:
 
-<!-- TODO: Choose *tbd* accessibility testing library to test components in isolation ([wtr](https://modern-web.dev/docs/test-runner/commands/#accessibility-snapshot) is my current top choice, but I'm also exploring other options mentioned in [this PR comment thread](https://github.com/StackExchange/Stacks/pull/1195#discussion_r1037616024)) -->
-- Integrate current Mabl testing for our documentation pages into our CI pipeline
+- All components should include accessibility tests using the [`chai-a11y-axe` plugin](https://open-wc.org/docs/testing/chai-a11y-axe/#chai-tdd-ui) in unit test files.
 
 ### Contexts to consider
 
 Components can be rendered in different combinations of light/dark and high contrast modes. We should test components in all of these contexts to ensure they meet the success criteria.
+
+#### A note on viewport widths
+
+Though accessibility issues can arise at any viewport width, accessibility issues tend arise primarily from invalid markup and color contrast issues. Because of this, we should focus on testing components at the default viewport width and use discression when adding tests for different viewport widths.
 
 ### Additional Guidelines
 
--- a/adrs/0004-accessibility-testing-strategy.md
+++ b/adrs/0004-accessibility-testing-strategy.md
@@ -25,8 +25,6 @@ Introduce the following automated accessibility testing for Stacks:
 
 Components can be rendered in different combinations of light/dark and high contrast modes. We should test components in all of these contexts to ensure they meet the success criteria.
 
-<!-- TODO: Is it neccessary to test against different viewport dimensions or specific input methods? If so, mention them here and specify *how* we should test in these contexts -->
-
 ### Additional Guidelines
 
 - All tests and linters runs in a GH workflow for every commit and pull requests on develop branch.
--- a/adrs/0004-accessibility-testing-strategy.md
+++ b/adrs/0004-accessibility-testing-strategy.md
@@ -5,7 +5,7 @@
 **Related PR:** [#1217](https://github.com/StackExchange/Stacks/pull/1217)
 
 ## Issue
-We currently rely on manual testing, third-party audits, and user reports to identify and fix accessibility issues. The only semi-automated testing we currently Furthermore, while we've generally aspired to WCAG 2.1 AA compliance, we haven't codified the accessibility success criteria for our components.
+We currently rely on manual testing, third-party audits, and user reports to identify and fix accessibility issues. The only semi-automated accessibility testing we do is via Mabl and is initiated manually. Furthermore, while we've generally aspired to WCAG 2.1 AA compliance, we haven't codified the accessibility success criteria for our components.
 
 **A defined success criteria and automated accessibility testing will reduce the effort needed to build and maintain accessibile components.**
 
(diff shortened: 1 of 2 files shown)


adrs/0004-accessibility-testing-strategy.md
```diff
@@ -0,0 +1,40 @@
+# Stacks Accessibility Testing Strategy
+
+**Date:** 2022-12-21
+
+**Related PR:** [#1217](https://github.com/StackExchange/Stacks/pull/1217)
+
+## Issue
+We currently rely on manual testing, third-party audits, and user reports to identify and fix accessibility issues. The only semi-automated testing we currently Furthermore, while we've generally aspired to WCAG 2.1 AA compliance, we haven't codified the accessibility success criteria for our components.
```
giamir: > The only semi-automated testing we currently ...

Something is missing here I guess. 🙂

adrs/0004-accessibility-testing-strategy.md
```diff
@@ -0,0 +1,40 @@
...
+
+**A defined success criteria and automated accessibility testing will reduce the effort needed to build and maintain accessibile components.**
+
+## Decision
+
+Introduce the following success criteria for the accessibility Stacks components:
+
+- All components should meet or exceed [WCAG 2.1](https://www.w3.org/TR/WCAG21/) level AA success criteria
```
giamir, line 16: Do we have something written down about this requirement in other org documents? It would be nice to cross link to those.

adrs/0004-accessibility-testing-strategy.md
```diff
@@ -0,0 +1,40 @@
...
+- Components rendered within high contrast mode should meet or exceed [WCAG 2.1 AAA success criteria for contrast](https://www.w3.org/WAI/WCAG21/Understanding/contrast-enhanced.html)
+
+Introduce the following automated accessibility testing for Stacks:
+
+<!-- TODO: Choose *tbd* accessibility testing library to test components in isolation ([wtr](https://modern-web.dev/docs/test-runner/commands/#accessibility-snapshot) is my current top choice, but I'm also exploring other options mentioned in [this PR comment thread](https://github.com/StackExchange/Stacks/pull/1195#discussion_r1037616024)) -->
```
giamir: Taking an a11y tree snapshot it's probably not super useful per se (at least I would not know what to do with it 🙂). We need an engine that checks for rules violation in the page like `axe-core` (for example via this chai extension https://open-wc.org/docs/testing/chai-a11y-axe/). I agree that more discovery is required here.

adrs/0004-accessibility-testing-strategy.md
```diff
@@ -0,0 +1,40 @@
...
+- Integrate current Mabl testing for our documentation pages into our CI pipeline
```
giamir: I would love to get an overview of how we use Mabl. I cannot even access it with my SO account. Maybe you or Ben could onboard me with a small session on the topic?

adrs/0004-accessibility-testing-strategy.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Contexts to consider
+
+Components can be rendered in different combinations of light/dark and high contrast modes. We should test components in all of these contexts to ensure they meet the success criteria.
+
+<!-- TODO: Is it neccessary to test against different viewport dimensions or specific input methods? If so, mention them here and specify *how* we should test in these contexts -->
```
giamir: I think we should start with the most used viewport and don't overcomplicate our life in the beginning. 🙂

adrs/0004-accessibility-testing-strategy.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Additional Guidelines
+
+- All tests and linters runs in a GH workflow for every commit and pull requests on develop branch.
+
+- Tests will be gradually added to existing components as developers need to change them. New components should always be accompanied by tests.
```
giamir, line 35: Will the a11y tests be written in a separate file (`s-tooltip.a11y.test.ts`)? Don't have strong opinions here at the moment. In my previous project we kept them together with the "unit" one since we were using this chai extension: https://open-wc.org/docs/testing/chai-a11y-axe/ and things like focus traps they are kind of like testing the behavior (even if it is a11y behavior in that context - the line can be bit blurry).

giamir (commented):
Thanks so much @dancormier for starting this effort. ❤️
I left some comments throughout the ADR but overall I agree with what I have read.
One thing that might be worth mentioning/adding is that automated a11y testing is just a starting point. Manual a11y testing is also important while building/editing components. In fact we could add a simplified a11y WCAG checklist to our PR templates. Thoughts?

adrs/0004-accessibility-testing-strategy.md
dancormier: We for sure have internal docs that mention WCAG 2.1 level AA as the goal. Most notably would be the Deque audit that the org has been working against. I won't link our non-public docs here but we can talk next week and see what receipts and references we feel comfortable providing here.

adrs/0004-accessibility-testing-strategy.md
dancormier: Ha. It's _almost_ a metaphor. Thanks for catching this.

adrs/0004-accessibility-testing-strategy.md
dancormier: I've contacted Helpdesk to get the ball rolling on giving you access to Mabl and I'd be glad to give you a quick overview of the tool and how we currently use it.

adrs/0004-accessibility-testing-strategy.md
dancormier: Good call. I'll update the ADR.

adrs/0004-accessibility-testing-strategy.md
dancormier: Hmm. I can kinda go either way with this. On the one hand, having distinct a11y test files will reduce noise within test files and make it obvious when we do/don't have a11y tests. On the other hand, a single test file per component seems simpler and would help with the blurry lines. And now, after typing the word "simpler" I'm leaning towards a single file.

adrs/0004-accessibility-testing-strategy.md
giamir: 👍

adrs/0004-accessibility-testing-strategy.md
giamir: 🙏

adrs/0004-accessibility-testing-strategy.md
dancormier: Good point. Since we mainly provide CSS classes atm, it's a little tricky to idealize how we can best test for accessibility. We can't enforce how a consumer structures HTML, so we really need to a) structure our component (and their styles) to encourage semantic markup, b) provide examples and guidance in the docs, and c) ensure our component styles (and interactivity, where applicable) meets our accessibility expectations. Not all of this is reasonably testable though.

For our components, I think we could really only test for legibility (min contrast, min font size). As for the documentation, I can think of a couple non-mutually-exclusive options on how to test to setup consumers for success:

- Integrate existing Mabl tests in CI to test our docs and report failures above a certain threshold
- Add tests for each example of a component to ensure our examples meet a11y expectations

---

### Testing examples

> We need an engine that checks for rules violation in the page like axe-core (for example via this chai extension https://open-wc.org/docs/testing/chai-a11y-axe/).

I think this would work for testing the examples we include in our docs.

### Testing components

Legibility seems a little tricky to test dependably. AFAICT, the accessibility testing included in `@open-wc/testing` doesn't seem to accurately test for contrast (though I'd love for this to be proven wrong. I've yet to find an easy drop-in solution for contrast checking, but we could grab computed values in the visual tests and compare them with something like https://github.com/bbc/color-contrast-checker. Basically, something like:

```js

var ccc = new ColorContrastChecker();

const backgroundColor = window.getComputedStyle(button).backgroundColor;
const color = window.getComputedStyle(button).color;
const fontSize = window.getComputedStyle(button).fontSize;

expect(ccc.isLevelAA(backgroundColor, color, fontSize)).to.be.true;
```

Feels kinda cumbersome and potentially frail, so I'm not sold on that approach.

giamir: The chai-a11y-axe plugin used by open-wc is running axe-core under the hood. It looks like axe-core scans should report a11y violations for color contrasts but I agree with you that we need to investigate further to be 100% sure. 
https://github.com/dequelabs/axe-core/blob/develop/lib/rules/color-contrast.json

We could also manually run the axe core scan. Perhaps we would have a better degree of configurability
https://github.com/dequelabs/axe-core#getting-started

dancormier:
I'm going ahead and merging this ADR and we can update it if needed. Btw, there's an example test here: https://github.com/StackExchange/Stacks/pull/1239

PR #1222, opened 2023-01-04
Add support for linking to future Razor component docs

b-kelly:
Sample PR that adds support for linking out to a separate Razor components docs site. Two different approaches have been taken here. We can easily keep one or both. Both approaches likely need some design love.

## Page badge
Added to Activity indicator, Buttons, Modals pages as examples:

image

## Codeblock "tab"
A fake "tab" that simply links you to the correct sub-section in the external docs. Good for pointing to a specific area on a page. Added to the Activity indicator page as an example.

image

### DIFF
--- a/docs/.eleventy.js
+++ b/docs/.eleventy.js
@@ -6,6 +6,7 @@ const highlightPlugin = require("./plugins/highlight");
 const iconPlugin = require("./plugins/icons");
 const tipPlugin = require("./plugins/tip");
 const markdownPlugin = require("./plugins/markdown");
+const codeblockTabsPlugin = require("./plugins/codeblock-tabs");
 
 module.exports = function(eleventyConfig) {
   eleventyConfig.setQuietMode(true); // Reduce the console output
@@ -15,6 +16,7 @@ module.exports = function(eleventyConfig) {
   eleventyConfig.addPlugin(iconPlugin);
   eleventyConfig.addPlugin(headerPlugin);
   eleventyConfig.addPlugin(tipPlugin);
+  eleventyConfig.addPlugin(codeblockTabsPlugin);
 
   // Version shortcode
   eleventyConfig.addLiquidShortcode("version", function() {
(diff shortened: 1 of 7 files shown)


giamir:
Thanks @b-kelly for putting together this PR so quickly. ❤️
My suggestion would be to **start with only the page badge for now** because of the following reasons:
- Simplicity (less deep links scattered across the documentation pages to maintain and less work trying to match up the razor playground page sections with the documentation ones - I believe that there might be some divergence on how we present the information in the playground and that would be fine in my opinion since we will focus on describing a component based api rather than how to nest markup, etc...)
- (Fake) tabs not being a good UX fit for deep linking (even if it's a temporary measure)

This of course is just my opinion based on the limited capacity we currently have for continuing Razor work. We might think of reevaluating this when we will have to document our WC library (although personally I would still like to keep a storybook playground/documentation).

@dancormier What are your thoughts?

### Side note
I think that we could benefit from having a conversation with @shanefoster about the general information architecture of stackoverflow.design. We should probably have a page describing our DS implementations at a glance (although I am not sure where it would exactly fit in the current architecture).

dancormier:
> My suggestion would be to start with only the page badge for now

I agree with @giamir. The badge approach seems most maintainable and like an easy-to-digest patter for visitors.

Ignore this: I'm _slightly_ bothered by the badges being both links to external (Razor, Figma) and internal (JavaScript) resources, but it's a pre-existing pattern and very inconsequential in the grand scheme. I just noticed myself going "wait, is `[JavaScript]` a link… it is! But to where? Oh."

b-kelly:
Page badges it is then! I'll go ahead and clean up this PR to remove the "tabs" markup

giamir:
@b-kelly: heads up I am taking over work in this branch as part of the Razor docs story.
To be precise I will just create a new branch based on this one so that we don't lose historical context.
Will close this PR when my new one is ready.

giamir:
Closing this PR. Work will be completed in this one:
https://github.com/StackExchange/Stacks/pull/1240

PR #1223, opened 2023-01-05
Update Stacks-Icons devDependency to latest breaking version

b-kelly:
Updates the docs to use Stacks-Icons v5.0.2. Also attempts to fix missing logo usage by manually embedding the affected icons. After meeting w/ Legal, we're ok to use these logos in this project (since it is written by SO, we don't need written permission _from_ SO to use our own logos). I've also tweaked wording (w/ content from Legal) on the `/brand/logo` page to increase transparency around how consumers are allowed to use our logo (without having to click through and read the legal guidelines).


b-kelly:
@StackExchange/stacks Ready for review! Should be pretty quick. I just need a second pair of eyes to make sure I didn't miss anything.

docs/plugins/icons.js
```diff
@@ -1,50 +1,44 @@
...
-
-      // If we need to change the size, do that too
-      if (dimension != null) {
-        svg = svg.replace('width="18" height="18"', 'width="' + dimension + '" height="' + dimension + '"');
-      }
-
-      return svg;
+    // embed svgs in the assets folder
+    eleventyConfig.addLiquidShortcode("embed", async function(name, classes) {
+      const file = await fs.readFile(`assets/img/icons/${name}.svg`, "utf8");
```
giamir, line 40: nit: I wonder if we should simply convert all the `var` in this file into `const` and `let` since we are at it. It probably helps with consistency in the long term. :)

giamir (approved):
Everything looks in order to me. Thanks @b-kelly for taking care of this! ❤️

dancormier (approved):
So TIL, we Stacks-Icons only removed all trademarked _Stack Overflow_ icons/logos and _not_ third party icons/logos (such as Facebook, Jira, etc)… unless I'm missing something really obvious. Kinda embarrassing to _just now_ get this since it's been in the works for a while now.

Assuming the above is accurate, this gets a ✅  from me.

docs/plugins/icons.js
b-kelly: If we add in the `@stackoverflow/eslint-config` package, then it'll start enforcing lots of best practices. I'm going to leave that change for a rainy day (follow-up PR some time in the future).

b-kelly:
> not third party icons/logos (such as Facebook, Jira, etc)

@dancormier That is correct. We're adhering to the branding guidelines for each of those icons, so we're not in violation of their usage.

b-kelly:
I'll merge this in after I fix up the `package.json` conflicts...

PR #1232, opened 2023-01-10
chore: use component-based file structure

dancormier:
This PR aims to restructure component code (styles, controllers, tests) so they reflect our components as closely as possible.

### Notes
- **Notice/Banner/Toast**: all of the styles for these components rely a Less function contained in `notice/index.less`. Eventually, `.s-banner` should be come a modifier on `.s-notice` and `.s-toast` shouldn't need the notice-based styles applied as it should act solely as a container for notices. We're not ready for that yet, so I figure some interdependence between these components is ok for now.
- **Folder naming**: components folders should reflect the name of the component (without the `s-` prefix). So for example, the `.s-avatar` component should be defined entirely within the folder `avatar`. In the event two components overlap to the extent that we need to colocate both, we should use an `_` to delineate component names (`input_textarea` contains both components `.s-input` and `.s-textarea` for example).

---

### Context / Benefits

We should colocate all code for a given component. This would make the maintenance of the project easier. 

### Acceptance Criteria

- Each component lives in its own folder
- Each component folder contains style, test, js when applicable

### Out of scope

- Move component docs in the same folder

### TODO

- [x] Create a ticket to update file structure to reflect "subcomponent" relationships (see [comment](https://github.com/StackExchange/Stacks/pull/1232#pullrequestreview-1330252822)) See #STACKS-281
- [ ] **Core followup**: when shipping, update [17 references in Core](https://github.com/search?q=%2Fstacks%2Flib%2F+repo%3AStackEng%2FStackOverflow&type=code) from `/lib/css` to `/lib`


webpack.config.js
```diff
@@ -58,7 +58,7 @@ const baseConfig = (isProd, minify) => ({
     },
     plugins: [
         new MiniCssExtractPlugin({
-            filename: `css/[name].css`,
```
giamir: I think this would be a breaking change. I would recommend to leave the same file structure for the bundles since there are too many places where we refer to that directly: https://github.com/search?q=dist%2Fcss%2Fstacks&type=code

You will need to readjust the `package.json` and `stacks-documentation.less` accordingly too.

lib/components/popover/tooltip-controller.ts
```diff
@@ -1,5 +1,5 @@
-import * as Stacks from "../stacks";
-import { BasePopoverController, PopoverController } from "./s-popover";
+import * as Stacks from "../../stacks";
+import { BasePopoverController, PopoverController } from "./controller";
```
giamir: This folder structure is a bit confusing at the moment to me. Do you think `tooltip` should be its own component/folder and have a dependency on the popover controller file?

giamir (changes requested):
Thanks Dan for putting together this PR. It mostly looks good to me I would like to exchange a bit about current naming convention before merging (and there is a requested change for the css distribution file). 
I know naming things can be quite subjective so take it with with a grain of salt (I would not mind to settle for what we have now if you make a good case for it).

Currently the naming structure for a component looks like this:
 ```
 component/
├─ index.less
├─ controller.ts
├─ s-component.test.ts
├─ s-component.a11y.test.ts
├─ s-component.visual.test.ts
 ```
I believe that we are mixing up few conventions together.
I would recommend a more [standard approach](https://github.com/shoelace-style/shoelace/tree/next/src/components/avatar) where we repeat the name of the component (this also has some benefits when searching for file by name across large codebases):
```
 component/
├─ component.less
├─ component.ts
├─ component.test.ts
├─ component.a11y.test.ts
├─ component.visual.test.ts
```

As of today the less entry point is named `stacks.less` and the ts one is `index.ts`. It would be nice to be able to spot the entry points at a glance following the same naming convention (e.g. `entry.less` and `entry.ts` or `index.less` and `index.ts`). Thoughts?

Can we merge `stacks-dynamic.less` and `stacks-static.less` in the file currently named `stacks.less` (make 1 file out of 3)? I think the separation only cause cognitive overload without clear benefits (we can still leave the inline comments if they are still relevant).

**Side note / Out of scope**
I think at some point we should take some time to reevaluate what is a component and what is a "subcomponent". For example I don't fully understand why input-fill and input-icon are their own components - I guess they make sense only in the context of input and therefore they should be encapsulated in the input folder. Similar thoughts on table-container and table.

Sorry, it turned out to be a wall of text.
Let me know if I can support you with the changes or if you would like me to clarify anything further. Thanks! 🙂

lib/components/popover/tooltip-controller.ts
dancormier: Tooltip is tricky since, from a styling viewpoint, it's a modifier on `.s-popover`. I've opted to use styling classes as the source of truth. IMO, the controller kinda reflects that too considering it extends `BasePopoverController`.

I don't really feel like tooltip is its own component as much as it's a modified version of popover. I'm open to ways we can make this less confusing, but I also want to make sure that the file structure reflects components that share a majority of their code (like with `input_textarea` or `checkbox_radio` for example).

webpack.config.js
dancormier: Ah, good catch! I'll revert this change.

I have a question about `stacks-documentation.less`: That references `/lib/stacks.less` (previously `/lib/css/stacks.less`). Are you suggesting I revert that as well? I _could_ but it's only referenced in [two places in Core](https://github.com/search?q=lib%2Fcss%2Fstacks.less+repo%3AStackEng%2FStackOverflow&type=code) and I figured we could get rid of that `css` directory just housing less files.

dancormier:
> I would recommend a more [standard approach](https://github.com/shoelace-style/shoelace/tree/next/src/components/avatar)

Good call! I'll change to that file naming convention.

> As of today the less entry point is named `stacks.less` and the ts one is `index.ts`. It would be nice to be able to spot the entry points at a glance following the same naming convention (e.g. `entry.less` and `entry.ts` or `index.less` and `index.ts`). Thoughts?

I like that idea, but I might want to consider it out of scope for the sake of this PR since there's already so much changing here. For the record, I tend to prefer `index.*` but `entry.*` seems fine to me too 🙂

> Can we merge `stacks-dynamic.less` and `stacks-static.less` in the file currently named `stacks.less` (make 1 file out of 3)? I think the separation only cause cognitive overload without clear benefits (we can still leave the inline comments if they are still relevant).

I think we can do this. It would _technically_ be a breaking change since [we directly reference those files in Core](https://github.com/search?q=lib%2Fcss%2Fstacks-static.less++repo%3AStackEng%2FStackOverflow&type=code) (I'm honestly not sure why), but it should be easy enough to update.

> I think at some point we should take some time to reevaluate what is a component and what is a "subcomponent". For example I don't fully understand why input-fill and input-icon are their own components - I guess they make sense only in the context of input and therefore they should be encapsulated in the input folder. Similar thoughts on table-container and table.

That's a good point. Some components are _obviously_ intended to be paired with another specific component. It would make sense to clearly reflect that relationship. For this PR, I took a blunt approach of defining a component as anything base `.s-*` class that isn't _directly_ dependent on another component to have styling applied. I do consider defining components as "subcomponents" out of scope but worth considering. I've added a TODO in the PR description to add a ticket for that before merging.

> Sorry, it turned out to be a wall of text.

No apology necessary! I appreciate the review!

balpha:
> > Can we merge `stacks-dynamic.less` and `stacks-static.less` in the file currently named `stacks.less` (make 1 file out of 3)? I think the separation only cause cognitive overload without clear benefits (we can still leave the inline comments if they are still relevant).
> 
> I think we can do this. It would _technically_ be a breaking change since [we directly reference those files in Core](https://github.com/search?q=lib%2Fcss%2Fstacks-static.less++repo%3AStackEng%2FStackOverflow&type=code) (I'm honestly not sure why), but it should be easy enough to update.

I can tell you why I created this separation back in the day; whether or not this is still relevant I don't know.

Before CSS custom properties were a thing, Stacks was configured via LESS variables. [This file, which no longer exists](https://github.com/StackExchange/Stacks/blob/27833a6575a085d666807e32d323945e6ebe54e5/lib/css/base/_stacks-configuration-dynamic.less), shows what that looked like. Note the comment at the top in particular.

Those parts of Stacks that were configurable in such a way would generate different CSS depending on which site it's compiled for. Buttons on https://rpg.stackexchange.com/ are brown, buttons on https://cooking.stackexchange.com/ are red, buttons on https://superuser.com/ are blue.

So in the LESS files that built the individual Stack Exchange site's `primary.css`, you would see something like this (simplified):

```less
@button-background-color: #c55a21;

@include "path/to/stacks-dynamic.less";
```

which would eventually have you end up with

```css
.s-btn {
  background-color: #c55a21;
}
```

in `Sites/rpg/primary.css`.

Clearly it was necessary for such CSS to be compiled individually per-site.

But the vast majority of Stacks CSS was independent of theming even back then (think e.g. of all the atomic spacing classes -- there's nothing styleable there).

We *could* have compiled the *full* Stacks into each site's `primary.css`, but that would mean that a user visiting Stack Overflow first, then Server Fault, would have to download all that CSS twice, and they would have to re-download for each additional site, and they would have to redownload all of that every time *something* in the CSS changes, even if it didn't involve Stacks at all.

That's why I split Stacks LESS into those parts that don't change based on theming configuration (that's the `static` part), and those that do (the `dynamic` part).

The former is only compiled once and served identically for every site:

the HTML source of the Role-playing games Stack Exchange homepage, highlighting stacks.css being included as a stylesheet

(Yes, the name doesn't make this quite obvious -- it's called `stacks.css`, but it's compiled from `stacks-static.less`.)

The latter is compiled into the site's `primary.css` together with all the Stack Exchange CSS that is not in Stacks.

I don't know if we can get rid of this split today (with most stuff being themed by custom properties), and if so, I also don't know what needs to be refactored in Q&A to make that happen.

webpack.config.js
giamir: yeah, apologies forget about the `stacks-documentation.less` I mentioned. I probably confused `lib` with `dist`.
Interesting that you found 2 places in core where we rely on it. We should make sure that in core everything refer to the `dist` folder when we upgrade. 🙂

lib/components/popover/tooltip-controller.ts
giamir: Makes sense.
I saw you agreed to change the naming convention to the one I suggested so I think the import will become a bit clearer already (`import { PopoverController } from "./popover"`)

giamir:
Thanks @balpha for the detailed explanation. Given that now we leverage css custom properties for theming we should rearrange things so that they make sense for this new approach (this will reduce chances of newcomers getting confused by the leftover of the previous implementation).

> I think we can do this. It would technically be a breaking change since [we directly reference those files in Core](https://github.com/search?q=lib%2Fcss%2Fstacks-static.less++repo%3AStackEng%2FStackOverflow&type=code) (I'm honestly not sure why), but it should be easy enough to update.

@dancormier shall we create an operational maintenance task to investigate how to merge those 2 files and link to this PR to don't lose the context?

Also somehow I just realized now that [we are shipping all the content of the lib folder as part of our package](https://github.com/StackExchange/Stacks/blob/develop/package.json#L11). This means we are exposing our "internal" file structure to the outside world. We have [17 places in Core](https://github.com/search?q=%2Fstacks%2Flib%2F+repo%3AStackEng%2FStackOverflow&type=code) which will require a path adjustment when we upgrade. We should remember to do that. 🙂

dancormier:
> Thanks @balpha for the detailed explanation. 

+1 thanks for all the context @balpha.

> Given that now we leverage css custom properties for theming we should rearrange things so that they make sense for this new approach (this will reduce chances of newcomers getting confused by the leftover of the previous implementation).
> 
> > I think we can do this. It would technically be a breaking change since [we directly reference those files in Core](https://github.com/search?q=lib%2Fcss%2Fstacks-static.less++repo%3AStackEng%2FStackOverflow&type=code) (I'm honestly not sure why), but it should be easy enough to update.
> 
> @dancormier shall we create an operational maintenance task to investigate how to merge those 2 files and link to this PR to don't lose the context?

I'm a step ahead here for once! We already have a ticket (it's #STACKS-208). I'll link this PR in that ticket for reference.

> Also somehow I just realized now that [we are shipping all the content of the lib folder as part of our package](https://github.com/StackExchange/Stacks/blob/develop/package.json#L11). This means we are exposing our "internal" file structure to the outside world. We have [17 places in Core](https://github.com/search?q=%2Fstacks%2Flib%2F+repo%3AStackEng%2FStackOverflow&type=code) which will require a path adjustment when we upgrade. We should remember to do that. 🙂

Yep! Adding a TODO in the PR description for this.

lib/tsconfig.json
```diff
@@ -9,5 +9,5 @@
         "skipLibCheck": true,
         "declaration": true
     },
-    "include": ["ts/**/*", "test/test-utils.ts"]
+    "include": ["components/**/*", "controllers.ts", "test/test-utils.tsls.ts"]
```
giamir: I wonder if a simple `**/*.ts` would be sufficient here. I did not try but the idea would be include any typescript file regardless of their location

lib/controllers.ts
```diff
@@ -0,0 +1,33 @@
+// export all controllers *with helpers* so they can be bulk re-exported by the package entry point
```
giamir, line 1: I am a bit unsure about the location of this file. Conceptually to me it is not too different from something like the `stacks-dynamic.less` and that resides directly in the `lib` folder. So maybe this one should too for the sake of consistency? Thoughts?

lib/components/btn/s-btn.a11y.test.ts
```diff
@@ -1,5 +1,5 @@
-import { runComponentTests } from "./test-utils";
-import "../ts/index";
+import { runComponentTests } from "../../test/test-utils";
+import "../../index";
 
 describe("s-btn", () => {
```
giamir: I just realized that we named this directory `btn` (and similarly `btn-group`). I understand that the abbreviation is what we use in the css classes but my opinion is that we should not tight the API naming we use necessarily to the file names. If you look at the razor components or even the documentation website we spell out the entire name: button and button-group. Shall we do the same here?

lib/components/activity-indicator/s-activity-indicator.a11y.test.ts
```diff
@@ -0,0 +1,21 @@
+import { runComponentTests } from "../../test/test-utils";
+import "../../index";
+
+describe("s-activity-indicator", () => {
```
giamir: Is there a specific reason why for the test files we kept the `s-` prefix? For consistency I would prefer to drop the `s-` entirely when naming files.

giamir (approved):
Great work @dancormier. I left few comments about a couple of consistency issues - nothing major. If you agree feel free to make the small changes and then merge.

I think one important tasks (to don't forget) when we release these changes would be to document the breaking changes in the release notes. Essentially have a few lines for consumers to explain the reference changes they need to make when upgrading. Writing things down will also help us out when we do upgrade the library in Core. 🙂

lib/tsconfig.json
dancormier: Good call! Not only is it sufficient, it removes a very silly typo!

> "test/test-utils.tsls.ts"

I must've fallen on my keyboard 🤦

lib/components/activity-indicator/s-activity-indicator.a11y.test.ts
dancormier: No specific reason, just an oversight. I'll drop the prefix.

lib/controllers.ts
dancormier: This file _does_ reside directly in the `lib` folder. Am I misunderstanding what your suggestion?

lib/components/btn/s-btn.a11y.test.ts
dancormier: Good call. I'll make this change

lib/controllers.ts
giamir: Somehow the new github tree UI tricked me to think this was in the components folder. Please ignore this. Sorry.

dancormier:
I deem this **_GOOD ENOUGH_** 🏆 

Merging!

PR #1233, opened 2023-01-11
update Stimulus to latest

balpha:
Aside from the actual upgrade to Stimulus 3, this requires a few other changes. Most importantly, TypeScript compilation can no longer target ES5 because Stimulus dropped support for Internet Explorer. This technically makes this a breaking change (but it should be the only one; feature-wise, Stimulus 3 is backwards-compatible to Stimulus 2).

The switch to ES6 also had an effect on how imports are reordered during transpilation, which required a minor change to the documentation code to ensure that jQuery is available before global.search.js is executed.

Final note, just because I got confused by it myself initially: Stimulus 2 temporarily still exists as a (transitive) dev dependency because Stacks depends on itself (via stacks-editor).

### DIFF
--- a/lib/tsconfig.json
+++ b/lib/tsconfig.json
@@ -1,8 +1,8 @@
 {
     "compilerOptions": {
         "strict": true,
-        "target": "es5",
-        "lib": ["dom", "es6", "dom.iterable", "scripthost"], // es6 stuff is polyfilled by stimulus
+        "target": "es6",
+        "lib": ["dom", "es6", "dom.iterable", "scripthost"],
         "outDir": "../dist",
         "sourceMap": true,
         "moduleResolution": "node",
--- a/docs/assets/js/global.jquery.ts
+++ b/docs/assets/js/global.jquery.ts
@@ -1,0 +1,2 @@
+// @ts-expect-error
+global.$ = require("jquery");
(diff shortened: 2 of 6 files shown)


b-kelly:
I'll be reviewing this by EOD (ET) today

lib/tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
     "compilerOptions": {
         "strict": true,
-        "target": "es5",
-        "lib": ["dom", "es6", "dom.iterable", "scripthost"], // es6 stuff is polyfilled by stimulus
+        "target": "es6",
```
b-kelly, line 4: I don't see an issue targeting `es6`, since our primary consumer (Stack Overflow's Core codebase) targets `es2019` in its tsconfig

b-kelly (approved):
lgtm, thanks for updating this for us! Looking over both Stimulus changelogs and the local dev environment, I don't see any cause for concern here.

Regarding Stacks (docs, not library) depending on itself via stacks-editor, that's pretty much correct. In the near future we'll be switching over to a monorepo w/ the docs separated out into it's own "package", which will solve this issue. In the meanwhile, it doesn't seem to be causing any harm since stacks-editor is a devDependency

b-kelly:
I'm opting to **not** mark this as a breaking change due to the fact that the browsers we officially support all have es6 support.

PR #1234, opened 2023-01-11
fix: allow independent stacks-dynamic/static compilation

dancormier:
`button.less` relies on styles from `.s-link()`. Because `.s-link` styles are included via `stacks-static.less` and `.s-btn` styles are included via from `stacks-dynamic.less`, we couldn't render `stacks-dynamic.less` independently (as we do in Core). h/t @balpha

Now that all components use CSS custom properties instead of Less variables for custom theming, we no longer need to include these component styles in `stacks-dynamic.less`. This PR moves any remaining component styles from `stacks-dynamic.less` to `stacks-static.less`.

### DIFF
--- a/lib/css/stacks-dynamic.less
+++ b/lib/css/stacks-dynamic.less
@@ -17,13 +17,6 @@
 //  --  SET BASIC STYLES FOR BODY
 @import "base/body.less";
 
-//  --  COMPONENTS
-@import "components/buttons.less";
-@import "components/link-previews.less";
-@import "components/notices.less";
-@import "components/tags.less";
-@import "components/pagination.less";
-
 //  --  LESS CONSTANTS AND MIXINS
 @import "exports/exports.less";
 
--- a/lib/css/stacks-static.less
+++ b/lib/css/stacks-static.less
@@ -22,6 +22,7 @@
 @import "components/award-bling.less";
 @import "components/avatars.less";
 @import "components/badges.less";
+@import "components/buttons.less";
 @import "components/empty-states.less";
 @import "components/block-link.less";
 @import "components/breadcrumbs.less";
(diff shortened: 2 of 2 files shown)


giamir (approved):
FWIW LGTM.
I think @balpha should give the final thumbs up here though. 🙂

### Question out of scope for this fix:
If this static/dynamic distinction doesn't make sense anymore since we are using custom properties how much effort would it be to consolidate all these imports in a single file?
I believe this is legacy context that creates unneeded cognitive overload for (new) people. 🙂

balpha:
> If this static/dynamic distinction doesn't make sense anymore since we are using custom properties how much effort would it be to consolidate all these imports in a single file?
I believe this is legacy context that creates unneeded cognitive overload for (new) people. 🙂

To give some context on this (historical and otherwise):

The reason for the static/dynamic split is that when building CSS files for individual Stack Exchange sites, we wanted as much of the Stacks CSS shared across the network, but anything that is themeable per-site (so that e.g. https://rpg.stackexchange.com/ can have brown buttons) could not be shared.

Back when theming happened purely through LESS variables, the only way to do this was to have themeable components in a separate LESS file that is then compiled into the site's `primary.css`. Anything that isn't themeable can be shared; that's the file `stacks.css` that's included in all Stack Exchange sites. Despite the name, that file isn't the full Stacks, but only the static part.

Now that everything is using custom properties, this distinction can probably be simplified or dropped completely, but I haven't kept up enough with the details (both on the Stacks side and on the PubPlat side) to opine on it.

balpha:
The good news is that core LESS compiles fine again (thus ✅)

The bad news is that while testing this, I discovered another issue that is probably also a fallout of #1206.

For testing purposes I switched to the RPG theme, and look at the sad contrast on that "ask question" button:

image

The CSS causing this is

```
a:visited, .s-link:visited {
    color: var(--_li-fc-visited);
}
```

which does not appear in [production](https://rpg.stackexchange.com/) (even if the link *is* indeed visited).

balpha (approved):

dancormier:
@giamir @balpha Thank you for the reviews.

> If this static/dynamic distinction doesn't make sense anymore since we are using custom properties how much effort would it be to consolidate all these imports in a single file?

When I started this task I could've sworn we had a ticket in our backlog to do just that, but I couldn't find one. In any case, I suspect it would be pretty low effort. I'll add a ticket!

> look at the sad contrast on that "ask question" button

@balpha good catch! I'm curious to look back and see why a similar style _wasn't_ applying before https://github.com/StackExchange/Stacks/pull/1206, but I'm assuming it was some cascade trickiness. I'll investigate! Also, thank you for all the historical context!

PR #1239, opened 2023-01-17
test(button): add accessibility tests for s-btn variations

dancormier:
See also https://github.com/StackExchange/Stacks/pull/1217, https://github.com/StackExchange/Stacks/pull/1241

---

This PR adds accessibility tests for (nearly) all permutations of the `.s-btn` component styles. I'm considering it a dry run to see if we can reasonably test components in the context of multiple themes, variants, modifiers, and children. `.s-btn` seemed like a good first example as it would surface issues with our accessibility testing approach suggested in https://github.com/StackExchange/Stacks/pull/1217. 

I get this is kinda wild and maybe over-the-top, but now that the styles are cleaned up and scannable, it should be easier to mirror component CSS as I've done in this PR and use nested loops to our advantage. For what it's worth, I'd anticipate most other component tests to be simpler.

> **Note**
> There are currently ~~5~~ _many_ failures and there _should be_ ~~5~~ _many_ failures. Many combinations of dark + danger fail.

## TODO
- [x] Set theme classes on body tags
- [x] ~~Define tests for `.s-btn-group`~~ (this should be a separate test)
- [x] Define tests for child elements
- [ ] Find a way to ignore tests
- [ ] Move accessibility tests to `*.a11y.test.ts` files
- [ ] Run accessibility tests in a single browser
- [ ] Figure out how to conditionally test against WCAG AAA for high contrast
- [ ] Consider creating abstractions that will work for all components (possibly outside this PR)


giamir:
Thanks @dancormier for driving this work!
 I like that the tests are catching actual a11y issues! This is awesome! 🎉
 
Here are some thoughts from my end:
- Pipeline stability and consistency
For me this is paramount  though which means I would not want to merge this PR until we find a way to mark the component permutations having a11y issues and skip them. I can imagine an array of strings matching the test names we want to skip.  The array will always be visible in the text and be a bit annoying (if long) which will remind us we need to take care of those when we jump in the test file.
- Pipeline execution time
I thought this would be worse: 79 tests run in 16s across 3 browsers in my machine (chrome and firefox took only 8s, webkit is slower). Given the programmatic nature of these tests though (they all run the same axe-core checks) I am still tempted to separate them in its own group/files (e.g. `s-button.a11y.test.ts`). This would help also to keep at bay the current semi-convoluted implementation.
- Test implementation
There is no doubt this file is a bit complex. I think we should invest on finding a good abstraction for ourselves and keep the complexity encapsulated in a couple of functions away from the testing files. The way I would approach this is by asking myself - what would be the ideal API for me developer to create those tests cases with minimal effort (without even thinking about the implementation). That said I am not against starting with what you have done so far and iterate on that (especially if we group those a11y tests in its own file - see my previous comment).
```
// I will need more thinking but something around those lines
a11yTest({
  name: "button",
  classes: [ ] // a data structure to define permutations smartly 🤔
  other: [ ] // same permutations structure 
  template: `<button class="${classes}" type="button">${other}</button>`
   skip: {
      classes: [ 
          // list of permutations to skips (for classes)
       ]
    }
})
```

Great progress here @dancormier. Well done!
I am ok to merge as long as pipeline stability won't be affected and we group this category of a11y tests in its own files.
We should keep an eye on how those tests evolve and invest time to find the right abstraction for us not too far in the future. 🙂

dancormier:
Thank you for the thoughtful notes @giamir! I've added some TODOs to the PR description based on your suggestions. I'm figuring that all the remaining TODOs can wait until a future PR, with the exception of ignoring tests. I'll find a way to ignore some tests before merging this.

giamir:
@dancormier Thanks. I would recommend to open a small operational maintenance ticket with the remaining todos so that we don't forget the work. 🙂

dancormier:
@giamir I wound up making separate tickets for each remaining TODO to keep the remaining issues small and help them move across our board quicker. They're STACKS-228 to STACKS-232 if you'd like to check them out.

I figure I'll pick up `STACKS-231: Separate accessibility tests into `*.a11y.test.ts` files and update this PR before merging.

giamir:
@dancormier Thanks for breaking these tasks down! Tickets look good to me and I would recommend quickly go through them during our next refinement on Friday. I believe they are already in a good state to mark them as refined. 🙂

> I figure I'll pick up STACKS-231: Separate accessibility tests into *.a11y.test.ts` files and update this PR before merging.

In general I would prefer if we prioritize the tickets in the planned column first. If this ticket though is small (e.g. half a day of work) and will allow us to don't keep this PR hanging I am ok to treat it as an exception. 
By the way how separating the tests will prevent the pipeline from failing? Do you plan to exclude the `a11y` group from the pipeline all together for the time being?

dancormier:
> By the way how separating the tests will prevent the pipeline from failing? 

@giamir Separating things out into `.a11y` files won't affect the pipeline. It was just a quick housekeeping chore I figured I could pop in, but it's probably best to do the work when it's planned 🙂  I'll leave that task for later.

> Do you plan to exclude the `a11y` group from the pipeline all together for the time being?

We _could_ but I wasn't planning on it. I'm personally a big fan of testing being run automatically to catch regressions because, just speaking for myself, I forget to manually run tests at those critical moments when they'd be useful. Even in these early stages, I think these button tests could catch a regression so I'd like to keep them running.

giamir:
@dancormier Oh yeah, I agree with running them as long as there are not known test failures which we are unable to skip and they make the pipeline constantly red. When the pipeline is always red then it loses its purpose - people start ignoring it. :)

We can also think about having a dedicated workflow for the a11y test group in future if we see they are taking a long time to run. For now it's fine to keep things simple in one workflow as long as we keep it green.

dancormier:
@giamir, in my effort to ensure tests pass, I went ahead and separated the accessibility tests into `.a11y.test.ts` files and created an `a11y` test group. I swear I wasn't intending to change from what I said in my last comment, but it wound up working out that way 😂

Anyways, these tests all pass in this branch _in GitHub Actions_, but when I run them locally I get failures that seem unrelated to accessibility tests. This only happens when running _all_ tests with `npm run test` with no suffix.

```
…
lib/test/s-tooltip.visual.test.ts:

 ❌ s-tooltip > should not introduce visual regressions (failed on Firefox)
      Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
        at createTimeoutError (node_modules/@web/test-runner-mocha/dist/autorun.js:1:242929)
…
Chromium: |██████████████████████████████| 2/2 test files | 404 passed, 0 failed
Chromium: |██████████████████████████████| 8/8 test files | 107 passed, 0 failed
Firefox:  |██████████████████████████████| 8/8 test files | 92 passed, 15 failed
Webkit:   |██████████████████████████████| 8/8 test files | 107 passed, 0 failed

Finished running tests in 49.3s with 15 failed tests.
```

I won't merge this in until we figure out what causes this issue. @giamir, can you see if this also happens locally for you?

At this point, I'm thinking it's a timeout issue and I'm passing a threshold on my machine that doesn't get hit elsewhere.

giamir:
@dancormier everything runs smoothly in my machine (~~in CI it appears to be all good too~~ actually [the pipeline seems to struggle as well](https://github.com/StackExchange/Stacks/actions/runs/4075023521/jobs/7020879993) with the 2000ms mocha timeout).
I think it's ok to push the mocha timeout threshold to 5000ms (instead of 2000ms). This won't affect the time our tests take in the speediest devices anyway and it will give some breath to less powerful machines. 
https://modern-web.dev/docs/test-runner/test-frameworks/mocha/#configuring-mocha-options

Let me know if that makes a difference. 🙂

dancormier:
Bumping the timeout to 5000 has me down to 2 failing tests on average 🎉

image

If I bump it up to 2000 the failures stop but that seems pretty extreme 😬 For now, 5000 works! Thanks @giamir!

PR #1240, opened 2023-01-18
Link to Razor Components Docs

giamir:
This branch is based off this initial [spike PR ](https://github.com/StackExchange/Stacks/pull/1222)@b-kelly started.

To be merged only after Razor Docs cleanup is live:
https://github.com/StackEngLabs/Stacks-RazorComponents/pull/11


docs/product/components/badges.html
```diff
@@ -23,7 +24,7 @@
 <span class="s-badge">
     <span class="s-award-bling s-award-bling__bronze">
         Altruist
-    </span>
```
b-kelly: nit: revert whitespace only changes

docs/product/components/banners.html
```diff
@@ -9,14 +10,14 @@
 
 <!-- Override stacks classes for demos -->
 <style>
-.s-banner[aria-hidden=false] {
```
b-kelly: nit: revert whitespace-only changes for this file

docs/product/components/editor.html
```diff
@@ -1,6 +1,7 @@
 ---
 layout: page
 title: Editor
+razor: https://razor.stackoverflow.design/components/editor-facade
```
b-kelly: Technically speaking, the editor itself does not have RC support, so I'm not 100% sure if we want to mark it here.

No changes necessary, just pontificating.

b-kelly (changes requested):
Overall, lgtm except for the additional whitespace/formatting changes on a few files

giamir:
Thanks @b-kelly for the review. WRT to the white spaces you mentioned those are due to how our current editorconfig file is set. We are instructing our editor to insert a new line on save if not present. 
https://github.com/StackExchange/Stacks/blob/develop/.editorconfig#L9

I would either remove the rule all together or merge the whitespaces since everyone would have the same issue when they try to save those files.

More info here:
https://spec.editorconfig.org/#supported-pairs

giamir:
Forget my previous comment. I haven't read correctly your comments. Will address those whitespaces changes.

docs/product/components/editor.html
giamir: I removed it for now. We can always add the badge later on.

giamir:
@b-kelly I addressed your comments. If I don't hear anything tomorrow (my) morning will merge this PR. Thanks

b-kelly (approved):
lgtm, thank you for the changes!

PR #1256, opened 2023-01-30
test(btn): resolve .is-loading test flakiness

dancormier:
Currently, the test for `.s-btn.is-loading` fails in the GitHub Actions workflow but not locally (aka, _works on my machine_).


giamir:
@dancormier I had similar issues with the tooltip visual tests. The solution was to create a wrapper with a fixed height and width. If this happens often we could even extract the wrapper setup in a helper function:
https://github.com/StackExchange/Stacks/blob/develop/lib/test/s-tooltip.visual.test.ts#L12

dancormier:
@giamir ah, of course! I should've thought of that. Thanks!

PR #1279, opened 2023-02-24
fix(card): target `:last-child` for margin reset

dancormier:
This PR targets `.s-card :last-child` instead of the previous `.s-card p:last-of-type` for a margin reset. This will allow us to use whatever element we feel is appropriate for the last child and still have the margin reset applied.

cc and h/t @balpha

> **Note**
> This is a breaking styling change and will require us to comb through Core when updating the Stacks dependencies to ensure styling is applied as expected.

### DIFF
--- a/lib/css/components/cards.less
+++ b/lib/css/components/cards.less
@@ -17,7 +17,7 @@
     }
 
     // CHILD ELEMENTS
-    p:last-of-type {
+    :last-child {
         margin-bottom: 0;
     }
 
--- a/lib/css/components/cards.less
+++ b/lib/css/components/cards.less
@@ -17,7 +17,7 @@
     }
 
     // CHILD ELEMENTS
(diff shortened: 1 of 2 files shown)


lib/css/components/cards.less
```diff
@@ -17,7 +17,7 @@
     }
 
     // CHILD ELEMENTS
-    p:last-of-type {
+    :last-child {
```
balpha:
```suggestion
    > :last-child {
```

giamir (approved):
Apart from @balpha's comment on the direct descendant selector LGTM.

lib/css/components/cards.less
dancormier: Ah, good catch! I was mistaken and thought that the `:last-child` _only_ applied to direct descendants.

PR #1280, opened 2023-02-24
fix(table): improve sortable table accessibility

dancormier:
This PR is built on top of #1144. It includes [a merge of develop with conflict fixes](https://github.com/StackExchange/Stacks/commit/2f0c3c363a34dfe95bc4c4cf501c2eb85fc7cb2e) the addition of [a test](https://github.com/StackExchange/Stacks/commit/87271fc55fcb641e6bdacb41c00200f5e8c42f2f) (currently failing, but it's a start), the [porting of styles to the refactored `tables.less`](https://github.com/StackExchange/Stacks/commit/4e37924c38a5845fa4823ccbc6a8429a070629a7). Once the test is in a good place, we should be able to merge this work.

@allejo I couldn't commit directly to your branch, so I made this one. Feel free to cherry pick from here or lift anything that's useful into your branch.


dionrhys:
_Just noticed this new PR for sortable tables - copying over my PR comments from #1144 so we don't miss some doc + styling fixes:_

> I'm really excited by this change! We'll be wanting sortable tables in an internal project and having sorting be keyboard and screen-reader accessible would be a great thing.
> 
> Here's another link for reference: <https://adrianroselli.com/2021/04/sortable-table-columns.html>. It's marked out-of-scope for this PR but this link also presents various different ways to have [sortability](https://adrianroselli.com/2021/04/sortable-table-columns.html#ScreenReaders) and [sort change](https://adrianroselli.com/2021/04/sortable-table-columns.html#Announce) to be announced for screen readers. If we do enhance the feature with instructions/announcements here or in future, the HTML will need to provide localised/translated strings for these.
> 
> ---
> 
> I think we should update the documentation in this PR to ensure devs write in the initial `aria-sort` attribute for the non-JS table examples where columns are already sorted. (It'd also apply to the JS examples, but those columns aren't pre-sorted in the examples)
> 
> Something like this for example:
> 
> ```diff
> --- a/docs/product/components/tables.html
> +++ b/docs/product/components/tables.html
> @@ -634,14 +634,14 @@ description: Tables are used to list all information from a data set. The base s
>  <section class="stacks-section">
>      {% header "h2", "Sortable tables" %}
>      <p class="stacks-copy">To indicate that the user can sort a table by different columns, add the <code class="stacks-code">s-table__sortable</code> class to the table.</p>
> -    <p class="stacks-copy">The <code class="stacks-code">&lt;th&gt;</code> cells should include arrows to indicate sortability or the currently applied sorting. In addition, the column that is currently sorted should be indicated with the <code class="stacks-code">is-sorted</code> class on its <code class="stacks-code">&lt;th&gt;</code>.</p>
> +    <p class="stacks-copy">The <code class="stacks-code">&lt;th&gt;</code> cells should include arrows to indicate sortability or the currently applied sorting. In addition, the column that is currently sorted should be indicated with the <code class="stacks-code">is-sorted</code> class and the <code class="stacks-code">aria-sort</code> attribute with the <a href="https://www.w3.org/TR/wai-aria-1.1/#aria-sort">appropriate value</a> on its <code class="stacks-code">&lt;th&gt;</code>.</p>
>      <div class="stacks-preview">
>  {% highlight html %}
>  <div class="s-table-container">
>      <table class="s-table s-table__sortable">
>          <thead>
>              <tr>
> -                <th scope="col" class="is-sorted">Listing @Svg.ArrowDownSm</th>
> +                <th scope="col" class="is-sorted" aria-sort="descending">Listing @Svg.ArrowDownSm</th>
>                  <th scope="col">Status @Svg.ArrowUpDownSm</th>
>                  <th scope="col">Owner @Svg.ArrowUpDownSm</th>
>                  <th scope="col" class="ta-right">Views @Svg.ArrowUpDownSm</th>
> @@ -659,7 +659,7 @@ description: Tables are used to list all information from a data set. The base s
>                  <table class="s-table s-table__sortable">
>                      <thead>
>                          <tr>
> -                            <th scope="col" class="is-sorted">
> +                            <th scope="col" class="is-sorted" aria-sort="descending">
>                                  Listing
>                                  {% icon "ArrowDownSm" %}
>                              </th>
> ```
> 
> Screen readers can announce the initial state of the sorting even if interactive sorting isn't enabled for the table.
> 
> ---
> 
> I've noticed that the styling doesn't quite match the previous appearance, especially noticeable in the font using the default `<button>` font instead of the bolded header font. The cursor is the regular arrow at the moment (instead of cursor) and it's also not possible to select and copy the header text any more:
> 
> Screenshot of JS table example with button elements that use default font and can't be selected
> 
> I think it just needs some CSS tweaks to get it looking and behaving like before when the text was directly within the `<th>`. The `s-btn__unset` class has rules `font: unset` and `user-select: auto` that I think fix these. Plus I think `cursor: pointer` can be added to fix that cursor. Applying these new rules seem to fix things up: 
> 
> Screenshot of JS table example with button elements that retain font and text selectability of the encompassing TH
> 
> (It might need additional tweaking for things I haven't noticed are not identical to before)
> 
> I'd love it if the headers had a focus shadow/outline like we see on other elements (right now it's default for button), but I have no idea how to get those showing up without being cropped by the boundaries of the `<th>` element.
> 
> ---
> 
> Now that there's a test framework for Stacks, hopefully we can add in tests to verify no breaking changes here (e.g. test `ensureHeadersAreClickable` keeps working with future changes).
> 
> Great work on implementing this! 🥳 Let me know if there's anything I can help with re. testing or anything.

lib/tsconfig.test.json
```diff
@@ -0,0 +1,4 @@
+{
+    "extends": "./tsconfig.json",
+    "include": ["test/"]
```
giamir: @allejo Heads up, if you are planning to continue working on this PR you might want to update the branch with the develop since recently we have changed the file structure significantly and this file for example won't be needed.

allejo: Appreciate the heads up! Just noticed the changes as I brought in `develop` to fix the web-test-runner issue.

allejo:
@StackExchange/stacks I'll defer to you all on whether or not this PR is in a good spot to review/test/merge. I'm comfortable with the tests I've written so far and think they cover a good chunk of the component's functionality. I need to step away from this PR to work on a few other things on my plate so I don't have time to write tests for the actual sorting logic right now.

Where I left off:

- [x] Tests written for `buildIndex` and `getCellSlot`
- [x] Tests written for making sure the correct SVG indicators are shown
- [x] Tests written to make sure the right column (and no other columns) is marked with `aria-sort`
- [ ] Tests written for confirming the sorting logic in the table cells
- [ ] Address Dion's comments about the accidental styling changes

Notes:

- I created PR https://github.com/open-wc/open-wc/pull/2652 to fix chai losing its typing when being imported; a temporary patch was pushed by Giamir here: a8fc6002b59d5385c87d943de39fd00ca51a3b46; keep an eye on this PR for when to remove the  temporary patch
- I create PR https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64954 to improve `@types/chai` to support type narrowing which should allow things in this PR to run correctly without babying TypeScript. e.g.
   ```ts
   let somethingNullable: string | null = "hi Stacky!";

   assert.isNotNull(somethingNullable);

   somethingNullable; // This is now a `string` instead of `string | null`
   ```

---

I've removed myself from the reviewers list since I wrote the majority of this code and don't feel comfortable reviewing it myself.

lib/components/table/table.less
```diff
@@ -122,10 +122,24 @@
...
+            button {
+                appearance: none;
+                background-color: transparent;
+                border: 0;
+                padding: var(--su8);
+                text-align: left;
+                width: 100%;
+            }
+
+            &[data-s-table-target="column"] { // If this column is sortable, then the padding will come from the button
```
giamir: I would prefer to not rely on a data attribute here because it makes the presentational layer coupled to the stimulus implementation.
Ideally I would use the `:has` selector (but it is not available yet in firefox)
```less
&:has(button)
```

That said I think we have a larger issue to solve here since this style will essentially lock all sortable header to a fixed padding ignoring eventual `spacing` set by the table (e.g. `s-table__lg`). So perhaps the solution here is to style the button in a way that take up all the height defined by the padding applied to the `th` element.

@dancormier any suggestion?

lib/components/table/table.ts
```diff
@@ -145,54 +126,124 @@ export class TableController extends Stacks.StacksController {
...
+            }
+        });
+    };
+
+    /**
+     * Transform legacy header markup into the new markup.
+     *
+     * @param headerEl
+     */
+    private ensureHeadersAreClickable = (headerEl: HTMLTableCellElement) => {
```
giamir: I feel strongly about NOT introducing a permanent markup transformer in Stacks to support legacy usage.
If we can automate the change at runtime we can do this also with a codemod that run once in Core before the library get upgraded.
We could look into [codemods](https://github.com/rajasegar/awesome-codemods) to automate everything but I suspect it will go faster to semi-automatically adjust (e.g. replace with regex in editor) the [20 occurrences](https://github.com/search?q=repo%3AStackEng%2FStackOverflow+click-%3Es-table%23sort&type=code) in Core.

If we are concerned about breaking changes and semantic versioning we can make sure to support both elements (button and th) as valid targets until we are ready to ship v2.0 (where we can finally deprecate th).
If we decide to do that we should create some sort of a roadmap for Stacks 2.0.0 where we keep track of todos for the major release and also come up with a schedule (e.g. every 6 months we release a major).
Given that Stacks is primarily used by Core (we are not React or Bootstrap) I don't feel strongly about following semantic versioning by the book honestly (and for example communicate this small breaking change in a minor release `v1.10.0`). Alternatively we could release `v2.0.0` and accept that the major numbers will change relatively often (in my previous team when I left we were at `12.x.x` because we did not wait for breaking changes to pile up before cutting a release - as soon as they were ready we released and bump that major number).  @dancormier any thoughts?

giamir (changes requested):
Thanks @allejo to make progress on this PR.
I have reviewed the source code and left a couple of comments that needs to be addressed by our team before being able to merge.

I haven't spent much time reviewing the tests. At a glance they look quite complex, but so it's the algorithm. Thanks for starting the effort though.

I would like to close up this PR soon, let me catchup with @dancormier and see how we can address the remaining items.

lib/components/table/table.less
dancormier: I think there's some CSS we can apply that will address both issues. Basically, we can remove the padding override targeting the data attribute and update the button styling to something like this:

```
button {
    appearance: none;
    background-color: transparent;
    border: 0;
    display: flex;
    gap: var(--su-static4);
    font-weight: inherit;
    margin: calc(var(--_ta-th-p) * -1);
    padding: var(--_ta-th-p);
    text-align: left;
    width: calc(100% + calc(var(--_ta-th-p) * 2));
}
```

This will ensure that the sizing responds to the size modifiers.



I've pushed this change in https://github.com/StackExchange/Stacks/pull/1280/commits/10b76fba352b927c703eb868ed9fcb345fc655d2. Let me know if it seems reasonable to you.

PS: I honestly don't _love_ using negative margins in general, but since they're used relative to a pseudo-private custom property for this component, it seems safe.

lib/components/table/table.ts
dancormier: I'm trying to recall my thoughts from when we discussed this a couple weeks ago 😂 Feels like forever ago. Anyways…

> I feel strongly about NOT introducing a permanent markup transformer in Stacks to support legacy usage.
If we can automate the change at runtime we can do this also with a codemod that run once in Core before the library get upgraded.
We could look into [codemods](https://github.com/rajasegar/awesome-codemods) to automate everything but I suspect it will go faster to semi-automatically adjust (e.g. replace with regex in editor) the [20 occurrences](https://github.com/search?q=repo%3AStackEng%2FStackOverflow+click-%3Es-table%23sort&type=code) in Core.

I agree. I think the markup transformer is a little too much magic for Stacks. I _think_ we could do the find+replace without too much complication. Only big issue I could see is if any tables are partially rendered with JS, but I figure we could always open a PR and find out.

>  I don't feel strongly about following semantic versioning by the book honestly

 As for the semver discussion, I remember starting from a place of "I want to follow semver strictly" to "strict semver can cause complication for a library like ours". I doubt we'd benefit from waiting for several breaking changes as the overhead for our small team would be a lot, so we'd instead be doing major releases for any minor breaking change. So, I'm good to ship the occasional breaking change in minor versions as long as we communicate it and provide support.

lib/components/table/table.test.ts
```diff
@@ -0,0 +1,363 @@
...
+                    </tr>
+                </tbody>
+            </table>
+        </div>
+    `;
+};
+
+describe("s-table", () => {
+    [true, false].forEach((useButton) => {
+        describe("Column sorting indicators", () => {
```
allejo:
```suggestion
    [true, false].forEach((useButton) => {
        describe(`Column sorting indicators (${useButton ? '' : 'not '}using legacy markup)`, () => {
```

Can you do something like this so that we can read the output of the test results a bit better about when they/if fail

dancormier: Absolutely! Funny thing is I had something similar in place junked it while trying to debug an issue.

lib/components/table/table.less
giamir: @dancormier You are indeed a CSS wizard. 😀

lib/components/table/table.ts
```diff
@@ -17,18 +17,12 @@ export class TableController extends Stacks.StacksController {
...
     sort(evt: PointerEvent) {
         // eslint-disable-next-line @typescript-eslint/no-this-alias
         const controller = this;
         const sortTriggerEl = evt.currentTarget;
-        if (!(sortTriggerEl instanceof HTMLButtonElement)) {
-            throw "invalid event target";
-        }
-        const colHead = sortTriggerEl.parentElement as HTMLTableCellElement;
+        const triggerIsButton = sortTriggerEl instanceof HTMLButtonElement;
+        const colHead = (triggerIsButton ? sortTriggerEl.parentElement : sortTriggerEl) as HTMLTableCellElement;
```
giamir: Can we write a small comment here reminding us that this conditional is here because of backward compatibility with the old API where we did not ask the consumer to wrap sortable column heads in a button. Also `TODO: only support button as trigger in next major release`.

lib/components/table/table.test.ts
```diff
@@ -0,0 +1,366 @@
...
+
+                    if (columnId === colId) {
+                        expect(columnEl).to.have.attribute("aria-sort", order);
+                    } else {
+                        expect(columnEl).to.not.have.attribute("aria-sort");
+                    }
+                });
+            };
+
+            const expectTheseIndicatorsToBeVisible = (visibleIds: SvgId[]) => {
```
giamir, line 152: Small Rant, no action required:
My personal feeling is that we test too much implementation details here. Ideally we should test things from a user perspective, reaching out to element via accessible queries (e.g. `getByRole('button', { name: "Season"  })`) and then check for the presence of the aria-sort attribute. 
Testing the SVG visibility is a bit border line but I see why we would do that - we could mess it up.

docs/product/components/tables.html
```diff
@@ -757,7 +757,7 @@
                 {% for item in tables.data-attributes %}
                     <tr>
                         <th scope="row"><code class="stacks-code">{{ item.attribute }}</code></th>
-                        <td><code class="stacks-code">{{ item.applies }}</td>
+                        <td><code class="stacks-code">{{ item.applies }}</code></td>
```
giamir, line 760: [Comment related to the docs snippet above]
Do you think that this part of the documentation (see screenshot) should be updated too with a the new markup button to avoid confusion?
At the moment it seems to use the old markup
Screenshot 2023-05-09 at 17 12 55

giamir (approved):
@dancormier Thanks for bringing this PR to an end. I have left few final minor comments to address. After that feel free to merge.
Great work (also to you @allejo for initiating this in the first place) 🎉

docs/product/components/tables.html
dancormier: Yes, it absolutely should be updated. Good catch! On it

dancormier: On second look, I take it back. The above example comes before the "JavaScript sorting" section and is static. The purpose of this example is to show a _sorted_ table and, IMO, shows that Stacks can provide the appearance of sorting without the ability for a user to interact with that sorting. I'm going to leave this example as-is but let me know if you have a strong opinion and think it should be updated.

lib/components/table/table.less
dancormier: file photo of CSS wizard Dan Cormier

docs/product/components/tables.html
giamir: @dancormier One thing that confuses me about that example is that it gives the impression to the user the table header can be interacted with: not just because of the carets icons but also because of the cursor becoming a `pointer` when hovering on top of it. 
I believe that there is not a real usecase for what we showcase there and it's going to trick people into forgetting to add the buttons. A "sortable table" can always be interacted with and should use buttons. A "sorted table" should not have the double carets symbols in the column which are not sorted and should not show a pointer as cursor (not even sure if that's a concrete use case).

PR #1289, opened 2023-03-07
fix(a11y): apply browser-default focus styles to reset links/buttons

dancormier:
`.s-btn__link`, ~~`.s-btn__unset`~~ and `button.s-link` don't have focus styles, but they should! This PR remedies that.

### Before (has focus)
focused button.s-link element with no visible indication it's focus

### After (has focus)
focused button.s-link element with a focus ring to indicate it's focused

h/t @collincchoy for reporting! See this [codepen](https://codepen.io/collincchoy/pen/OJoxNbr) for a demonstration of the problem

### DIFF
--- a/lib/css/components/buttons.less
+++ b/lib/css/components/buttons.less
@@ -140,10 +140,10 @@
     &&__unset {
         &:focus,
         &:focus-visible {
-            outline-style: auto;
+            outline: revert;
         }
 
-        outline: initial;
+        outline: revert;
     }
 
     &&__link {
--- a/lib/css/components/buttons.less
+++ b/lib/css/components/buttons.less
@@ -140,10 +140,8 @@
     &&__unset {
         &:focus,
         &:focus-visible {
-            outline: revert;
+            outline-style: auto;
         }
-
-        outline: revert;
     }
 
     &&__link {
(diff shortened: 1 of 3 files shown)


lib/css/components/buttons.less
```diff
@@ -140,10 +140,10 @@
     &&__unset {
         &:focus,
         &:focus-visible {
-            outline-style: auto;
+            outline: revert;
         }
 
-        outline: initial;
+        outline: revert;
```
collincchoy: `s-btn__unset` already works as I'd expect it to? sorry if that was confusing in my report/codepen 😅 

not sure if this change is to address something else tho or if this change was just for consistency. 

I tried the preview link for the PR and at least AFAIK tabbing still triggers the same desired visual behavior as before so doesn't seem to me like it _hurt_

dancormier: Ah, yeah I was hasty and confused myself. I _think_ we could safely make this style consistent between `__link` and `__unset` but we use `__unset` in a surprising number of places in Core. In the spirit of keeping fixes as targeted as possible, I've pushed a new commit that only changes the link elements. Good catch!

PR #1290, opened 2023-03-07
test: abstract test implementation

dancormier:
See STACKS-228

This PR creates testing utility functions aimed at simplifying baseline test writing for Stacks components. I've introduced the following abstractions for use in test files:

- **`runComponentTests`**: This is the primary test utility function to test all variations possible with the arguments passed. It accepts arguments describing a components markup, variations, the type of test needed (`a11y` or `visual`), and options.
- **`runComponentTest`**: This function is used to run a single visual or accessibility test.

Other functions could be exported, but I've opted only to export the ones currently used for the sake of simplicity and to encourage use of the above functions.

### Current tests being run in this PR (as of 2023/03/13 5PM EST)

|             | a11y | visual |
|-------------|------|--------|
| Active test | 495  | 1209   |
| Skipped     | 185  | 96     |

### TODO
- [x] Name util functions sensibly… I marked this as done, but I'm open to a better naming strategy
- [x] Re-enable dark mode tests

### TODO post-merge stories to create
- [ ] Fix various `.s-btn` styling issues*
    - [ ] Various contrast issues in high contrast mode
    - [ ] Several `.s-btn--badge` contrast issues
        s-btn-dark-primary-filled-badge
    - [ ] `.s-btn__dropdown.is-loading` (pseudo-element collision)
        s-btn-light-primary-filled-dropdown-is-loading

---

*What's great is that creating these test actually surfaced styling bugs present in the `.s-btn` component 🎉


lib/ts/test-utils.ts
```diff
@@ -0,0 +1,209 @@
...
+        testHighContrast: true,
+        includeNullVariant: false,
+        includeNullModifier: true,
+    },
+}: {
+    baseClass: string;
+    variants?: string[];
+    modifiers?: TestModifiers;
+    options?: TestOptions;
+}) => {
```
giamir: Potentially add a return type for the function
```
TestVariation[]

type TestVariation = {
  testId: string // shall we call this just id?
  classes: string
  theme: string // maybe we can have a more restrictive type for this (union)
}
```

giamir: I see you have named it `TestProps`, I think we can probably keep it consistent with the name of the function `getTestVariations` -> `TestVariation[]`.

lib/ts/test-utils.ts
```diff
@@ -0,0 +1,209 @@
...
+};
+
+export const makeTest = ({
+    description,
+    element,
+    testid,
+    theme = [],
+    type,
+}: {
+    description?: string;
```
giamir: Skimming through the code I did not see the description field use. Premature optimization? These are automatically generated tests maybe we should not allow the user to override our automatically generated descriptions.

lib/test/s-activity-indicator.visual.test.ts
```diff
@@ -0,0 +1,35 @@
+import { html } from "@open-wc/testing";
+import { getTestVariations, makeTest, makeTestElement } from "../ts/test-utils";
+import "../ts/index";
+
+describe("s-activity-indicator", () => {
+    getTestVariations({
```
giamir: I wonder if we can abstract even further and hide some more implementation details.
Some pseudo code
```
import { generateVisualTests, generateA11yTests } from 'test-utils'

generateVisualTests({
  ...variationsParameters, // baseClass, variants, etc...
  extraVariations: string[] // to capture things like ["", "new"]
  template: ({ classes,  with a fn we could pass here pretty much anything including a prefilled version of the makeTestElement fn if necessary }) => html`... ${class}...`
})
```

I hope this makes sense. The base functions you have created are great and should be reused. I am suggesting to create an "abstraction of the abstraction" to make things even simpler for the consumers.

giamir (commented):
Looks great @dancormier! Awesome work! 
I have added some comments for some additional improvements/suggestions.
Reach out if something is not clear. 
Thanks! 🙂

lib/ts/test-utils.ts
dancormier: > Potentially add a return type for the function

👍 good call! I'll add that

> testId: string // shall we call this just id?

I think I'd like to stick with `testid`. The Testing Library function [getByTestId](https://testing-library.com/docs/queries/bytestid/#api) targets the`data-testid` attribute, so I went with `testid` (note lack of camelCasing). Also, the util function `makeTestElement` accepts attributes, so it could wind up with a collision if we ever included an `id`. 

> theme: string // maybe we can have a more restrictive type for this (union)

Another good idea! On it.

lib/ts/test-utils.ts
dancormier: I have it only on the `.s-btn.is-loading` visual test which _hopefully_ will be replaced by generated tests soon. Someone could make an individual test using `makeTest` and pass a description, but yeah it does seem kinda like a premature optimization. I'll remove it!

lib/test/s-activity-indicator.visual.test.ts
dancormier: I added `runComponentTests` (note the plural) to abstract out some common patterns. I _could_ add `runA11yComponentTests` and `runVisualComponentTests` but that doesn't seem too much more convenient then specifying a `type` argument. Curious to get your thoughts on the abstractions!

Sidenote: Naming things is tough so I'd like to take a moment when this is close to being merged to reconsider the util function naming.

lib/ts/test-utils.ts
```diff
@@ -0,0 +1,334 @@
...
+type TestTypes = "visual" | "a11y";
+
+interface ComponentTestVariationArgs {
+    baseClass: string;
+    variants?: string[];
+    modifiers?: ComponentTestModifiers;
+    options?: TestOptions;
+}
+
+interface ComponentTestArgs extends ComponentTestVariationArgs {
```
giamir: I think it is pretty important to document all the options part of the public APIs. Even if we found the perfect name a description to clarify what the option does goes a long way and will hopefully prevent the consumer from having to read the implementation to understand how to use something.

```
 /**
   * additional html attributes applied to the element under test   
   * (e.g. { role: "button",  id: "id" } -> <element role="button" id="id">  )
   */
  attributes?: Record<string, string>
```

lib/test/s-btn.a11y.test.ts
```diff
@@ -1,123 +1,30 @@
...
+            primary: ["filled", "outlined"],
+            secondary: [...["xs", "sm", "md"], ...["dropdown", "icon"]],
+            global: ["is-loading"],
+            standalone: [
+                ...["link", "unset"],
+                ...["facebook", "github", "google"],
+            ],
+        },
+        attributes: {
+            role: "button",
```
giamir: Why do we need to add the role button here? I thought the button element has that role applied by default.

lib/ts/test-utils.ts
```diff
@@ -0,0 +1,334 @@
...
+    const excludeTest = excludedTestids.some((pattern) => {
+        if (pattern instanceof RegExp) {
+            return pattern.test(testid);
+        } else {
+            return pattern === testid;
+        }
+    });
+
+    if (excludeTest) {
+        it.skip(`${type}: ${testid} (skipped)`, () => {
```
giamir, line 436: We are mixing the concept of skipping (temporary measure / test will eventually need to be fixed) and excluding (this is something more permanent and we should not even see the test being skipped it.skip). I would recommend to have separate options for the 2 so that we can see easily how many tests need to be actual fixing. I really like what you have done here though. Very smart. :)

giamir (changes requested):
Awesome work @dancormier. I think the most important part left is about taking some time to describe the "options" of the public methods as I mentioned in one of the comment below. Apart from that it looks good to be merged to me.

I see that we are skipping visual tests: this I guess is the consequence of non differentiating between "permanent excluding" and "temporary skipping". In my opinion we should never get in a situation where we feel like we need to skip visual tests. I left a comment below about this too.

Running all the tests in my machine takes now around 145s. We have around 4000 test cases in total. This doesn't bother me too much since I can easily run only the unit tests.

A lot of visual regression tests means a lot of baseline images. Again I am not too worried here because we are using git LFS. That said when I pulled your branch it took around 15s to download all the baseline images. This is a one off operations though for people pulling the repo and we don't keep a git history of the images (like we do for files). It should be fine. If we see issues we can always rethink our approach.

I think the test-utils file should live in the test folder as well (not the ts folder) until you merge the file restructuring PR.

I noticed that there is an eslint warning in the test utils file, it would be nice to take care of it before merging.

Great work! After addressing those couple of points feel free to merge directly.

lib/test/s-btn.a11y.test.ts
dancormier: We don't need it! I'm guessing that I added that confusing it with `type: "button"` which we generally include on buttons. I'll update it. Thanks for catching this.

lib/ts/test-utils.ts
dancormier: Fair point! I've split this group into `excludedTestids` and `skippedTestids`.

> In my opinion we should never get in a situation where we feel like we need to skip visual tests.

For now, only the `s-btn` _accessibility_ tests skip anything tests and include a `TODO` to fix issues. We could just allow these tests to fail, but I wanted to exclude them so that we can start from passing and highly regressions when they're introduced. If you disagree, let me know and I'll reconsider this approach.

lib/ts/test-utils.ts
dancormier: I've added jsdom definitions to anything that could wind up in a test

dancormier:
@giamir I'm going ahead and merging this. Thanks for the reviews and let me know if anything I added can be improved

PR #1296, opened 2023-03-13
fix(a11y): remove collapsed expandable content from accessibility tree

balpha:
When testing keyboard focus in something we're building, @emgrasmeder noticed that the collapsed content of an `s-expandable` can still grab focus. That's because as of #1164, the content is only hidden via `opacity: 0`, which hides it *visually*, but keeps it in the visibility tree.

It looks removing `visibility: hidden` in that change was purely by accident, since the `visibility` transition that delays the hiding is still there.


giamir (approved):
LGTM. I took the liberty of adding a test. Those are the kind of regressions which are hard to spot and tests can help us in the long run. 
@dancormier Could you do a final review and merge? Thanks.

Thanks @balpha and @emgrasmeder for the report.

dancormier (approved):
@balpha You're right, `visibility` got dropped accidentally in the refactor. Thanks for the fix, thanks to @emgrasmeder for catching it, and thanks to @giamir for the test!

PR #1312, opened 2023-03-27
Native search results [EXPERIMENTAL] 

KyleMit:
## Summary

Works in two phases:

1. Creates `/search.json` file with all data at time of build
2. On Search, fetches that file, filters and displays matches

Includes the following hierarchy for matches

```
enum MatchType {
    TitleExactMatch,
    SectionTitleExactMatch,
    TitlePartialMatch,
    DescriptionPartialMatch,
    SectionTitlePartialMatch,
    CodePartialMatch,
    ParagraphPartialMatch
}
```

### Render full page description if no other hits

image

### Knows classes within the page

image

### Renders subtitles inline next to page title, but prioritizes subtitle match over content match

image

### Prioritizes title over other matches

image


giamir:
@KyleMit I am closing this PR for now. We have reached out to Algolia customer support and we were able to figure out a solution to manage the Algolia app as a team. That said I think this is a great proof of concept and a good starting point if we will decide at some point to stop using Algolia. 
I added your Stack Overflow email address to the Algolia app in case you want to play around with the index configuration.

PR #1344, opened 2023-04-25
test(block-link): add a11y, visual tests

dancormier:
This PR adds visual and accessibility tests for the `s-block-link` component.

### DIFF
--- a/lib/components/block-link/block-link.visual.test.ts
+++ b/lib/components/block-link/block-link.visual.test.ts
@@ -8,7 +8,7 @@ const blockLinkTemplate = ({ component, testid }: any) => html`<div
     data-testid="${testid}"
 >
     ${component}
-</div>`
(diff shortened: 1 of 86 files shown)


lib/components/block-link/block-link.visual.test.ts
```diff
@@ -0,0 +1,61 @@
...
+        template: blockLinkTemplate,
+    });
+
+    // Base + danger
+    runComponentTests({
+        type: "visual",
+        baseClass: "s-block-link",
+        modifiers: {
+            primary: ["danger"],
+        },
```
dancormier, line 33: This should be a variant.

dancormier: Nvm, for the purposes of this test, primary works better to generate all variations. Keeping it as-is.

PR #1346, opened 2023-04-26
test(button-group): add a11y, visual tests

dancormier:
This PR adds visual and accessibility tests for the `s-btn-group` component.

> ~~**Note**~~
> ~~When constructing these tests, I realized we had used `--` to construct the `radio` modifier class (`.s-btn-group--radio`) instead of `__`. `--` is reserved for denoting _child_ elements of a component. This PR resolves that by changing the class to `.s-btn-group__radio`, which will result in a (easy to address) breaking change in the one place we use it: [stacks-editor](https://github.com/StackExchange/Stacks-Editor/blob/2124fd6ec1fb0b6569a4f9d2fea8686984120292/src/stacks-editor/editor.ts#L348).~~

~~@giamir since we'll need to update Core to support the [new button group design](https://github.com/StackExchange/Stacks/pull/1630), would this be a good opportunity to also include the minor class syntax breaking change noted above?~~

In the button group redesign implementation, the `radio` modifier` on `s-btn-group` no longer provides any unique styling. Consumers _should_ remove any references to this modifier, but including it will not affect anything materially.


lib/components/button-group/button-group.a11y.test.ts
```diff
@@ -0,0 +1,66 @@
...
+        },
+        children: {
+            default: btns
+                .map((btn) => getBtn({ ...btn, isRadio: true }))
+                .join(""),
+        },
+        options: {
+            includeNullModifier: false,
+        },
+        skippedTestids: ["s-btn-group-dark-radio"],
```
giamir: What's the problem with this variant?

giamir (changes requested):
Somehow the button group radio in the docs is now broken for me (cannot even focus on it). I haven't investigated further though.

For the tests I can see that the setup for the a11y and the visual tests is pretty similar. I would recommend to create something like `button-group.test.setup.ts` so that there is no need to duplicate many line of codes. You could also tests the `form` variations in the a11y tests if you want which probably makes sense (???).
If you go with the `.setup.ts` file approach we may want to add the `*.setup.ts` in[ this tsconfig](https://github.com/StackExchange/Stacks/blob/develop/lib/tsconfig.build.json#L3) to avoid creating unnecessary type for the production build.

I understand that we are introducing a small breaking change here both in Stacks Classic and Stacks Editor (which will depend on consumers upgrading the stacks classic css they use in their page).
Do you feel comfortable merging this change as part of v 2.1.0 or shall we wait for a future v2.2.0? I am asking because I would like to don't rush this through if possible due to enterprise release fixes we need to ship.

Apart from those minor things, the rest looks good. 
Thanks @dancormier for adding the tests 🎉

lib/components/button-group/button-group.a11y.test.ts
dancormier: I don't remember 😂 but it's no longer an issue with the updated design 🎉

giamir:
> @giamir since we'll need to update Core to support the https://github.com/StackExchange/Stacks/pull/1630, would this be a good opportunity to also include the minor class syntax breaking change noted above?

~Yep, I think that makes sense. Let's make sure to call it out in the release notes.~

Class should be removed all together.

lib/components/button-group/button-group.a11y.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+describe("button group", () => {
+    runA11yTests({
+        baseClass: "s-btn-group",
+        children: {
+            default: btns.map((btn) => getBtn({ ...btn })).join(""),
+        },
+    });
+
+    // s-btn-group--radio
```
giamir: Should probably change to `s-btn-group__radio`

lib/components/button-group/button-group.a11y.test.ts
```diff
@@ -0,0 +1,64 @@
...
+const getBtn = ({
+    name = "",
+    isRadio,
+    isSelected,
+}: {
+    name: string;
+    isRadio?: boolean;
+    isSelected?: boolean;
+}): string => {
+    const baseClasses = "s-btn s-btn__muted s-btn__outlined";
```
giamir: I thought we don't need `s-btn__outlined` anymore

lib/components/button-group/button-group.visual.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+const btns = [
+    { name: "Newest", isSelected: true },
+    { name: "Frequent" },
+    { name: "Active" },
+    { name: "Bountied" },
+    { name: "Unanswered" },
+];
+
+const form = (children: number[]) => `
```
giamir: We can remove those variants since they don't add much value, screenshots are identical

lib/components/button-group/button-group.visual.test.ts
```diff
@@ -0,0 +1,137 @@
...
+                default: group.children,
+            },
+            template: btnGroupTemplate,
+            options: {
+                includeNullModifier: false,
+            },
+        });
+    });
+
+    // s-btn-group--radio
```
giamir: Should probably change to `s-btn-group__radio`. Class will probably be removed all together.

docs/_data/button-groups.json
```diff
@@ -6,7 +6,7 @@
       "description": "Base button group style."
     },
     {
-      "class": ".s-btn-group--radio",
+      "class": ".s-btn-group__radio",
```
giamir: Unused class. Removed from docs as agreed in Slack. :)

screenshots/Chromium/baseline/s-btn-group-dark.png
giamir, line 1: This variation has a problem. Need investigation

giamir (changes requested):
Few changes to make before we are able to merge this one. 
Will circle back here when you finish the cleanup. :) 
Thanks

dancormier:
@giamir I've updated this PR with your suggested changes. Since your last review I:

- Removed all references to `s-btn-group--radio` and `s-btn-group__radio`
- Abstracted most button group test code to `button-group.test.setup.ts`
- Removed all but one `form` test (tests that wrap a `button` with `form` element)
- Added tests for buttons with badges
- Simplified the children tested

> [!NOTE]
> The default, form, and radio test images render identically. This is as intended to test those unique children within the button group.

giamir (approved):
Outstanding work. Reads much better than before and great catch on those a11y failures. 🎉

I took the liberty to remove an unnecessary `eslint-disable` rule from the test setup. 
Good to be merged :)

PR #1360, opened 2023-05-12
Removing shadows from sidebar

PiperLawson:

PR #1361, opened 2023-05-12
Top bar changes

PiperLawson:
For cprod only: Made light background white. Added border to white background

For cprod and teams: Removed shadow. Added height.

### DIFF
--- a/lib/atomic/spacing.less
+++ b/lib/atomic/spacing.less
@@ -26,6 +26,7 @@ body {
     --su-static24: 24px;
     --su-static32: 32px;
     --su-static48: 48px;
+    --su-static56: 56px;
     --su-static64: 64px;
     --su-static96: 96px;
     --su-static128: 128px;
--- a/lib/atomic/spacing.less
+++ b/lib/atomic/spacing.less
@@ -26,7 +26,6 @@ body {
     --su-static24: 24px;
     --su-static32: 32px;
     --su-static48: 48px;
-    --su-static56: 56px;
     --su-static64: 64px;
     --su-static96: 96px;
     --su-static128: 128px;
--- a/lib/exports/constants-colors.less
+++ b/lib/exports/constants-colors.less
@@ -242,8 +242,8 @@
     --theme-tag-hover-border-color: transparent;
 
     // Topbar
-    --theme-topbar-height: var(--su-static48);
-    --theme-topbar-background-color: var(--black-025);
+    --theme-topbar-height: var(--su-static56);
+    --theme-topbar-background-color: var(--white);
 
     // Topbar Search input
     --theme-topbar-search-color: var(--black-700);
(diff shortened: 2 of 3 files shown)


lib/atomic/spacing.less
```diff
@@ -26,6 +26,7 @@ body {
     --su-static24: 24px;
     --su-static32: 32px;
     --su-static48: 48px;
+    --su-static56: 56px;
```
dancormier: We can calculate this value with existing variables by doing something like `calc(var(--su-static48 ) + var(--su-static8))` to keep the scale consistent
```suggestion
```

lib/exports/constants-colors.less
```diff
@@ -242,8 +242,8 @@
     --theme-tag-hover-border-color: transparent;
 
     // Topbar
-    --theme-topbar-height: var(--su-static48);
-    --theme-topbar-background-color: var(--black-025);
+    --theme-topbar-height: var(--su-static56);
```
dancormier:
```suggestion
    --theme-topbar-height: calc(var(--su-static48 ) + var(--su-static8)); // 56px
```

dancormier (changes requested):
> For cprod only: Made light background white. Added border to white background
> 
> For cprod and teams: Removed shadow. Added height.


All of these changes will apply everywhere. Is that your intention @PiperLawson?

<details>
<summary>Before vs After</summary>

### Before
image

### After
image

</details>

I believe this will be a breaking change in Core since page top padding is relative to topbar height. It's not a _huge_ breaking change but we'd need to make sure it is addressed when this ships.

[Giamir beat me to mentioning it](https://github.com/StackExchange/Stacks/pull/1360#pullrequestreview-1424742107): We format commit messages (which are prefilled by the PR title) using [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) formatting. This PR might be merged with a commit message like `fix(topbar): update background and spacing`. I use [this cheat sheet](https://kapeli.com/cheat_sheets/Conventional_Commits.docset/Contents/Resources/Documents/index) to jog my terrible memory.

giamir:
>  It's not a huge breaking change but we'd need to make sure it is addressed when this ships.

I think it might require some thorough testing in Core because as far as I remember I saw banners being placed into the page based on the fixed height of the topbar. Certainly mergeable but would require some extra eyes. 👀

PiperLawson:
Can you explain this to me: 
> top padding is relative to topbar height

Where can I see how this might change things?

dancormier:
> Can you explain this to me:
> 
> > top padding is relative to topbar height
> 
> Where can I see how this might change things?

I had this a little backwards.

This is kinda similar to the background color variable where it's exposed via the theming API. Themes could override this value by setting the `--theme-topbar-height` variable, _but_ we have [this shim file](https://github.com/StackEng/StackOverflow/blob/main/StackOverflow/Content/less/views/_topbar-stacks-shim.less) that does the overriding instead. FWIW, _I think_ is there to account for announcement banners and similar items that can affect page layout.

PR #1362, opened 2023-05-12
Ignored post summary state

PiperLawson:
Updating this so I at least have one win with meta

### DIFF
--- a/lib/components/post-summary/post-summary.less
+++ b/lib/components/post-summary/post-summary.less
@@ -48,14 +48,14 @@
     // VARIANTS
     &&__deleted,
     &&__ignored {
-        --_ps-has-answers-bc: var(--black-300);
+        --_ps-has-answers-bc: var(--black-100);
         --_ps-has-answers-bg: transparent;
-        --_ps-has-answers-fc: var(--black-500);
+        --_ps-has-answers-fc: var(--black-350);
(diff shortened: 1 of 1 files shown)


giamir:
@PiperLawson could we get some references about why we are making those changes for future references. Is there a meta post we can link to this PR? Thanks

PiperLawson:
@giamir 
https://meta.stackexchange.com/questions/382047/my-brain-cannot-ignore-the-ignored-questions-due-to-contrast-change-please-chan/385181#385181

dancormier (changes requested):
I'm hesitant to approve this PR for accessibility reasons. This [fails color contrast guidelines](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) because the post summary can still be interacted with.

There is a way around _half_ of the issue: Most accessibility contrast checking tools check the values of the colors used. They ignore any modifications to `opacity`, so you can achieve a similar appearance just by dropping the opacity to something like `0.75`. 

## Examples

### Original

image

### Updated color values (this PR)
image


### Original w/ opacity: 0.75

image

I don't _love_ this approach for accessibility reasons, but if we have to ship this change, altering opacity would at least prevent us from failing some automated accessibility tests 🤷‍♂️

dancormier:
superseded by https://github.com/StackExchange/Stacks/pull/1370

PR #1363, opened 2023-05-12
refactor(less): use spacing unit variables where possible

dancormier:
This cleanup is to create some code consistency. As we hope to get some contributors outside the Stacks team, we need to lead by example. This PR replaces (most) existing static pixel values with our spacing unit custom properties. My hope is that this will encourage contributors to also use spacing units.

This should result in no visual changes to Stacks.

---

### A little post script on _why_ this matters, broadly speaking

> **Note**
> This feels like a good opportunity to outline some of the reasoning behind spacing units. This is just my (@dancormier) personal take. Feel free to read **none of this** if you want. I honestly just felt like getting it out of my head 🙂

First and foremost, for consistency. We use spacing units all over and I'd like the expectation to be that static pixel values are the _exception_ (they kinda already are but this PR makes them even more _exceptional_).

Speaking more broadly/philosophically, Stacks operates with the philosophy that a little markup, styling, and interactivity abstraction/restriction leads to a better developer experience, more consistent design, and improved maintainability. For instance, we have atomic color classes that help our UIs have a consistent look and reduce dev effort by providing a subset of colors to choose from, plus they allow for quick color adjustments across our entire codebase by just changing a few values in Stacks. Taking this further, high contrast mode wouldn't have been feasible without the centralized color definitions within Stacks.

**So, how does this relate to the spacing unit custom properties and why are they important?**

Applying spacing units to our components encourages us to adhere to a scale, justify when don't, and broadly will allow us to make the same broad changes as we do with other values (just like the color + high contrast example in the paragraph above). Our spacing units come in two forms: static and dynamic. The static ones us static pixel values (think `--su-static24: 24px;`), which the dynamic ones take that static value and multiply it by a base spacing unit value (like `--su24: calc(var(--su-static24) * var(--su-base));`) opening the door to potential dynamic spacing values. With these in place, it becomes more trivial to adjust the spacing across Stacks (and it's consumers) UI.

**How this relates to my feelings toward creating values within Stacks**

I believe it's a healthy practice to tend toward values that adhere to a predefined scale. I prefer we define scales for anything that will take them (line-height, spacing/sizing, colors, radii, weights, animation) using with custom properties so we can ease the task of making broad alterations and provide an inlet to creating an API to expose to consumers if we see fit. For a bonus, this has the potential for greater ease of choice and customization is of our UI _by the user_, which is an accessibility [best](https://www.markettailor.io/blog/role-of-personalization-in-increasing-website-accessibility) [practice](https://www.getstark.co/blog/why-personalization-is-key-to-your-accessibility-toolbox).


lib/atomic/misc.less
```diff
@@ -149,14 +149,14 @@
     // Borrowed from Wordpress
     // https://make.wordpress.org/accessibility/2015/02/09/hiding-text-for-screen-readers-with-wordpress-core/
     border: 0;
-    clip: rect(1px, 1px, 1px, 1px);
+    clip: rect(var(--su-static1), var(--su-static1), var(--su-static1), var(--su-static1)); // rect(1px, 1px, 1px, 1px)
```
giamir, line 152: I see sometimes you add comments sometimes you don't. Is there a rule you use to decide that (e.g. if it is just a `var` is self-explanatory enough)?

lib/base/reset-normalize.less
```diff
@@ -69,7 +69,7 @@ main { /* 1 */
  */
 
 figure {
-  margin: 1em 40px;
+  margin: 1em calc(var(--su-static32) + var(--su-static8)); // 1em 40px
```
giamir, line 72: Question for my understanding: why we don't store `1em` in a variable too? Are we just saying here use the same font size of your parent to the determine the vertical margin?

Side note for my benefits:
How do we decide when to use `px` over `em` over `rem`? Is there some historical document I can read about the decision?

docs/_includes/header.html
```diff
@@ -1,4 +1,4 @@
-<header class="s-topbar stacks-topbar ps-fixed h64 js-stacks-topbar print:d-none">
+<header class="s-topbar stacks-topbar ps-fixed js-stacks-topbar print:d-none">
```
giamir: Is this intentional? It doesn't seem related to this PR. Maybe it could be in its separate commit.

giamir (approved):
Looks good to me @dancormier.
I left some comments primarily for my benefits.

Feel free to merge when you are ready.
I also highly appreciated the rationale section in the PR description. It helps me a lot understanding why we use unit variables. 
What do you think about just copy/pasting that content and make it a Stacks ADR for posterity?

lib/atomic/misc.less
dancormier: Yep, I add comments if I use a `calc`. I figure there's enough room for misinterpretation that it warrants a comment saying what the computed value should be. With a `var`, it should wind up just being the value indicated by the `var`.

docs/_includes/header.html
dancormier: Good catch! This is a remnant from some premature optimization to support variable spacing. I have some local changes that I neglected to commit that support some spacing theming in the docs, but this change slipped by me. I'll remove it.

lib/base/reset-normalize.less
dancormier: > why we don't store `1em` in a variable too?

I think we'd want an existing scale first. Prior to creating the `--su` variables, we had `@su` Less variables (see [PR](https://github.com/StackExchange/Stacks/pull/873#issue-1165843841)).

> Are we just saying here use the same font size of your parent to the determine the vertical margin?

Seems that way. This seems like a reasonable enough reset, but honestly it predates me at SO 🤷‍♂️

> How do we decide when to use `px` over `em` over `rem`?

`px` and static and `em`/`rem` are not 🤷‍♂️. Ignoring this one case, I think we've tended towards `px` values for greater control. For instance, if we used `em`/`rem` all over and we adjust the root font size, the sizing of many elements would be impacted. I think we tend towards `px` for the sake of control and I think that still persists even with these `--su` variables since we actively differentiate between ones that could actually be variable (`--su[x]`) and static ones (`--su-static[x]`).

With all that said, I think I know where this is headed 🙂. We _could_ use `rem` instead of `px` values then just change the base font size to adjust spacing (without the need for all of these spacing units). It's an interesting idea but probably one for us to shove back into the dark recesses of our mind until we have time to consider it completely.

> Is there some historical document I can read about the decision?

There's no historical Stacks document that I'm aware of that provides justification or context for usage of `em`/`rem`/`px`/etc 🤷‍♂️.

dancormier:
> Feel free to merge when you are ready. I also highly appreciated the rationale section in the PR description. It helps me a lot understanding why we use unit variables. What do you think about just copy/pasting that content and make it a Stacks ADR for posterity?

Gladly! I'll add an ADR to this PR.

dancormier:
I've added an ADR to this PR that leverages much of what I wrote in the PR description. I could see it being broadened out to outline a philosophy towards having scales similar to our spacing units for most values used across our codebase, but for the sake of time/scope, I've kept it to just the spacing units for now.

lib/base/reset-normalize.less
giamir: Thanks for all the detailed replies. They helped. :)

PR #1402, opened 2023-06-26
fix(toggle-switch): apply lower opacity when disabled

dancormier:
This PR applies the disabled element opacity to `s-toggle-switch` components. Previously, the disabled state for `input[type="checkbox].s-toggle-switch` would result in no visual changes.

### Before

image

### After

image

---

> **Note**
> This should be good to merge after we containerize visual regression tests


giamir:
@dancormier out of curiosity why we changed so many unrelated baseline images as part of this PR?

dancormier:
> @dancormier out of curiosity why we changed so many unrelated baseline images as part of this PR?

I'm not sure.Looks like when I run `npm run test:visual:update` it makes _minor_ changes to every image.

image

I'm gonna dig into commits between the last time we've updated the images and now to see if there's any change that might have caused this..

dancormier:
I went back to the last time baseline images were updated (https://github.com/StackExchange/Stacks/commit/a813110cedb915d9b80d985c17dc87d4fa8a5153), tried another update, and saw the same thrashing. My initial theory is that an OS or browser update has modified how these components get rendered, but that theory might not explain why tests are failing via GitHub actions.

I'm going to open a new PR that includes regenerated baseline images and see if it resolves test failures and consider merging in those updated images. Expect a PR review request @giamir

Edit: I just noticed they're all Chromium images that are getting regenerated, so I'm guessing a browser update included a slight rendering update 🤷‍♂️

PR #1403, opened 2023-06-27
chore: regenerate Chromium visual regression test images

dancormier:
When updating baseline visual regression test images, all Chromium images change slightly. My working theory is that there's been a slight change to rendering in Chromium. In any case, this PR updates Chromium baseline images so that we can prevent some test thrashing.
See https://github.com/StackExchange/Stacks/pull/1402#issuecomment-1609785241


giamir (approved):
LGTM. Probably one more reason to dockerize things :)

PR #1409, opened 2023-07-06
fix(button): use inherit when --_bu-bg is undefined

dancormier:
This PRs updates `s-btn` styling to inherit background color from a parent. This is to resolve an issue where buttons with a transparent background would, well… be transparent, causing any element place behind it to show through.

### Before
image

### After
image

cc and h/t @kylejrp

### DIFF
--- a/lib/components/button/button.less
+++ b/lib/components/button/button.less
@@ -499,8 +499,9 @@
     user-select: none;
 }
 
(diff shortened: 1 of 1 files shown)


giamir (approved):
@dancormier LGTM.  Nice catch.

PR #1414, opened 2023-07-12
build: minify docs HTML output

dancormier:
This PR does what it says on the tin. Just something I whipped up in the process of dealing with another docs issue and figured I'd capture.

I think this is pretty safe to merge, but it's not urgent and we're probably best off running through the docs to make sure this change didn't cause any unexpected issues.

### DIFF
--- a/docs/.eleventy.js
+++ b/docs/.eleventy.js
@@ -7,6 +7,7 @@ const iconPlugin = require("./plugins/icons");
 const bannerExamplePlugin = require("./plugins/banner-example");
 const tipPlugin = require("./plugins/tip");
 const markdownPlugin = require("./plugins/markdown");
+const htmlmin = require("html-minifier");
 
 module.exports = function(eleventyConfig) {
   eleventyConfig.setQuietMode(true); // Reduce the console output
@@ -39,4 +40,19 @@ module.exports = function(eleventyConfig) {
   // Ignore liquid parsing on these files
   eleventyConfig.ignores.add('email/templates/code');
   eleventyConfig.ignores.add('email/templates/examples');
+
+  eleventyConfig.addTransform("htmlmin", function(content, outputPath) {
+    if (outputPath && outputPath.endsWith(".html")) {
+      const shouldMinify = !content.includes("---");
+      if (shouldMinify) {
+        let minified = htmlmin.minify(content, {
+          useShortDoctype: true,
+          removeComments: true,
+          collapseWhitespace: true
+        });
+        return minified;
+      }
+    }
+    return content;
+  });
 }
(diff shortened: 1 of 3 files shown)


docs/.eleventy.js
```diff
@@ -39,4 +40,19 @@ module.exports = function(eleventyConfig) {
   // Ignore liquid parsing on these files
   eleventyConfig.ignores.add('email/templates/code');
   eleventyConfig.ignores.add('email/templates/examples');
+
+  eleventyConfig.addTransform("htmlmin", function(content, outputPath) {
+    if (outputPath && outputPath.endsWith(".html")) {
+      const shouldMinify = !content.includes("---");
```
giamir, line 46: It is not clear at a glance why we exclude content containing 3 dashes. It might be worth a comment or rename the `shouldMinify` variable to hint in the right direction.

docs/.eleventy.js
```diff
@@ -39,4 +40,19 @@ module.exports = function(eleventyConfig) {
   // Ignore liquid parsing on these files
   eleventyConfig.ignores.add('email/templates/code');
   eleventyConfig.ignores.add('email/templates/examples');
+
+  eleventyConfig.addTransform("htmlmin", function(content, outputPath) {
+    if (outputPath && outputPath.endsWith(".html")) {
```
giamir, line 45: This is just a personal opinion: I tend to prefer [guard clauses](https://maximegel.medium.com/what-are-guard-clauses-and-how-to-use-them-350c8f1b6fd2) over nested if statements.
The block below could be somenthing like (pseudo)
```
if (!html) {
  return content
}

if (contains 3 dashes thing I did not fully understand) {
  return content
}

return minify(...)
```
Again this is not really an objective comment (feel free to discard).

giamir (commented):
Everything run fine in my local.

We could merge but in general I am a bit afraid of premature optimizations like this one.
We introduced a dependency and some code: do we get enough value out of it to justify those costs?

I will leave it to you to make a judgment call, I am ok to have this merged after the comment about the variable naming has been addressed. 🙂

docs/.eleventy.js
dancormier: That line for sure could use a comment. We exclude `---` because that's what wraps the front matter of the liquid templates. Without this exclusion, the minification is attempted on the front matter which breaks the build.

dancormier:
> Everything run fine in my local.
> 
> We could merge but in general I am a bit afraid of premature optimizations like this one. We introduced a dependency and some code: do we get enough value out of it to justify those costs?

IMO, the costs are low but so are the benefits. I didn't fully expect to merge this and it's probably not worth too much more consideration, so I'm happy to see this closed and have this PR as a little historical record if we ever change our minds.

> 
> I will leave it to you to make a judgment call, I am ok to have this merged after the comment about the variable naming has been addressed. 🙂

Closing w/o merging! Thanks for checking it out

PR #1415, opened 2023-07-14
feat(color): Add new color palette

dancormier:
> **Note**
> This PR is targeted to merge into `stacks-next`, our long-lived branch intended for our v2 release

**tl;dr**: This PR adds the new color palettes and updates how they're generated.

---

## The `new` suffix

All added colors include the `new` suffix. The new color should come immediately before any color stop or h/s/l/ 

```Less
/* Legacy */
--white: …;
--black-200: …;
--theme-base-primary-color-h: …;
.fc-theme-primary-400 { … }

/* New */
--white-new: …;
--black-new-200: …;
--theme-base-primary-color-new-h: …;
.fc-theme-primary-new-400 { … }
```

## New colors

All updated color palettes from [the Figma](https://www.figma.com/file/cCU6A56sHwZluraDoyg5o7/Colors%3A-Accessible-Color-Palette?type=design&node-id=1303-30963&mode=design&t=IitSCpq8f7TzKmwD-0) have added. See the new [Colors docs page](https://deploy-preview-1415--stacks.netlify.app/product/base/colors-new/) for more details.

## Theme generation

Theme colors are now reduced to 6 stops (down from 14), plus a base class/variable that maps to the 400 stop (e.g.: `--theme-primary-new: var(--theme-primary-new-400);`). All theme colors now fall back to  The generation of theme variables/classes now only relies on h/s/l variables, removing any need for r/g/b variables.

See the new [Theming docs page](https://deploy-preview-1415--stacks.netlify.app/product/guidelines/theming-new/) for more details.

### Base/dark h/s/l/ variables

TODO: Explain how now they're mutually exclusive of one another. Basically, color variables will not fall back to custom light values in dark mode and instead will fall back to default Stacks Classic theme values. This may affect Teams/enterprise, but that needs to be verified.

### TODO

- [x] [New color palette](https://www.figma.com/file/cCU6A56sHwZluraDoyg5o7/Colors%3A-Accessible-Color-Palette?type=design&node-id=1303%3A30963&mode=design&t=8WSgFJeINP8imMmA-1) with updated stops is created (including new less functions to generate “themed“ palettes) and includes the new suffix
- [x] Create a colors doc page Colors (new) and to showcase new colors
- [x] Create a Theming (New) page to showcase the new themed variables (this page will be finalized in a follow up story)


netlify.toml
```diff
@@ -80,6 +80,10 @@
   from = "/base/colors"
   to = "/product/base/colors"
 
+[[redirects]]
```
giamir: @dancormier I don't think this redirect rule is necessary, this is a brand new page. In fact I do think we should check in netlify if those rules are actually still executed (e.g. there is still some references somewhere in the world pointing to them) otherwise we should consider removing them. 🙂

lib/exports/colors-new/colors.less
```diff
@@ -0,0 +1,54 @@
+@import (reference) "./color-mixins.less";
+
+// EXISTING ISSUES
+// - theme primary vars not generated properly 🤷‍♂️ (--_ private var issue via Less)
```
giamir: @dancormier how confident are we about making those variables "pseudo" private? Do we have enough evidence they won't need to be referred in Core?
What are the consequences of making these variables private from a documentation perspective? [Today we are documenting them as if people could access them](https://stackoverflow.design/product/base/colors/). 

In the spirit of working on iterative small steps, what do you think about dropping the transition to "pseudo" private variable at least for this PR and revaluate at some point down the line (still before releasing 2.0)? 

I am suggesting this also because my understanding is that some of those issues reported here are likely due to that.

lib/exports/colors-new/color-sets.less
```diff
@@ -0,0 +1,419 @@
+// White
```
giamir, line 4: @dancormier Getting many style linting problems here about not known CSS properties in the mixins. My understanding is that we are using mixins in here to remedy the lack of better data structures in Less. Since this is intentional we can add this comment at the top of this file:
```
/* stylelint-disable property-no-unknown */
```

netlify.toml
dancormier: Yeah, that was me being heavy-handed with duplicating _everything_ we have for the legacy colors. I'll get rid of this change at the very least.

lib/exports/colors-new/colors.less
dancormier: That comment is outdated and needs to be removed altogether. I've already fixed the issue mentioned there and dropped the pseudo private prefix from the colors and functions but still have a note or two lying around. I'll do a clean up soon.

docs/product/guidelines/theming-new.html
```diff
@@ -0,0 +1,537 @@
...
+    </ol>
+    <p class="stacks-copy">There are also <a class="s-link" href="{{ "/product/guidelines/conditional-classes#dark-mode" | url }}">conditional classes</a> that can be applied to override assumed dark mode colors, force light mode, or to force dark mode. Forcing modes can be good for previews in admin-only situations.</p>
+</section>
+
+<section class="stacks-section">
+    {% header "h2", "Theming API" %}
+    <p class="stacks-copy">Stacks provides an API to apply custom themes.</p>
+
+    {% header "h3", "Programmatic theme generation" %}
+    <p class="stacks-copy">Use <code class="stacks-code">.create-custom-theme-hsl-rgb-variables(@color, @tier, @modeCustom)</code> to create a custom theme.</p>
```
giamir: Can you remind me why we use `base` and `dark` for themes instead of `light` and `dark`? Is it to reinforce that custom themes might not specify a dark mode and the "base" one won't necessary be "light"?

docs/product/guidelines/theming-new.html
```diff
@@ -0,0 +1,537 @@
...
+            </tbody>
+        </table>
+    </div>
+
+    <div class="stacks-preview _preview-none mb24">
+{% highlight css %}
+.theme-custom.themed {
+    .create-custom-theme-hsl-rgb-variables(hsl(172, 37%, 48%), primary);
+    .create-custom-theme-hsl-rgb-variables(hsl(259, 29%, 55%), secondary);
+    .create-custom-theme-hsl-rgb-variables(hsl(172, 37%, 48%), primary, dark);
```
giamir: `base` and `dark` mode colors are the same in this example. Is this something that makes sense in a real world scenario? If not, we should change the values to don't confuse the consumers.

docs/product/guidelines/theming-new.html
```diff
@@ -0,0 +1,537 @@
...
+{% highlight css %}
+/* Input */
+.theme-custom.themed {
+    .create-custom-theme-hsl-rgb-variables(hsl(172, 37%, 48%), primary);
+}
+
+/* Output */
+.theme-custom.themed {
+    /* HSL variables */
+    --theme-base-secondary-color-new-h: 172;
```
giamir: Based on the input I would expect these variables to contain `primary` instead of `secondary`.

docs/product/guidelines/theming-new.html
```diff
@@ -0,0 +1,537 @@
...
+    .create-custom-theme-hsl-rgb-variables(hsl(259, 29%, 55%), secondary);
+    .create-custom-theme-hsl-rgb-variables(hsl(172, 37%, 48%), primary, dark);
+    .create-custom-theme-hsl-rgb-variables(hsl(259, 29%, 55%), secondary, dark);
+}
+{% endhighlight %}
+    </div>
+
+        {% header "h4", "Output" %}
+
+    <p class="stacks-copy">Explain how the above function the variables below and how a consumer can opt to forgo the function and just set the variables. Make sure to describe the base/dark and primary/secondary variants.</p>
```
giamir: It looks like this is explained already in the paragraph that follows:
`Manual addition of theme variables`

docs/product/guidelines/theming-new.html
```diff
@@ -0,0 +1,537 @@
...
+    --theme-primary-new-custom-500: hsl(var(--theme-base-primary-color-new-h), var(--theme-base-primary-color-new-s), calc(var(--theme-base-primary-color-new-l) - 5%));
+    --theme-primary-new-custom-600: hsl(var(--theme-base-primary-color-new-h), var(--theme-base-primary-color-new-s), calc(var(--theme-base-primary-color-new-l) - 10%));
+}
+{% endhighlight %}
+    </div>
+
+    {% header "h4", "Manual addition of theme variables" %}
+    <p class="stacks-copy">If you need to apply a theme without using the above function, you can do so by manually adding the variables above to your CSS. The most common use for this approach is when the theme needs to change client-side, such as when allowing the user to change and preview a theme dynamically.</p>
+
+    <p class="stacks-copy">With this approach, we recommend targeting new h/s/l and r/g/b color variables on a parent element that includes <code class="stacks-code ws-nowrap">.themed</code> class.</p>
```
giamir: Love the sliders. ❤️

Out of curiosity why we use the RGB values to generate the first 3 color stops (100, 200, 300) and HSL values for the last 3 (400, 500, 600)?
There is a part of me that don't understand why we are asking the user to pass both RGB and HSL values. 
Thanks in advance for clarifying.

docs/product/base/colors-new.html
```diff
@@ -0,0 +1,239 @@
...
+
+<!-- TODO find a good place for this info -->
+<section class="stacks-section">
+    {% header "h2", "Theming" %}
+    <p class="stacks-copy">
+        Stacks’s approach to theming is powered by CSS variables that are split into their HSL and RGB components. Our default primary color is based on Stack Overflow’s branded <code class="stacks-code">orange-400</code> represented as <code class="stacks-code">hsl(27, 90%, 55%)</code>. We can store these as separate values, apply various color operations, and then recombine them into on-demand color palettes to use for theming.
+    </p>
+</section>
+
+<section class="stacks-section">
```
giamir, line 121: TODO: needs to update the section about "Aliases" both for body text and danger error mapping.

giamir (commented):
@dancormier PR looks good to me overall. Love the new theming page. 🙂

I took the liberty to add css snapshots also for the generated color css atomic classes.
By doing that I realize that in the new file we are missing the status class aliases (danger, etc...). [diff](https://diffy.org/diff/6ff98a13749a7)

I also noticed that we haven't created new variable or classes for typography color aliases
https://github.com/StackExchange/Stacks/blob/develop/lib/atomic/typography.less#L70-L75

Do you think we should look into aliases before merging this PR? 
Or do you want to work on them as part of the next story?

I left some additional minor comments through the code.

Thank you for moving this forward!

docs/product/guidelines/theming-new.html
dancormier: > Is it to reinforce that custom themes might not specify a dark mode and the "base" one won't necessary be "light"?

That might be the historical context. I'm honestly not sure what the original motivation was but as it stands, the `light` mode is the de facto "base"/"default" mode. I maintained `base` here for two (related) reasons:

- It directly corresponds to the variables that it creates `--theme-@{modeCustom}-@{tier}-color-new`
- Maintaining parity with what already exists to reduce liability


It's a little bit of us bending to change fewer things for our biggest customer (Core), though I'm not totally opposed to changing it to be `light` and `dark`. 

We'd have to either update the mixin with a conditional so if the mode if `light`, it uses `base` in the variable name or we'd have to ensure consumers update their references to `--theme-base-[…]` to be `--theme-light-[…]`. I'd probably prefer the latter since it creates more consistency in Stacks and reinforces the notion that their is no "base" or "default" mode (which is not currently the case, but something I'd like to move towards).

docs/product/guidelines/theming-new.html
dancormier: It _honestly_ does not make a lot of sense to me either. I'd be open to changing this but generally, I yielded on the site of keeping these sorts of things the same to limit risk (and partially out of laziness since I didn't want to have to rewrite our `.native-tint` and `.native-shade` Less functions).

Do you think it would be worth it to change now?

docs/product/guidelines/theming-new.html
dancormier: > Is this something that makes sense in a real world scenario?

Not really!

> If not, we should change the values to don't confuse the consumers.

On it!

dancormier: Well actually… this could make sense. It's what I'm using for the theming in the docs. Because the dark color generation kinda just flips the order of light to dark, it does work just fine.

In any case, I think consumers will _mostly_ supply different colors (or at least shades) to the `base` vs `dark` modes, so I'll make them distinct here.

docs/product/guidelines/theming-new.html
giamir: Can I suggest to create a list of tradeoffs we are making for reducing liability / maintaining parity?
Nothing fancy - adding something to [this document](https://docs.google.com/document/d/18JkYEVioPhPmHGZaV-ba8S-3rl_zdbMIhpqtsX7iqYk/edit?usp=sharing) should be enough.
I am suggesting this because we could reflect on those later this month when we are close to release 2.0 and decide if it's worth adjusting them as part of this major bump or if we are ok to accept them as "tech debt" for the foreseeable future.

So my suggestion is to make a note about the ideal API we would like and leave things as they are for this PR. Potentially revisit later when we are a bit closer to releasing 2.0.
How does it sound?

docs/product/guidelines/theming-new.html
giamir: Similarly to the comment I made above we should at least have a place where we track all these tradeoffs we make to reduce liability and amount of work. 
I think it's ok to leave things as they are for this PR but it will be important to have a conversation about those "questionable" APIs **before** releasing 2.0. 
Can you add a point for this in [this document](https://docs.google.com/document/d/18JkYEVioPhPmHGZaV-ba8S-3rl_zdbMIhpqtsX7iqYk/edit?usp=sharing)? Thanks 🙂

dancormier:
> […] we are missing the status class aliases (danger, etc...)
> […] we haven't created new variable or classes for typography color aliases
> 
> Do you think we should look into aliases before merging this PR? Or do you want to work on them as part of the next story?

Good catch! Those classes and variables should be in this PR. I'll add them today.

- [x] status class aliases (danger, etc...) https://github.com/StackExchange/Stacks/pull/1415/commits/12f409493262bf7aae5fbd97d98ef2f3f750c167
- [x] `--fc-[dark|medium|light]` variables https://github.com/StackExchange/Stacks/pull/1415/commits/31eb1358955953499826bb71442e0f17ea68ea5e
- [x] `.fc-[dark|medium|light]` classes https://github.com/StackExchange/Stacks/pull/1415/commits/12f409493262bf7aae5fbd97d98ef2f3f750c167

lib/atomic/color-new.less
```diff
@@ -0,0 +1,7 @@
+@import (reference) "../exports/colors-new/color-mixins.less";
+
+& {
+    @set: .sets-mode()[default];
+    .create-colors(@set, classes);
+    .create-colors(.sets-theme-default(), classes);
```
dancormier, line 6: This generates `.fc`, `.bg` theming classes.

lib/exports/colors-new/__snapshots__/colors.less.test.ts.snap
```diff
@@ -0,0 +1,752 @@
...
+body:not(.theme-dark),
+body.theme-dark .theme-light__forced,
+body.theme-system .theme-light__forced,
+body:not(.theme-dark).themed,
+body.theme-dark .theme-light__forced.themed,
+body.theme-system .theme-light__forced.themed,
+body:not(.theme-dark) .themed,
+body.theme-dark .theme-light__forced .themed,
+body.theme-system .theme-light__forced .themed {
+    --white-new: hsl(0, 0%, 100%);
```
dancormier, line 89: We generate color variables _per mode_.

lib/exports/colors-new/__snapshots__/colors.less.test.ts.snap
```diff
@@ -0,0 +1,752 @@
...
+    --highlight-new-bg: hsl(0, 0%, 96.5%);
+    --highlight-new-color: var(--black-750);
+    --highlight-new-comment: hsl(210, 8%, 43.5%);
+    --highlight-new-deletion: var(--red-600);
+    --highlight-new-keyword: hsl(206, 98.5%, 29%);
+    --highlight-new-literal: hsl(27, 99%, 36%);
+    --highlight-new-namespace: hsl(27, 99%, 36%);
+    --highlight-new-punctuation: var(--black-600);
+    --highlight-new-symbol: hsl(306, 43%, 35%);
+    --highlight-new-variable: hsl(80, 80.5%, 26.5%);
```
dancormier: > **Note**
> `--highlight-*` variables referencing legacy colors need to be remapped.
> This should be done in a follow-up PR.

lib/exports/colors-new/__snapshots__/colors.less.test.ts.snap
```diff
@@ -0,0 +1,752 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`colors > should output all the css generated by the new colors/theming generation mixins 1`] = `
+"body {
```
dancormier, line 4: In `body`, we have the baseline state alias (error, success, etc) and component-specific theme variables defined.

docs/product/guidelines/theming-new.html
```diff
@@ -0,0 +1,460 @@
...
+</section>
+
+<section class="stacks-section">
+    {% header "h2", "Theming API" %}
+    <p class="stacks-copy">Stacks provides an API to apply custom themes.</p>
+
+    {% header "h3", "Programmatic theme generation" %}
+    <p class="stacks-copy">Use <code class="stacks-code">.create-custom-theme-hsl-variables(@color, @tier, @modeCustom)</code> to create a custom theme.</p>
+        
+    <p class="stacks-copy">This function generates two sets of CSS variables: 1) independent h/s/l and r/g/b color variables and 2) variables at each colors stop that reference the h/s/l and r/g/b variables. Provide this function the arguments defined below to generate theme colors which will apply across Stacks.</p>
```
giamir: NIT: remove r/g/b color variables references from the text since they are not generated anymore

docs/product/guidelines/theming-new.html
giamir: This has now been addressed. We only use hsl.

docs/product/guidelines/theming-new.html
```diff
@@ -0,0 +1,460 @@
...
+    .create-custom-theme-hsl-variables(hsl(172, 37%, 48%), primary);
+}
+
+/* Output */
+.theme-custom.themed {
+    /* HSL variables */
+    --theme-base-primary-color-new-h: 172;
+    --theme-base-primary-color-new-s: 37%;
+    --theme-base-primary-color-new-l: 48%;
+    /* Color variables based on HSL and RGB variables */
```
giamir: NIT: Remove RGB from comment

docs/product/base/colors-new.html
```diff
@@ -0,0 +1,228 @@
...
+            </div>
+        {% endif %}
+    {% endfor %}
+</section>
+
+<!-- TODO find a good place for this info -->
+<section class="stacks-section">
+    {% header "h2", "Theming" %}
+    <p class="stacks-copy">
+        Stacks’s approach to theming is powered by CSS variables that are split into their HSL and RGB components. Our default primary color is based on Stack Overflow’s branded <code class="stacks-code">orange-400</code> represented as <code class="stacks-code">hsl(27, 90%, 55%)</code>. We can store these as separate values, apply various color operations, and then recombine them into on-demand color palettes to use for theming.
```
giamir: NIT: RGB needs to be removed by this paragraph I believe.

docs/product/base/colors-new.html
```diff
@@ -0,0 +1,228 @@
...
+                            <div class="ws-nowrap">.fc-new-error</div>
+                        </div>
+                    </td>
+                    <td class="p8">
+                        <div class="d-flex g4">
+                            <div class="bg-new-danger p8 fc-white bar-sm ws-nowrap">.bg-new-danger</div>
+                            <div class="bg-new-error p8 fc-white bar-sm ws-nowrap">.bg-new-error</div>
+                        </div>
+                    </td>
+                    <!-- TODO mention deprecation of border-color alias classes (which we documented in legacy colors but didn't generate 😬) -->
```
giamir: Why the border-color alias deprecation is only for the danger/error state?
I see that we are still documenting the border color classes for success and warning.

Why the deprecation for that aliases was decided on the first place?
I see that we are documenting them also in the borders doc: https://stackoverflow.design/product/base/borders/#red

## Side note
It looks like we should create a Borders (new) page too as part of this PR. 
I can see there are plenty of references to the old color palette / stops
What do you think?

docs/product/base/colors-new.html
```diff
@@ -0,0 +1,228 @@
...
+        Stacks’s approach to theming is powered by CSS variables that are split into their HSL and RGB components. Our default primary color is based on Stack Overflow’s branded <code class="stacks-code">orange-400</code> represented as <code class="stacks-code">hsl(27, 90%, 55%)</code>. We can store these as separate values, apply various color operations, and then recombine them into on-demand color palettes to use for theming.
+    </p>
+</section>
+
+<section class="stacks-section">
+    {% header "h2", "Aliases" %}
+    {% header "h3", "Body text" %}
+    <div class="stacks-preview">
+{% highlight html %}
+<p class="fc-light">…</p>
```
giamir: Was there a specific reason why we have decided to don't create aliases for body text (e.g. `fc-new-light`, etc...) and then showcasing them here?

lib/exports/colors-new/theme.less
```diff
@@ -0,0 +1,86 @@
+.theme-variables() {
```
giamir, line 1: For consistency should we add the `-new` suffix to these variables as well?
(of course we will need to drop it in the next story but it might be a good reminder that we will actually have to add the `legacy` suffix to the old theme variables).

What is the reason we haven't switched to new color variables here? 
Do we need clarification from design?

lib/exports/colors-new/color-sets.less
```diff
@@ -0,0 +1,614 @@
...
+}
+.set-focus-hc() {
+    default: hsla(206, 100%, 40%, 0.9);
+    success: hsla(140, 40%, 40%, 0.9);
+    warning: hsla(47, 76%, 46%, 0.9);
+    error: hsla(358, 62%, 47%, 0.9);
+    muted: hsla(210, 8%, 55%, 0.95);
+}
+
+// highlight
```
giamir, line 405: I could not find any documentation about these highlights variables, is this something we refer only in the `code-block` and `prose` styles or also in Core or the editor for example? Try to understand if it makes sense to expose these extra variables or we could encapsulate them directly in those 2 components for the long run.

If the variables are here to be consumed outside stacks I think we should document them somewhere in our docs (maybe a small section in the Colors (new) page)

Any reason why we could not assign the variables to the new color set? Waiting confirmation from design?

giamir (approved):
Great work @dancormier. 🎉 

I left few comments across the PR but overall I think we could merge it.

The most important point I had is if we should create a `Borders (new)` page before merging, to document how borders are expected to be used with the new color palette. (I also noticed we don't create some `bc` classes/variables anymore - like bronze-darker, etc... - I suppose that is intentional).

There are also few additional NITs.

lib/exports/colors-new/__snapshots__/colors.less.test.ts.snap
giamir: I wrote a comment about this in my review. I guess this should be done in a follow up PR because we need design to confirm the mapping `old` -> `new` color?

lib/exports/colors-new/color-sets.less
dancormier: > I could not find any documentation about these highlights variables, is this something we refer only in the `code-block` and `prose` styles or also in Core or the editor for example? Try to understand if it makes sense to expose these extra variables or we could encapsulate them directly in those 2 components for the long run.

AFAICT, we only refer to these in the the `code-block` and `prose` styles. We _technically_ [redefine and reference them in Core](https://github.com/StackEng/StackOverflow/blob/main/StackOverflow.Api.V2/Content/all.css#L1008-L1100) but they all have equivalent [values](https://github.com/StackExchange/Stacks/blob/dcormier/colors-2023/lib/exports/constants-colors.less#L418-L430) and [references](https://github.com/StackExchange/Stacks/blob/dcormier/colors-2023/lib/components/code-block/code-block.less#L12-L84) in Stacks.

I think we could safely consider them private and encapsulate them in those two components. I might want to leave that for a follow up PR though just to minimize moving parts.

> Any reason why we could not assign the variables to the new color set? Waiting confirmation from design?

I left these as-is since _most_ of the values referenced are of direct color references (`hsl(…)` instead of `var(…)`). I've gone ahead and changed what I could to map to the new color palette, but many colors here are one-offs and don't map to any Stacks colors (new or legacy). I'd like to get design's opinion on this before making too many changes here, though I don't think it needs to hold up this PR.

---

@PiperLawson considering most colors used for [code blocks](https://deploy-preview-1415--stacks.netlify.app/product/components/code-blocks/#language-examples) are custom and not used elsewhere in Stacks, are we planning on updating the colors or do we want to keep those as-is?

docs/product/base/colors-new.html
dancormier: > Why the border-color alias deprecation is only for the danger/error state? I see that we are still documenting the border color classes for success and warning.

This is an oversight on my part. I intended to remove documentation for `.bc-*` alias classes for any states, not just danger/error. After further consideration though, we probably should hold on deprecating this for consistencies sake. I'm going to add the danger/error border color aliases back and ensure the classes are generated.

> Why the deprecation for that aliases was decided on the first place? I see that we are documenting them also in the borders doc: https://stackoverflow.design/product/base/borders/#red

> ## Side note
> 
> It looks like we should create a Borders (new) page too as part of this PR. I can see there are plenty of references to the old color palette / stops What do you think?

Yep, this is a good catch! I'll add generation for the border color classes using the new colors and the new Borders docs page.

docs/product/base/colors-new.html
dancormier: We did, I just didn't document them! I've now updated the documentation with these classes

lib/exports/colors-new/theme.less
dancormier: I don't think these theme variables should get `new` or `legacy` suffixes, since they never define color values; they only reference them. What I decided to do in https://github.com/StackExchange/Stacks/pull/1415/commits/afa19ede76398b3b75ab99a245082c2d690bac68 is just remove duplicates of these variables so they're only generated from this file.

I see switching the values of these variables to the new color variables here as work to when we get to "[Update Stacks Components Styles to use new Color Palette](https://stackoverflow.atlassian.net/browse/STACKS-349)".

giamir: Understood, it makes sense. 👍

lib/exports/colors-new/color-sets.less
giamir: I also think we should not block this PR because of them. We can revisit those colors later on this month if necessary after we get feedback from design.

giamir (approved):
Thanks @dancormier for taking care of the borders!
We should merge this PR today. Awesome work! 🎉

PR #1422, opened 2023-07-19
test(visual-regression): run on a docker container

giamir:
[STACKS-189](https://stackoverflow.atlassian.net/browse/STACKS-189)

## Context 
Our visual regression tests run in all sort of environments (ubuntu in CI, macOS, windows). Different OSes render browser screenshots slightly different therefore we are forced to set [a generous threshold](https://github.com/StackExchange/Stacks/blob/develop/web-test-runner.config.mjs#L49) for the comparison algorithm which sometimes is not even enough (test comparison was failing in windows machines).

## Technical Notes
Despite all baseline images being updated the only file being added by this PR is [this powershell script](https://github.com/StackExchange/Stacks/blob/541177bdf4517d093259ee0b8f6d8723f7408d99/run-test-visual.ps1) (initially [bash](https://github.com/StackExchange/Stacks/blob/ecf45b2b1650ae546b93f12b03d5fb0796fc499f/run-test-visual.sh) - moved to pwsh for better windows compatibility).
The script is only 19 lines and includes the Dockerfile inline. 
Relevant files to run the tests are bind mounted to the container.
There are minor things I could not get to work as I wanted (although they are not a showstopper):
-  Use image ids instead of relying on tags (apparently docker for windows doesn't like referencing to image ids in the docker run command)
- Having to actually create and remove a dockerfile on the fly instead of simply injecting the docker file in the build command (I could do that in [bash](https://github.com/StackExchange/Stacks/blob/ecf45b2b1650ae546b93f12b03d5fb0796fc499f/run-test-visual.sh) but not in pwsh) 🤔

## How to test
- Freshen up your dependencies with `npm i`
- Make sure you have [`docker`](https://docs.docker.com/engine/install/) running in your machine (whether that is windows, macos or a linux distribution)
- Make sure you have [`git-lfs`](https://docs.github.com/en/repositories/working-with-files/managing-large-files/installing-git-large-file-storage) and [`pwsh`](https://learn.microsoft.com/en-us/powershell/scripting/install/installing-powershell?view=powershell-7.3) installed in your machine too
- Run `npm run test:visual` (go grab a coffee - tests will take between 3 to 6min to run depending on your hardware - they are comparing 4K+ images)


giamir:
@dancormier I am merging this after applying feedback from the Core Engineering team. It should work in all environments. Let me know if you encounter any unforeseen issue when you rebase your branch.

PR #1440, opened 2023-08-07
Implement legacy naming for old color palette

dancormier:
> **Note**
> This PR is targeted to merge into `stacks-next`, our long-lived branch intended for our v2 release

**tl;dr**: This PR removes the `new` prefix from all new colors, adds the `legacy` prefix to old colors, refines much of the theme generation, and updates docs. It builds from the previous related PR: https://github.com/StackExchange/Stacks/pull/1415

### Color naming

- New colors no longer include the `new` suffix
- All old colors now include the `legacy` suffix
- All references to legacy, new colors updated
- Internally, legacy and new colors are now referred to as `v1` and `v2` respectively

### Docs

- New docs pages have dropped the `(new)` label
- Legacy docs pages have gained the `(legacy)` label and have been moved to `v1/` path
- New utility color classes (`.fc-*`, `.bc-*`, `.bg-*`) are now entirely documented on the [colors docs page](https://deploy-preview-1440--stacks.netlify.app/product/base/colors/#classes)

### Other

- Moved new color files out of `new` directories, moved legacy into `v1` directories
- Classes previously tied to legacy color generation (such as `*-transparent`) have been moved to new color generation
- Resolved any issues with color generation

## The most relevant changes

> **Note**
> The only changes to all files within `lib/components` should be updating references to legacy colors

Since this PR touches so many files, I figured it may be helpful to highlight files with the most relevant and significant changes:

- [lib/atomic/color.less](https://github.com/StackExchange/Stacks/pull/1440/files#diff-2c75d563f3eae108a7163233a5135d98be0572d91e16b444bfc9cf88435f5b40)
    - [lib/atomic/__snapshots__/color.less.test.ts.snap](https://github.com/StackExchange/Stacks/pull/1440/files#diff-ece4f4e3bf606b04a253b8ec80107a9241127fd1cba05920ce9ba20bcc2f6a54)
- [lib/exports/color.less](https://github.com/StackExchange/Stacks/pull/1440/files#diff-a3ba1b63e7acab9bc351c8b6a6298f4fecd1f04a5732e013fbdf992364764ddc)
    - [lib/exports/__snapshots__/color.less.test.ts.snap](https://github.com/StackExchange/Stacks/pull/1440/files#diff-b5298c2cb59175aea95b33e42b4aa0df9a39bc68d001e7387ea8f625b812f6e2)
[...]

PR #1449, opened 2023-08-16
Apply v2 colors to components

dancormier:
This PR update components to use the v2 color palettes, per the [stops mappings](https://docs.google.com/document/d/1XbS8mJfsXYqu-LunHZDXDqk4MOSz-04aVhzx_-qj_mk/edit#heading=h.67twajeucjws).

> **Note**
> I've opted to keep the remapping of color references in the docs separate. See https://github.com/StackExchange/Stacks/pull/1450 for the docs updates.

> **Note**
> I've create a branch with fixes for the issues under the **"Design notes"** section below. See https://github.com/StackExchange/Stacks/pull/1451 (it  _somehow_ has more APCA failures than this branch 🤷‍♂️)

> **Note**
> For this PR, here are the results of `npm run test:a11y` as of Aug 21, 2023:
>    > Chromium: |██████████████████████████████| 16/16 test files | 787 passed, 333 fa
iled
>    > 
>    > Finished running tests in 35.6s with 333 failed tests.

[accessibility test run at 20230821-112331](https://gist.github.com/dancormier/c86a503d31744905d2d947ba56951cd2)

## TODO

- [x] Fix focus ring variables (The main `--focus-ring` value is a static `hsla` value but should use theme colors)
- [x] Update atomic classes
- [x] Replace all Less variable color references
- [x] Regenerate baseline test images (waiting until design signoff)
- [x] Remove `skippedTestids` from accessibility tests

## Breaking changes

- `--theme-primary-color` has been changed to `--theme-primary`
- `--theme-secondary-color` has been changed to `--theme-secondary` 

## Technical notes

- Only ~70 of the > 4000 changed files are code changes (the rest are baseline visual regression tests images)
- For replacing Less color variable references that don't have an equivalent CSS variable, I've opted to reference the equivalent value available in the color sets (e.g. `.set-black()[300]`). This is often necessary in situations where we operate on the color and need to pass what Less considers a color value. I think this is our best option but I figured it was worth mentioning.
    - We'll also need to consider how we want to address this same issue in Core. Ideally, we should be able to switch nearly all Less variable color references to use CSS variables, but I question whether we should all _direct_ references to colors within defined color sets (like `.set-black()[300]`)
- I've moved the atomic legacy border utility classes to a separate file `lib/atomic/v1/border.less` for easy removal when the time comes.
- I've moved the atomic legacy typography utility classes to a separate file `lib/atomic/v1/typography.less` for easy removal when the time comes.
- All references to `@white-legacy` and `@black-legacy` have been updated to use `--_white-static` and `--_black-static`. These CSS variables were added in #1440 (see [`lib/exports/color.less`](https://github.com/StackExchange/Stacks/pull/1440/files#diff-a3ba1b63e7acab9bc351c8b6a6298f4fecd1f04a5732e013fbdf992364764ddcR6-R7)). They map to white and black color values and do not change based on mode. These were used in places where we wanted to keep the color black/white no matter the mode without the need/ability to map colors per mode.
- This PR resolves an issue introduced in #1440 where `-ring` was missing from `--focus-ring[*]` variables.
- This PR resolves an issue introduced in #1440 where default `focus-ring` values were not being overridden when using a custom theme. That has been [resolved in `lib/exports.color-sets.less`](https://github.com/StackExchange/Stacks/pull/1449/files#diff-ea72389fc03dddb312af7fc98958fa06757fdf175252829b9468c4a187e6f9a4R390-R397).

## Design notes (light mode)

> **Note**
> Below, I've called out potential color issues noticed in light mode. More issues exist, but this seems like a good place to start considering that updating which colors are used below could have an impact on dark/high contrast modes.
> **Update: many dark and high contrast mode issues are resolved in a follow up branch #1451**)

- Many theme values which were differentiated now have equivalent values. This causes issues like hover states being identical to base and/or active states. [See `lib/exports/theme.less`](https://github.com/StackExchange/Stacks/pull/1449/files#r1298480296)
- `.has-warning` text, checkbox border too dark
    <details>
        image
        image
    </details>
- [**Label**](https://deploy-preview-1449--stacks.netlify.app/product/components/labels/#status): status badges have low contrast (required, new, beta)
    <details>
        <summary>Screenshot</summary>
        image
    </details>
- [**Activity indicator**](https://deploy-preview-1449--stacks.netlify.app/product/components/activity-indicator/#variations): warning style too dark?
    <details>
        <summary>Screenshot</summary>
        image
    </details>
- [**Badges**](https://deploy-preview-1449--stacks.netlify.app/product/components/badges/#badge-states): some state badges are low contrast
    <details>
        <summary>Screenshot</summary>
        image
    </details>
- [**Badges**](https://deploy-preview-1449--stacks.netlify.app/product/components/badges/#user-badges): staff badge is low contrast
    <details>
        <summary>Screenshot</summary>
        image
    </details>
- [**Banner/notice/toast**](https://deploy-preview-1449--stacks.netlify.app/product/components/banners/): success, warning "important" banners are low contrast
    <details>
        <summary>Screenshot</summary>
        image
    </details>
- [**Button**](https://deploy-preview-1449--stacks.netlify.app/product/components/buttons/#secondary): base filled selected button is low contrast
    <details>
        <summary>Screenshot</summary>
        image
    </details>
- [**Post summary**](https://deploy-preview-1449--stacks.netlify.app/product/components/post-summary/#hotness): hotness classes are dark and difficult to differentiate
    <details>
        <summary>Screenshot</summary>
        image
    </details>
- [**Prose**](https://deploy-preview-1449--stacks.netlify.app/product/components/prose/): `pre` block background color possibly too pale?
    <details>
        <summary>Screenshot</summary>
        image
    </details>
- [**Sidebar widget**](https://deploy-preview-1449--stacks.netlify.app/product/components/sidebar-widgets/#alternative-colors): Alt color widget headings may be too dark?
    <details>
        <summary>Screenshot</summary>
        image
    </details>


lib/exports/theme.less
```diff
@@ -1,80 +1,80 @@
...
-    --theme-body-font-color: var(--black-legacy-800);
-    --theme-background-color: var(--white-legacy);
+    --theme-body-font-color: var(--black-600);
+    --theme-background-color: var(--white);
     // Links
-    --theme-link-color: var(--theme-secondary-legacy-400);
-    --theme-link-color-hover: var(--theme-secondary-legacy-350);
-    --theme-link-color-visited: var(--theme-secondary-legacy-700);
+    --theme-link-color: var(--theme-secondary-400);
+    --theme-link-color-hover: var(--theme-secondary-400); // TODO was 350, now same as base link color
```
dancormier, line 6: > **Note**
> Many default theme component colors have collapsed into the same values. I'd like design to check this out and consider the best course of action.

lib/components/modal/modal.less
```diff
@@ -1,11 +1,12 @@
 .s-modal {
-    --_mo-bg: fade(@black-legacy-900, 50%); // Background remains fixed
+    // TODO Replace this Less variable with a CSS variable or other solution
+    --_mo-bg: fade(.set-black()[600], 50%); // Background remains fixed
```
dancormier, line 3: Across the codebase, I've converted references to Less color variables that don't have equivalent CSS variables to reference values available in the color sets. Here, a static color value is needed in order to operate on the color.

lib/atomic/misc.less
```diff
@@ -303,12 +303,13 @@
...
-.bs-i-sm.bs-hover:hover { box-shadow: inset 0 var(--su-static1) var(--su-static8) 0 fade(@black-legacy-200, 75%) !important; }
-.bs-i-md.bs-hover:hover { box-shadow: inset 0 var(--su-static1) var(--su-static8) 0 fade(@black-legacy-350, 80%) !important; }
-.bs-i-lg.bs-hover:hover { box-shadow: inset 0 var(--su-static1) var(--su-static12) 0 @black-legacy-350 !important; }
+// TODO do we still need them? Can we deprecate/remove these weird hover classes?
+.bs-sm.bs-hover:hover { box-shadow: 0 var(--su-static2) var(--su-static8) .set-black()[150] !important; }
+.bs-md.bs-hover:hover { box-shadow: 0 var(--su-static4) var(--su-static8) .set-black()[225] !important; }
+.bs-lg.bs-hover:hover { box-shadow: 0 var(--su-static4) var(--su-static12) fade(.set-black()[350], 70%) !important; }
+.bs-i-sm.bs-hover:hover { box-shadow: inset 0 var(--su-static1) var(--su-static8) 0 fade(.set-black()[300], 75%) !important; }
+.bs-i-md.bs-hover:hover { box-shadow: inset 0 var(--su-static1) var(--su-static8) 0 fade(.set-black()[350], 80%) !important; }
+.bs-i-lg.bs-hover:hover { box-shadow: inset 0 var(--su-static1) var(--su-static12) 0 .set-black()[350] !important; }
```
dancormier, line 312: A few notes here:

- I believe **none of this** is documented in Stacks. We should either a) document it or b) get rid of these classes.
- We have a few references to `.bs-hover` in Core but _I think_ they could be replaced with conditional classes (like `h:bs-sm`). I'm not sure why we have these hover classes.
- AFAICT, these `.bs-i-[*]` classes are unused in Core. I'd feel pretty comfortable getting rid of them.

lib/components/modal/modal.less
giamir: @dancormier In our previous PRs we talked a bit about recording tradeoffs in a document. I cannot remember if you ended up starting that document or not. This is the type of thing we should list in there. I can see how it might be useful to have a story at some point to create static color variables like `var(--static-black-600)`.

lib/atomic/misc.less
giamir: Maybe this is something to add to our famous document too. :) 
At the end of this initiative we could create few stories out of it and prioritize it at our own pace. 
We could even decide to work on some of them before cutting the 2.0 final release.

lib/components/modal/modal.less
giamir: It looks like replacing fade with native css won't be exactly a walk in the park:
https://stackoverflow.com/questions/40010597/how-do-i-apply-opacity-to-a-css-color-variable

giamir (approved):
It looks good to me @dancormier. 🎉  Great work! 
Feel free to merge this PR and move the story in ready for showcase once you have created a follow up tasks to resolve all the new APCA violations. We should initiate a conversation with design about those as soon as we have a good understanding of the contrast problems.

lib/components/modal/modal.less
dancormier: Yep, for now I think our only reasonable course of action here is to reference the color directly in this `fade` function. Seeing as this one modifies a color value's opacity, I don't see a better short-term solution.

Long-term, we could use something like the proposed [relative color syntax](https://www.w3.org/TR/css-color-5/#relative-colors) to adjust opacity, but that won't be for a while.

lib/atomic/misc.less
dancormier: Added to our "Color Redefinition: Breaking changes, trade offs, details" doc under "[DEPRECATION: Box-shadow atomic hover classes](https://docs.google.com/document/d/1dqIymDut_dRHblyvuMhgpxeFlA6-dh9VPAdb3kT7vSA/edit#heading=h.w6a14mf4uf1v)"

PR #1481, opened 2023-09-07
docs: add migration guide

dancormier:
Work in progress

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -119,8 +119,8 @@ Stacks v2 includes new CSS custom properties to replace a handful of commonly re
 
 Stacks v1 | Stacks v2
 :-:|:-:
-@black | var(--_black-static)
-@white | var(--_white-static)
+`@black` | `var(--_black-static)`
+`@white` | `var(--_white-static)`
 
 <!-- TODO any other values worth mentioning above? -->
 
@@ -130,7 +130,7 @@ In rare circumstances, you may need to reference a color value directly in order
 
 Stacks v1 | Stacks v2
 :-:|:-:
-@black-350 | .set-black()[350]
+`@black-350` | `.set-black()[350]`
 
 ## Using legacy colors
 
@@ -194,9 +194,11 @@ To update to the new custom theme variable generation function, simply search fo
 
 ## Other deprecations
 
-<!-- TODO add any other deprecation here -->
+*TODO add any other deprecation here*
 
 ## Other removals and breaking changes
+
+*TODO add any other removals and breaking changes here*
 
 <!-- TODO Not removed as of Thu Sep 7, 2023. For details, see https://docs.google.com/document/d/1dqIymDut_dRHblyvuMhgpxeFlA6-dh9VPAdb3kT7vSA/edit#heading=h.w6a14mf4uf1v
 
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -178,7 +178,7 @@ $1-$2-legacy
 
 ## Custom theme generation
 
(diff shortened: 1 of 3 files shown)


MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,240 @@
+# Migrating from Stacks v1 to v2
+
+With the release of Stacks v2, we've made some breaking changes to the library. This purpose of this guide is to help you prepare your code be compatible with the changes in Stacks v2.
```
giamir:
```suggestion
With the release of Stacks v2, we've made some breaking changes to the library. The purpose of this guide is to help you prepare your code be compatible with the changes in Stacks v2.
```

MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,240 @@
...
+    - [Add `-legacy` suffix to colors](#add--legacy-suffix-to-colors)
+    - [Add `-legacy` to aliased values](#add--legacy-to-aliased-values)
+- [Custom theme generation](#custom-theme-generation)
+  - [Replacing `.generate-calculated-themed-variables()`](#replacing-generate-calculated-themed-variables)
+- [Other deprecations](#other-deprecations)
+- [Other removals and breaking changes](#other-removals-and-breaking-changes)
+
+## Upgrading dependencies
+
+<!-- TODO outline the need to update Stacks-Editor, Stacks-Icons, etc -->
```
giamir: I was reading in the markdown preview and I could not see this comment. Maybe it would be less confusing to leave the TODO in cursive so that people understand that there is not a mistake in this section but it is in fact a work in progress. 🙂

MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,240 @@
...
+```
+
+##### Replace
+```regexp
+$1-$2-legacy
+```
+
+## Custom theme generation
+
+Stacks v2 has deprecated the Less function `.generate-calculated-themed-variables()` in favor of the newly added `.create-custom-theme-hsl-variables()` function. This new function works differently behind the scenes, but it accepts the same arguments and generates the same CSS custom properties as it's deprecated equivelent.
```
giamir:
```suggestion
Stacks v2 has deprecated the Less function `.generate-calculated-themed-variables()` in favor of the newly added `.create-custom-theme-hsl-variables()` function. This new function works differently behind the scenes, but it accepts the same arguments and generates the same CSS custom properties as its deprecated equivelent. 
```

MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,240 @@
...
+## Custom theme generation
+
+Stacks v2 has deprecated the Less function `.generate-calculated-themed-variables()` in favor of the newly added `.create-custom-theme-hsl-variables()` function. This new function works differently behind the scenes, but it accepts the same arguments and generates the same CSS custom properties as it's deprecated equivelent. 
+<!-- TODO See the [theming documentation]() for more information. -->
+
+### Replacing `.generate-calculated-themed-variables()`
+
+To update to the new custom theme variable generation function, simply search for `.generate-calculated-themed-variables` in your codebase and replace it with `.create-custom-theme-hsl-variables`.
+
+## Other deprecations
```
giamir, line 195: Also for these last 2 sections we could put a WIP sign for the content since they look odd in the markdown preview (almost as if we forgot something).

giamir (approved):
@dancormier The draft looks good to me. I think there is enough information for (potentially) product engineers starting the integration on Monday. We can add more details to this document as the work progress. By the time we will merge `stacks-next` in `develop` it will be ready. 

The only part that could probably be completed already is the first TODO about Stacks Icons and Stacks Editor RCs going hand in hand with Stacks Classic v2. A short paragraph would be more than enough in my opinion.

Thanks for moving this forward. 🎉

PR #1482, opened 2023-09-08
test(code-block): add a11y; visual tests

dancormier:
This PR adds visual and accessibility tests for the `.s-code-block` component.

---

_Usually_, I'll request a review from @StackExchange/stacks, but just merge it before getting a review, but I would like you to check this out before I merge, @giamir. I suspect there's a way to get pass a chunk of code and have highlight-js run before the tests execute, but I sidestepped figuring that out just to get some tests in place. If you don't know of any simple way to achieve this, I don't think it's worth spending too much time on, but I figured this is a PR that would benefit from a quick review.


giamir:
@dancormier Hey Dan, I actually have limited context about how we use highlightjs in Core. Furthermore I think [our code-block documentation](https://stackoverflow.design/product/components/code-blocks/) need some upgrades because it doesn't have crucial information about how to setup consumer codebases to use highlightjs in combination with our styles (maybe we could create an operational maintainance ticket for that).

Regarding the test setup I took the liberty of adding a commit with some suggestions. I decided to push it directly in this repository but please feel free to revert it (I was lazy to open a new branch).
https://github.com/StackExchange/Stacks/pull/1482/commits/43f78af2aa3f157ab47329e6376aae30615c2bbf

I am not sure if that was what you were looking for: the benefits of my change is that the fixtures are a bit more readable and we execute highlight.js instead of hardcoding its output in a string.
I noticed the way we run highlight.js in our docs is via [this plugin](https://github.com/b-kelly/eleventy-plugin-highlightjs) but it is not an option to use that in our test setup.

dancormier:
> @dancormier Hey Dan, I actually have limited context about how we use highlightjs in Core. Furthermore I think [our code-block documentation](https://stackoverflow.design/product/components/code-blocks/) need some upgrades because it doesn't have crucial information about how to setup consumer codebases to use highlightjs in combination with our styles (maybe we could create an operational maintainance ticket for that).

Good catch. I'll add a ticket

> Regarding the test setup I took the liberty of adding a commit with some suggestions. I decided to push it directly in this repository but please feel free to revert it (I was lazy to open a new branch). [43f78af](https://github.com/StackExchange/Stacks/commit/43f78af2aa3f157ab47329e6376aae30615c2bbf)
> 
> I am not sure if that was what you were looking for: the benefits of my change is that the fixtures are a bit more readable and we execute highlight.js instead of hardcoding its output in a string. I noticed the way we run highlight.js in our docs is via [this plugin](https://github.com/b-kelly/eleventy-plugin-highlightjs) but it is not an option to use that in our test setup.

It's just what I was hoping for! I might have a couple of very minor tweaks, but your changes really help. Thank you!

PR #1487, opened 2023-09-11
test(a11y): add ability to pass custom APCA threshold to axe-apca

dancormier:
Addresses [STACKS-388](https://stackoverflow.atlassian.net/browse/STACKS-388).

This PR modifies `registerAxeAPCA` to:
1. Accept `custom` as a conformance level
2. Accept an optional custom threshold function for the second argument

It also updates `lib/test/test-utils.ts` to use a custom APCA threshold of Lc 60 for all fontSize < 32.

### DIFF
--- a/lib/test/axe-apca/README.md
+++ b/lib/test/axe-apca/README.md
@@ -42,9 +42,9 @@ To set custom thresholds for APCA checks, follow these steps:
 
 
 ```js
-const customAPCAThreshold = (fontSize) => {
+const customConformanceThresholdFn = (fontSize) => {
     return parseFloat(fontSize) >= 32 ? 45 : 60;
 };
 
-registerAxeAPCA('custom', customAPCAThreshold);
+registerAxeAPCA('custom', customConformanceThresholdFn);
 ```
--- a/lib/test/axe-apca/README.md
+++ b/lib/test/axe-apca/README.md
@@ -42,8 +42,11 @@ To set custom thresholds for APCA checks, follow these steps:
 
 
 ```js
-const customConformanceThresholdFn = (fontSize) => {
-    return parseFloat(fontSize) >= 32 ? 45 : 60;
+const customConformanceThresholdFn = (fontSize, fontWeight) => {
+    const size = parseFloat(fontSize);
+    const weight = parseFloat(fontWeight);
+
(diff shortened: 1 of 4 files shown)


lib/test/test-utils.ts
```diff
@@ -328,7 +332,7 @@ const runComponentTest = ({
                     // and use APCA bronze level instead
```
giamir: This comment needs updating (e.g. we aim for bronze but at the moment we had to settle for less strict thresholds)

lib/test/axe-apca/README.md
```diff
@@ -32,3 +32,19 @@ axe.configure({
...
+### Using custom APCA thresholds
+
+To set custom thresholds for APCA checks, follow these steps:
+
+1. Use `custom` as the first argument when calling `registerAxeAPCA`.
+1. Provide a function as the second argument, optionally accepting `fontSize` and `fontWeight` arguments.
+
+
+```js
+const customConformanceThresholdFn = (fontSize) => {
```
giamir: it might be worth for documentation purposes to also pass the `fontWeight` as second argument even if we don't use it.

giamir (approved):
Looks great @dancormier. Thanks for generalising this. 🎉
Feel free to merge. 

The only comment I have is that by relaxing the threshold we might have actually quite a few skipped tests now that they don't need to be skipped anymore.
I am not sure how relevant it is at this stage is to unskip them given that we will eventually merge the stacks-next branch. The only benefits I see is if we want to have a comparison on number of custom apca failures for v1 vs v2. I will leave that decision up to you.

dancormier:
> The only comment I have is that by relaxing the threshold we might have actually quite a few skipped tests now that they don't need to be skipped anymore.
> I am not sure how relevant it is at this stage is to unskip them given that we will eventually merge the stacks-next branch. The only benefits I see is if we want to have a comparison on number of custom apca failures for v1 vs v2. I will leave that decision up to you.

Before merging in stacks-next (or retroactively), I plan on getting a final v1 vs v2 test failure comparison. At this stage, that number is heavily in flux as we add new tests in tandem with our color initiative work, so I don't feel capturing a number now would be a good point of comparison. Similarly, we could comb through and see which tests now pass, but I don't quite think the juice is worth the squeeze on that one.

PR #1496, opened 2023-09-21
don't apply color override to links in the `s-user-card--type` area

johnmwright:
Allow the links to set set their own style/inherit from parent styles

The `.s-user-card--type` style is currently used exclusively for Collectives membership badges and there are cases where we want the link color to inherit from the parent elements instead of always being gray.

### DIFF
--- a/lib/components/sidebar-widget/sidebar-widget.less
+++ b/lib/components/sidebar-widget/sidebar-widget.less
@@ -11,7 +11,7 @@
 
     // MODIFIERS
     &:not(.s-anchors) {
-        a:not(.button):not(.post-tag):not(.s-btn):not(.s-sidebarwidget--action):not(.s-user-card--link) {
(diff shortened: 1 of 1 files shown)


dancormier:
@johnmwright this PR seems to affect only the user card _within a sidebar widget_. Is this your intention or do you want to have this apply to _all_ `s-user-card`s?

johnmwright:
The specific case I have where it's doing the wrong thing is in a sidebar due to the style that I'm changing overriding the font color.  I basically just don't want that style to be applied.

dancormier:
@johnmwright I've resolved the issue in https://github.com/StackExchange/Stacks/commit/5cf01e52ec9f5063e74c0c22871455105f49cb44

In this PR, the change targets `.s-sidebarwidget a:not(.s-user-card--type > .s-link)`, which results in no material change because `:not` only accepts [a single selector](https://www.w3.org/TR/selectors-3/#negation), since it's relative to a give target prefixing the `:not` (`.s-sidebarwidget a` in this case). 

In the commit I made, we'll instead ensure that `a` elements within `.s-user-card--type` inherits color unless they include `.s-link`. Expect this fix in the next Stacks release.

PR #1499, opened 2023-09-22
test(post-summary): add a11y, visual test

dancormier:
This PR adds visual and accessibility tests for the `.s-post-summary` component.


lib/components/post-summary/post-summary.a11y.test.ts
```diff
@@ -0,0 +1,183 @@
...
+                        text: "featured-request",
+                        type: "required",
+                    },
+                ],
+                title: "Short title",
+            }),
+        },
+        template,
+        // TODO resolve test failures
+        skippedTestids: [
```
giamir, line 10: Do you think these failures deserve to be logged in our a11y board?

lib/components/post-summary/post-summary.a11y.test.ts
```diff
@@ -0,0 +1,183 @@
...
+import { runA11yTests } from "../../test/a11y-test-utils";
+import { getChildren } from "../../test/post-summary-test-utils";
+import type {
+    BadgeType,
+    TruncationSizes,
+} from "../../test/post-summary-test-utils";
+import "../../index";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const template = ({ component, testid }: any) => html`
```
giamir: My overall impression is that there is a lot of repetition between the visual and a11y test files and more things can be moved in a setup file like the test utils you have created.

lib/test/post-summary-test-utils.ts
giamir, line 1: I like that you have extracted shared setup functions in this file. I even think you could extract more.
Regarding the placement of this file though I would prefer to keep it close to where it is used. I think I have left a similar comment in another PR recently.
The idea would be to standardize around something like this `post-summary.test.setup.ts` and then make sure the `*.setup.ts` files are removed from  the tsconfig.build configuration. Thoughts?

screenshots/Chromium/baseline/s-post-summary-dark-badge-danger-filled.png
giamir, line 1: Is this the wanted effect? should the elements be aligned horizontally?

screenshots/Chromium/baseline/s-post-summary-dark-stats-answered-accepted.png
giamir, line 1: Similarly here. Is it ok to keep the elements horizontal?

giamir (approved):
Thanks for adding tests for this component @dancormier. 🙏
Overall it looks good and could be merged in my opinion. It's a complex component with a complex setup and there are some minor refactoring that could always be done but the screenshots looks good. It's not so easy for me to understand the many variations people could potentially use this component with even by looking at the docs page.  😅

The main comment I had was to keep those complex test setups close to the tests file instead of the shared `/test` folder (unless we plan to share them with other components).

**Sidenote**
I believe that if we would use Stacks Svelte Components our test setup would look less convoluted because the markup rendering logic will be abstracted away by a simplified API. As we continue building the Svelte library, in future we could consider a simplification of these complex test setups by using components. 🙂

lib/components/post-summary/post-summary.a11y.test.ts
dancormier: 💯 This is something I've been thinking about for our test broadly. Most of the a11y and visual tests for a given component overlap > 80%. I'll probably want to discuss a uniform way to do this across our components.

For now, I'll port much of the post-summary tests into the test utils and I'll create a ticket to investigate creating a pattern for merging overlapping test code.

@giamir If you think the pattern of a test-utils file that is imported into the a11y/visual test files is sufficient and doesn't require more consideration, please say so and I'll close the ticket.

lib/components/post-summary/post-summary.a11y.test.ts
dancormier: Yep, I'll add them

lib/components/post-summary/post-summary.a11y.test.ts
giamir: I think we can leave the ticket. In the meantime the test-utils is probably good enough (and maybe we won't even be able to come up with something simpler than that).

lib/components/post-summary/post-summary.a11y.test.ts
dancormier: - Deleted [A11Y-27](https://stackoverflow.atlassian.net/browse/A11Y-27), [A11Y-28](https://stackoverflow.atlassian.net/browse/A11Y-28)
- Ignored [A11Y-34](https://stackoverflow.atlassian.net/browse/A11Y-34), [A11Y-35](https://stackoverflow.atlassian.net/browse/A11Y-35)
- Watched [A11Y-36](https://stackoverflow.atlassian.net/browse/A11Y-36)

lib/test/post-summary-test-utils.ts
dancormier: Addressed in https://github.com/StackExchange/Stacks/pull/1499/commits/d202db94155252c39edaf3ff363f9d7fe1218747

screenshots/Chromium/baseline/s-post-summary-dark-badge-danger-filled.png
dancormier: Yes, these elements are responsive and layout horizontally below the `md` breakpoint. Eventually, it would be nice to test at different breakpoints but for now I figure this is fine

### >= md

image

### < md

image

PR #1514, opened 2023-10-10
chore(changelog): add changelog with details about v2 release

dancormier:
This PR adds a `CHANGELOG.md` which includes changes in v2.

> **Note**
> I tried to _roughly_ stick to the format produced by [Release Please](https://github.com/googleapis/release-please). I figure this is the format we'll likely use in the future. Let me know if I need to alter the changelog formatting at all please.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,34 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+## [2.0.0](https://github.com/StackExchange/Stacks/compare/v1.10.6...v2.0.0) (2023-10-06)
+
+### ⚠ BREAKING CHANGES
+
+> **Note**
+> This is a breaking change and will require your references to color atomic classes be remapped.
+Please see the **[migration guide](https://github.com/StackExchange/Stacks/blob/develop/MIGRATION_GUIDE.md)** for detailed information on these changes and guidance on migrating.
+
+#### Improvements and additions
+
+- ◑ Color contrast improvements to meet or exceed [APCA](https://git.apcacontrast.com/documentation/README) Lc 60 in standard light and dark modes and WCAG 2.2 AAA in high contrast modes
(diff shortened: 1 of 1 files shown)


giamir:
@dancormier Thanks for adding this. 

I was reflecting a bit about adding this file manually like that and I started to think we should not do it until we have a tool such as release-please automatizing this process.
I am saying this because if we start to now maintain a changelog file we will need to remember to do it each time we release. We are already writing sort of a changelog manually in the Github Release and that is probably sufficient for our status quo. 
Apologies for adding as an acceptance criteria to create a changelog.md (I was probably thinking about adding the changelog somewhere - the GH release being sufficient)

**So I am suggesting to close this PR if you agree and wait until we integrate a tool such as release-please.**

## Sidenote
I was reading the documentation of [release-please](https://github.com/googleapis/release-please) as part of this review. I think that would be the correct tool for us if we want to stick 100% to conventional commits for our changelog generations. I think this works particularly well for small packages, _for a larger repo like ours it might be a bit trickier to fit all necessary changelog info in git commit messages_. What are your thoughts on this?

I think we should consider also [changesets](https://github.com/changesets/changesets). Changesets works similarly to release-please but does not rely on conventional commits. Instead authors need to proactively describe their changesets in yml files that then get picked up by the changesets GH action to open release PRs similarly to what release-please does.

From their docs (they are not totally wrong)
```
Git is a bad place to store this information, as it discourages writing detailed change descriptions - 
you want to allow people to provide as much documentation for the change as they want.
```

I am not advocating to use changesets over release-please. I am merely suggesting to look into it before making a decision. A downside of using changesets over release-please could be the introduction of some extra complexity (yml files, changesets cli, ...).

I believe the story for extracting axe-apca in its own repo could get an additional acceptance criteria to test changesets for releases. This could inform a potential Team ADR to pick one tool over the other one for automatic releases.
[I went ahead and added the additional acceptance criteria.](https://stackoverflow.atlassian.net/browse/STACKS-387)
Let me know if you have any concerns.

PR #1515, opened 2023-10-10
chore(colors): update high contrast colors

dancormier:
@CGuindon here's the high contrast mode colors updated to match the [Figma](https://www.figma.com/file/cCU6A56sHwZluraDoyg5o7/Colors%3A-Accessible-Color-Palette?type=design&node-id=2170%3A15294&mode=design&t=8rJUwUAaQ1NoQUMq-1). Let me know if the values seem correct and, if so, I'll merge and cut a new release.


CGuindon (approved):

giamir:
@dancormier if we want to follow [conventional commit standards](https://www.conventionalcommits.org/en/v1.0.0/), this commit type should have been marked as `fix`. It doesn't matter now because we are bumping the npm version manually but it is something to keep in mind if we want to move to a tool like release-please in the future. 🙂

Sidenote: I ended up here while reviewing [this changelog PR](https://github.com/StackExchange/Stacks/pull/1514).

PR #1519, opened 2023-10-13
test(axe-apca): extract package in its own repo

giamir:
## Summary

[STACKS-387](https://stackoverflow.atlassian.net/browse/STACKS-387)

This PR extracts axe-apca rules in its own independent [npm package](https://www.npmjs.com/package/axe-apca) and [GH repo](https://github.com/StackExchange/axe-apca).

## How to test

- A11y tests should work exactly as before (CI workflow should pass)


dancormier (approved):
Ran a11y checks and they all passed successfully. Reran with a change that should result in a failure and that went as expected too.

Thanks for handling this @giamir

PR #1543, opened 2023-10-24
test(visual): run 3 browsers concurrently

dancormier:
As we add more visual regression test, the time it takes to run our test suite increases. This PR alters our test config to speed up our test runs by setting `concurrentBrowsers` to `3` (up from the default of 2).

This change means that all three browsers we test against are run concurrently, resulting in a significant decrease in test run duration.

---

I tried to get some data and found the following:

- Increasing the concurrent browsers from the default (2) to 3 reduces test run duration significantly (~13% when testing all components)
- Altering the number of test files to run concurrently has a negligible impact
    - The higher the number, the fast the run _generally_ but I expect any increase would only reduce test duration by less than 5%

See also [Test Runner: CLI and Configuration](https://modern-web.dev/docs/test-runner/cli-and-configuration/)

### Data

As I made changes to the test config, I cataloged the impact those changes had. All data is from `npm run test:visual` and the "Test file(s)" column shows which tests were run to get the data.

Test file(s) | Time (s) | concurrentBrowsers | concurrency | % change
-- | -- | -- | -- | --
tag | 53 | default (2) | default (2) | N/A
tag | 69 | 1 | default (2) | 130%
tag | 40 | 3 | default (2) | 75%
tag | 40 | 3 | 1 | 75%
tag | 38 | 3 | 20 | 72%
tag | 37 | 3 | 100 | 70%
tag | 69 | 1 | 200 | 130%
tag | 38 | 3 | 200 | 72%
tag | 40 | 10 | 200 | 75%
button | 180 | default (2) | default (2) | N/A
button | 133 | 3 | default (2) | 74%
button | 135 | 3 | 50 | 75%
button | 133 | 3 | 200 | 74%
button | 132 | 3 | 500 | 73%
all | 484 | default (2) | default (2) | N/A
all | 423 | 3 | default (2) | 87%


giamir:
I had to upgrade playwright and web test runner dependencies before being able to test this PR.

See https://github.com/StackExchange/Stacks/pull/1544

After that I could see similar benefits in my machine by changing the concurrentBrowsers option to 3. 

I would recommend to first merge [this PR](https://github.com/StackExchange/Stacks/pull/1544) and then rebase this one on it. I did not merge that PR directly on develop because it came with several baseline image changes and I thought it would be better for you to double check too - don't wait for me to merge it though. 🙂

Great work on improving the setup. ❤️

giamir (approved):

PR #1546, opened 2023-10-26
fix(link): update link, anchor grayscale hover/visited

dancormier:
cc @CGuindon 

This PR does two things:

- Fix: unify the anchor and link `__grayscale` styles
- Update `__grayscale` styles to be base `500`, hover `600`, visited `400`


dancormier:
@CGuindon 👋

PR #1547, opened 2023-10-27
fix(a11y): make focus ring meet 3:1 requirements

dancormier:
> **Note**
> Expect this PR to fail until I update the test snapshots (which I'll do when it's in a stable state)

This PR modifies the focus ring we use on focused elements to meet or exceed [WCAG 2.2 Success Criterion 2.4.11: Focus Appearance (Minimum)](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum)

# Outline of changes

## Update all focus state CSS vars to meet or exceed accessibility standard
[color-sets.less diff](https://github.com/StackExchange/Stacks/pull/1547/files#diff-ea72389fc03dddb312af7fc98958fa06757fdf175252829b9468c4a187e6f9a4)

A common approach to satisfying [WCAG 2.2 Success Criterion 2.4.11: Focus Appearance (Minimum)](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum) is to add a focus ring that includes a high contrast outline between the element and the focus ring. This is common among [Microsoft](https://react.fluentui.dev/?path=/docs/components-button-button--default), [Ant Design](https://ant.design/components/button), [Carbon](https://carbondesignsystem.com/components/button/usage/), and the  [U.S. Web Design System](https://designsystem.digital.gov/components/button/).

<details>
<summary>Screenshot of the U.S. Web Design System</summary>

image

</details>

## Activity indicator, stepped progress bar

These components relied on `--focus-ring`. Since the style of `--focus-ring` is now much darker, it's being applied as a pseudo-element on `activity-indicator` and the `stepped` version of `progress` with a lower opacity to achieve the same effect as before.

## Provide themed fallback focus styles

Previously, focus states relied on `hsla(206,100%,40%,0.15)` or `--theme-secondary-custom-focus-ring`. If the custom property was not provided, it would result in the default value being used despite a primary theme value being available. 

<details>
<summary>Example of themed button with default focus ring style</summary>

image

</details>

This PR allows focus ring to fall back from `theme-secondary-custom-500` to `theme-primary-custom-500` to theme-secondary-500`.

## Added focus styles to `pagination` and `tag` components

These elements were missing our custom focus states. This PR adds them.

## Added focus styles to `topbar` children

TBD

## `@focus-styles()` and `@focus-visible-styles()`
[constants-helpers.less diff](https://github.com/StackExchange/Stacks/pull/1547/files#diff-dd4021b008e0ec1764a2aa34ff408c687d5c12b6a5742bb120b4ca91ff161259)

Since the existing `@focus-styles()` function operated on `:focus-visible`, I renamed it to reflect that and free the name up. With `@focus-styles()` available, we can provide an abstraction for common focus styles that don't apply to `:focus-visible`. This means that many components that previously included custom `:focus` CSS can rely on `@focus-styles()`.

## Add `--_focus-ring-color`

Inputs (all of them including textareas, select element, etc) have validation states that restyle both the input and the focus ring. This PPCP is to be used to update the focus ring color in a given context.

## Add `.bs-ring-inset` with conditional prefixes

Since some elements live within a container that hides overflow, a normal box shadow cannot always be used to show a focus state. I've added this class to allow for an inset focus ring to be added to elements under those conditions.

<details>
<summary>Example of UI that would require an inset focus ring</summary>

image

</details>

## Minor improvements and updates to doc

- Added `.bs-ring-inset` to box-shadow documentation
- Updated box-shadow documentation to render examples programmatically
- Changed tag documentation to use `span` when example includes a nested button
    - This was changed from the `.s-tag` being an anchor with a nested `span` for `.s-tag--dismiss`. Since `.s-tag--dismiss` is expected to always be interactive, it should be rendered as a `button` element, which cannot be placed within an anchor element.  From [W3C "`a` Additional constraints and admonitions"](https://w3c.github.io/html-reference/a.html#a-constraints):
        > The interactive element `a` must not appear as a descendant of the `button` element. 

# Success Criterion [2.4.11 Focus Appearance (Minimum)](https://www.w3.org/TR/WCAG22/#focus-appearance-minimum) (Level AA)
> When a [user interface component](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum#dfn-user-interface-component) has keyboard focus, the [focus indicator](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum#dfn-focus-indicator):
>
>   - [Encloses](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum#dfn-encloses) the visual presentation of the user interface component;
>   - Has a contrast ratio of at least 3:1 between its pixels in the focused and unfocused states;
>   - Has a contrast ratio of at least 3:1 against adjacent colors.
>
>    Exceptions:
>
>    - The focus indicator is determined by the user-agent and cannot be adjusted by the author, or
>    - The focus indicator and the indicator's background color are not modified by the author, or
>    - An area of the focus indicator meets all the following:
>        - At least as large as the area of a 1 [CSS pixel](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum#dfn-css-pixel) thick [perimeter](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum#dfn-perimeter) of the unfocused component, or is at least as large as a 4 CSS pixel thick line along the shortest side of the [minimum bounding box](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum#dfn-minimum-bounding-box) of the unfocused component, and
>        - has a contrast ratio of at least 3:1 against the same pixels in the unfocused state, and
>        - has a contrast ratio of at least 3:1 against adjacent non-focus-indicator colors, or is no thinner than 2 CSS pixels.
>
>    Where a user interface component has active sub-components, if a sub-component receives a focus indicator, these requirements can be applied to the sub-component instead.
>
>    Note: Examples of sub-components that may receive a focus indicator are menu items in an opened drop-down menu. However, it may also be possible to indicate user interaction for such sub-components by relying strictly on a visual indication of which item is selected. Where selectable sub-components have no differentiated focus indicator, the visual indicator for sub-component selection is measured against [1.4.11 Non-text Contrast](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html) requirements, not against this Criterion.
>
>    Note: Contrast calculations can be based on colors defined within the technology (such as HTML, CSS and SVG). Pixels modified by user agent resolution enhancements and anti-aliasing can be ignored.

### Questions

- ~~Should we retain the `muted` version of the focus ring or replace it with the default?~~ retain it as `black-500`
- ~~Which components need focus rings added?~~ `tag`, `pagination`, elements within `topbar`
- How should we go about testing focus states?

cc @CGuindon @andrewmeacham

### DIFF
--- a/docs/_data/atomics.json
+++ b/docs/_data/atomics.json
@@ -26,7 +26,7 @@
     },
     {
       "class": "bs-ring",
-      "output": "box-shadow: 0 0 0 var(--su-static4) var(--focus-ring);",
+      "output": "box-shadow: var(--focus-ring-divider), 0 0 0 var(--_focus-ring-size, var(--su-static4)) var(--_focus-ring-color, var(--focus-ring));",
       "hover": true,
       "focus": true
     }
--- a/docs/_data/atomics.json
+++ b/docs/_data/atomics.json
@@ -29,6 +29,12 @@
       "output": "box-shadow: var(--focus-ring-divider), 0 0 0 var(--_focus-ring-size, var(--su-static4)) var(--_focus-ring-color, var(--focus-ring));",
       "hover": true,
       "focus": true
+    },
+    {
+      "class": "bs-ring-inset",
+      "output": "box-shadow: 0 0 0 var(--_focus-ring-size, var(--su-static4)) var(--_focus-ring-color, var(--focus-ring)), var(--focus-ring-divider-inset);",
+      "hover": true,
+      "focus": true
     }
   ],
   "box-sizing": [
--- a/docs/product/base/box-shadow.html
+++ b/docs/product/base/box-shadow.html
@@ -51,24 +51,15 @@ description: Box shadow atomic classes allow you to change an element’s box sh
 <div class="bs-lg">…</div>
 <div class="bs-xl">…</div>
 <div class="bs-ring">…</div>
+<div class="bs-ring-inset">…</div>
 {% endhighlight %}
         <div class="stacks-preview--example fs-caption ff-mono">
             <div class="d-grid grid__4 lg:grid__3 sm:grid__2 g16">
-                <div class="flex--item fd-column p12 bs-sm bar-sm h96">
-                    .bs-sm
-                </div>
-                <div class="flex--item fd-column p12 bs-md bar-sm h96">
-                    .bs-md
-                </div>
-                <div class="flex--item fd-column p12 bs-lg bar-sm h96">
-                    .bs-lg
-                </div>
-                <div class="flex--item fd-column p12 bs-xl bar-sm h96">
-                    .bs-xl
-                </div>
-                <div class="flex--item fd-column p12 bs-ring bar-sm h96">
-                    .bs-ring
-                </div>
+                {% for atomic in atomics.box-shadow %}
+                    <div class="fd-column p12 bar-sm h96 {{ atomic.class }}">
+                        .{{ atomic.class }}
+                    </div>
+                {% endfor %}
             </div>
         </div>
     </div>
--- a/docs/product/components/tags.html
+++ b/docs/product/components/tags.html
@@ -36,7 +36,7 @@ description: Tags are an interactive, community-generated keyword that allow com
 {% highlight html %}
 <div class="d-flex g4">
     <a class="s-tag" href="#">jquery</a>
-    <a class="s-tag" href="#">javascript <span class="s-tag--dismiss">@Svg.ClearSm</span></a>
+    <span class="s-tag" href="#">javascript <button class="s-tag--dismiss">@Svg.ClearSm</button></span>
     <a class="s-tag" href="#"><img class="s-tag--sponsor" src="https://i.stack.imgur.com/tKsDb.png" width="16" height="18" alt="Google Android"> android</a>
     <a class="s-tag is-selected" href="#">razor <span class="s-tag--dismiss">@Svg.ClearSm</span></a>
 </div>
@@ -44,9 +44,9 @@ description: Tags are an interactive, community-generated keyword that allow com
         <div class="stacks-preview--example">
             <div class="d-flex g4 fw-wrap">
                 <a class="s-tag" href="#">jquery</a>
-                <a class="s-tag" href="#">javascript <span class="s-tag--dismiss">{% icon "ClearSm" %}</span></a>
+                <span class="s-tag" href="#">javascript <button class="s-tag--dismiss">{% icon "ClearSm" %}</button></span>
                 <a class="s-tag" href="#"><img class="s-tag--sponsor" src="https://i.stack.imgur.com/tKsDb.png" width="16" height="18" alt="Google Android"> android</a>
-                <a class="s-tag is-selected" href="#">razor <span class="s-tag--dismiss">{% icon "ClearSm" %} </span></a>
+                <span class="s-tag is-selected" href="#">razor <button class="s-tag--dismiss">{% icon "ClearSm" %}</button></span>
             </div>
         </div>
     </div>
@@ -56,7 +56,7 @@ description: Tags are an interactive, community-generated keyword that allow com
 {% highlight html %}
 <div class="d-flex g4">
     <a class="s-tag s-tag__moderator" href="#">status-completed</a>
-    <a class="s-tag s-tag__moderator" href="#">status-bydesign <span class="s-tag--dismiss">@Svg.ClearSm</span></a>
+    <span class="s-tag s-tag__moderator" href="#">status-bydesign <button class="s-tag--dismiss">@Svg.ClearSm</button></span>
     <a class="s-tag s-tag__moderator" href="#">status-planned</a>
     <a class="s-tag s-tag__moderator is-selected" href="#">status-deferred</a>
 </div>
@@ -64,7 +64,7 @@ description: Tags are an interactive, community-generated keyword that allow com
         <div class="stacks-preview--example">
             <div class="d-flex g4 fw-wrap">
                 <a class="s-tag s-tag__moderator" href="#">status-completed</a>
-                <a class="s-tag s-tag__moderator" href="#">status-bydesign <span class="s-tag--dismiss">{% icon "ClearSm" %}</span></a>
+                <span class="s-tag s-tag__moderator" href="#">status-bydesign <button class="s-tag--dismiss">{% icon "ClearSm" %}</button></span>
                 <a class="s-tag s-tag__moderator" href="#">status-planned</a>
                 <a class="s-tag s-tag__moderator is-selected" href="#">status-deferred</a>
             </div>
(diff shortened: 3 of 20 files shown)


CGuindon:
@dancormier 

> Should we retain the muted version of the focus ring or replace it with the default?

I think you left Muted (keeping the grey) but bumped it up to use stop 500 as well... so would that always show up as grey on themes or is it pulling from a theme color?

CGuindon (changes requested):
Just started looking quickly and noticed we might have to add the extra outline to a few other elements like the nav items:

Screenshot 2023-10-27 at 12 46 55 PM

CGuindon:
> Which components need focus rings added?

Just finished getting through everything mentioned specifically in the bug sheet + a few others I noticed. Everything we need to add the focus-ring class to should be in the ticket now. I can start helping with this on Monday. I'm going to wrap up a few things on the Cprod side for the rest of the afternoon and then come back to this fresh next week.

dancormier:
> @dancormier
> 
> > Should we retain the muted version of the focus ring or replace it with the default?
> 
> I think you left Muted (keeping the grey) but bumped it up to use stop 500 as well... so would that always show up as grey on themes or is it pulling from a theme color?

Originally, I kept it around with `black-500` as the color. I removed it in https://github.com/StackExchange/Stacks/pull/1547/commits/4c6048a93f955b6900769664ca1688e90bc5f0ee base on this from the ticket:

> The focus muted state should no longer be used — move to have everything use focus default

@CGuindon Still good on removing it or would you like to keep it as `black-500` (or another stop)?

dancormier:
> Just started looking quickly and noticed we might have to add the extra outline to a few other elements like the nav items:
> Screenshot 2023-10-27 at 12 46 55 PM

The topbar had some focus overrides that affected `s-navigation`. I removed those overrides in [this commit](https://github.com/StackExchange/Stacks/pull/1547/commits/402bf1c21b220e3e6e514b6a0f5519aa727be79e), so it should be all good

CGuindon:
@dancormier Let's keep muted style for now — I think we're ok leaving it since it also got bumped to use black-500. I'll take a look through the PR.

CGuindon (changes requested):
(testing on deploy preview)

Not sure if this is just in Stacks or will apply to the sites as well but I would make everything in the top bar use the default focus ring (vs some using default and some using muted). I _think_ everything should be blue up top if it gets switched to default. I don't mind the orange outline as much but I don't think we need to be adding another additional color. (default:theme blue, muted: grey and then the warning ones).

What determines it using the orange instead of the blue? tied to the background color somehow?

Screenshot 2023-10-30 at 12 58 58 PM
 
Screenshot 2023-10-30 at 12 59 03 PM

Screenshot 2023-10-30 at 12 58 42 PM

CGuindon:
The `? icon` in the text editor seems to have some transparency compared to the other icons in there so it's appearing lighter than blue-500. 
Screenshot 2023-10-30 at 1 25 12 PM

CGuindon:
In stacks today (focus on tag):
Screenshot 2023-10-30 at 1 29 20 PM

In the deploy-preview (focus on tag):
Screenshot 2023-10-30 at 1 28 25 PM

There's a difference happening with the hover and focus state of tags with the X icon. It seems that I can't hover/select/focus the tag itself when the icon is present, I can only focus on the X icon. So this would mean you couldn't link to the tag page when this is used, you'd only be able to remove the tag — which is a change from what we have today I think.

dancormier:
> The ? icon in the text editor seems to have some transparency compared to the other icons in there so it's appearing lighter than blue-500.

@CGuindon In the Stacks docs, that button is _styled as_ disabled. It really shouldn't be focusable at all. I have a [PR on the Stacks-Editor](https://github.com/StackExchange/Stacks-Editor/pull/260) to address this issue.

---

> In stacks today (focus on tag): Screenshot 2023-10-30 at 1 29 20 PM
> 
> In the deploy-preview (focus on tag): Screenshot 2023-10-30 at 1 28 25 PM
> 
> There's a difference happening with the hover and focus state of tags with the X icon. It seems that I can't hover/select/focus the tag itself when the icon is present, I can only focus on the X icon. So this would mean you couldn't link to the tag page when this is used, you'd only be able to remove the tag — which is a change from what we have today I think.

What your seeing here is an update to the docs to replace invalid HTML. I added some clarification to the PR description:

> Changed tag documentation to use `span` when example includes a nested button
>    - This was changed from the `.s-tag` being an anchor with a nested `span` for `.s-tag--dismiss`. Since `.s-tag--dismiss` is expected to always be interactive, it should be rendered as a `button` element, which cannot be placed within an anchor element.  From [W3C "`a` Additional constraints and admonitions"](https://w3c.github.io/html-reference/a.html#a-constraints):
>        > The interactive element `a` must not appear as a descendant of the `button` element. 

It has no material effect on consumers of tag. It's really just a docs change (that should be implemented in Core, but that's another story)

dancormier:
> (testing on deploy preview)
> 
> Not sure if this is just in Stacks or will apply to the sites as well but I would make everything in the top bar use the default focus ring (vs some using default and some using muted). I _think_ everything should be blue up top if it gets switched to default. I don't mind the orange outline as much but I don't think we need to be adding another additional color. (default:theme blue, muted: grey and then the warning ones).
> 
> What determines it using the orange instead of the blue? tied to the background color somehow?
> Screenshot 2023-10-30 at 12 58 58 PM Screenshot 2023-10-30 at 12 59 03 PM Screenshot 2023-10-30 at 12 58 42 PM

## Muted focus ring in topbar

There's two elements that get that the muted focus ring in the topbar are "navigation" items and the muted buttons.

### Navigation

I haven't modified how the [navigation component](https://deploy-preview-1547--stacks.netlify.app/product/components/navigation/) focus states look in general, so they've retained the muted focus ring. It's focus ring style is unmodified in the context of the topbar.

### Muted button

Same with the [muted button](https://deploy-preview-1547--stacks.netlify.app/product/components/buttons/#muted). In the topbar, buttons get whichever focus ring style they normally have. This is proabably only an issue in the main topbar in the Stack docs pages since we don't generally stick muted buttons up there.

### Possible solution

I figure there's a few ways to address this:

#### Add topbar-specific overrides

We could set the focus ring color for those elements within the context of the topbar. I'm not a huge fan of this idea since we could wind up playing whack-a-mole since we can't always guarantee what elements could be shoved into the topbar.

#### Remove the muted focus ring in favor of the default style

We _almost_ removed the muted focus ring style. If we want muted elements to use the default focus ring style, replacing the muted style is one option.

#### Use the default focus ring style on navigation items

Navigation items use muted but could be switched to use the default focus ring. This wouldn't address the muted button in the header issue though.

## The orange outline

> I don't mind the orange outline as much

@CGuindon Some themed elements get a matching themed outline currently (navigation may be the only one that gets the "primary" theme color by default). Do you want to change this so that it gets the secondary color for the focus ring to match up with most other components?

CGuindon:
@dancormier 

> Navigation items use muted but could be switched to use the default focus ring. This wouldn't address the muted button in the header issue though.

I'm tempted to go this route, do you recall the reasoning behind why an element should get default vs muted? I think if buttons have default than nav items should as well?

> Do you want to change this so that it gets the secondary color for the focus ring to match up with most other components

Let's leave this as is.

PR #1550, opened 2023-11-02
fix(a11y): update orange values to make pagination and navigation compliant to WCAG non-text contrast SC

giamir:
This PR addresses [STACKS-440](https://stackoverflow.atlassian.net/browse/STACKS-440).

## Problem

Pagination and Navigation components selected state don't comply with the [WCAG2.2 non-text contrast Success Criterion](https://www.w3.org/TR/WCAG22/#non-text-contrast) (3:1).

## Solution

Tweak slightly the default orange palette to reach the 3:1 WCAG Contrast Ratio Requirement.

## Extending Test Utils
This PR also extends our testing utils to run `additionalAssertions` for dedicated components.
This allows us to add targeted a11y tests checking that the we are compliant with the [WCAG2.2 non-text contrast Success Criterion](https://www.w3.org/TR/WCAG22/#non-text-contrast).
@dancormier I would appreciate if you could review how I have extended the test utils. I do think this extension could help us also to test for the [Focus Appereance SC](https://www.w3.org/TR/WCAG22/#focus-appearance) in targeted components answering the open question in [this WIP PR](https://github.com/StackExchange/Stacks/pull/1547).

## Open Questions

- Our `highcontrast-light` mode still fails the [WCAG2.2 non-text contrast Success Criterion](https://www.w3.org/TR/WCAG22/#non-text-contrast). See failing tests (WCAG Contrast Ratio 2.6:1) in this PR. 
@CGuindon I guess this should be fixed before we merge this in?
- We are currently testing for this SC only against the Pagination and Navigation component because they were mentioned in the original ticket. I do believe more components need to actually be checked/tested though. What about buttons and button groups?
- [Tangent] I was wondering if we can assume that any element containing text need to have a WCAG Color Contrast Ratio of at least 3:1 (regardless of APCA). I am asking because if that is the case we might want to create an axe-core rule that check against that.


CGuindon:
- I'll look into HC today to see if we can change something to fix this (at least on pagination and navigation)

> What about buttons and button groups?

Our buttons don't use the orange color, just the blues and blacks so we don't have to worry about those components. The only other component I can think of that uses orange might be the voting arrows but it's a different UI (has borders and different fill as well) so it ends up passing for different reasons. Nothing else I can think of uses a straight orange background only to signal an interactive change to users.

>I was wondering if we can assume that any element containing text need to have a WCAG Color Contrast Ratio of at least 3:1 (regardless of APCA). I am asking because if that is the case we might want to create an axe-core rule that check against that.

I don't think that would work... I believe there are a few color combos that hit our APAC 60 or 70 rule but don't meet AA 3:1 ratio. I could be wrong but I recall coming across that scenario while exploring certain color alterations.

dancormier (commented):
Thanks for implementing these changes @giamir! I have a few questions about the `WCAGNonTextContrast` function you added. (Ignoring a11y failures for now), the changes seem solid.

> **Note**
> I'm not explicitly approving since I have some questions below, we have some a11y test failures, and I assume visual regression test images will need to be regenerated before merging. Considering this is time-sensitive, I'm with this being merged once the tests are passing. If answering my questions takes too much time, I'm fine with getting responses post-merge.

CGuindon (changes requested):
Ok, I've played around with the orange colors and it's a tricky one to say the least for HC mode.

The best path forward I can see is to make change the orange color stops in Light HC mode to hit the AAA standards we need. 
- I've kept the orange-300 to match our default orange-400 (similar to what it is currently, I think this is used for borders and illustrations in some cases so I'm hoping we can keep some orange there where the AAA is not needed)
- The new orange-400 in Light HC is very close to our default orange-600 — this is the minimum I needed to hit both APAC LC60 and WCAG AAA 7:1 contrast with white text. This will also give us our minimum 3:1 ratio with the background color on white background for the nav and pagination elements.
- Which means, I added a new darker orange-600 so that we would have a difference between orange-400 and oranges-500/600 (the same value in HC) for the HC Light mode. 

Those changes should gives us #fff text on #8A440A which is a ratio of 7.21:1 and LC89 for the pagination on HC Light.

I haven't seen any other elements from the MS sheet that had issues with our orange so I'm going to assume if we can at least fix the nav and pagination components we're good for the Nov 8th release.

Screenshot 2023-11-02 at 12 57 43 PM

lib/exports/color-sets.less
```diff
@@ -106,8 +106,8 @@
     200: hsl(23, 87%, 97%);
     300: hsl(27, 90%, 55%);
     400: hsl(27, 90%, 55%);
-    500: hsl(27, 80%, 28%);
-    600: hsl(27, 80%, 28%);
+    500: hsl(27, 87%, 27%);
```
CGuindon: @giamir I don't think there's supposed to be a change to the Dark mode colors — only Light and HC light based on my ticket. Dark was ok.

I'm also seeing some different values for orange dark mode everywhere. We did a saturation increase which I'm not seeing in this individual values either: https://stackexchange.slack.com/archives/C05RZ6Y8NKD/p1698164022556159?thread_ts=1698162819.819939&cid=C05RZ6Y8NKD

CGuindon: It's possible I grabbed the color-sets.less file from the wrong place. I would pull fresh from what's in stacks today and change the orange values only.

giamir: Sometimes Github is collapsing files and causes confusion.
Screenshot 2023-11-03 at 09 02 22

These are the only changes I made yesterday in terms of source code files.
Screenshot 2023-11-03 at 08 57 50
As you can see only `orange` `orange-hc` has changed as mentioned in the ticket.
This branch is based off develop and it contains all the latest changes already.

giamir:
@CGuindon 

> The best path forward I can see is to make change the orange color stops in Light HC mode to hit the AAA standards we need.

Honestly I am a bit confused around what we are after. I really would like us to start being **very specific and detailed** about the WCAG rules we are targeting (for example by including links to Success Criteria from the WCAG docs).
I thought that this fix was related to failures related to [Success Criterion 1.4.11 Non-text Contrast](https://www.w3.org/TR/WCAG22/#non-text-contrast).

From what you are writing we are now referring also to 2 other success criteria:
- [Success Criterion 1.4.3 Contrast (Minimum)](https://www.w3.org/TR/WCAG22/#contrast-minimum). This is the one necessary to reach AA level for text and images of text (4.5:1 for body text)
- [Success Criterion 1.4.6 Contrast (Enhanced)](https://www.w3.org/TR/WCAG22/#contrast-enhanced). This is the one necessary to reach AAA level for text and images of text (7:1 for body text)

<br><br>

So for me we either categorize the pagination and navigation components as non-text content ([see definition](https://www.w3.org/TR/WCAG22/#dfn-non-text-content)) or as regular text.

<hr>

If we categorize them as **non-text content** then this criterion apply:
- [Success Criterion 1.4.11 Non-text Contrast](https://www.w3.org/TR/WCAG22/#non-text-contrast)

This means we need to have a 3:1 contrast ratio (AA). **There are no other criterion to satisfy for AAA**.

<hr>

If we categorize them as **text content** than this 2 criteria apply:
- [Success Criterion 1.4.3 Contrast (Minimum)](https://www.w3.org/TR/WCAG22/#contrast-minimum)
- [Success Criterion 1.4.6 Contrast (Enhanced)](https://www.w3.org/TR/WCAG22/#contrast-enhanced)

This means we need to have a 4.5:1 contrast ratio for our regular modes (light and dark) and a 7:1 for our highcontrast modes.

<hr>

<br><br>

I see how all of this might sound pedantic but I think it is really important to be precise about this stuff to:
- Avoid communication misunderstanding on implementation
- Be confident on our claims (e.g. our pagination component is WCAG2.2 AA compliant)
- **Be able to test the changes so that we can prevent regressions**

<br>

So in a nutshell my question is: **what are we fixing?**
- A. a non-text contrast issue (which would require 3:1 contrast ratio)
- B. a text contrast issue (which would require 4.5:1 contrast ratio for regular modes and 7:1 for highcontrast modes)

giamir:
> Thanks for implementing these changes @giamir! I have a few questions about the WCAGNonTextContrast function you added. (Ignoring a11y failures for now), the changes seem solid.

@dancormier Somehow I cannot see the questions you have (in case you have written them down somewhere). Or maybe you want ask them sync in person?

giamir:
Open conversations have been addressed in a sync way. You can find the outcomes in [this commit](https://github.com/StackExchange/Stacks/pull/1550/commits/4867855ebe3fdc5ded09f52e4953bda5903c9197).

PR #1576, opened 2023-12-04
docs(adr): add draft ADR for library release automation

dancormier:
[STACKS-417](https://stackoverflow.atlassian.net/browse/STACKS-417).

This PR adds an ADR to describe the decision to use Changesets as a way to automate away much of the manual effort currently necessary to cut releases (across _all_ Stacks libraries).

> **Note**
> I considered implementing Changesets into this repo out of scope for the sake of this PR and instead decided to focus solely on the ADR. IMO, we should handle updating the release process _after_ this ADR is agreed upon and merged.


giamir:
> I considered implementing Changesets into this repo out of scope for the sake of this PR and instead decided to focus solely on the ADR. IMO, we should handle updating the release process after this ADR is agreed upon and merged.

Absolutely, this is out of scope for STACKS-417. We have other placeholder stories in place which should be updated linking to this ADR. 🙂

adrs/0006-automatic-library-release.md
```diff
@@ -0,0 +1,46 @@
+# Automatic Library Release
+
+**Date:** 2023-12-04
+
+**Related PR:** [#1576](https://github.com/StackExchange/Stacks/pull/1576)
+
+## Issue
+
+At the time of writing, many Stacks libraries are released manually. This is a time-consuming process that is prone to human error. We should automate this process to save time and reduce errors.
```
giamir: On top of the process being time-consuming and error-prone there is also an aspect about using several personal account credentials with publish permissions (which doesn't follow [PoLP](https://en.wikipedia.org/wiki/Principle_of_least_privilege)). Might be worth mentioning.

adrs/0006-automatic-library-release.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Decision
+
+We will use [Changesets](https://github.com/changesets/changesets) to manage versioning and automate the release process of our libraries in conjunction with GitHub Actions.
+
+### Reasoning
+
+#### Ease of use
+
+The primary implentation of Changesets is its [CLI package](https://www.npmjs.com/package/@changesets/cli) which will help us manage the versioning and changelog entries for your packages. We can apply the [Changesets bot](https://github.com/apps/changeset-bot) to any repository verify that a given PR has a changeset associated with it and remind the author to create one if it does not.
```
giamir:
```suggestion
The primary implentation of Changesets is its [CLI package](https://www.npmjs.com/package/@changesets/cli) which will help us manage the versioning and changelog entries for our packages. We can apply the [Changesets bot](https://github.com/apps/changeset-bot) to any repository to verify that a given PR has a changeset associated with it and remind the author to create one if it does not.
```

giamir (approved):
Overall it looks good to me. 🙂

I believe the only thing I would add is a small section explaining why we picked changesets over "other tools" (e.g. [release-please](https://github.com/googleapis/release-please/tree/main), [sematic-release](https://github.com/semantic-release/semantic-release), etc... - there might be more).
We don't need to necessarily do deep research on them but at least mention what we have quickly evaluated. For example semantic-release doesn't support monorepo setups well. Release please if I recall correctly doesn't have an automated way to publish to npm (but I might be wrong). 

One more thing worth highlighting in the ADR is that despite changesets incentivize describing changes independently from the commit messages we are still committed to follow conventional-commits. You can probably take inspiration from what I have wrote [here](https://github.com/StackExchange/apca-check#publishing).

Thanks @dancormier. ❤️

dancormier:
@giamir thank you for your review! i made the changes you suggested and will merge this PR shortly. If you see any recent changes that need improvement, please let me know and I'll open a new PR.

PR #1595, opened 2024-01-03
fix(bundling): correctly generate minified css bundle

giamir:
[STACKS-366](https://stackoverflow.atlassian.net/browse/STACKS-366)

## Context 

While merging dependabot PRs I noticed that we got a lot of warning when running npm run build.

```
[w:bundle]   (875:3) from "postcss-calc" plugin: Parse error on line 1:
[w:bundle]   ...r(--theme-topbar-height, calc(var(--su-static48) + var(--su-static8))) * var(--...
[w:bundle]   ------------------------------------------------------------------------^
[w:bundle]   Expecting end of input, "ADD", "SUB", "MUL", "DIV", got unexpected "RPAREN"
[w:bundle]
[w:bundle]   Code:
[w:bundle]     transform:translate3d(0,calc(var(--theme-topbar-height, calc(var(--su-static48) + var(--su-static8))) * var(--_no-x-offset)),0)
```

```
[w:bundle]   WARNING in ./lib/stacks.less (./lib/stacks.less.webpack[javascript/auto]!=!./node_modules/css-loader/dist/cjs.js??ruleSet[1].rules[1].use[1]!./node_modules/postcss-loader/dist/cjs.js??ruleSet[1].rules[1].use[2]!./node_modules/less-loader/dist/cjs.js!./lib/stacks.less)
[w:bundle]   Module Warning (from ./node_modules/postcss-loader/dist/cjs.js):
[w:bundle]   (1817:5) from "postcss-svgo" plugin: SvgoParserError: <input>:1:1: Non-whitespace before first tag.
[w:bundle]
[w:bundle]   > 1 | %3Csvg width='11' height='11' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M1...
[w:bundle]       | ^
[w:bundle]
[w:bundle]
[w:bundle]   Code:
[w:bundle]     --_ch-bg-image:url("data:image/svg+xml,%3Csvg width='11' height='11' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M10 3.41L8.59 2 4 6.59 2.41 5 1 6.41l3 3z' fill='hsl(210, 3.0000000000000027%, 15%)'/%3E%3C/svg%3E")
```

The command, between other things, creates a minified version of our css bundle using `cssnano`.

`cssnano` runs its default presets that includes `postcss-svgo` and `postcss-calc`. Those plugins were responsible for the warnings and the incorrect minification of our data-uri svg images.

More specifically `postcss-svgo` does not play well with [data-uris](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs) not containing a semicolon after the mediatype:
`data:image/svg+xml,%3Csvg ...` will yield this warning `Non-whitespace before first tag` and incorrectly optimize the svg. For now I managed to make svgo behave correctly by simply adding a semicolon after the mediatype (this should not be necessary according to the data uri scheme). Other solutions could be:
- Switching off svgo from cssnano all together
- Decode our svg in utf-8 (e.g. from `%3C` to `<` - like what they do in the [postcss-svgo example docs](https://www.npmjs.com/package/postcss-svgo))

## How to test

To test that the bundle is minified correctly and the confetti animation works as expected you can serve (`npx serve`) this content in an temporary `test.html` file at the root level of this repo.
```
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Test</title>
    <link href="dist/css/stacks.min.css" rel="stylesheet">
</head>
    <body>
        <div class="hmn2 bg-confetti-animated">…</div>
    </body>
</html>

```

No warning should be showed when running `npm run build`. Tests and Linters should pass.

This PR also updates all remaining dependencies. All minor updates apart from the vitest update which is the reason of why the snapshot files has changed (if you look carefully you can notice that only some backward slashes have been removed which is a good thing and no real regression is happening in the PR).

[Here you can find a diff checker link comparing stacks.min.css before and after the PR ](https://www.diffchecker.com/w56KorMi/)


lib/components/banner/banner.less
```diff
@@ -37,8 +38,7 @@
     inset: 0 0 auto 0;
     padding: var(--su12);
     position: fixed;
-    -webkit-transform: translate3d(0, calc(var(--theme-topbar-height, calc(var(--su-static48) + var(--su-static8))) * var(--_no-x-offset)), 0);
```
giamir, line 40: @dancormier I dropped the vendor prefix since I believe we don't support old version of browsers. Let me know if that makes sense.

lib/components/banner/banner.less
```diff
@@ -3,6 +3,7 @@
 // This would reduce the amount of CSS we ship to the client and simplify our codebase
 .s-banner {
     --_no-x-offset: 0;
+    --_no-height: var(--theme-topbar-height, calc(var(--su-static48) + var(--su-static8)));
```
giamir: @dancormier It looks like this is a legitimate use case for PPCP. postcss-calc was thrown off by the previous long inlined expression contained in the transform property.

dancormier: Nice catch!

BTW, I renamed these PPCPs to reflect the axis being translated (based on the syntax described [here](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/translate3d#ty)) in [this commit](https://github.com/StackExchange/Stacks/pull/1595/commits/2c80beaa5419218e2d252062d4399fef7c53331b). I think using `x` was a mistake and this commit corrects that and unifies the syntax with the newly-added PPCP.

lib/components/banner/banner.less
dancormier: Makes sense!

dancormier (approved):
Works well! Thanks for this fix! I'm perfectly happy turning off svgo from css-nano

> **Note**
> I made a minor naming change in https://github.com/StackExchange/Stacks/pull/1595/commits/2c80beaa5419218e2d252062d4399fef7c53331b that should result in no material changes to the build.

PR #1609, opened 2024-01-22
refactor(focus-ring): use focus-ring CSS vars for focus states only

dancormier:
[STACKS-532](https://stackoverflow.atlassian.net/browse/STACKS-532)

@giamir I'm curious to get your thoughts on two things:

1) Naming: I went with `translucent` because I figured that was descriptive enough but I'm wondering if you can think of a better name to go with for the added css variables
2) When I update the less snapshots locally, each instance of `:` in selectors gets prefixed with `\\\\`. When the workflow runs on GitHub, it expects "\\". Any idea of why this happens and how to resolve? (This is a minor issue so please don't spend too much time on this!)

### DIFF
--- a/lib/atomic/misc.less
+++ b/lib/atomic/misc.less
@@ -267,7 +267,7 @@
 //  $  OUTLINE
 //  ----------------------------------------------------------------------------
 .outline-none { outline: 0 !important; }
-.outline-ring { outline: solid var(--su-static4) var(--focus-ring); }
+.outline-ring { outline: solid var(--su-static4) var(--translucent-secondary); }
 
 .f\:outline-ring {
     &:focus,
--- a/docs/_data/atomics.json
+++ b/docs/_data/atomics.json
@@ -1120,8 +1120,8 @@
     },
     {
       "class": "outline-ring",
-      "output": "outline: solid var(--su-static4) var(--focus-ring);",
-      "define": "<p class='mb0'>Adds an outline that mimics our button focus styling.</p>",
+      "output": "outline: solid var(--su-static4) var(--translucent-secondary);",
+      "define": "<p class='mb0'>Adds an outline using our translucent secondary color.</p>",
       "focus": true
     }
   ],
--- a/docs/product/base/outline.html
+++ b/docs/product/base/outline.html
@@ -4,6 +4,10 @@ title: Outline
 description: 
 ---
 <section class="stacks-section">
+    {% tip, "warning", "mb24" %}
+        <strong>Warning:</strong> <code>.outline-ring</code> is does not meet our accessibility standards for focus state styling and should be used for decorative purposes only.
+    {% endtip %}
+
     {% header "h2", "Classes" %}
     <div class="overflow-x-auto mb48" tabindex="0">
         <table class="wmn3 s-table s-table__bx-simple">
@@ -12,7 +16,6 @@ description:
                     <th class="s-table--cell2" scope="col">Class</th>
                     <th class="s-table--cell4" scope="col">Output</th>
                     <th scope="col">Definition</th>
-                    <th scope="col" class="ta-center"><a class="s-link s-link__inherit" href="{{ "/product/guidelines/conditional-classes#focus" | url }}">Focus?</a></th>
                 </tr>
             </thead>
             <tbody>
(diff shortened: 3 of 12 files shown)


lib/atomic/misc.less
```diff
@@ -267,7 +267,7 @@
 //  $  OUTLINE
 //  ----------------------------------------------------------------------------
 .outline-none { outline: 0 !important; }
-.outline-ring { outline: solid var(--su-static4) var(--focus-ring); }
+.outline-ring { outline: solid var(--su-static4) var(--translucent-secondary); }
```
giamir, line 270: I think the doc for this atomic class will need to be updated:
https://stackoverflow.design/product/base/outline

Furthermore we might want to put a disclaimer that `outline-ring` should not be used to highlight focus states because it is not WCAG compliant. Essentially saying that this atomic class is for  "decorative-only" use and should not convey any semantic meaning.

giamir: If it was not for the breaking changes I would even go as far as suggesting to rename the class to something like `.outline-decorative` or `.outline-decorative-ring`

giamir (changes requested):
Overall looks good to me after we take care of updating the atomic class docs. 

> Naming: I went with translucent because I figured that was descriptive enough but I'm wondering if you can think of a better name to go with for the added css variables

I think `translucent` is good given that we don't want to tight the color to a semantic value/meaning at this stage I believe.

> When I update the less snapshots locally, each instance of : in selectors gets prefixed with \\\\. When the workflow runs on GitHub, it expects "\". Any idea of why this happens and how to resolve? (This is a minor issue so please don't spend too much time on this!)

I remember upgrading from vitest 0.x.x to 1.x.x resolved some of those weird backslashes appearing in the snapshots. Have you tried to nuke the node_modules and do a fresh install? In my machine I cannot reproduce that issue (only downgrading vitest would cause the problem). 🤷

dancormier:
> I remember upgrading from vitest 0.x.x to 1.x.x resolved some of those weird backslashes appearing in the snapshots. Have you tried to nuke the node_modules and do a fresh install?

Nuking and reinstalling deps fixed the issue 🎉 @giamir thank you!

lib/atomic/misc.less
dancormier: > I think the doc for this atomic class will need to be updated

Currently, it still mimics our focus style (though it references `--translucent-secondary`). We'll probably want it to mimic our focus style after we update to the new style broadly. Considering this, 

> If it was not for the breaking changes I would even go as far as suggesting to rename the class to something like .outline-decorative or .outline-decorative-ring

Ultimately, instead of renaming I'd like to get rid of this atomic class altogether. We reference it in only [a single place](https://github.com/StackEng/StackOverflow/blob/231517c19550c525ed2477285d7884ecea6e5c81/StackOverflow/_Scripts/PostCollections/CreateEditPage.ts#L177) and I think it's a class has very little utility.

**Related**: Every once in a while, I come across some minor breaking change I think would be worth including in the next major release whenever it comes along. Do you think it's reasonable to add Jira issues for these scenarios with a tag like `breaking-change` that way we could find all proposed breaking changes when planning for a major release?

giamir: > Ultimately, instead of renaming I'd like to get rid of this atomic class altogether. We reference it in only [a single place](https://github.com/StackEng/StackOverflow/blob/231517c19550c525ed2477285d7884ecea6e5c81/StackOverflow/_Scripts/PostCollections/CreateEditPage.ts#L177) and I think it's a class has very little utility.

Oh yes, let's remove that. :) 

> Related: Every once in a while, I come across some minor breaking change I think would be worth including in the next major release whenever it comes along. Do you think it's reasonable to add Jira issues for these scenarios with a tag like breaking-change that way we could find all proposed breaking changes when planning for a major release?

That makes sense to me, once we have few of those we could start planning a major release. 
Feel free to go ahead with that. Maybe add also the label `stacks-classic` since in our backlog we have tickets for several Stacks libraries these days.

dancormier: I've created [STACKS-540](https://stackoverflow.atlassian.net/browse/STACKS-540) for removal of `.outline-ring`

PR #1611, opened 2024-01-25
test(visual): add improved baseline images update workflow

giamir:
This PR change the way we upgrade our baseline images so that only the failing snapshots get regenerated.
It should also provide a better way to look at diff directly in vscode.

Run this command `git config diff.lfs.textconv cat` and install [this vscode extension](https://marketplace.visualstudio.com/items?itemName=RayWiis.png-image-diff) for optimal experience.

Chekout the [updated readme](https://github.com/StackExchange/Stacks/blob/b35aee76c1ececae3132aa2f6a157a4787add7e5/README.md#visual-regression-tests) for more info.


dancormier (approved):
I tested this out and it works as expected. This is a nice enhancement. Thank you for this update @giamir!

PR #1614, opened 2024-01-25
fix(tag): improve focus style to meet WCAG 2.2 AA

dancormier:
Addresses [STACKS-541](https://stackoverflow.atlassian.net/browse/STACKS-541). See also #1612, #1613 for related PRs.

This commit also adds some minor styling updates to account for button-based `.s-tag--dismiss` elements.

These changes should result in no changes to existing `.s-tag--dismiss` elements, but the docs now display the semantically proper usage of a button element for the `X` button with the parent `.s-tag` being rendered as a span and consumers can now use a button element for the `.s-tag--dismiss` child without needing to apply reset classes.


docs/product/components/tags.html
```diff
@@ -36,17 +36,17 @@
 {% highlight html %}
 <div class="d-flex g4">
     <a class="s-tag" href="#">jquery</a>
-    <a class="s-tag" href="#">javascript <span class="s-tag--dismiss">@Svg.ClearSm</span></a>
+    <span class="s-tag">javascript <button class="s-tag--dismiss s-btn s-btn__unset">@Svg.ClearSm</button></span>
```
giamir: `s-btn s-btn__unset` are not necessary

docs/product/components/tags.html
```diff
@@ -36,17 +36,17 @@
...
     <a class="s-tag" href="#"><img class="s-tag--sponsor" src="https://i.stack.imgur.com/tKsDb.png" width="16" height="18" alt="Google Android"> android</a>
-    <a class="s-tag is-selected" href="#">razor <span class="s-tag--dismiss">@Svg.ClearSm</span></a>
+    <span class="s-tag is-selected">razor <button class="s-tag--dismiss s-btn s-btn__unset">@Svg.ClearSm</button></span>
```
giamir: `s-btn s-btn__unset` are not necessary

giamir (commented):
Overall looks good to me. Great work @dancormier. I found the black/light-blue between the dismiss button and the tag in the following picture an odd combination but I would let @CGuindon comment on that. Probably necessary to get to the desired 3:1 contrast.

Screenshot 2024-01-26 at 15 00 05

The only observation I have is to clarify if a use case where the tag is an anchor and it also contains a dismiss button is a valid one?
```
<a class="s-tag" href="#">javascript <button class="s-tag--dismiss">@Svg.ClearSm</button></a>
```

if no, we should probably mention in the `s-tag--dismiss` documentation that the dismiss action can only be used in non focusable tags.

If it's valid, we should add an example.

CGuindon (changes requested):
Oh yes, I missed the focus state on the clear icon version of tags. In trying to achieve the 3:1 ratio, would either of these be possible (my preference is the second one):

1. Add an extra (3rd) ring layer of white on the outside — since we have 6px padding in between icon and text, this would mean the text will be flush against that last white ring layer.
2. Add additional padding around the Clear Icon when in focus to push the rings to the 'outside' of the tag element. This would allow us to have 3 edges that _touch_ the white background color and maintain our 3:1 ratio on 3 sides. Only the inner left border wouldn't meet that ratio in certain cases but I think that's ok.

[Figma](https://www.figma.com/file/3OlPL6x0chNbMIPjP2vxas/Stack-Overflow-3.0---Site-Modernization?type=design&node-id=3473%3A2750&mode=design&t=YOtlRCtH27zazs2J-1)

Screenshot 2024-01-26 at 8 22 19 AM

dancormier:
@CGuindon We could make one of those changes but I wanted to show what they'd look like in practice before I make these changes, plus run an alternative option by you as well.

# Option 1

> 1. Add an extra (3rd) ring layer of white on the outside — since we have 6px padding in between icon and text, this would mean the text will be flush against that last white ring layer.

For option 1, the outermost outline would overlap the text unless we change the margin of the dismiss button.

image

I also think this technically might not meet the [WCAG SC 2.4.13 Focus Appearance](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance):

> has a contrast ratio of at least 3:1 between the same pixels in the focused and unfocused states. 

Since these outlines would be placed on top of the tag element, they may not always contrast 3:1 with the pixels they're replacing (at least with the base selected tag and possibly with others).

# Option 2

> 2. Add additional padding around the Clear Icon when in focus to push the rings to the 'outside' of the tag element. This would allow us to have 3 edges that touch the white background color and maintain our 3:1 ratio on 3 sides. Only the inner left border wouldn't meet that ratio in certain cases but I think that's ok.

Option 2 would cause fewer problems IMO. If we go with option two, we'd need to change the size of the dismiss button and compensate for that change with negative margins to maintain the same layout.

image

This gets the outlines outside the tag, but this may cause hover to look kinda weird in some cases:
 
image
image

Outside of that minor issue, I think this solution would work fine.

# Alternative option

Since it seems like what we have currently in this PR _almost_ meets our contrast target, there's another option we could consider. We could use a slightly darker `theme-secondary` color for the outline on dismiss buttons within selected tags. This would allow us to maintain the current button size, not have to adjust the layout, and hit our 3:1 target contrast ratio.

The lowest effective contrast ratio I could find is between the blue ring and the base selected tag background, which is a [1.87:1 ratio](https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=92C2F2) (All other color combinations exceed the 3:1 ratio). If we use the `theme-secondary-500` instead of `theme-secondary-400` for selected tags dismiss button, we should exceed the 3:1 ratio in all cases.

## Selected state tags with `theme-secondary-500` outline
### Base
#### Light: [3.63:1](https://webaim.org/resources/contrastchecker/?fcolor=92C2F2&bcolor=155CA2)
image

#### Dark: [3.36:1](https://webaim.org/resources/contrastchecker/?fcolor=3C7AB8&bcolor=252627)
image

### Moderator
#### Light: [3.71:1](https://webaim.org/resources/contrastchecker/?fcolor=F6B179&bcolor=155CA2)
image

#### Dark: [3.33:1](https://webaim.org/resources/contrastchecker/?fcolor=A46837&bcolor=252627)
image

### Required
#### Light: [11.34:1](https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=0D3B6A)
image

#### Dark: [12.43:1](https://webaim.org/resources/contrastchecker/?fcolor=252627&bcolor=D8EBFD)
image

### Muted
#### Light: [4.8:1](https://webaim.org/resources/contrastchecker/?fcolor=D6D9DC&bcolor=155CA2)
image

#### Dark: [5.65:1](https://webaim.org/resources/contrastchecker/?fcolor=B2D6FB&bcolor=494D50)
image

All high contrast versions would also meet or exceed the 3:1 ratio.

---

@CGuindon let me know what you think and I'll implement whatever you think the best option is.

dancormier:
> The only observation I have is to clarify if a use case where the tag is an anchor and it also contains a dismiss button is a valid one?
> 
> ```
> <a class="s-tag" href="#">javascript <button class="s-tag--dismiss">@Svg.ClearSm</button></a>
> ```
> 
> if no, we should probably mention in the `s-tag--dismiss` documentation that the dismiss action can only be used in non focusable tags.
> 
> If it's valid, we should add an example.

@giamir From [W3C HTML reference](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element):

> … there must be no interactive content descendant, `a` element descendant, or descendant with the `tabindex` attribute specified.

It's invalid to have nested focusable elements (so a `button` cannot be nested in an `a` tag). I've updated the examples to show this and I've updated the description for `.s-tag--dismiss` to describe the appropriate markup.

CGuindon:
@dancormier Alternate option gives me pause because of theming... and in some cases we don't have the 3:1 ratio on both sides so I think Option 2 is safest and doesn't feel that bad for the hover state weirdness.

Option 2

dancormier:
I've updated the dismiss button so the theme-secondary outline is shown outside of the tag element as shown in option two [here](https://github.com/StackExchange/Stacks/pull/1614#pullrequestreview-1846086702). In order to do this, I needed to shift the location of the dismiss button horizontally slightly (I believe it winds up being 1px to the right).



@CGuindon @giamir Can you two give this a rereview when you get a moment and let me know if this seems ok?

CGuindon (approved):
I can live with this — ooooohhhh that focus state looks niiiiiiiiiiice

lib/components/tag/tag.less
```diff
@@ -18,6 +18,7 @@
     --_ta-pl: var(--_ta-px);
     --_ta-pr: var(--_ta-px);
     --_ta-px: var(--su4);
+    --_ta-dismiss-padding: calc(var(--_ta-px) - 1px);
```
giamir: should we use `var(--su-static1)` ?

giamir (approved):
Looks great to me. 
I am also happy to see that the only baseline images affected are the one featuring the dismiss button. 🎉

PR #1621, opened 2024-01-29
fix(block-link): improve focus style to meet WCAG 2.2 AA

dancormier:
Resolves [STACKS-545](https://stackoverflow.atlassian.net/browse/STACKS-545)

> **Note**
> The above linked ticket shows updates as they relate to the `s-menu` component. The focusable elements shown within the `s-menu` component are `s-block-link`s, so this PR updates the `s-block-link` focus style.

### DIFF
--- a/lib/components/block-link/block-link.less
+++ b/lib/components/block-link/block-link.less
@@ -25,16 +25,9 @@ a.s-block-link,
     &&__left,
     &&__right {
         &.is-selected {
-            &:focus:not(:focus-visible),
-            &:focus-visible {
-                outline: none;
+            &:not(:focus-visible) {
+                box-shadow: inset var(--_li-block-bs-offset-x, 3px) 0 0 var(--_bl-bs-color);
             }
-
-            &:focus-visible {
-                box-shadow: inset var(--_li-block-bs-offset-x, 3px) 0 0 var(--_bl-bs-color), 0 0 0 var(--su-static4) var(--focus-ring-muted);
-            }
-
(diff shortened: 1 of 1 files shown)


CGuindon (commented):
Screenshot 2024-01-30 at 7 40 31 AM

I'm assuming adding padding here won't do anything overall — devs would need to make sure the element has a container somewhere in the core code? 

Is there a better way to show an example of 'no container' that would show at least 2px padding from the text?

dancormier:
@CGuindon That example is kinda misleading. Each link has markup like this:

```html
<a href="…" class="s-block-link px0">…</a>
```

Note the `px0`. That class overrides any left or right padding to be `0px`, so a dev would have to specifically apply an atomic class to override the padding in this case. If we ever take a good look at our docs pages, I think this example would be a good one to get rid of.

I found instances of this in two files ([1](https://github.com/StackEng/StackOverflow/blob/main/StackOverflow/Views/SubCommunities/Articles/Edit.cshtml#L276), [2](https://github.com/StackEng/StackOverflow/blob/main/StackOverflow/Views/TopBar/Inbox.cshtml#L36)), so I don't think it's too big of a deal and it's probably something that can be resolved pretty easily in Core by finessing the padding a little.

giamir (approved):
Code changes look good to me. Thanks @dancormier.

I  see that there is a small divergence on the focus ring radius between the mockup

Screenshot 2024-01-30 at 17 30 35

and the final result

Screenshot 2024-01-30 at 17 30 28

CGuindon (approved):

PR #1622, opened 2024-01-29
refactor(topbar): use pseudo-private custom properties

dancormier:
[STACKS-556](https://stackoverflow.atlassian.net/browse/STACKS-556)

This PR is a refactor of the topbar component styles to use our pseudo-private custom properties structure.

I'm considering this PR blocked by https://github.com/StackExchange/Stacks/pull/1624.

> **Note**
> Some Core Less files directly reference `.s-topbar__dark` and `.s-topbar__light`, so this PR includes those styles outside of the `.s-topbar` block. _Eventually_, Core should be updated to _not_ act as though these styles are exported.

> **Note**
> include tests for custom theme properties


giamir:
Out of scope for v2.1.0

lib/components/topbar/topbar.less
```diff
@@ -1,23 +1,69 @@
...
-    // Redefine the variables for extra contrast in high-contrast mode
     .highcontrast-mode({
+        --_tb-notice-td: underline;
+
+        &.s-topbar__dark {
+            --focus-theme: .theme-dark-default()[secondary];
+        }
+
+        &__dark {
+            // TODO HACK remove everything below once light/dark topbars are inheriting forced themes correctly
```
giamir, line 48: Do we have a tech debt ticket already for this explaining what should be the ideal API/behavior? If not I would recommend to create one and add it to the backlog (even if ultimately it might end up in the low prio pile).

lib/components/topbar/topbar.less
```diff
@@ -1,23 +1,69 @@
...
+
+            .s-badge {
+                border-color: currentColor;
+            }
+        }
+
+        &__light {
+            // TODO HACK remove everything below once light/dark topbars are inheriting forced themes correctly
+            --theme-topbar-item-color: .set-black()[600];
+            --theme-topbar-item-color-hover: .set-black()[600]; // TODO hover is now identical color to base color
```
giamir, line 62: Is this something that @CGuindon could consult us with a solution? If we are actively not happy about this behavior we should track it in a ticket. If it's not important and we are ok with keeping it that way it should not be a TODO in the first place. 🙂

giamir (approved):
Thanks @dancormier for doing this refactor. I checked as much as I could - this is a very complex component/pattern with many design choices undocumented (e.g. [responsive styles for small screens](https://github.com/StackExchange/Stacks/blob/dcormier/refactor-topbar-ppcp/lib/components/topbar/topbar.less#L92), [JS behavior expected to be applied by consumer](https://github.com/StackExchange/Stacks/blob/80ecef3f4f4c870c580dff255e672e6aa475a9da/docs/assets/js/entry.topbar.js#L7), etc...).

Overall I could not find anything off / different from how it was before (so the refactor works). 🎉

Out of curiosity: did the visual regression tests helped you during the refactor? Did they catch anything?

Thank you and great work as always. 👍

dancormier:
> Out of curiosity: did the visual regression tests helped you during the refactor? Did they catch anything?

Yes!

I introduced a [minor regression](https://github.com/StackExchange/Stacks/pull/1622/commits/c01d88aff0ffe4a6e48ad9339fd12dd34dc97cf6) in the course resolving merge conflicts which were caught when I ran the visual regression tests. The tests also gave me a bit more confidence in my changes and saved me lots of time so I could focus on testing the things not considered by the tests (element interactive states and custom theming, for example).

lib/components/topbar/topbar.less
dancormier: > Do we have a tech debt ticket already for this explaining what should be the ideal API/behavior? 

I don't believe so. That `TODO` was created over two years ago and I'm not entirely confident I know what "inheriting forced themes _correctly_" looks like. My best guess is that we're overriding a couple of custom theme vars in HC mode to ensure that they take precedence over once set in our consumers' code. I'll create a ticket outlining my best understanding of this (**update**: [ticket created](https://stackoverflow.atlassian.net/browse/STACKS-569)).

**Tangent**: I would really like to deprecate these `dark`/`light` topbar variants in favor of instructing consumers to [apply `.theme-[dark|light]__forced` when appropriate](https://stackoverflow.atlassian.net/browse/STACKS-408) or at least use our forced light/dark mode styles to set the colors used on our dark/light topbar variants. If we deprecate, consumers would need to do some work to ensure the topbar renders as they need, since things like the custom popovers will need to be adjusted to _not necessarily_ render in the forced mode.

lib/components/topbar/topbar.less
dancormier: This `TODO` was added in the course of the color redefinition work. When we included more shades of black/gray in our library, these two variables had similar but distinct values. When we moved to the v2 colors, the two values both converted to `black-600`, rendering them identical.

I'm not sure if we _need_ to have these be distinct values but this was marked as a `TODO` with the assumption that this was sub-optimal. I'll add a ticket to the design backlog to investigate this (**update**: added to _in triage_ [here](https://stackoverflow.atlassian.net/browse/STACKS-570))

dancormier:
@giamir thank you for reviewing these changes. I know that it's a tricky one to review but I appreciate you taking the time to do so. Since your review, I [expanded the visual test slightly](https://github.com/StackExchange/Stacks/commit/d651b8ea167ce173ccefb457ae9497d87325de7b) to include a topbar with custom theme variables set to give us more confidence that those variables will work as expected.

PR #1625, opened 2024-01-31
fix(test): add doctype to test HTML output

dancormier:
Without a doctype, we were essentially rendering our visual regression tests in quirks mode. This PR adds a doctype and includes regenerations of baseline images.


giamir (approved):
Works well in my machine. Nice catch @dancormier!
Merging.

PR #1628, opened 2024-02-01
fix(topbar): improve focus style to meet WCAG 2.2 AA

dancormier:
Addresses https://stackoverflow.atlassian.net/browse/STACKS-549

This PR updates focus styles for all elements _generally_ included within the topbar component.

[Deploy preview](https://deploy-preview-1628--stacks.netlify.app/product/components/topbar/)

### `.s-topbar--item`, `.s-topbar--logo`

**From [STACKS-549](https://stackoverflow.atlassian.net/browse/STACKS-549)**:

> Rounded corners will match the original non-focused component.
> 

To match the mockup, I added a border radius of `var(--br-sm)` of items (and of the logo for consistency). @CGuindon let me know if this is ok or if it needs modification.

| | focused | hovered | focused + hovered |
|-|---------|---------|-------------------|
| item | image | image | image |
| logo | image | image | image |

### Search

The old focus styles looked out of place and it was easy enough to add a matching focus ring to the search input and select element. @CGuindon is this alright or should I revert/modify?

| | focused |
|-|---------|
| select | image |
| input | image |

#### Ring placement

For the search input and select elements, I opted to render the focus ring _outside_ the element. This stands in contrast with other topbar elements, which all have the focus ring rendered inside the element. I did this because these elements rely on the border being visible to read as an input/select. When I tried adding the ring inside the element, the focus state was kinda tough to distinguish from the unfocused state, especially in high contrast mode.

| | light mode | light HC mode |
|-|---------|-|
| select | image | image |
| input | image | image |

### Code structure

Since this component doesn't yet adhere to our PPCP structure, I decided it was best to keep all of the focus code grouped for easy parsing (and easy porting when we make the switch to PPCP).

### DIFF
--- a/lib/components/topbar/topbar.less
+++ b/lib/components/topbar/topbar.less
@@ -34,7 +34,7 @@
         // Topbar items
         --theme-topbar-item-color: var(--black-600);
         --theme-topbar-item-color-hover: var(--black-600);
-        --theme-topbar-item-background-hover: var(--black-200);
+        --theme-topbar-item-background-hover: var(--black-300);
         --theme-topbar-item-color-current: var(--black);
     });
 
@@ -68,7 +68,6 @@
         --_tb-focus-color-inner: .set-black()[600]; // set to match .s-topbar__dark --theme-topbar-background-color;
         --_tb-focus-color-outer: var(--theme-dark-secondary-custom-400, .theme-dark-default()[secondary]);
     }
-
 
     // focus styles
     a&--logo,
--- a/lib/components/topbar/topbar.less
+++ b/lib/components/topbar/topbar.less
@@ -5,6 +5,9 @@
 }
 
 .s-topbar {
+    --_tb-focus-color-inner: var(--white);
+    --_tb-focus-color-outer: var(--theme-secondary-400);
+
     min-width: auto;
     width: 100%;
     z-index: var(--zi-navigation-fixed);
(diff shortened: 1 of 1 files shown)


CGuindon:
@dancormier 
> For the search input and select elements, I opted to render the focus ring outside the element. 

Yes, I agree. I was assuming we'd have to go outside for all input fields so in this case the search bar matches that logic and even though it's a button beside the input, it probably looks weirder to have it go inside then outside with side-by-side elements so I think your approach makes the most sense. Outside for both search elements.

CGuindon (approved):
Beauty!

giamir (approved):
Looks good to me. If we haven't created a ticket yet to migrate this component to PPCP we should. Apart from that I could not find anything out of the ordinary.

One question related to the search input. Shouldn't that match the focus behavior we will eventually add to all input components?
What I am asking is if the picture below represents how we will end up styling inputs going forward?
Screenshot 2024-02-02 at 14 53 32

I don't see why the topbar input should have a special style 🙂

dancormier:
@giamir I've created a [ticket](https://stackoverflow.atlassian.net/browse/STACKS-556) for the PPCP migration and will follow through on applying the input focus styles in an upcoming PR ([STACKS-553](https://stackoverflow.atlassian.net/browse/STACKS-553)).

PR #1629, opened 2024-02-02
fix(form-elements): improve focus style to meet WCAG 2.2 AA

dancormier:
[STACKS-553](https://stackoverflow.atlassian.net/browse/STACKS-553)

This PR update the focus styles for any input/textarea-based elements including:

- `s-input`
- `s-textarea`
- `s-checkbox`
- `s-radio`
- `s-select`
- `s-uploader`

### `.focus-styles()` mixin

I took this opportunity to add the `--focus-neutral` and `--focus-theme` color custom properties and generalize some common focus styles with the `.focus-styles()` mixin. This will allow us to set focus styles for a given component using just the mixin and modify the colors of the focus styles contextually when needed (particularly useful for the `.s-topbar__light` and `.s-topbar__dark` variants).

> **Note**
> This PR incidentally addresses a minor bug in the `s-btn` and `s-pagination` components. Previously, the themed focus ring rendered as 3px and this PR updates that so it renders 2px as expected.

### DIFF
--- a/lib/components/topbar/topbar.less
+++ b/lib/components/topbar/topbar.less
@@ -5,9 +5,6 @@
 }
 
 .s-topbar {
-    --_tb-focus-color-inner: var(--white);
-    --_tb-focus-color-outer: var(--theme-secondary-400);
-
     min-width: auto;
     width: 100%;
     z-index: var(--zi-navigation-fixed);
@@ -42,50 +39,39 @@
 
     // Overrides for focus style colors in forced light variant
     &&__light {
-        --_tb-focus-color-inner: .set-white()[default]; // forces white for inner focus ring color
+        --focus-neutral: .set-white()[default]; // forces white for inner focus ring color
     }
 
     .dark-mode({
         &.s-topbar__light {
-            --_tb-focus-color-outer: var(--theme-dark-secondary-custom-200, .set-theme-secondary-default()[200]);
+            --focus-theme: var(--theme-dark-secondary-custom-200, .set-theme-secondary-default()[200]);
         }
     });
 
     .highcontrast-dark-mode({
         &.s-topbar__light {
-            --_tb-focus-color-outer: .set-theme-secondary-default()[200];
+            --focus-theme: .set-theme-secondary-default()[200];
         }
     });
 
     // Overrides for focus style colors in forced dark variant
     .highcontrast-mode({
         &.s-topbar__dark {
-            --_tb-focus-color-outer: .theme-dark-default()[secondary];
+            --focus-theme: .theme-dark-default()[secondary];
         }
     });
 
     &&__dark {
-        --_tb-focus-color-inner: .set-black()[600]; // set to match .s-topbar__dark --theme-topbar-background-color;
-        --_tb-focus-color-outer: var(--theme-dark-secondary-custom-400, .theme-dark-default()[secondary]);
+        --focus-neutral: .set-black()[600]; // set to match .s-topbar__dark --theme-topbar-background-color;
+        --focus-theme: var(--theme-dark-secondary-custom-400, .theme-dark-default()[secondary]);
     }
 
     // focus styles
     a&--logo,
     & &--content &--item:not(&--item__unset),
-    &--notice,
-    .s-navigation .s-navigation--item {
+    &--notice {
         &:focus-visible {
-            box-shadow: inset 0 0 0 var(--su-static2) var(--_tb-focus-color-outer), inset 0 0 0 var(--su-static4) var(--_tb-focus-color-inner);
-            outline: var(--su-static2) solid transparent;
-        }
-    }
-
-    // TODO investigate removing once .s-input and .s-select have updated focus styles
-    & .s-topbar--searchbar .s-topbar--searchbar--input-group .s-input,
-    & .s-topbar--searchbar .s-select > select {
-        &:focus {
-            box-shadow: 0 0 0 var(--su-static2) var(--_tb-focus-color-inner), 0 0 0 var(--su-static4) var(--_tb-focus-color-outer) !important;
-            outline: var(--su-static2) solid transparent !important;
+            .focus-styles(true);
         }
     }
 
--- a/lib/components/topbar/topbar.less
+++ b/lib/components/topbar/topbar.less
@@ -83,8 +83,7 @@
     // TODO investigate removing once .s-input and .s-select have updated focus styles
     & .s-topbar--searchbar .s-topbar--searchbar--input-group .s-input,
     & .s-topbar--searchbar .s-select > select {
-        &:focus-visible {
-            // border-color: var(--_tb-focus-color-outer) !important;
+        &:focus {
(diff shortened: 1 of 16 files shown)


lib/exports/mixins.less
```diff
@@ -110,6 +110,40 @@
...
+/**
+ * Focus styles for the given context.
+ *
+ * Usage example:
+ * .focus-styles(true, true);
+ *
+ * @inset: boolean - whether the focus style be placed inside the element.
+ * @border: boolean - whether the element's border color change to match the focus style.
+ */
+.focus-styles(@inset: false, @border: false) {
```
giamir, line 122: When I am in a component less file and I see `.focus-style(true, true)` I have no clue what true and true means and I would need to come here to figure this out. 
Something like this would be more readable in my opinion:
```less
.component {
    .focus-styles({
        inset: true;
        border: false;
    });
}
```

What do you think?

I believe it could be achieved with something like this but I have not tested it myself. 🙂

```less
.focus-styles(@options: {}) {
    @inset: extract(@options, inset);
    @border: extract(@options, border);
    
    ...
```

lib/exports/mixins.less
```diff
@@ -110,6 +110,40 @@
...
+    & when (@inset = false) {
```
giamir: No action required here but I wanted to mention that personally I find easier to read when there are less level of nesting. For example:
```less
    & when not (@inset) and not (@border) {
        box-shadow: 0 0 0 var(--su-static2) var(--focus-neutral), 0 0 0 var(--su-static4) var(--focus-theme);
    }

    & when not (@inset) and (@border) {
        border-color: var(--focus-neutral) !important;
        box-shadow: 0 0 0 var(--su-static1) var(--focus-neutral), 0 0 0 calc(var(--su-static4) - var(--su-static1)) var(--focus-theme);
    }

    & when (@inset) and not (@border) {
        box-shadow: inset 0 0 0 var(--su-static2) var(--focus-theme), inset 0 0 0 var(--su-static4) var(--focus-neutral);
    }

    & when (@inset) and (@border) {
        border-color: var(--focus-theme) !important;
        box-shadow: inset 0 0 0 var(--su-static1) var(--focus-theme), inset 0 0 0 calc(var(--su-static4) - var(--su-static1)) var(--focus-neutral);
    }
```
Wanted to double check what's your preference. I am happy to leave things as they are. 🙂

lib/components/topbar/topbar.less
```diff
@@ -42,51 +39,39 @@
...
     & &--content &--item:not(&--item__unset),
-    &--notice,
-    .s-navigation .s-navigation--item {
-        &:focus-visible {
-            box-shadow: inset 0 0 0 var(--su-static2) var(--_tb-focus-color-outer), inset 0 0 0 var(--su-static4) var(--_tb-focus-color-inner);
-            outline: var(--su-static2) solid transparent;
-        }
-    }
-
-    // TODO investigate removing once .s-input and .s-select have updated focus styles
```
giamir, line 83: Nice 🎉

giamir (approved):
Great work @dancormier. Everything works as expected. I would also add `s-select` to the list of components you mentioned in the PR description. 

I only left a couple of suggestions about code (Less) readability. Feel free to apply them or ignore them. They are not blocking this PR from being merged. 

Thanks 🙏

lib/exports/mixins.less
dancormier: > When I am in a component less file and I see `.focus-style(true, true)` I have no clue what true and true means and I would need to come here to figure this out. Something like this would be more readable in my opinion:
> 
> ```css-less
> .component {
>     .focus-styles({
>         inset: true;
>         border: false;
>     });
> }
> ```
> 
> What do you think?
> 
> I believe it could be achieved with something like this but I have not tested it myself. 🙂
> 
> ```css-less
> .focus-styles(@options: {}) {
>     @inset: extract(@options, inset);
>     @border: extract(@options, border);
>     
>     ...
> ```


I would like that, but Less does not support object literal arguments 🙁 I remember wanting this when I was creating all the color mixins and had to include one property per argument.

It tried your suggestion (and a few variations) in the Less Preview online playground:

```Less
.focus-styles(@options: {}) {
    @inset: extract(@options, inset);
    @border: extract(@options, border);

    & when (@inset = false) {
        color: red;
    }
    & when (@inset = true) {
        color: blue;
    }
    & when (@border = false) {
        border-color: yellow;
    }
    & when (@border = true) {
        border-color: green;
    }
}

.component {
    background: purple;

    .focus-styles({
        inset: false;
        border: false;
    });
}
```

and got the following output:

```css
.component {
  background: purple;
}
```

FWIW, I originally included arguments that accepted strings like `bordered/unbordered` and `inside/outside`, but I scrapped them in favor of booleans. Would you prefer something like that? I'm happy to explore it if you think that's better.

lib/exports/mixins.less
dancormier: Good call! I've made this change in https://github.com/StackExchange/Stacks/pull/1629/commits/1f6e0fa717b82a2f294df2e88f2fe33e5bdcea95

CGuindon (changes requested):
The double ring looks really good!

I noticed a few pattern inconsistencies with the input depending on the mode and variant. Unless there's an accessibility reason I'm unaware of, I think we can keep original the border line color when in focus. If we have a red border line for a danger input, when it's in focus, that border should stay red — even in HC mode. The outer ring is what will always be blue. This seems to be the case sometimes but in most HC cases the border color is changing to black or white or blue as well. I assume this might be leftover from the previous focus style?

In the examples below, the only "correct" one is Light Danger input — all the others have a border color that changes. Do we need to keep that change?

Focus inconsistencies

lib/exports/mixins.less
giamir: I think I am ok to stick to booleans. I guess Less is starting to show its age. 
For my curiosity if you would start today how would you recommend to accomplish this behavior? PlainCSS + PostCSS plugin?

dancormier:
Ah, good catch @CGuindon! Your assumption is correct, the changing border color is a remnant from the previous focus style. As of https://github.com/StackExchange/Stacks/pull/1629/commits/acc29780e03bbba126c83ab49b4b420d5e8068a0, the border color stays the same color in the focused and non-focused states in both HC and non-HC modes.

lib/exports/mixins.less
dancormier: That's a good question! My gut says we could do pretty well with PostCSS + a few plugins (probably for [nesting](https://www.npmjs.com/package/postcss-nesting), [Sass/Less-like variables](https://github.com/postcss/postcss-simple-vars), and [mixins](https://github.com/postcss/postcss-mixins)) but I haven't researched this much so I'm not 100% sure it's the way to go.

> PlainCSS + PostCSS

I'm guessing by PlainCSS, you intended plain as in _vanilla_ and not [PlainCSS](https://ilyadzh.github.io/plain-css), right?

CGuindon (approved):

lib/exports/mixins.less
giamir: Yes. vanilla.  Of course there is a PlainCSS framework. 😅

PR #1637, opened 2024-02-06
feat(focus): add atomic focus classes

dancormier:
This PR adds atomic focus style classes to the library. I think this is necessary in order to allow consumers to style elements that otherwise wouldn't receive the desired focus style.

https://deploy-preview-1637--stacks.netlify.app/product/base/interactivity/#focus


CGuindon:
How do I test this @dancormier ?

giamir (approved):
@dancormier The PR looks good to me. Great work!
I would recommend to take the occassion to add a less snapshot test for the misc file similarly to what we do for colors [here](https://github.com/StackExchange/Stacks/blob/develop/lib/atomic/color.less.test.ts). 

I was also wondering if we already know that we will have use cases for all the 4 classes in Core. Creating larger API areas = More maintenance efforts.  Anyway I am also happy to expose all 4 (actually 8 with the conditional one) at once.

dancormier:
> I was also wondering if we already know that we will have use cases for all the 4 classes in Core. Creating larger API areas = More maintenance efforts. Anyway I am also happy to expose all 4 (actually 8 with the conditional one) at once.

I _think_ these classes will be useful for custom built components within consumer libraries, but I'm not too sure of a concrete example. At first, I was assuming it it would be used for the post summaries that show up as search suggestion but I those can probably be handled with removing `.outline-none` and adding `.s-block-link` to those elements.

Eventually, I expect Stacks to handle `a[href]`, `button` and similar elements (or maybe just `*:focus` 🤔) which I'd expect would reduce or eliminate the value of these utility classes. @giamir Does this require a broader conversation (and a delay on merging)?

dancormier:
> How do I test this @dancormier ?

Sorry, I should have mentioned how this can be tested @CGuindon. You can see these atomic classes in action in the docs at https://deploy-preview-1637--stacks.netlify.app/product/base/interactivity/#focus

giamir:
> @giamir Does this require a broader conversation (and a delay on merging)?

I am a fan of adding things only when we have a real world use case for it. Predicting future needs is difficult so I will hold off from merging those classes until somebody in Core will present us with a use case (or we find one ourselves). 🙂

CGuindon (approved):

dancormier:
**Update**: I found a use case for these atomic classes! Currently, the editor [programmatically applies `bs-ring` to the editor DOM element](https://github.com/StackExchange/Stacks-Editor/blob/c8d9869abd5abfd7785f0f162c0958143d91928d/src/stacks-editor/editor.ts#L311) (by default, a `div`) when the editor is focused. Now that we have these new focus styles, we need to programmatically apply focus styling to that element but we currently do not have a convenient way to do so. Atomic focus classes would resolve this issue.

lib/atomic/misc.less.test.ts
```diff
@@ -0,0 +1,16 @@
+import { describe, it, expect } from "vitest";
+import { renderLess } from "../test/less-test-utils";
+
+describe("atomic: misc", () => {
+    it("should output all atomic css classes", async () => {
+        const css = await renderLess(`
+            @import (reference) "./lib/base/internal.less";
```
giamir: @dancormier I think that those `@import (reference)` if they are necessary to compile the file correctly they should be added to the source code and not to the test file (I am suggesting to do what we have done for colors [here](https://github.com/StackExchange/Stacks/blob/develop/lib/exports/color-mixins.less#L1-L2)). This gives helpful information to the reader of the source code highlighting that the file have dependencies on other less files and cannot be used/compiled in isolation. So essentially I would prefer if possible to follow the same pattern ES Modules are using to import other modules. Let me know if that makes sense to you.

giamir:
I am happy to merge this in if we have a use case now for them. I only had a minor comment around import statements in the less file. Thanks @dancormier 🙂

lib/atomic/misc.less.test.ts
dancormier: Yep, good call. Thanks for the suggestion

Resolved in https://github.com/StackExchange/Stacks/pull/1637/commits/e0ca9757fd015fda57c3bf2b197ad5f0eaee0395

PR #1638, opened 2024-02-07
fix(test): ensure `includeNullModifier` value is honored

dancormier:
After having visual failures of _[test(select): add a11y, visual tests #1498](https://github.com/StackExchange/Stacks/pull/1498)_ after it had been updated with the latest test changes in develop, I realized something was not working as expected.

Turns out that we're not fully honoring the `includeNullModifier` option for secondary and global modifiers (this bug predates the recent test changes). This PR fixes that bug.


giamir (approved):
LGTM. Thanks 🙏

**Sidenote**
I might be going a bit meta here but this specific test utils is getting so complex that might deserve its own set of unit tests. 😅

PR #1640, opened 2024-02-09
fix(button): ensure visited link button renders legibly

dancormier:
In the course of reviewing https://github.com/StackEng/StackOverflow/pull/18700#issuecomment-1936535576, a bug was discovered where a visited link with `.s-btn` would be illegible in HC mode. This was due to a selector being mis-scoped in  https://github.com/StackExchange/Stacks/pull/1613. This PR resolves that mistake.

> Questions page sort button text color doesn't change when selected/visited in HC mode
questions_sort

**Sidenote:** @giamir This is a situation where visual regression tests for interaction states may have caught the bug before it was shipped. A nice real-world case that proves the value of interaction state visual regression test.

### DIFF
--- a/docs/product/components/button-groups.html
+++ b/docs/product/components/button-groups.html
@@ -61,27 +61,27 @@ description: Button groups are a collection of buttons. This component is used i
     <div class="stacks-preview">
 {% highlight html %}
 <div class="s-btn-group">
-    <button class="s-btn s-btn__muted s-btn__outlined" role="button">Newest</button>
-    <button class="s-btn s-btn__muted s-btn__outlined" role="button">Frequent</button>
+    <button class="s-btn s-btn__muted s-btn__outlined" type="button">Newest</button>
+    <button class="s-btn s-btn__muted s-btn__outlined" type="button">Frequent</button>
     <form>
-        <button class="s-btn s-btn__muted s-btn__outlined is-selected" role="button" aria-current="true">Votes</button>
+        <button class="s-btn s-btn__muted s-btn__outlined is-selected" type="button" aria-current="true">Votes</button>
     </form>
-    <button class="s-btn s-btn__muted s-btn__outlined" role="button">Active</button>
+    <button class="s-btn s-btn__muted s-btn__outlined" type="button">Active</button>
     <form>
-        <button class="s-btn s-btn__muted s-btn__outlined" role="button">Unanswered</button>
+        <button class="s-btn s-btn__muted s-btn__outlined" type="button">Unanswered</button>
     </form>
 </div>
 {% endhighlight %}
         <div class="stacks-preview--example">
             <div class="s-btn-group">
-                <button class="s-btn s-btn__muted s-btn__outlined" role="button">Newest</button>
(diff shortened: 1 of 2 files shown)


giamir (approved):
@dancormier Thanks for this fix. Unfortunately I had a hard time manually testing this change. Initially I did not understand why we were using the `:visited` pseudoclass on button. It looks like this is for button groups with anchor elements (instead of actual button elements). I took the liberty to create a small extra section in the docs about this use case where the s-btn is an `a` tag so that I could test easily.

> Sidenote: @giamir This is a situation where visual regression tests for interaction states may have caught the bug before it was shipped. A nice real-world case that proves the value of interaction state visual regression test.

Yes, tests would also be nice as a follow up steps but I think in this specific instance we were missing also the docs describing the use case. 

**Sidenote:**

There is a small style bug when we wrap the first or the last button group items in a `form` element (see screenshot).

Screenshot 2024-02-12 at 11 00 25

This was already there before but you might want to consider creating another small PR to address the issue or log a bug ticket.
I was also confused about the use case for wrapping button groups items in a `form` element and I also find the `role="button"` applied to many examples redundant. What am I missing? 🤔

dancormier:
> @dancormier Thanks for this fix. Unfortunately I had a hard time manually testing this change. Initially I did not understand why we were using the `:visited` pseudoclass on button. It looks like this is for button groups with anchor elements (instead of actual button elements). I took the liberty to create a small extra section in the docs about this use case where the s-btn is an `a` tag so that I could test easily.

@giamir Sorry for the difficulty testing. Thank you for reviewing and adding the anchor-based example.

> > Sidenote: @giamir This is a situation where visual regression tests for interaction states may have caught the bug before it was shipped. A nice real-world case that proves the value of interaction state visual regression test.
> 
> Yes, tests would also be nice as a follow up steps but I think in this specific instance we were missing also the docs describing the use case.
> 
> **Sidenote:**
> 
> There is a small style bug when we wrap the first or the last button group items in a `form` element (see screenshot).
> Screenshot 2024-02-12 at 11 00 25

Yep, that's a long-standing docs bug. I did a quick scan of Core to see if there's an instance of use applying form elements like this and I didn't find any. think we could safely remove the child `form` elements from the `s-btn-group` examples. Do you agree @giamir?

> This was already there before but you might want to consider creating another small PR to address the issue or log a bug ticket. I was also confused about the use case for wrapping button groups items in a `form` element and I also find the `role="button"` applied to many examples redundant. What am I missing? 🤔

That's cruft. It should instead be `type="button"`. From [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#description):

> Adding role="button" tells the screen reader the element is a button, but provides no button functionality. Use [<button>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) or [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) with type="button" instead.

I'll take this branch as an opportunity to make that change.

giamir:
> Do you agree @giamir?

Yes let's remove the `form` elements from the examples. Thank you.

PR #1646, opened 2024-02-12
fix(form-elements): add focus-within focus styling

dancormier:
In the process of reviewing https://github.com/StackEng/StackOverflow/pull/18700, we identified an issue with non-focusable elements styled to look like inputs with `.s-input` ([tag editor input for example](https://github.com/StackEng/StackOverflow/pull/18700#issuecomment-1936407723)). These elements contain invisible or restyled input elements. We want to ensure that the elements with `.s-input` (or similar) receive focus styling when child elements are focused.

**Related**: https://github.com/StackExchange/Stacks/pull/1629

> **Note**
> In this PR, I added `:focus-within` styling to all Stacks component form elements _except_ `s-uploader`. That element has a unique structure that doesn't necessitate handling `:focus-within`.

### DIFF
--- a/docs/product/components/inputs.html
+++ b/docs/product/components/inputs.html
@@ -341,8 +341,6 @@ description: Input elements are used to gather information from users.
 
     <div class="stacks-preview">
 {% highlight html %}
-<label class="s-label" for="tag-selector">Tags</label>
-<div class="s-input">
     <span>
         <span class="s-tag">
             svelte
--- a/docs/product/components/inputs.html
+++ b/docs/product/components/inputs.html
@@ -333,3 +333,64 @@ description: Input elements are used to gather information from users.
         </div>
     </div>
 </section>
+
+<section class="stacks-section">
+    {% header "h2", "Input fills" %}
+    <p class="stacks-copy">Input fills are used to visually connect input text boxes with related content.</p>
+
+    {% header "h3", "Prepended inputs" %}
+    <div class="stacks-preview">
+{% highlight html %}
+<div class="d-flex gy4 fd-column">
+    <label class="flex--item s-label" for="website-url">Website URL</label>
+    <div class="d-flex">
+        <div class="flex--item s-input-fill order-first">https://</div>
+        <div class="d-flex fl-grow1 ps-relative">
+            <input class="flex--item s-input blr0" type="text" id="website-url" placeholder="www.stackoverflow.com" />
+        </div>
+    </div>
+</div>
+{% endhighlight %}
+        <div class="stacks-preview--example">
+            <div class="d-flex gy4 fd-column">
+                <label class="flex--item s-label" for="website-url">Website URL</label>
+                <div class="d-flex">
+                    <div class="flex--item s-input-fill order-first">https://</div>
+                    <div class="d-flex fl-grow1 ps-relative">
+                        <input class="flex--item s-input blr0" type="text" id="website-url" placeholder="www.stackoverflow.com" />
+                    </div>
+                </div>
+            </div>
+        </div>
+    </div>
+
+{% header "h2", "Nested inputs" %}
+<p class="stacks-copy">An input can be nested within a container that has the <code class="stacks-code">.s-input</code> class applied to display styled elements as if they're within an input.</p>
+
+<div class="stacks-preview">
+{% highlight html %}
+<label class="s-label" for="tag-selector">Tags</label>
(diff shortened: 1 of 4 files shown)


giamir (commented):
Based on MDN definition _the :focus-within [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS) [pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes) represents an element that is itself matched by the [:focus](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus) pseudo-class or has a descendant that is matched by :focus_. If I understand that correctly we can simply apply `:focus-within` and that would also take care of `:focus`.

Overall the changes are pretty simple and I am ok to merge them. I probably don't fully understand the creative ways we use those classes in Core though. See sidenote for more details. 🙂


**Sidenote**
Do we incentivize the use of `s-input`, `s-checkbox`, etc... classes to elements that are not in fact `input`, `input type="checkbox"`, etc...?
If we do so we should have at least a use case example in the docs in my opinion documenting when that makes sense.

I am asking because an `input type="checkbox"` under regular circumstances would not have any focusable descendants and therefore there would not be a need to use `:focus-within` - In fact it would likely confuse the average reader unless there is at least a comment explaining why we use that over `:focus`.

dancormier:
> **Sidenote** Do we incentivize the use of `s-input`, `s-checkbox`, etc... classes to elements that are not in fact `input`, `input type="checkbox"`, etc...? If we do so we should have at least a use case example in the docs in my opinion documenting when that makes sense.
> 
> I am asking because an `input type="checkbox"` under regular circumstances would not have any focusable descendants and therefore there would not be a need to use `:focus-within` - In fact it would likely confuse the average reader unless there is at least a comment explaining why we use that over `:focus`.

This is more of a desire path situation. I wouldn't say that we explicitly incentivize the pattern of wrapping inputs in parents that include the input component class, but we have a few instances of it in Core (an example could be found on the [question ask page](https://stackoverflow.com/questions/ask)).

image

```html
<div class="js-tag-editor tag-editor multi-line s-input">
    <span>
        <span class="s-tag rendered-element">java<button class="s-tag--dismiss baw0 js-delete-tag" type="button" title="Remove tag"><svg …></svg></button></span>
    </span>
    <input type="text" id="tageditor-replacing-tagnames--input" class="s-input js-tageditor-replacing" …>
</div>
```

We could consider making this more specific and only applying the focus styles to `:focus-within` for `s-input` (and maybe `s-textarea`) instead of all the inputs. We could also consider adding an example tag input in the docs to show the recommended structure. @giamir What do you think? I'm happy to modify this PR however you recommend.

giamir:
>  @giamir What do you think? I'm happy to modify this PR however you recommend.

I would say to keep the `:focus-witihin` only for the elements we know that there is a use case in Core and for those ones add a small blurp in the docs (or at a minimum a comment in the less code). Thank you 😊

dancormier:
@giamir I've removed the `:focus-within` styling from all but the `s-input` and `s-textarea` components. I also added some [user-facing documentation](https://deploy-preview-1646--stacks.netlify.app/product/components/inputs/#nested-inputs) and a comment in the less code in https://github.com/StackExchange/Stacks/pull/1646/commits/87f10c41e4e139c17a65ff5bf83b58b78cbfa780

docs/product/components/inputs.html
```diff
@@ -333,3 +333,64 @@
         </div>
     </div>
 </section>
+
+<section class="stacks-section">
+    {% header "h2", "Input fills" %}
```
giamir: This is probably a copy/paste error. We don't need to repeat this section. :)

giamir (commented):
Thanks @dancormier for improving the docs with this use case. Apart from a small copy/paste error everything looks good to me.

I also noticed that the example doesn't look right in dark mode (see screenshot). I believe we could apply some atomic classes to make sure the page stays in harmony.

Screenshot 2024-02-21 at 08 59 48



Thank you. 🙂

docs/product/components/inputs.html
dancormier: D'oh! Good catch. Addressed in https://github.com/StackExchange/Stacks/pull/1646/commits/27f84808d8d9ef27ca7f1ed4f5675a5061c24f55

giamir (approved):
Looks good to go now. Thanks for addressing the comments. 🙂

PR #1652, opened 2024-02-20
Add `.pb0` definition

### LINKED ISSUE #1653: Responsive class `.pb0` does not work
Noticed `sm:pb0` etc does not work. Looks like an oversight in setting up the classes

https://github.com/StackExchange/Stacks/blob/64f0bf227e4bbcc3edf50e11fdbfce87df81f569/lib/atomic/spacing.less#L210-L224

abovedave:
Adds a definition for responsive class `pb0`. Closes #1653

### DIFF
--- a/lib/atomic/spacing.less
+++ b/lib/atomic/spacing.less
@@ -214,6 +214,8 @@ body {
     1 2 4 6 8 12 16 24 32 48 64 96 128
 );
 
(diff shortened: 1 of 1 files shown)


dancormier:
@abovedave thanks for bring this bug to our attention!

We addressed it (as well as other atomic spacing class omissions) in https://github.com/StackExchange/Stacks/pull/1657, which was just merged and should go out on this upcoming Wednesday.

PR #1670, opened 2024-03-06
docs(a11y): add accessibility docs page

dancormier:
[STACKS-472](https://stackoverflow.atlassian.net/browse/STACKS-472)
This PR adds an accessibility page to our docs site.

> [!TIP]
> ~~In the process of creating this docs page, I made a few QoL improvements, most notably with the addition of an 11ty `stacks-markdown` plugin so we can write our docs pages using markdown when we see fit (this is distinct from our `markdown` plugin since it is intended accept raw markdown to output Stacks docs markup). There are also minor refactors of some partials and the heading plugin (`header.js`).~~
> These changes have been moved to [dcormier/docs-markdown-support branch](https://github.com/StackExchange/Stacks/tree/dcormier/docs-markdown-support)

> [!NOTE]
> I omitted the "known issues" section shown in the [figma mock](https://www.figma.com/file/3OlPL6x0chNbMIPjP2vxas/Stack-Overflow-3.0---Site-Modernization?node-id=3747%3A15709&mode=dev) since it only contained an entry for the button group component which was only relevant to the previous button group style.

---

Question for @CGuindon: Should we make any changes to the [Color fundamentals > Accessibility standards](https://deploy-preview-1670--stacks.netlify.app/product/base/color-fundamentals/#accessibility-standards) section?

---

### TODO

- [x] Update left nav
    - [x] See the [figma spec](https://www.figma.com/file/3OlPL6x0chNbMIPjP2vxas/Stack-Overflow-3.0---Site-Modernization?node-id=3747%3A15709&mode=dev)
    - [x] Move "Current color" to `BASE` section


docs/plugins/stacks-markdown.js
dancormier, line 1: This file is a new plugin for Stacks documentation markdown. It converts the markdown input to HTML, modifying headings, paragraphs, and code elements with our standard Stacks documentation classes.

> [!NOTE]
> This plugin includes an option for `inline` markdown. This results in the `p` tags being replaced with `span`s so text can be placed in whatever container is most appropriate (usually a parent `p.stacks-copy` as seen in [accessibility.html#L128-L132](https://github.com/StackExchange/Stacks/blob/STACKS-472/docs-a11y-page/docs/product/guidelines/accessibility.html#L128-L132)), allowing us to use other common markdown syntax for things like links and inline code without the need for cumbersome markup.

docs/_includes/accessibility/comparison.html
dancormier, line 1: This is the markup for "comparison" blocks like this:

image

docs/product/base/color-fundamentals.html
dancormier, line 1: This file contains two types of changes:

- Rename `color-row-detail.html` to `example-row-detail.html`
- Update markup to use and support the newly generalize `color-swatch.html` partial

CGuindon:
@dancormier Yup! I have a note in the ticket (last bullet) but basically remove the [entire](https://deploy-preview-1670--stacks.netlify.app/product/base/color-fundamentals/#accessibility-standards) section from that page since it will now live on the new page. 

I also have some left nav hierarchy changes mentioned in the design specs — open to feedback on these.

CGuindon (changes requested):
Did a quick visual pass over:

1. The link is correct but I just realized our copy had a typo, the HC link should for " Success Criterion 1.4.3 Contrast (Enhanced)" should "1.4.**6**"

Screenshot 2024-03-15 at 1 58 25 PM
Screenshot 2024-03-15 at 1 57 56 PM

---

2. Any chance we can easily get the color swatches to be vertically aligned in the middle? If it's too annoying, I can live with it. 
Screenshot 2024-03-15 at 2 06 18 PM

---

3. I noticed the little text labels on the images disappear when I re-size the screen to small, is that on purpose/normal?
Screenshot 2024-03-15 at 2 08 01 PM

---

4. Left nav changes and moving the Current Color section

docs/product/guidelines/accessibility.html
```diff
@@ -0,0 +1,359 @@
...
+        <button class="s-btn s-btn__filled" role="presentation">Button</button>
+        <div class="bar-pill ta-center px8 py2 bg-green-400 fc-white">Robotics</div>
+    {% endcapture %}
+
+    {% capture incorrectElement %}
+        <button class="s-btn s-btn__filled bg-theme-secondary-200" role="presentation">Button</button>
+        <div class="bar-pill ta-center px8 py2 bg-green-400 fc-green-300">Robotics</div>
+    {% endcapture %}
+
+    {% render 'accessibility/comparison.html',
```
giamir: **Sidenote (No action required)**

I look at something like this and I cannot stop but thinking how much more elegant/readable would be to use a Svelte component. This gives me the Razor vibe 😀

docs/plugins/stacks-markdown.js
giamir: I am ok to merge this in but not having tests certainly makes me a bit uncomfortable about it. Especially because I see how this converter might need to evolve/change over time.

giamir: Have you considered treating everything as markdown without having to add a dedicated shortcode?
Could we process all the markdown in a view via the converter by default?
Maybe we could start by adding a boolean in the front matter if we are afraid of side effects in other pages.

I am suggesting this because I don't particularly like constantly opening and closing shortcodes like this `{% stacks-markdown %} ... {% endstacks-markdown %}`. They also make syntax highlighting pretty poor compared to how it was before with just plain html (which had the problem of being cluttered with a lot of repeated tags and classes).

giamir (commented):
Great work Dan. Looks good overall. 🎉

I am happy to merge it in although I am not entirely sold on the `stacks-markdown` plugin as it is now. I understand that it will save us from repeating classes and tags over and over but I dislike that we need to open and close shortcodes and we lose syntax highlighting in the process. I made a suggestion about exploring to run the converter you wrote against the whole markdown found in the view file without the usage of shortcodes (see review comment).

Also in my opinion we should make an effort to test that converter even if it's just for our docs. I can see how we want to change/add things there over time and having test cases will help us to do that with the confidence of not breaking things. Let me know if you need support setting those tests up (I would recommend to use vitest if you want to give it a shot - e.g. `"test:docs:unit": "vitest path/to/docs"`) 🙂

docs/product/guidelines/accessibility.html
dancormier: I absolutely agree. I would love to rearchitect our docs to use SvelteKit for reasons like that one (and many more)

dancormier:
> Great work Dan. Looks good overall. 🎉
> 
> I am happy to merge it in although I am not entirely sold on the `stacks-markdown` plugin as it is now. I understand that it will save us from repeating classes and tags over and over but I dislike that we need to open and close shortcodes and we lose syntax highlighting in the process. I made a suggestion about exploring to run the converter you wrote against the whole markdown found in the view file without the usage of shortcodes (see review comment).
> 
> Also in my opinion we should make an effort to test that converter even if it's just for our docs. I can see how we want to change/add things there over time and having test cases will help us to do that with the confidence of not breaking things. Let me know if you need support setting those tests up (I would recommend to use vitest if you want to give it a shot - e.g. `"test:docs:unit": "vitest path/to/docs"`) 🙂

You're the voice of reason here, @giamir. I've removed the `stacks-markdown` from this PR (in commit https://github.com/StackExchange/Stacks/pull/1670/commits/46e0cdbfdeac6d570625ef084a0b709c86938a10) so we can keep this PR scoped reasonably and get the accessibility page merged without hesitation.

FWIW, I think you're right to not be totally sold on the `stacks-markdown` plugin. I think we can explore it further when the time is right. For now, I've ported the plugin to the [dcormier/docs-markdown-support branch](https://github.com/StackExchange/Stacks/tree/dcormier/docs-markdown-support) so we have the code if/when we feel like revisiting.

Let me know if this seems reasonable.

dancormier:
Thanks for the review @CGuindon. I've made changes that should address the issues you've found.

> Did a quick visual pass over:
> 
> 1. The link is correct but I just realized our copy had a typo, the HC link should for " Success Criterion 1.4.3 Contrast (Enhanced)" should "1.4.**6**"
> 
> Screenshot 2024-03-15 at 1 58 25 PM Screenshot 2024-03-15 at 1 57 56 PM

✅ Fixed in https://github.com/StackExchange/Stacks/pull/1670/commits/04c76bc9022d1b52c1d414e0f2f0ebdf7583c6b9

---

> 2. Any chance we can easily get the color swatches to be vertically aligned in the middle? If it's too annoying, I can live with it.
> 
> 
> Screenshot 2024-03-15 at 2 06 18 PM

✅ Fixed in https://github.com/StackExchange/Stacks/pull/1670/commits/55ee4694d351e9c2511d75fb873cc9f064597ea2

---

> 3. I noticed the little text labels on the images disappear when I re-size the screen to small, is that on purpose/normal?
> 
> 
> Screenshot 2024-03-15 at 2 08 01 PM

The disappearing labels are a remnant from the Color Fundamentals page build. Since the color rows in [the Usage section](https://stackoverflow.design/product/base/color-fundamentals/#backgrounds) shift to a vertical orientation in smaller viewports, it didn't make sense to show those labels in that partial. It also just kinda helped sidestep having to deal with ensuring the examples wrapped/flowed gracefully.

image

image

I've changed the partial to show those labels at all screen sizes unless we override it in https://github.com/StackExchange/Stacks/pull/1670/commits/087774b38a93c2a92dc5f88e366a896386fbc681. Let me know how this looks to you. I'm happy to make modifications if you think we should.

---

> 4. Left nav changes and moving the Current Color section

✅ Completed in https://github.com/StackExchange/Stacks/pull/1670/commits/e51495d3f7723d06d48ffea15290366e2424bc49

giamir:
> FWIW, I think you're right to not be totally sold on the stacks-markdown plugin. I think we can explore it further when the time is right. For now, I've ported the plugin to the [dcormier/docs-markdown-support branch](https://github.com/StackExchange/Stacks/tree/dcormier/docs-markdown-support) so we have the code if/when we feel like revisiting.

It works for me. Maybe you can also create a quick companion "operational maintenance" jira ticket if it makes sense (it will probably end up in the low priority pile but at least we don't lose track of the improvement). Thank you. 😊

giamir (approved):

giamir:
@dancormier @CGuindon Is the Accessibility navigation item under the correct category?
https://deploy-preview-1670--stacks.netlify.app/product/guidelines/accessibility/

Screenshot 2024-03-20 at 09 53 58

dancormier:
> @dancormier @CGuindon Is the Accessibility navigation item under the correct category? https://deploy-preview-1670--stacks.netlify.app/product/guidelines/accessibility/

Yep, I originally missed the reorganization of that navigation in the [Figma](https://www.figma.com/file/3OlPL6x0chNbMIPjP2vxas/Stack-Overflow-3.0---Site-Modernization?node-id=3522%3A4090&mode=dev) and made those changes in https://github.com/StackExchange/Stacks/commit/e51495d3f7723d06d48ffea15290366e2424bc49

CGuindon:
> > @dancormier @CGuindon Is the Accessibility navigation item under the correct category? https://deploy-preview-1670--stacks.netlify.app/product/guidelines/accessibility/
> 
> Yep, I originally missed the reorganization of that navigation in the [Figma](https://www.figma.com/file/3OlPL6x0chNbMIPjP2vxas/Stack-Overflow-3.0---Site-Modernization?node-id=3522%3A4090&mode=dev) and made those changes in [e51495d](https://github.com/StackExchange/Stacks/commit/e51495d3f7723d06d48ffea15290366e2424bc49)

@giamir @dancormier I also wanted to change the name of that top section to "Develop" — since everything in here is a _Guideline_. Thoughts or concerns?

Screenshot 2024-03-20 at 3 58 07 PM

CGuindon (changes requested):
Commented on this [commit](https://github.com/StackExchange/Stacks/commit/55ee4694d351e9c2511d75fb873cc9f064597ea2) directly since I can't push any changes (permission thing I think?). Otherwise the other changes look good.

dancormier:
### [Color swatch bottom margin](https://github.com/StackExchange/Stacks/pull/1670#pullrequestreview-1950425284)

@CGuindon Updated!

> Commented on this [commit](https://github.com/StackExchange/Stacks/commit/55ee4694d351e9c2511d75fb873cc9f064597ea2) directly since I can't push any changes (permission thing I think?). Otherwise the other changes look good.

### [Docs navigation changes](https://github.com/StackExchange/Stacks/pull/1670#issuecomment-2010840773)

> I also wanted to change the name of that top section to "Develop" — since everything in here is a Guideline. Thoughts or concerns?

@CGuindon I think that'll work fine. I originally changed the wrong heading but have fixed it in https://github.com/StackExchange/Stacks/pull/1670/commits/d092389d26a14810a4e2fa89070c2159dd723641

I made [another PR](https://github.com/StackExchange/Stacks/pull/1684) to update urls to better reflect the nav structure. This isn't _entirely_ necessary but it feels like a good excuse to perform that cleanup. @giamir @CGuindon when you get a moment, let me know what you think of that PR and I'll either merge it into this PR or close it out.

CGuindon (approved):

PR #1671, opened 2024-03-08
fix(button): ensure button gains bg color when focused

dancormier:
This PR ensures that the button gains a background color when focused as expected.

[🔗 Button component in deploy preview](https://deploy-preview-1671--stacks.netlify.app/product/components/buttons/)

---

In the process of building the accessibility docs page, I noticed that button variants with no modifiers were not showing the intended background color when focused:

### Expectation

image
image
image

### Reality

image
image
image

### DIFF
--- a/lib/components/button/button.less
+++ b/lib/components/button/button.less
@@ -411,8 +411,6 @@
     &:not(&--radio:checked + label):not(&__link):not(&__unset):not(&__facebook):not(&__github):not(&__google):not(.is-selected) {
         &:focus-visible {
             &.s-btn__filled {
-                &:not(:hover) .s-btn--number {
-        &:focus-visible,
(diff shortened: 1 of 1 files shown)


giamir (approved):
Nice catch. 🎉

**Sidenote**
It would be nice to have those interactive state tests. Do we already have a tech debt ticket for exploring introducing those? If not we should create one. 🙂
I think that if we add them in a separate file (e.g. `button.interactive.visual.test.ts`) we won't have a big increase in the time it takes to run them.

dancormier:
@giamir thanks for the speedy review! I've added [an issue](https://stackoverflow.atlassian.net/browse/STACKS-578) for testing interaction states as you suggested

PR #1677, opened 2024-03-11
fix(notice): improve contrast of inline code

hamvocke:
The `s-notice` component currently has insufficient color contrast when used in combination with inline `code` elements. Here's what it looks like in dark mode, for example:

image

Screenshot 2024-03-11 091710

I'm opening this PR to give us a playground to discuss and fix this problem. I tried to play around with various color combinations for the background and text colors myself but quickly realized that I've got no idea what we're aiming for from a visual perspective. I could tweak the colors until I hit the right color contrast, obviously, but I really need some guidance on the visual design, too.


CGuindon:
I had [previously](https://stackoverflow.atlassian.net/browse/STACKS-488) figured out the color combos needed for the comment box — so these should work for the blue notice. To do  the other colors, we'll some design time and that's not going to happen until the end of this quarter (we have bug bash I can assign this to someone). You can also verify if the the other colors work with the same color stops but I bet there will be issues with red and yellow.

Screenshot 2024-03-11 at 1 59 40 PM

dancormier:
Merging this since the changes here are really just from https://github.com/StackExchange/Stacks/pull/1678 which has been reviewed and approved. 🚢

PR #1664, opened 2024-03-01
chore(colors): remove v1 color output and references

dancormier:
[STACKS-354](https://stackoverflow.atlassian.net/browse/STACKS-354)

This PR removes all references and output of all legacy v1 colors, with the exception of the migration guide.

@CGuindon added as reviewer for visibility

PR #1684, opened 2024-03-21
docs(navigation): remap paths to reflect nav sections

dancormier:
This PR updates the paths for `/product/*` urls to reflect the nav group they're in (with the exception of `v1` docs pages, which [should be gone soon](https://github.com/StackExchange/Stacks/pull/1664)). I've included redirects in `netlify.toml` and have updated links within our docs.

> [!NOTE]
> The target branch is `STACKS/472-docs-a11y-page` ([PR](https://github.com/StackExchange/Stacks/pull/1670))

@CGuindon added as reviewer for visibility


netlify.toml
```diff
@@ -243,3 +243,51 @@
 [[redirects]]
   from = "/product/guidelines/releasing/"
   to = "https://github.com/StackExchange/Stacks#release-a-new-version-of-stacks"
+
+[[redirects]]
```
giamir, line 247: Duplicate redirect

lib/test/a11y-test-utils.ts
```diff
@@ -13,7 +13,7 @@ type A11yTestArgs = TestVariationArgs & {
 };
 
 // register Stack APCA conformance threshold function
-// see also https://stackoverflow.design/product/base/color-fundamentals/#accessibility-standards
+// see also https://stackoverflow.design/product/foundation/color-fundamentals/#accessibility-standards
```
giamir: This reference should point to the accessibility page now: `https://stackoverflow.design/product/foundation/accessibility/#visual-accessibility`

giamir (approved):
Looks good @dancormier. Those urls now make much more sense. 🎉
I have left a couple of NIT comments.

Also I would recommend you to do a CMD+F in the `netlify.toml` file for `product/guidelines/` and `/product/resources/`. There are a bunch of redirects that can be either mapped directly to the new `/develop/` `/foundation/` or they can simply be deleted. 

If you don't want to do that cleanup as part of this PR can we create a new Jira ticket to clean up in general the toml file. This should be pretty straightforward if we can look at the traffic logs in netlify.

Thank you.

CGuindon (approved):

PR #1685, opened 2024-03-21
feat(button): add vote button variant

dancormier:
This PR adds a `vote` variant of the button component to more tightly control and customize vote button styling via Stacks.

Deploy preview: https://deploy-preview-1685--stacks.netlify.app/product/components/buttons/#vote

### DIFF
--- a/docs/product/components/buttons.html
+++ b/docs/product/components/buttons.html
@@ -234,6 +234,92 @@ description: Buttons are user interface elements which allows users to take acti
 </section>
 
 <section class="stacks-section">
+    {% header "h2", "Vote" %}
+    <p class="stacks-copy">TODO describe</p>
+
+    <div class="stacks-preview">
+{% highlight html %}
+<button class="s-btn s-btn__vote" type="button">
+    @Svg.ArrowUp
+    <span class="v-visible-sr">Upvote</span>
+</button>
+<button class="s-btn s-btn__vote" type="button">
+    @Svg.ArrowDown
+    <span class="v-visible-sr">Downvote</span>
+</button>
+<button class="s-btn s-btn__vote is-selected" type="button">
+    @Svg.ArrowUp
+    <span class="v-visible-sr">Upvoted</span>
+</button>
+<button class="s-btn s-btn__vote is-selected" type="button">
+    @Svg.ArrowUp
+    <span class="v-visible-sr">Downvoted</span>
+</button>
+
+{% endhighlight %}
+        <div class="stacks-preview--example">
+            <div class="overflow-x-auto">
+                <table class="wmn5 s-table s-table__bx-simple">
+                    <thead>
+                        <tr>
+                            <th scope="col">Type</th>
+                            <th scope="col" class="s-table--cell4">Class</th>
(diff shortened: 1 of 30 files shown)


giamir:
@dancormier @CGuindon What do you think about "dog feed" this new button styles in our docs and replace the existing feedback buttons (see screenshot)? 
Screenshot 2024-03-22 at 13 11 58

giamir:
@CGuindon @dancormier I am not sure how to feel about the focus state changing the internal background color of the button with different colors depending if it is selected or not. 
Screenshot 2024-03-22 at 13 38 05
Screenshot 2024-03-22 at 13 38 23
Screenshot 2024-03-22 at 13 38 47
Screenshot 2024-03-22 at 13 38 59


Would not be simpler and more intuitive to treat this special vote button like we do toggles or tags? I mean just surrounding the element with an outline and leave the background colors unchanged.
Screenshot 2024-03-22 at 13 39 52
Screenshot 2024-03-22 at 13 39 44


I am thinking that if I struggle to understand which color should the vote button be when focused and I work in Stacks I can only imagine how difficult it would be for others to wrap their heads around it.

giamir:
@dancormier Could we remove the arrow icon button example in the icons section of the page? I think it could be confusing for people since now we have a dedicate vote variant.

Screenshot 2024-03-22 at 13 45 13

docs/product/components/buttons.html
```diff
@@ -233,6 +233,92 @@
     </div>
 </section>
 
+<section class="stacks-section">
+    {% header "h2", "Vote" %}
+    <p class="stacks-copy">TODO describe</p>
+
+    <div class="stacks-preview">
+{% highlight html %}
+<button class="s-btn s-btn__vote" type="button">
```
giamir, line 242: I think we should add the appropriate aria attributes to this example since it is used in most places as a toggle: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed

docs/product/components/buttons.html
```diff
@@ -233,6 +233,92 @@
...
+    @Svg.ArrowUp
+    <span class="v-visible-sr">Upvote</span>
```
giamir, line 244: Would it not be more elegant and shorter to use an `aria-label` in the button instead?

giamir:
@dancormier @CGuindon Is `vote` the right name for the variant? I think it would be best to describe the variant from a stylistic perspective instead of semantically. I am saying this because we use the same button for bookmarking already and I can imagine how it could be used for other purposes too in the future.
Screenshot 2024-03-22 at 13 59 33

`round` variant perhaps? I am sure we can come up with a good (domain-agnostic) name.

giamir:
@dancormier @CGuindon I think this is a good starting point to standardize the vote (or round or whatever 🙂) buttons in Stacks but I think it could benefit from us reflecting some more on our choices before shipping it. We certainly should not treat this work as "side" work especially because I imagine it will be quite painful to integrate and test this changes across different places in Core. I would recommend to plan this extraction work for one of our next sprints and address anything urgent in Core with a temporary fix. We can talk more in our planning today and it's totally ok if you disagree. 🙂

PR #1699, opened 2024-04-03
fix(checkbox): checked state forced color contrast

dancormier:
[STACKS-599](https://stackoverflow.atlassian.net/browse/STACKS-599)

---

This PR fixes the lack of contrast in the checked checkbox component when the user has enabled forced color mode.

### Before

stackoverflow design_product_components_checkbox_

### After

localhost_8080_product_components_checkbox_

### DIFF
--- a/lib/components/checkbox_radio/checkbox_radio.less
+++ b/lib/components/checkbox_radio/checkbox_radio.less
@@ -66,6 +66,17 @@
         }
     });
 
+    @media (forced-colors: active) {
+        &:checked {
+            --_ch-bg: ButtonText !important;
+            --_ch-bg-image: url("data:image/svg+xml;,%3Csvg width='11' height='11' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M10 3.41L8.59 2 4 6.59 2.41 5 1 6.41l3 3z' fill='ButtonFace'/%3E%3C/svg%3E") !important;
+        }
+        &:indeterminate {
+            --_ch-bg: ButtonText !important;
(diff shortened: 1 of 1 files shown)


giamir (changes requested):
Based on my testing in WHCM (Desert mode) we can achieve what we want without having to redefine the bg-image
This is sufficient:
```
   @media (forced-colors: active) {
        &:checked {
            --_ch-bg: ButtonText !important;
        }
        &:indeterminate {
            --_ch-bg: ButtonText !important;
        }
    }
```

I am happy to merge the simplified version above in. 

**Sidenote**
While I was in Windows I took some time to explore our DS in WHCM and I could see several components that are not rendering well (including toggles, tags, navigation not showing selected items, etc...).  We should probably create a rough inventory of those issues as part of the upcoming spike story.

dancormier:
> Based on my testing in WHCM (Desert mode) we can achieve what we want without having to redefine the bg-image This is sufficient:
> 
> ```
>    @media (forced-colors: active) {
>         &:checked {
>             --_ch-bg: ButtonText !important;
>         }
>         &:indeterminate {
>             --_ch-bg: ButtonText !important;
>         }
>     }
> ```
> 
> I am happy to merge the simplified version above in.

Without the background image redefinition, we'll have contrast issues if Stacks dark mode is enabled in addition to forced color mode.

image

I understand this would be an unusual circumstance but I figured it was a good idea to account for all possible cases. What do you think @giamir?

> **Sidenote** While I was in Windows I took some time to explore our DS in WHCM and I could see several components that are not rendering well (including toggles, tags, navigation not showing selected items, etc...). We should probably create a rough inventory of those issues as part of the upcoming spike story.

Yep, there's definitely room for improvement across our components. The [ticket](https://stackoverflow.atlassian.net/browse/STACKS-559) has "A rough inventory of user forced color issues in our components (e.g. toggles, tags, etc…) is added to the document" in the success criteria so I'll make sure to include a list of components (assuming I perform the spike sprint).

PR #1706, opened 2024-04-11
Refactor skip link to nest it within .s-topbar

dancormier:
This PR is an attempt to add the `.s-topbar--skip-link` component as a child of `.s-topbar` while retaining the style shown in https://github.com/StackExchange/Stacks/pull/1692.

@giamir I've added notes in a review of this code that should provide some context. I'm happy to discuss this PR with you whenever you'd like.

### DIFF
--- a/docs/_includes/layouts/page.html
+++ b/docs/_includes/layouts/page.html
@@ -9,7 +9,7 @@
     <div class="d-flex mx-auto w100 wmx12">
         {% include 'navigation.html' %}
 
-        <main id="content" class="d-flex fl-grow1 ps-relative pt96 pb24 pl48 md:pl24 sm:pl16 sm:pr16">
+        <main id="content" class="d-flex fl-grow1 ps-relative t64 py24 pl48 md:pl24 sm:pl16 sm:pr16">
             {% if hide-menu != true %}
                 <div class="flex--item order-last ml32 sm:d-none print:d-none" style="width: 14rem;">
                     <div class="ps-fixed t64 b0 py12 overflow-y-auto" style="width: 14rem;">
--- a/lib/components/topbar/topbar.less
+++ b/lib/components/topbar/topbar.less
@@ -407,7 +407,7 @@
         background-color: var(--theme-secondary-100);
         border-bottom: var(--_tb-bt);
         display: block;
-        outline: var(--su-static2) solid transparent;
+        outline: none;
         padding: var(--su12);
         text-align: center;
     }
--- a/lib/components/topbar/topbar.less
+++ b/lib/components/topbar/topbar.less
@@ -1,4 +1,5 @@
 .s-topbar {
+    --_tb-bt: var(--theme-topbar-accent-border, 3px solid var(--theme-primary));
     // CHILD COMPONENT CUSTOM PROPERTIES
     // Item
     --_tb-item-bg: unset;
@@ -121,6 +122,16 @@
     }
 
     // CHILD ELEMENTS
+    &:before {
+        border-top: var(--_tb-bt);
+        content: '';
+        display: block;
+    }
+
+    &:has(> &--skip-link:focus):before {
+        display: none; // Hide `:before` element if .s-topbar--skip-link is focused
+    }
+
     & a&--logo {
         &:focus-visible {
             .focus-styles(true);
@@ -135,7 +146,7 @@
     & &--container {
         align-items: center;
         display: flex;
-        height: 100%;
+        height: var(--theme-topbar-height, calc(var(--su-static48) + var(--su-static8)));
         margin: 0 auto;
         max-width: 100%;
         width: var(--s-full);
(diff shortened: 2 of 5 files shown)


lib/components/topbar/topbar.less
```diff
@@ -1,4 +1,5 @@
 .s-topbar {
+    --_tb-bt: var(--theme-topbar-accent-border, 3px solid var(--theme-primary));
```
dancormier, line 2: Since both `.s-topbar` and `.s-topbar--skip-link` use this style, I've abstracted it to a variable.

lib/components/topbar/topbar.less
```diff
@@ -135,7 +136,7 @@
     & &--container {
         align-items: center;
         display: flex;
-        height: 100%;
+        height: var(--theme-topbar-height, calc(var(--su-static48) + var(--su-static8)));
```
dancormier, line 139: The topbar height has been moved into the `.s-topbar--container` element. This allows the topbar to have flexible height and accommodate the skip link.

> [!NOTE]
> if we really wanted to handle this case, we could apply height to .s-topbar with `.s-topbar:not(:has(.s-topbar--container)) { height: …; }`, but I don't think it's worth doing this.

lib/components/topbar/topbar.less
```diff
@@ -398,6 +399,20 @@
         padding: var(--_tb-searchbar-p);
     }
 
+    & &--skip-link {
+        &:not(:focus) {
+            &:extend(.v-visible-sr);
+        }
```
dancormier, line 405: By applying the `.v-visible-sr` styles only when the skip link _is not_ focused, it removes the need to override these styles when the skip link is focused.

lib/components/topbar/topbar.less
```diff
@@ -424,13 +439,15 @@
         }
     }
 
+    &:has(> &--skip-link:focus) {
+        border-top: none; // Hide the border-top when the skip link is focused
+    }
+
```
dancormier, line 445: This is where the magic happens 🪄

Here, we're saying saying when `.s-topbar` includes a _focused_ direct child of `.s-topbar--skip-link`, remove the top border from `.s-topbar`. This works because the skip link includes a bottom border that takes the place of the topbar's top border.

This will be our first time using `:has`! (btw, I checked https://browsers.stackoverflow.design to make sure we can use it).

lib/components/topbar/topbar.less
```diff
@@ -398,6 +399,20 @@
         padding: var(--_tb-searchbar-p);
     }
 
+    & &--skip-link {
+        &:not(:focus) {
+            &:extend(.v-visible-sr);
+        }
+
+        background-color: var(--theme-secondary-100);
+        border-bottom: var(--_tb-bt);
```
dancormier, line 408: The skip link now includes a bottom border that replicates the top border on the topbar. When the skip link is focused, the topbar will remove it's top border in favor of this border. (see the next comment on line ~442)

lib/components/topbar/topbar.less
```diff
@@ -424,13 +439,15 @@
...
+    &:has(> &--skip-link:focus) {
+        border-top: none; // Hide the border-top when the skip link is focused
+    }
+
     // STATIC COMPONENT STYLES
     align-items: center;
     background-color: var(--theme-topbar-background-color, var(--white));
     border-bottom: var(--theme-topbar-bottom-border, var(--su-static1) solid var(--black-225));
-    border-top: var(--theme-topbar-accent-border, 3px solid var(--theme-primary));
-    display: flex;
```
dancormier, line 432: AFAICT, we don't need to set a `display: flex` on the topbar. If we want to play it safe, we could keep `display: flex` here and override it with `:has` (my new favorite trick!) to change the display in the event the topbar includes `--container` and `--skip-link`.

docs/_includes/layouts/page.html
```diff
@@ -9,7 +9,7 @@
     <div class="d-flex mx-auto w100 wmx12">
         {% include 'navigation.html' %}
 
-        <main id="content" class="d-flex fl-grow1 ps-relative pt96 pb24 pl48 md:pl24 sm:pl16 sm:pr16">
+        <main id="content" class="d-flex fl-grow1 ps-relative t64 py24 pl48 md:pl24 sm:pl16 sm:pr16">
```
giamir, line 12: @dancormier I will reintroduce this as it was changed. Please refer to this comment for more details: https://github.com/StackExchange/Stacks/pull/1692#discussion_r1561284307

lib/components/topbar/topbar.less
giamir: Smart. Great trick 🙂

lib/components/topbar/topbar.less
giamir: Makes sense

lib/components/topbar/topbar.less
giamir: That's pretty clever - I did not think about that. Nice work. 
Thank you so much for helping me out here. 🙂

PR #1723, opened 2024-04-29
fix(a11y): indicate required form controls using an asterisk

giamir:
[STACKS-582](https://stackoverflow.atlassian.net/browse/STACKS-582)

- [Input accessibility docs](https://deploy-preview-1723--stacks.netlify.app/product/components/inputs/#accessibility) (from @CGuindon's [draft](https://docs.google.com/document/d/1AHzVBvjiRgPY9p8NAq78HQUitwQr_GlLZ-kSy2IR_mc/edit))
- [Label status accessibility docs](https://deploy-preview-1723--stacks.netlify.app/product/components/labels/#status) (also from @CGuindon's draft)

@CGuindon @dancormier Shall we have a section where we document the new class a bit more formally like we do for validations classes?

Screenshot 2024-04-29 at 17 02 23

@dancormier I would like to get your feedback on the class naming `s-required-symbol` and location `label.less`. 
I did not prefix it with `s-label` since the class could be used elsewhere (e.g. in the legend explaining the abbreviation)

TODOs:
- [x] Make a decision about the name of the class (currently `s-required-symbol`)
- [x] Add visual and a11y test variant
- [x] Test screen reader behavior with NVDA in Windows


lib/components/label/label.less
```diff
@@ -96,6 +96,14 @@
     padding: 0 var(--su2); // Helps the label visually line up with inputs
 }
 
+// TODO: is this the right place for this?
+.s-required-symbol {
+    color: var(--red-400);
+    font-size: 125%;
+    font-weight: normal;
+    text-decoration: none !important;
+}
```
dancormier: I noticed a height difference in labels with and without the required symbol due to the font-size.

<details>
<summary>Gif example</summary>



</details>

I think we can prevent this layout shift while retaining the larger font size by setting line-height to 0 on `.s-required-symbol`.

```suggestion
.s-required-symbol {
    color: var(--red-400);
    font-size: 125%;
    font-weight: normal;
    line-height: 0;
    text-decoration: none !important;
}
```

docs/product/components/labels.html
```diff
@@ -93,7 +93,7 @@
 
 <section class="stacks-section">
     {% header "h2", "Status" %}
-    <p class="stacks-copy">When you need to flag labels as required or optional, use the following flags. Use the full word “Required” or “Optional” for these flags instead of using asterisks. If a majority of a form’s inputs are required, it isn’t necessary to show that they’re required until after the form has been submitted with errors. It may be more appropriate to only mark the non-required fields as optional.</p>
+    <p class="stacks-copy">When you need to flag labels as required or optional, use the following flags. Use the full word “Required” or “Optional” for these flags instead of using asterisks. If a majority of a form’s inputs are required, use the asterisk symbol to indicate which fields are required and include a legend. See <a href="/product/components/inputs/#accessibility">Input Accessibility</a> for more details.</p>
```
dancormier, line 96: I noticed the [Jira ticket](https://stackoverflow.atlassian.net/browse/STACKS-582) says:

> _Alternate approach_
> Add a required text next to the input label (not recommended unless the preferred approach is not possible)

This seems to slightly contradict the guidance added to the docs since the Jira ticket suggests to avoid the longhand "Required" label unless the `*` approach is not possible. 

@giamir what you added here seems more appropriate IMO, since it avoids the need for a legend when there are few required inputs. I wanted to call it out though to make sure we reconcile the discrepancy and so we don't ignore a reasoning for one approach over another.

---

@CGuindon any strong opinion on the recommendation on when to use `Required` vs `*`?

lib/components/label/label.less
```diff
@@ -96,6 +96,14 @@
     padding: 0 var(--su2); // Helps the label visually line up with inputs
 }
 
+// TODO: is this the right place for this?
```
dancormier: > I would like to get your feedback on the class naming s-required-symbol and location label.less.

I think this is a reasonable home for the required symbol styles. I _try_ to include each top-level component (i.e. any component that doesn't directly correspond to a parent component) in it's own less file, but I think this is a reasonable exception. If I consider where I'd think to look for this component, it would probably be here.

If for any reason we change our mind, it would be simple enough to relocate it, but I'm good with it here for now 🙂

---

> I did not prefix it with s-label since the class could be used elsewhere (e.g. in the legend explaining the abbreviation)

This also seems reasonable to me. No need to bind this directly to `.s-label`.

docs/product/components/inputs.html
```diff
@@ -71,6 +71,70 @@
...
+        <li>Enlarge the asterisk by <code class="stacks-code">125%</code> compared to the label text</li>
+        <li>Use <code class="stacks-code">red-400</code>
+            {%
+                render 'color-swatch.html',
+                color: "red-400",
+                size: "16",
+                classes: "mbn2"
+            %}
+            for the asterisk symbol</li>
+    </ul>
```
dancormier: When I first read this section, I felt like this was suggesting I apply a font-size and color to the asterisk _instead of_ apply a pre-built solution like the `.s-required-symbol` included in this PR.

> Shall we have a section where we document the new class a bit more formally like we do for validations classes?

Yes, I think that's a good call. I'd expect that to reduce confusion like I experienced. I'd also recommend mentioning the `.s-required-symbol` here as well so we encourage it's use. Maybe something like this added after we mention the styles:
> Stacks includes a `.s-required-symbol` component which includes the above styles.

What do you think @giamir?

dancormier (commented):
Great work @giamir. I think this PR shows an elegant and uncomplicated way to handle denoting required inputs. I have a few minor suggestions but nothing too intense.

lib/components/label/label.less
giamir: Nice catch @dancormier. Thank you

docs/product/components/inputs.html
giamir: I copied over from @CGuindon's [draft](https://docs.google.com/document/d/1AHzVBvjiRgPY9p8NAq78HQUitwQr_GlLZ-kSy2IR_mc/edit#bookmark=id.3jj27yq0o9py) without doing a thorough review of the copy. Will make sure to address your point and add the class detailed description. 

Thank you

giamir: I have adjusted the copy a bit for the inputs docs. Let me know if now it sounds better. 
https://deploy-preview-1723--stacks.netlify.app/product/components/inputs/#required-input-fields

docs/product/components/labels.html
CGuindon: Yup, that's older, I like what @giamir has here.

docs/product/components/inputs.html
CGuindon: LGTM! Thanks for the copy edits and review!

CGuindon (approved):

giamir:
@dancormier I have added the visual/a11y tests and took the occasion to refactor a bit. [You can find the diff here](https://github.com/StackExchange/Stacks/pull/1723/commits/ba4a5ee9f706429872e7eac3c42010fa79f67a0f). 
I would say that this PR is now ready to be merged after a final look from your end.

I have tested [the example](https://deploy-preview-1723--stacks.netlify.app/product/components/inputs/#required-input-fields-example) I put together in the inputs page in NVDA and this is what the narrator announces when I move from section to section with the arrow keys:
- Heading level 1 ask a question
- Required fields star
- Title star
- Edit type a title
- Body star
- Edit multiline type a question
- Ask team members
- Edit type a name

As you can see by default NVDA doesn't read the `title` attribute of the `abbr` tag. This seems to be expected behavior from the quick research I have done. There is a settings that users can change to have the narrator announcing the title attribute too. 
From a personal perspective I don't find the text announced optimal but the technique we are using is one recommended by WCAG so we should be good. Especially the `Required fields star` is a bit robotic. I can see how the text they use in the WCAG example would deliver a better experience for users using SRs: `Required fields are marked with an asterisk star`.
   Anyway I think what we have is an acceptable tradeoff that makes us WCAG compliant. Unless I get a red flag either from you or @CGuindon I am happy to proceed with the merge and integrate similar solution for the pages we know about in Core.

CGuindon:
I'm less familiar with screen reader norms but I trust you both and if this passes WCAG I think we can proceed. No red flags from me.

giamir:
@dancormier when you have a second give me the green flag here and I will merge. Thank you. 😊

dancormier (approved):

PR #1735, opened 2024-05-14
chore(links): update visited links colors and application

dancormier:
[STACK-481](https://stackoverflow.atlassian.net/browse/STACKS-481)

This PR updates the visited link style to `purple-500` (and `purple-600` when visited and hovering) where appropriate.

### Changes of note

- This PR changes the visited link style for _all_ anchors, with the exception anchors of Stacks components that should not ~~(see [this change](https://github.com/StackExchange/Stacks/pull/1735/files#diff-9cbc22c2132424697b77e86d7a6b8ab33b8e0b13f84670db8de0c846beaf1691R85-R88) for the list of unaffected elements)~~. I'll be testing this in Core, but would appreciate feedback on this approach in the meantime.
- I've applied styles based on the LVHA-order (`:link` — `:visited` — `:hover` — `:active`). See [the `:visited` MDN page](https://developer.mozilla.org/en-US/docs/Web/CSS/:visited) for more details

### Testing these changes

> [!TIP]
> The below seems to not work in incognito/private browser modes since (TIL) these modes do not honor visited link styles (at least in Mac Firefox and Chrome)

1. Visit the [link component docs page](https://deploy-preview-1735--stacks.netlify.app//product/components/links/#single-link-examples)
2. Observe the color of the "Visited" link (this one has forced visited style)
3. Go to `https://deploy-preview-1735--stacks.netlify.app//product/components/links/#` (note the `#` at the end. Observe that the appropriate links are now `purple-500`. Ensure this renders as expected in all modes.
4. Visit other components that include links to ensure they only gain this `purple-500` when expected

### DIFF
--- a/lib/components/breadcrumbs/breadcrumbs.less
+++ b/lib/components/breadcrumbs/breadcrumbs.less
@@ -25,11 +25,15 @@
         margin-top: var(--su2);
     }
 
-    & &--link {
-        color: var(--_br-link-fc);
-
+    & &--link,
+    & a&--link{
         &:hover {
             --_br-link-fc: var(--fc-medium);
+        }
+
+        &,
+        &:visited:not(:hover) {
+            color: var(--_br-link-fc);
         }
     }
 
--- a/lib/components/breadcrumbs/breadcrumbs.less
+++ b/lib/components/breadcrumbs/breadcrumbs.less
@@ -26,7 +26,7 @@
     }
 
     & &--link,
-    & a&--link{
+    & a&--link {
         &:hover {
             --_br-link-fc: var(--fc-medium);
         }
--- a/lib/components/breadcrumbs/breadcrumbs.less
+++ b/lib/components/breadcrumbs/breadcrumbs.less
@@ -25,16 +25,11 @@
         margin-top: var(--su2);
     }
 
-    & &--link,
-    & a&--link {
-        &:hover,
-        &:visited:hover {
-            --_br-link-fc: var(--fc-medium);
-        }
+    & &--link {
+        color: var(--_br-link-fc);
 
-        &,
-        &:visited {
-            color: var(--_br-link-fc);
+        &:hover{
+            --_br-link-fc: var(--fc-medium);
         }
     }
 
--- a/lib/components/breadcrumbs/breadcrumbs.less
+++ b/lib/components/breadcrumbs/breadcrumbs.less
@@ -27,12 +27,13 @@
 
     & &--link,
     & a&--link {
-        &:hover {
+        &:hover,
+        &:visited:hover {
             --_br-link-fc: var(--fc-medium);
         }
 
         &,
-        &:visited:not(:hover) {
+        &:visited {
             color: var(--_br-link-fc);
         }
     }
--- a/lib/components/breadcrumbs/breadcrumbs.less
+++ b/lib/components/breadcrumbs/breadcrumbs.less
@@ -28,7 +28,7 @@
     & &--link {
         color: var(--_br-link-fc);
 
-        &:hover{
+        &:hover {
             --_br-link-fc: var(--fc-medium);
         }
     }
--- a/lib/components/button/button.less
+++ b/lib/components/button/button.less
@@ -412,7 +412,7 @@
                 border-color: var(--_bu-bc-hover);
             }
 
-            &:visited:not(:hover):not(:focus) {
+            &:visited:not(:active):not(:focus) {
                 &.s-btn__filled {
                     background-color: var(--_bu-filled-bg);
                     border-color: var(--_bu-filled-bc);
--- a/docs/product/components/inputs.html
+++ b/docs/product/components/inputs.html
@@ -196,7 +196,7 @@ description: Input elements are used to gather information from users.
         <input class="s-input" id="example-warning" type="text" placeholder="" aria-describedby="example-warning-desc" />
         @Svg.Alert.With("s-input-icon")
     </div>
-    <p id="example-warning-desc" class="flex--item s-input-message mb0">Caps lock is on! <a>Having trouble entering your username?</a></p>
+    <p id="example-warning-desc" class="flex--item s-input-message mb0">Caps lock is on! <a href="#">Having trouble entering your username?</a></p>
 </div>
 {% endhighlight %}
         <div class="stacks-preview--example">
@@ -207,7 +207,7 @@ description: Input elements are used to gather information from users.
                         <input class="s-input" id="example-warning" type="text" value="AA" aria-describedby="example-warning-desc" />
                         {% icon "Alert", "s-input-icon" %}
                     </div>
-                    <p id="example-warning-desc" class="flex--item s-input-message mb0">Caps lock is on! <a>Having trouble entering your username?</a></p>
+                    <p id="example-warning-desc" class="flex--item s-input-message mb0">Caps lock is on! <a href="#">Having trouble entering your username?</a></p>
                 </div>
             </div>
         </div>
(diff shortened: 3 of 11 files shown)


CGuindon (changes requested):
Is it just me or is there no more hover color change on the Visited link? I think it use to go up/down a stop depending on dark/light mode. I'm looking at the Stacks docs now and in the PR env [here](https://deploy-preview-1735--stacks.netlify.app//product/components/links/#single-link-examples)

giamir:
> Is it just me or is there no more hover color change on the Visited link? I think it use to go up/down a stop depending on dark/light mode. I'm looking at the Stacks docs now and in the PR env [here](https://deploy-preview-1735--stacks.netlify.app//product/components/links/#single-link-examples)

I can reproduce the same. It looks like now visited links are always `purple-500` no matter if hovered on or not. 
Before they went from `theme-secondary-500` (visited, not hovered) to `theme-secondary-400` (visited, hovered). The [figma comp](https://www.figma.com/design/znU3ZzJJHTJzmvDS46rRUE/Visited-Link-Colors?node-id=26-5027&t=8XTocJYf9qIhtHHV-0) doesn't specify what to use for visited+hover status though.

lib/components/breadcrumbs/breadcrumbs.less
```diff
@@ -25,12 +25,16 @@
         margin-top: var(--su2);
     }
 
-    & &--link {
-        color: var(--_br-link-fc);
-
+    & &--link,
+    & a&--link{
```
giamir: I am not sure I understand what the `a` does here? `as-breadcrumbs--link`?

giamir (commented):
I did not expect this "simple" ticket to involve so many changes. 😱

Apart from the point raised from @CGuindon the other changes seems reasonable to me. I feel we are wrestling quite a bit CSS specificity (hence the many changes). I also do think there is a high probability we might have missed something but it is also a relatively low risk change so I am ok with going ahead with it. 

Thanks @dancormier for the thorough PR.

CGuindon:
I would keep the same logic of color stops `purple-400` for visited, hovered

lib/components/breadcrumbs/breadcrumbs.less
dancormier: This `&` in less always just plugs in the selector found in the outer scope, so in this case, it's targeting:

```css
  .s-breadcrumbs a.s-breadcrumbs--link
```

The `a` is just there to ensure it applies to breadcrumbs that are links.

dancormier:
> I would keep the same logic of color stops `purple-400` for visited, hovered

I've added `purple-400` for visited+hovered links in https://github.com/StackExchange/Stacks/pull/1735/commits/20f10fb1552d47b4d07bae87855a171792343e9c

### input message

I noticed tests were failing for [the s-input-message component](https://deploy-preview-1735--stacks.netlify.app/product/components/inputs/#validation-examples) and realized that the visited link style was being applied to the link within that element (for some reason, our visual tests render those links as if they're visited no matter what 🤷‍♂️). I found that we use a different color pattern than our usual 400/500/600 pattern for default/hover/visited.

@CGuindon would we be okay to punt on visited link styles for anchors within `.s-input-message`? I can add it but I'm hesitant to introduce more complication at this moment.

CGuindon:
@dancormier 
1. weird that the [the s-input-message component](https://deploy-preview-1735--stacks.netlify.app/product/components/inputs/#validation-examples) automatically uses visited links — but that's a problem for another day
2. Ideally these remain the colors they are now — warning message link stays red. If we punt on this component, would those links be purple or would they stay as they are now?

dancormier:
@CGuindon 

>     1. weird that the [the s-input-message component](https://deploy-preview-1735--stacks.netlify.app/product/components/inputs/#validation-examples) automatically uses visited links — but that's a problem for another day

To be clear, the links only show up as visited _in the visual tests images_, not by default when the component is rendered.

>     2. Ideally these remain the colors they are now — warning message link stays red. If we punt on this component, would those links be purple or would they stay as they are now?

They would stay as they are now (visited state will not change the color of the link within the input messages)

CGuindon:
I can see in the inspector that `_li-fc-hover-visited:` is set to `purple-400` but when I hover, the color isn't changing. Checking this in the PR env. 

Also... (and this is my fault for not catching before) but since we are going _darker_ for blue links on hover:
- default blue-400
- hover blue-500

shouldn't we go darker for purple as well on hover? 
- visited default purple-500
- visited hover-purple-600

Screenshot 2024-05-17 at 10 57 07 AM

dancormier:
@CGuindon @giamir I decided to ditch the `:not(.s-badge):not(…` approach I was taking before in https://github.com/StackExchange/Stacks/pull/1735/commits/28657166a01cb456c13498006683bf82fdfcd70b and instead go with what's in https://github.com/StackExchange/Stacks/pull/1735/commits/aecad2a75ea2796fb369cbcda02b39ca92713ade. 

This is the most significant change:

```less
a {
    &:visited {
        &:not([class*="s-"]),
        &.s-link,
        &.s-sidebarwidget--action,
        &.s-user-card--link {
            &:hover {
                color: var(--_li-fc-hover-visited);
            }

            color: var(--_li-fc-visited);
        }
    }
}
```

This is specifying to apply these styles to anchors that **don't** include a class starting with `s-`, with the exception of `s-link`, `s-sidebarwidget--action`, or `s-user-card--link`. IMO, this is way safer and simpler than the previous because it no longer requires this long list of `:not` exclusions that increases the specificity of the rule.

@giamir I'd like to get your thoughts on this approach and it's relative safety when compared to the previous approach when you have a moment.

---

The stuff below is nonsense, so feel free to ignore it.

### Story time (aka TIL about properties allowed in `:visited`)

The big problem was with specificity. If we applied styles to `a:visited`, most visited links would gain that style _even if_ we have other non-visited styles on that element (think a `.s-btn` on an anchor). This is _not_ what we want.

**I had a lightbulb moment**: what if we specificy CSS variables on `a:visited` so the variable is a high specificity but the style being applied would stay only on the `a` selector so it could be easily overwritten. 

```less
a,
.s-link {
  --_li-fc: var(--theme-link-color, var(--theme-secondary-400));
  …

  &:visited {
    --_li-fc-override: var(--purple-500);
  }

  color: var(--_li-fc-override, var(--_li-fc));
  …
}
```

This seemed like a winner. I implemented it and was… very confused. The CSS variable was being picked up by the browser. The style said it was applied in the inspector. But, when I looked at the page, the color defined as a CSS variable in `:visited` wasn't applying. What's the deal?

#### The TIL 

You can only set specific CSS properties within `:visited` for security/privacy reasons 🙃. [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/:visited#privacy_restrictions) gives more details, but for this case the big takeaway is:

> [!TIP]
> You cannot define or redefine CSS variables within the `:visited` selector.

C'mon now, CSS!

CGuindon (approved):
I see all the purples now 🎉!

lib/components/button/button.less
```diff
@@ -412,7 +412,7 @@
                 border-color: var(--_bu-bc-hover);
             }
 
-            &:visited:not(:hover):not(:focus) {
+            &:visited:not(:active):not(:focus) {
```
giamir, line 415: Is this change intentional?

giamir (approved):
> @giamir I'd like to get your thoughts on this approach and it's relative safety when compared to the previous approach when you have a moment.

I think `&:not([class*="s-"])` is pretty smart and not less safe than what we had before. Great work @dancormier! 🎉

lib/components/button/button.less
dancormier: Yep, that rule is _already_ within a `:hover` selector, so adding `:not(hover)` would make it so the rule here never applies. This is an existing bug that I resolved in the course of this PR.

PR #1738, opened 2024-05-20
docs(a11y): document min viewport dimensions

dancormier:
[STACKS-580](https://stackoverflow.atlassian.net/browse/STACKS-580)

This PR adds a section to the Stacks docs Accessibility page describing the minimum supported viewport size, adhering to [WCAG 2.2 1.4.10: Reflow](https://www.w3.org/WAI/WCAG22/Understanding/reflow.html).

---

### To test

1. Go to the new ["Viewport size" section](https://deploy-preview-1738--stacks.netlify.app/product/foundation/accessibility/#viewport-size) of the Stacks docs Accessibility page
2. Let me know if it sounds reasonable and is free from grammatical/spelling errors
3. Profit!

### DIFF
--- a/docs/product/foundation/accessibility.html
+++ b/docs/product/foundation/accessibility.html
@@ -324,11 +324,11 @@ description: A non-comprehensive guide to accessibility best practices when usin
     {% header "h2", "Viewport size" %}
 
     <p class="stacks-copy">
-        All Stack Overflow products must conform to the <a class="s-link" href="https://www.w3.org/WAI/WCAG22/Understanding/reflow.html">WCAG 2.2 <span style="text-decoration: underline var(--black) dashed" class="" data-controller="s-tooltip" data-s-tooltip-placement="top" title="Success Criteria">SC</span> 1.4.10: Reflow</a>. This requires that our product UIs support viewports as small as <b>320px x 256px</b> without requiring the user scrolling in multiple dimensions (unless an element requires it for usage or meaning). Very few users will ever use a viewport this small, but it's important to support it so users can zoom in up to 400% and still have a usable experience. At 400% zoom, a 320x256 viewport translates to 1280x1024, which is a common resolution for many users. Supporting this small viewport size ensures that users with low vision can still use our products effectively.
+        All Stack Overflow products must conform to the <a class="s-link" href="https://www.w3.org/WAI/WCAG22/Understanding/reflow.html">WCAG 2.2 <abbr title="Success Criteria">SC</abbr> 1.4.10: Reflow</a>. This requires that our product UIs support viewports as small as <b>320px x 256px</b> without requiring the user scrolling in multiple dimensions (unless an element requires it for usage or meaning). Very few users will ever use a viewport this small, but it's important to support it so users can zoom in up to 400% and still have a usable experience. At 400% zoom, a 320x256 viewport translates to 1280x1024, which is a common resolution for many users. Supporting this small viewport size ensures that users with low vision can still use our products effectively.
     </p>
 
     {% header "h3", "Exceptions" %}
     <p class="stacks-copy">
-        There are some exceptions to this rule. Tables, videos, and may require horizontal scrolling on small viewports. In these cases, it's acceptable to require scrolling in two dimensions, though we should avoid requiring two dimensional scrolling when possible. See the <a class="s-link" href="https://www.w3.org/WAI/WCAG22/Understanding/reflow.html#content-exceptions-for-reflow">WCAG 2.2 documentation on Reflow</a> for detailed guidance.
+        There are some exceptions to this rule. Some elements such as tables and videos may require horizontal scrolling on small viewports. In these cases, it's acceptable to require scrolling in two dimensions. See the <a class="s-link" href="https://www.w3.org/WAI/WCAG22/Understanding/reflow.html#content-exceptions-for-reflow">WCAG 2.2 documentation on Reflow</a> for detailed guidance.
     </p>
 </section>
--- a/docs/product/foundation/accessibility.html
+++ b/docs/product/foundation/accessibility.html
@@ -319,3 +319,16 @@ description: A non-comprehensive guide to accessibility best practices when usin
 
     <p class="stacks-copy">Some focusable elements and Stacks components currently do not include custom focus styling. These elements will instead render the browser-default focus indicators.</p>
 </section>
+
+<section class="stacks-section">
+    {% header "h2", "Viewport size" %}
+
+    <p class="stacks-copy">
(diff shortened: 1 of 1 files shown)


docs/product/foundation/accessibility.html
```diff
@@ -319,3 +319,16 @@
...
+    </p>
+
+    {% header "h3", "Exceptions" %}
+    <p class="stacks-copy">
+        There are some exceptions to this rule. Tables, videos, and may require horizontal scrolling on small viewports. In these cases, it's acceptable to require scrolling in two dimensions, though we should avoid requiring two dimensional scrolling when possible. See the <a class="s-link" href="https://www.w3.org/WAI/WCAG22/Understanding/reflow.html#content-exceptions-for-reflow">WCAG 2.2 documentation on Reflow</a> for detailed guidance.
```
giamir: `Table, videos and ...` did you forget something?

giamir: `though we should avoid requiring two dimensional scrolling when possible`
Wonder if we could simply drop this bit since it sounds like a repetition to me. Just keep:
`In these cases it's acceptable to require scrolling in two dimensions.`

giamir (approved):
Copy looks good to me. I only have 2 minor questions:

- Is using a tooltip (see SC for Success Criterion) to explain an abbreviation a common pattern described somewhere in Stacks? Semantically wouldn't the [`abbr` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr) be more appropriate? 

- Is it me or the visited links in dark mode are now barely visible (see screenshot)? I think we might be in violation of some a11y SCs there. I almost did not see the difference if it was not for the dashed SC part. cc @CGuindon 

Screenshot 2024-05-21 at 09 54 14


Thanks @dancormier for documenting this. Great work. 🎉

CGuindon:
Re: [comment here](https://github.com/StackExchange/Stacks/pull/1738#pullrequestreview-2067853207)

@giamir What's the violation you're thinking of? Not enough contrast between regular text color and link text color because we're using color alone to identify a link? I think we were hitting that issue before we switched to purple anyway (would have been a dark blue). I'd have to review full requirements here but I think this would be out of scope for this PR anyway... if there's an issue tied to a requirement do you want to open an a11y ticket?

CGuindon (approved):
That's some good copy writing! GOLD

giamir:
> @giamir What's the violation you're thinking of? Not enough contrast between regular text color and link text color because we're using color alone to identify a link? I think we were hitting that issue before we switched to purple anyway (would have been a dark blue). I'd have to review full requirements here but I think this would be out of scope for this PR anyway... if there's an issue tied to a requirement do you want to open an a11y ticket?

Yeah, I did not take the time to investigate further. I just had a gut feeling that somehow that the almost lack of differentiation between link text and static text probably violates something. 

From [WCAG SC 1.4.1](https://www.w3.org/WAI/WCAG22/Understanding/use-of-color.html) (it looks like the lack of differentiation between visited and non-visited colors won't be a problem)

> Most user agents provide users with a color-only cue that a link has been previously activated by them ("visited"). However, several technical constraints result in authors having very limited control over these color-only indications of visited links. The technical constraints are as follows:
> - User agents constrain the exposure of a link's visited state due to [privacy concerns](https://developer.mozilla.org/en-US/docs/Web/CSS/Privacy_and_the_:visited_selector). Author queries to user agents will indicate all links have not been visited.
> - Any available information on the visited state of a link would be inaccurate since it is both user and browser-dependent. Even if an author could accurately get information on previously activated links by a certain user, the author would be constrained to the current browser's preserved history, and would be unable to determine if the user had visited the page using a different browser (or if the history was not preserved, either from cache clearing or use of private sessions).
> - Authors can only use color to modify the :visited CSS pseudoclass style. The technology constrains any non-color styling. Due to palette limitations, an author cannot use color alone to achieve 3:1 contrast between link and non-link text as well as between visited and unvisited links while also achieving 4.5:1 contrast for all link and non-link text.
> - Authors also cannot set the visited state of links. The anchor element does not include a "visited" attribute; therefore the author has no ability to alter the state through an attribute setting. As such, authors cannot achieve [1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships) or [4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value) in regard to visited links.
>
> For these reasons, setting or conveying a link's visited status is not an author responsibility. Where color alone distinguishes between visited and unvisited links, it does not result in a failure of this Success Criterion, even where the contrast between the two link colors is below 3:1. Note that authors must continue to ensure that all text links meet contrast minimums against the page background (SC 1.4.3).

There is no violation for not differentiating enough between visited and not-visited links but I believe we are still in breach of that SC since: **Link underlines or some other non-color visual distinction are required (when the links are discernible to those with color vision)**  _from_ https://www.w3.org/WAI/WCAG22/Techniques/failures/F73

@dancormier Could you confirm my analysis? Then I will create an a11y ticket around those lines: 
_Link text is distinguishable from static text only via color_
This task will require design input @CGuindon and I suspect it won't be straightforward to come up with a solution.

docs/product/foundation/accessibility.html
dancormier: I did! Thanks for catching this!

dancormier:
> Is using a tooltip (see SC for Success Criterion) to explain an abbreviation a common pattern described somewhere in Stacks? Semantically wouldn't the abbr element be more appropriate?

You're right @giamir! `abbr` is the correct element here.

dancormier:
> @dancormier Could you confirm my analysis?

Yep, it seems you're correct and we'll need to either:

a) differentiate links (including visited ones) by some means other than color (underlining, bolding, etc), or
b) differentiate significantly by **both hue and lightness**. 

This is the particularly relevant part of [WCAG SC 1.4.1](https://www.w3.org/WAI/WCAG22/Understanding/use-of-color.html) IMO:

> If content is conveyed through the use of colors that differ not only in their hue, but that also have a significant difference in lightness, then this counts as an additional visual distinction, as long as the difference in relative luminance between the colors leads to a contrast ratio of 3:1 or greater. For example, a light green and a dark red differ both by color (hue) and by lightness, so they would pass if the contrast ratio is at least 3:1. Similarly, if content is distinguished by inverting an element's foreground and background colors, this would pass (again, assuming that the foreground and background colors have a sufficient contrast).

`purple-500` has a [1.09:1](https://webaim.org/resources/contrastchecker/?fcolor=D7D9FF&bcolor=E3E6E8) contrast ratio between visited links and surrounding text in dark mode according to the WCAG algorithm. Bumping the base value to `purple-400` still only gets us to [1.37:1](https://webaim.org/resources/contrastchecker/?fcolor=BDC1FE&bcolor=E3E6E8).

*With all of that said, I couldn't find a single instance of a big player adhering to this rule in their respective dark modes. I checked Google, Bing, Jira, GitHub, and Reddit. Some of these sites (Jira, GitHub) didn't even have a differentiation between visited and non-visited links. 

This made me second-guess my assessment that we're interpreting the requirements correctly. I reread it a few times and I can't see how else I'd interpret it. Assuming my interpretation of the rule is correct it seems like lots of sites don't hit this SC. I don't know how we want to use that information except to consider _why_ these big sites don't bother hitting a 3:1 or providing other differentiation for links compared to adjacent text (unless I'm missing something obvious, which is very possible).

> Then I will create an a11y ticket around those lines:
> Link text is distinguishable from static text only via color

@giamir yep, I think we need a ticket at least to investigate and consider this further.

---

**Related question/can of worms 🐛**: Considering that we use APCA to determine the contrast between _text and the background_, should we also use it for situations like this where we need to determine contrast between elements other than text and backgrounds where Lc 60 wouldn't apply?

PR #1739, opened 2024-05-22
feat(topbar): add popover child component

dancormier:
[STACKS-613](https://stackoverflow.atlassian.net/browse/STACKS-613)

This PR adds a Stacks topbar popover component so we can control the size and overflow behavior, primarily to meet the success criteria for [WCAG SC 1.4.0 Reflow](https://www.w3.org/WAI/WCAG22/Understanding/reflow.html). 

> [!NOTE]
> I still need to add tests to this PR, but I'm opening it for review to get eyes on it earlier rather than later.

## New classes

This PR introduces the following child component classes to the topbar:

- `s-topbar--searchbar--popover`: This class applies a few sensible defaults to the popover in this context (specifically it sets `padding` to `0` and min/max width)
- `.s-topbar--searchbar--popover-content`: This class is intended to wrap the content within the popover to ensure it can resize and scroll vertically (if necessary)

## Concessions

Since the topbar height is set with `--theme-topbar-height`, we know the height of the topbar for certain, but _we cannot know for sure_ what the height of the search group is when it's rendered below the topbar (in the case of small viewports). Because of this, I took an educated guess and yielded on the side of caution to ensure the popover is rendered within the viewport.

It would be unlikely consumers would manually alter the height of the search group in mobile in such a way that it's significantly larger than the default, potentially rendering a portion of the popover outside of the viewport boundaries. I think this is a safe approach but I'm willing to reconsider if objections arise.

## TODO

- [ ] Add tests specifically for the popover when it's shown
- [ ] **Before merging**: Remove excess popover content

## Screenshots

<details>
<summary>320 x 256</summary>

</details>

<details>
<summary>770 x 256</summary>

</details>

## To test

For ease of testing this PR, I've added the popover to the main topbar of the Stacks docs site. It's temporary and will be removed before merging this PR.

- Go to the [Stacks docs](https://deploy-preview-1739--stacks.netlify.app/)
- Focus the search input in the main topbar of the docs site
- Notice the addition of a popover anchored to the search input
- Make the browser real small and notice the popover resizes to stay within the boundries of the viewport and introduces vertical scrolling if needed.
  - **Tip**: [320 x 256](https://www.w3.org/WAI/WCAG22/Understanding/reflow.html) is the min supported viewport size

### Bonus if you're ambitious (I don't necessarily recommend it, though)

- In the [topbar docs page](https://deploy-preview-1739--stacks.netlify.app/product/components/topbar), test out some of the other topbars and their respective popovers
  - You'll notice some overlapping issues. Ignore that since it's a docs-only issue.
  - You'll also notice you need to scroll so the topbar you're interacting with is where a topbar _would_ be on the page (at the top, usually 😝) in order to see the expected behavior wrt height. This is why I included the popover in the main topbar. You're probably better off just testing it there.


docs/_includes/header.html
```diff
@@ -38,8 +38,63 @@
 
         <div class="s-topbar--searchbar w100 wmx3 sm:wmx-initial js-search">
             <div class="s-topbar--searchbar--input-group">
-                <input id="searchbox" type="text" placeholder="Search Stacks…" value="" autocomplete="off" class="s-input s-input__search" />
+                <!-- TODO remove added attributes before merging topbar popover -->
```
giamir: It looks like we do get the same issue reported by MSFT in our stacks docs search as well. Can we use the newly created classes to fix that issue in stackoverflow.design too as part of this PR? Is it possible with the [algolia library](https://www.npmjs.com/package/@docsearch/js) we are using (which I believe is also outdated 🙃 - still v2)?
Screenshot 2024-05-24 at 10 10 53

lib/components/topbar/topbar.less
```diff
@@ -367,6 +371,17 @@
             position: relative;
         }
 
+        .s-topbar--searchbar--popover {
```
giamir: Should those classes live here or be coupled to the popover ones instead?
To me this sound like a problem that any popover in a small viewport could have. Could we generalize the solution instead of coupling it exclusively to the topbar-searchbar component?

giamir (commented):
@dancormier I did a preliminary review. It looks good to me so far. I have added a couple of high level questions to make sure we have considered also other approaches. 
Let me know when things are ready for a final review.  😊

CGuindon (changes requested):
Ignore my previous comment (if you got a notification email) — apparently I didn't read anything you posted.

For the overlapping issue that _should_ be Stacks only, would this same issue appear on Teams? They also have both the helper and results — would those two popovers now overlap?

Also, I don't know how to test this with the overlapping of two popovers? I can't see the results on small screens.

Screenshot 2024-05-27 at 2 17 37 PM

dancormier:
@CGuindon 
> For the overlapping issue that _should_ be Stacks only, would this same issue appear on Teams? They also have both the helper and results — would those two popovers now overlap?

On Teams, there's just a single popover that gets its contents swapped out when results render. It shouldn't be an issue there.

> Also, I don't know how to test this with the overlapping of two popovers? I can't see the results on small screens.

I should've specified that the Algolia search suggestions popover is not a native Stacks popover so exclude that from your testing. I added a dummy popover to the search for demonstration purposes and plan to remove it once this PR is approved, and that's the one I recommend testing against (you can get only that popover to show by only focusing the search input)

docs/_includes/header.html
dancormier: Good suggestion! I couldn't find a simple way to use Stacks classes for that popover (I can't even [update the config](https://stackexchange.slack.com/archives/C27RWNQN9/p1611196118000600?thread_ts=1611184636.095000&cid=C27RWNQN9) without going through Algolia), so I took the easy way out in https://github.com/StackExchange/Stacks/pull/1739/commits/31d7055dcb54e0644dad9c9318499e6428be6661 and just added a couple of styles lifted from the topbar.

How do you feel about using those custom properties outside of the topbar? I was on the fence between that or duplicating the custom properties in the context of the custom algolia popover CSS.

lib/components/topbar/topbar.less
dancormier: I think that's probably the right call. I initially wanted to apply this fix as targeted as possible, but I think you're correct to point out that the same fix could just apply to the popover component itself.

I generalized this fix to the popover component broadly in https://github.com/StackExchange/Stacks/pull/1739/commits/f27ca29baecee8558687f996d6da482175b7265c. There, I still account for the height of the topbar and override some values for the popover in the context of the topbar.

docs/_includes/header.html
giamir: Since those are only applied to the docs I think it's ok. I think we need to make sure though that somebody doesn't get tempted to do that elsewhere. Duplicating with a comment is also a viable option. Perhaps a comment should be added anyway saying "don't do this, we are only doing this because we cannot apply our classes easily to the algolia popover".

lib/components/popover/popover.less
```diff
@@ -17,6 +17,9 @@
     --_po-arrow-after-r: unset;
     --_po-arrow-after-t: unset;
     --_po-arrow-after-bs: unset;
+    // content
+    --_po-topbar-height: var(--theme-topbar-height, calc(var(--su-static48) + var(--su-static8)));
```
giamir, line 22: @dancormier Is it safe to assume that all pages will have a sticky topbar of these dimensions? I am genuinely asking because I don't have an opinion yet.

docs/product/components/popovers.html
giamir, line 1: @dancormier I see that you have now changed the markup of all the popovers in the docs. Should this be considered a breaking change then (meaning popover markup should always have a wrapper element with the `s-popover--content` class applied)?
Or can we keep the application of the new class optional? I also don't have a strong opinion here. I am just not very excited at the idea of changing a lot of markup in Core. 😅

We should also make a note (e.g. tiny story) to add this new class in the stacks-svelte library markup too:
https://github.com/StackEng/Stacks-Svelte/blob/main/src/components/Popover/PopoverContent.svelte#L28  
Potentially even Razor: https://github.com/StackEng/Stacks-Razor/blob/main/Stacks.Razor/Components/Popover.razor#L10

giamir (commented):
I left few additional comments @dancormier. All minor stuff I would say. After addressing those and cleaning up the stub markup you have added I think we are in a good shape to merge. Thank you

CGuindon (approved):

lib/components/popover/popover.less
dancormier: I think it's safe _enough_ to assume. I think a vast majority of our pages include a sticky topbar and if a page doesn't have a sticky topbar, the worst that will happen is that the popover will leave a little more space than desirable when the viewport is tiny

docs/product/components/popovers.html
dancormier: I consider this an _additive_ change. I would like implementations of the popover component to include `s-popover--content` going forward but it wouldn't be worth it to retroactively apply this wrapper.

lib/components/popover/popover.less
giamir: Sounds good to me. A little comment about the assumption and the low impact if it is not respected could be added close to that variable in my opinion. I will leave it to you to decide. Either way consider this comment resolved. Thanks for taking the time to explain your reasoning.

docs/product/components/popovers.html
giamir: Makes sense, the only action for this comment is to create a tiny placeholder story so that we don't forget to change the markup of our svelte and Razor libraries later on.  Thank you.

giamir (approved):

docs/product/components/popovers.html
dancormier: Done! https://stackoverflow.atlassian.net/browse/STACKS-630

PR #1740, opened 2024-05-24
chore(links): update visited links colors w/ a11y considerations

dancormier:
This PR addresses [STACKS-481](https://stackoverflow.atlassian.net/browse/STACKS-481) (update visited link color) and [STACKS-646](https://stackoverflow.atlassian.net/browse/STACKS-646) (underline links within in paragraphs to meet [WCAG 2.2 SC 1.4.1](https://www.w3.org/WAI/WCAG22/Techniques/failures/F73)). It should also incidentally resolve [STACKS-480](https://stackoverflow.atlassian.net/browse/STACKS-480) (improve contrast between unvisited and visited links on Meta).

### Changes in this PR

- Updates the visited link style to `purple-500` (and `purple-600` when visited and hovering)
- Adds an underline to elements with `.s-link` when within a paragraph
    - It also updates the docs to mimic this style

### To test visited link color

1. Go to the "[Single link examples](https://deploy-preview-1740--stacks.netlify.app/product/components/links/#single-link-examples)" section of the links docs
2. Look at the "visited" link example
3. **Bonus**: Try it out in different color modes
4. **Double bonus**: Open a link in the docs and observe it change from blue to purple

### To test underline

We don't include an `.s-link` within a paragraph anywhere in our docs, but we did apply the same styling to links in the documentation copy. You can see those by doing the following:

1. Visit a docs page that includes links within copy (e.g. the [accessibility page](https://deploy-preview-1740--stacks.netlify.app/product/foundation/accessibility/))
2. Check out those underlined links 😎 

### Screenshots

<details>
<summary>Light mode</summary>

#### Default
Screen Shot 2024-07-30 at 17 38 14

#### High contrast

</details>

<details>
<summary>Dark mode</summary>

#### High contrast
Screen Shot 2024-07-30 at 17 41 25

</details>

---

### History

[STACK-481](https://stackoverflow.atlassian.net/browse/STACKS-481)

In https://github.com/StackExchange/Stacks/pull/1735, we updated visited link styles. @giamir [later pointed out](https://github.com/StackExchange/Stacks/pull/1738#issuecomment-2124075186) that we should consider [WCAG SC 1.4.1](https://www.w3.org/WAI/WCAG22/Understanding/use-of-color.html), which suggests that links within otherwise unlinked text needs to be differentiated from surrounding text either by some means other than color (underlining, bolding, etc) or differentiate significantly by both hue and lightness (see [comment](https://github.com/StackExchange/Stacks/pull/1738#issuecomment-2125690020)). Also see [A11Y-108](https://stackoverflow.atlassian.net/browse/A11Y-108)

Because link styles may require more work/consideration than was performed in https://github.com/StackExchange/Stacks/pull/1735, we decided to [revert that PRs merge](https://github.com/StackExchange/Stacks/commit/43b6c22248646628d7cc4663c639509f0985bfec) to avoid shipping a half-baked change. In this PR, we'll resolve any lingering accessibility issues with link color and then ship it out.

cc @giamir @CGuindon


docs/product/components/links.html
```diff
@@ -56,6 +56,11 @@
             </div>
         </div>
     </div>
+
+    {% header "h3", "Accessibility" %}
+
+    <p class="stacks-copy">Any link with adjacent static text cannot use color alone to differentiate it as a link. If a link is next to static text and the only visual indication that it’s a link is the color of the text (often <code class="stacks-code">blue-400</code>), it will require an underline in addition to the color. Reference <a href="https://www.w3.org/WAI/WCAG22/Techniques/failures/F73">WCAG SC 1.4.1</a> for more details.</p>
```
giamir, line 62: What about adding a small snippet as well here to make it even clearer for devs?

```
<span>Some static text with a <a href="#" class="s-link s-link__underlined">link</a> inside</span>
```

lib/components/link/link.less
```diff
@@ -1,10 +1,26 @@
 // TODO we *really* shouldn't be apply styles directly onto `<a>` like this, but
 // it's tech debt that'll take some doing in consumer's code to pay down.
+a {
+    &:visited {
+        &:not([class*="s-"]),
```
giamir, line 7: So in here we are saying that if an anchor element has a stacks class which is not `s-link`, `s-sidebarwidget--action` or `s-user-card--link` should not get the visited styles. Out of curiosity why is this necessary? What happen if we would just target all the anchor tags (including the ones that have stacks classes applied)?

Regardless this line might be worth a small comment.

giamir (commented):
Looks great @dancormier. Thanks for taking care of this. I only left a couple of minor comments.

The biggest recommendation I would have is to at least add the class="s-link" to the anchor elements in our accessibility page that don't have it yet because at the moment some links get the underline some don't. 
Screenshot 2024-07-31 at 14 04 42

If you can find a good regex I would suggest to do it for all our docs pages so that they are consistent for our devs to consult. Maybe copilot chat can help with finding the regex. 🙂

Thank you.

lib/components/link/link.less
dancormier: > Out of curiosity why is this necessary? 

To make sure we're on the same page, this ensures we only set visited styles on:

- `a.s-link`
- `a.s-sidebarwidget--action`
- `a.s-user-card--link`
- all other anchors `a` that don't include a stacks component class (identified by `.s-`)

These rules keep it (somewhat) narrowly scoped and only applies it to elements where we can be confident the visited style is appropriate.

> What happen if we would just target all the anchor tags (including the ones that have stacks classes applied)?

If we broadly targeted _all anchors_, this style could wind up being applied to components like `.s-btn` and `.s-tag`. If they don't have explicit `:visited` styles defined, this rule would potentially override any color set on the element in favor of the visited color, resulting in unexpected colors and possibly causing contrast issues.

For example, here's a visited moderator tag if we targeted `a:visited` here:

image


> Regardless this line might be worth a small comment.

This seems like a good call. When I came back to this PR, I had to take a moment to refresh myself on what why this rule was structured this way (basically, I had to ask myself the same questions you asked). I'll add a comment describing the reasoning.

docs/product/components/links.html
dancormier: After your review, I realized that I had a failed push after my last commit. The commit added after your review results in the `WCAG SC 1.4.1` link being underlined. Let me know if you think we should call it out more explicitly

CGuindon (approved):
LGTM! Will the purple change happen on SE Meta as well or would we need to make a theme change there specifically? Any any other themes that use blue as the link color?

PR #1752, opened 2024-06-05
docs(a11y): document usage of ARIA landmarks across our products

giamir:
[STACKS-614](https://stackoverflow.atlassian.net/browse/STACKS-614)

This PR documents the usage of ARIA landmarks in our [accessibility page](https://deploy-preview-1752--stacks.netlify.app/product/foundation/accessibility/#landmarks).

Furthermore it adjusts landmarks issues across our docs. 

## How to test

- Download the [Landmarks Browser Extension](https://matatk.agrip.org.uk/landmarks)
- Navigate to any page
- Observe we now have 6 landmarks (`banner`, including `global navigation` and `search` - `main navigation` - `main`, including `table of contents navigation`). Screen readers and keyboard-only users can easily jump between landmark sections.

_Sidenote: This PR is also adding the `aria-current="page"` to the selected item in the global and main navigations._

### DIFF
--- a/docs/_includes/layouts/home.html
+++ b/docs/_includes/layouts/home.html
@@ -4,7 +4,7 @@
 
 <body class="d-flex fd-column bg-black-100 theme-system">
     {% include 'theme.html' %}
-    <div class="d-flex fl-grow1 bg-theme-primary-400 fd-column ps-relative ai-center px64 sm:pl24 sm:pr24">
+    <header class="d-flex fl-grow1 bg-theme-primary-400 fd-column ps-relative ai-center px64 sm:pl24 sm:pr24">
         <svg class="stacks-intro-field theme-light__forced" width="716" height="519" viewBox="0 0 716 519">
             <defs>
                 <linearGradient id="a" x1="427.325" y1="187.632" x2="458.682" y2="132.104" gradientUnits="userSpaceOnUse">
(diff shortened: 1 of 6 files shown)


docs/_includes/layouts/home.html
```diff
@@ -192,9 +192,9 @@ <h1 class="flex--item fw-bold mln2 mb0 fs-display1 lh-xs stacks-title">Stacks</h
                 <p class="m0 fs-subheading"><em>Stacks</em> provides everything you need to quickly design, build, and ship coherent experiences across all of Stack Overflow—from the brand and product itself, down to how we send emails and write copy.</p>
             </div>
         </div>
-    </div>
+    </header>
     <div class="flex--item bg-black-100 p64 sm:pt24 sm:pl24 sm:pr24 sm:pb24">
-        <div class="d-grid grid__4 lg:grid__2 sm:grid__1 g16 w100 wmx12 mx-auto">
+        <nav class="d-grid grid__4 lg:grid__2 sm:grid__1 g16 w100 wmx12 mx-auto" aria-label="Global">
```
gabrieltron, line 197: question: how descriptive should the labels of navigation elements be? Or are descriptions like "Global" and "Main" fairly understood as a standard? Using NVDA, I can see this structure
image

Going by this structure, it is a bit hard for me to understand the information present in the page, so even if we have landmarks I'm not sure what those landmarks refer to when going through them 🤔

gabrieltron (approved):
LGTM! The added landmarks makes sense to me, as well as the guidance around using landmarks on product pages.

Left a question, more for my understanding on how we should be tackling navigation items. I'm approving as it looks good to me but beware that I'm unfamiliar with the stacks project. You might want an approval from someone who may be more familiar with the full picture as well

docs/_includes/layouts/home.html
giamir: I kept the navigation label simple after reading this section of the MDN article about landmarks
> Here, we use the nav element because it implicitly has the navigation role. Since a page could have multiple navigation elements, it is important to identify each of these landmarks uniquely. For this, we use the aria-label attribute and give the navigation landmark the name "Primary".
> Note When a screen reader describes the above navigation landmark, it will say "Primary navigation". This is why the word "navigation" is not included in the aria-label value.

https://developer.mozilla.org/en-US/blog/aria-accessibility-html-landmark-roles/#the_navigation_landmark

I think "Primary" and "Secondary" are pretty well understood terms. We use "Global" and "Main" because Global was already used in the docs. I am also happy to revisit the naming but I would not know necessarily what to use instead (perhaps Primary and Secondary?).

gabrieltron: Got it, thanks for sharing the reference! Makes sense to me. If we were to rename it, I'm also not sure which could be more descriptive name. Those are not landmarks with a straightforward way to give a detailed name like we do with "Table of contents", so on a second thought Global and Main sound good 🤔

CGuindon (approved):
Overall looks good to me. I noticed the Main navigation does appear twice though (because it has a duplicated as 'guessed') — not sure if that's supposed to happen or not.

Screenshot 2024-06-05 at 11 41 14 AM

giamir:
> Overall looks good to me. I noticed the Main navigation does appear twice though (because it has a duplicated as 'guessed') — not sure if that's supposed to happen or not.
> 
> Screenshot 2024-06-05 at 11 41 14 AM

The plugin has this feature where it tries to guess if an area is a landmark and it does not really work well. It should be turned off for us developers/designers I guess. You can do that by clicking the preferences section of the plugin

Screenshot 2024-06-06 at 09 39 16

PR #1764, opened 2024-06-21
fix(popover): prevent content child from clipping elements

dancormier:
[STACKS-635](https://stackoverflow.atlassian.net/browse/STACKS-635)

---

This PR ensures that anything extending (12px or less) from the `s-popover--content` element will still be visible. The largest impact will be on focused inputs within a popover.

### Before

image

### After

image


giamir (approved):
Looks good to me. Thanks for the fix. 🙂

PR #1799, opened 2024-08-08
Improve readonly input APCA Lc values for a11y

qheaden-stack:
# Summary

This PR modifies the font color for readonly inputs to improve their contrast, matching the recommendations by APCA for accessibility friendly contrast values.

Here are before and after screenshots of the readonly input in various color modes:

### Before & After for Light Mode:
readonly-input-old-style-light
readonly-input-new-style-light

### Before & After for Dark Mode:
readonly-input-old-style-dark
readonly-input-new-style-dark

### Before & After For Light Mode High Contrast:
readonly-input-old-style-light-highcontrast
readonly-input-new-style-light-highcontrast

### Before & After For Dark Mode High Contrast:
readonly-input-old-style-dark-highcontrast
After for Dark Mode High Contrast:
readonly-input-new-style-dark-highcontrast

### DIFF
--- a/lib/components/input_textarea/input_textarea.a11y.test.ts
+++ b/lib/components/input_textarea/input_textarea.a11y.test.ts
@@ -78,9 +78,7 @@ const otherModifiers = ["creditcard", "search"];
                               ...attributes,
                           },
                 children,
-                template,
-                // TODO revisit these skipped test ids
-                skippedTestids: [/readonly/],
+                template
             });
         });
 
--- a/lib/components/input_textarea/input_textarea.a11y.test.ts
+++ b/lib/components/input_textarea/input_textarea.a11y.test.ts
@@ -78,7 +78,7 @@ const otherModifiers = ["creditcard", "search"];
                               ...attributes,
                           },
                 children,
-                template
+                template,
             });
         });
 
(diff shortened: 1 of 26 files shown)


CGuindon (approved):
Tested in the PR env

lib/components/input_textarea/input_textarea.a11y.test.ts
```diff
@@ -79,8 +79,6 @@ const otherModifiers = ["creditcard", "search"];
                           },
                 children,
                 template,
-                // TODO revisit these skipped test ids
-                skippedTestids: [/readonly/],
```
dancormier, line 83: It's great to see this removed 🎉

dancormier (commented):
This looks good! The tests pass and the text contrast looks good while keeping it differentiated from the writable input. Thanks for making this fix @qheaden-stack!

The only thing I'm curious about is if we should increase the contrast on that lock icon:



I'm not sure we have any requirements specifically for supplementary icons (WCAG requires a 3:1 ratio for icons I believe, but we don't base our contrast targets on WCAG), but the icon is Lc 27 by APCA's measurement which seems _real_ low. 

@CGuindon any thoughts on bumping the lock icon from `300` to `400`? Here's the input with the lock at `400` (Lc 69):

image

The `350` stop is also an option if we wanna split the difference (Lc 36):

image

dancormier:
@CGuindon isn't available today so I'm gonna go ahead and merge this PR, but we can always tweak later if we decide to change the icon color.

CGuindon:
Illustration minimums are 350 — bumping to 400 is fine.

PR #1811, opened 2024-09-04
fix(tooltip): prevent hide when focus is within tooltip

dancormier:
https://stackoverflow.atlassian.net/browse/STACKS-667

---

This PR prevents the tooltip from hiding when focus is moved to an element within the tooltip.

### To test

- Go to the "Rich tooltips" section of the popover docs page
- Tab to the button in this section
- Once the tooltip shows, tab to the link within
    - The tooltip should remain visible
- Tab so the focus is no longer on the triggering button or within the tooltip
    - The tooltip should hide 

### TODO
- [x] Implement test


giamir:
@dancormier I took the liberty to swap the sleep functions in the tests with `waitFor` and `waitForElementToBeRemoved` since they are reducing significantly the likeness of test flakiness. We should try to stay away from sleep functions as much as possible.
https://github.com/StackExchange/Stacks/pull/1811/commits/fdb7870f0477882f6ede31cf0e0545ce8e1e36d5

giamir:
@dancormier Tangential question, I noticed that we recommend to add a `aria-hidden="true"` in our markup examples but then we don't have any js code in our stimulus controllers that dynamically change that value when the tooltip is displayed/hidden. Should we log a bug?

dancormier:
Thank you for the review @giamir!

> @dancormier I took the liberty to swap the sleep functions in the tests with `waitFor` and `waitForElementToBeRemoved` since they are reducing significantly the likeness of test flakiness. We should try to stay away from sleep functions as much as possible. [fdb7870](https://github.com/StackExchange/Stacks/commit/fdb7870f0477882f6ede31cf0e0545ce8e1e36d5)

Thank you for this! I entirely forgot about `waitFor`. I think I was in "how do I get this test to work how I'd expect" mode and never got to "how to make this test work and be as stable as possible".

> @dancormier Tangential question, I noticed that we recommend to add a `aria-hidden="true"` in our markup examples but then we don't have any js code in our stimulus controllers that dynamically change that value when the tooltip is displayed/hidden. Should we log a bug?

Yeah, I think so. I'll go ahead and add a ticket.

Tangentially, it seems like the popover broadly shouldn't include the `aria-hidden` attribute since we hide it with `display: none`. See [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden):

> `aria-hidden="true"` should not be added when:
>
>    - The HTML [hidden](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) attribute is present
>    - **The element or the element's ancestor is hidden with [display: none](https://developer.mozilla.org/en-US/docs/Web/CSS/display)**
>    - The element or the element's ancestor is hidden with [visibility: hidden](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility)

giamir (approved):

PR #1857, opened 2024-11-14
feat(button): add "featured" variant

dancormier:
[STACKS-680](https://stackoverflow.atlassian.net/browse/STACKS-680)

## TODO

> [!NOTE]
> I've opened this PR up for review but there's a couple of outstanding items that need to be resolved:

- [x] Verify name "featured"
- [ ] Write description in docs

---

This PR adds a "featured" (purple) variant to the button component.

<details>

## Light

image

## Dark

image

## Light high contrast

image

## Dark high contrast

image

</details>

## Testing

1. Go to `/product/components/buttons/#featured`
2. Verify the newly added _featured_ buttons look and act as expected

### High contrast filled background color

~~To meet accessibility requirements, the featured filled variant background color has been set to `--purple-500` (normally `--purple-400`).~~

We've modified the `purple-400` light HC mode color instead. See https://github.com/StackExchange/Stacks/pull/1857#pullrequestreview-2466079719

### Borders on filled, selected buttons

This PR includes a minor fix to button border colors when hovered and filled and/or selected. Previously, the value of `--bu-bc` (`currentColor`) was applied and now the border color applied is equal to `--bu-bg`, ensuring the border and button background are unified.

### Button group visual regression test images

There are six visual regression tests (all on `s-btn-group-highcontrast-*` in Chromium runs) that have been updated. They are visually identical to me but there could be extremely slight differences due to https://github.com/StackExchange/Stacks/pull/1857/commits/838e25ad90ab4b68e52b54c2e89e2b8b43bf4054.


docs/product/components/buttons.html
```diff
@@ -119,6 +120,61 @@
         </div>
     </div>
 
+
+    {% header "h3", "Featured" %}
+    <!-- TODO add featured button description -->
+    <p class="stacks-copy">Featured buttons are…</p>
```
dancormier: @CGuindon Here's where the description for the "featured" buttons would go. Feel free to make a commit or pass a description along to me and I'll plug it in.

docs/_data/buttons.json
```diff
@@ -35,6 +35,26 @@
         }
       ]
     }
+  ]
+  ,
+  "featured": [
```
dancormier, line 40: I went with `featured` for this variant name. 

I figured it's best to avoid coupling naming with visual that could change (`purple`) and `new` felt too restrictive since I could see this button being used just to grab some attention generally. `featured` felt like the best fit but I'm glad to change it if you feel differently @CGuindon.

lib/components/button/button.a11y.test.ts
```diff
@@ -14,8 +14,8 @@ describe("button", () => {
             /s-btn-(light|dark).*?badge/,
             // matches tests with a badge in highcontrast-light modes, excluding filled, danger, github, facebook, sm, or xs
             /s-btn-highcontrast-light-(?!.*(filled|danger|github|facebook|sm|xs)).*?badge/,
-            // matches tests with a badge in highcontrast-light modes, are muted and/or outlined, and are sm or xs
-            /s-btn-highcontrast-light-(?:muted-outlined-|muted-|outlined-)?(?:sm|xs).*?badge/,
+            // matches tests with a badge in highcontrast-light modes, are muted or featured and/or outlined, and are sm or xs
```
giamir, line 17: Out of curiosity what's the contrast ratio for the failing tests? 
By glancing at the UI it doesn't seem we are a lot off from the 7:1 target.

docs/_data/buttons.json
```diff
@@ -85,6 +105,16 @@
           "class": "s-btn__danger",
           "class2": "s-btn__filled"
         },
+        {
```
giamir, line 108: We are missing the `Featured, Clear` combination

docs/_data/buttons.json
```diff
@@ -127,6 +157,16 @@
...
```
giamir, line 164: We are missing the `Featured, Clear` combination

docs/_data/buttons.json
```diff
@@ -169,6 +209,16 @@
...
```
giamir, line 220: We are missing the `Featured, Clear` combination

giamir (commented):
Overall looks good, there is only a minor docs correction to do I left in comment.

Nice to see the PPCP abstraction making our lives easier here. 
Well done @dancormier ❤️

lib/components/button/button.a11y.test.ts
dancormier: Yeah, the featured variant seems to land at 6.61 in the failing cases. It's from the button badge opacity being set to `.8` in high contrast mode.

_Personally_, I'd be fine to bump that to `.85` or `1`, but we'd either need to add an exception for the `featured` variant for this or change the button badge opacity for all buttons in HC mode. I'd prefer the latter but I'd want to get approval from design before making either change. Thoughts?

giamir: I also think the latter is the better option. Then let's consider this out of scope for this PR and maybe let's have a chat with design about it at some point. Thanks

giamir (approved):

docs/product/components/buttons.html
CGuindon: I can't push to Stacks so here you go — let me know what you think. I looked at the other buttons and the color purple description in the color fundamentals:

> Featured buttons are a secondary button style, used to visually draw attention to something new or temporary, usually as part of onboarding or to announce a new feature. These should be used sparingly, and permanent placements should be avoided

CGuindon (commented):
Ok so the only thing I don't love is that in HC mode, the filled default state is the same color as the filled selected state (no difference). 

Because we had to bump the background color from 400 to 500 for HC filled, and because 500 and 600 are the same in HC, we end up with no difference there.

For Base and Danger we have a modified red-400 and blue-400 just for HC mode. Those colors were modified from Light/Dark so that they meet the 7:1 ratio (they're pretty close but not quite otherwise). And so for those variants we're able to keep the 400 stop as the background color for filled default and therefore have a color difference with the selected state.

So....

I'm proposing we modify HC modes purple 400 color to meet a 7:1 ratio with white.
Change the selected state back to using purple-400 but with this new color (in dark mode the purple 400 already passes 7:1 ratio.

- HC Light purple-400 would become #3B41BA ([Figma where I played with this](https://www.figma.com/design/X2txHfaagLc3lMSwnjBZnT/Colors%3A-Accessible-Color-Palette-v2?node-id=2170-14817&t=flVejKXMnu6uG3XU-1))
- HC Dark purple-400 would become  NO CHANGE NEEDED? since it already passes the 7:1 ratio?

---

Screenshot 2024-11-27 at 3 08 26 PM

---

Screenshot 2024-11-27 at 3 08 47 PM

PR #1859, opened 2024-11-22
docs(fundamentals): fix example-row-detail positioning

dancormier:
[STACKS-657](https://stackoverflow.atlassian.net/browse/STACKS-657)

---

This PR resolves positioning issues with the annotations used in color fundamentals docs page.

### Before

image
image

### After

image
image

### How to test

1. Spin this PR up locally (`npm run start`)
2. Head to the [_Color Fundamentals page_](http://localhost:8080/product/foundation/color-fundamentals)
3. Verify that the annotations under each of the [monochromatic color strips](http://localhost:8080/product/foundation/color-fundamentals/#backgrounds) are positioned as expected
4. **Bonus**: Since this fix necessitated changes to the [accessibility docs](http://localhost:8080/product/foundation/accessibility/#foundation-for-custom-approach) code, verify that annotations under examples are positioned as expected


docs/product/foundation/accessibility.html
```diff
@@ -108,7 +108,7 @@
                 </button>
                 {% render 'example-row-detail.html',
                     text: group.description,
-                    style: "margin-left: 50%;",
+                    classes: "mx-auto",
                     customHeight: 16
```
giamir: **Issue:** Looking at the `example-row-detail.html` partial I think that what we called `customHeight` in this file should be called `connectorHeight` instead. 

**Sidenote (out of scope)**
I think we could find a better name for this partial. It took me a while to understand what `Example Row Detail` even mean and I still have some doubt. Perhaps something like`annotation.html` or `connected-annotation.html` could work at least based on its usage here. Thoughts?

giamir (commented):
It looks good to me from a dev perspective. 
I have only left a small thing to address in a comment.

@CGuindon can give a final thumbs up from a design perspective.

docs/product/foundation/accessibility.html
dancormier: > **Issue:** Looking at the `example-row-detail.html` partial I think that what we called `customHeight` in this file should be called `connectorHeight` instead.

Ah, nice catch! I've resolved this in [this commit](https://github.com/StackExchange/Stacks/pull/1859/commits/ba66ab73a84bb05382d9d5b228b96da7166ab182).

> **Sidenote (out of scope)** I think we could find a better name for this partial. It took me a while to understand what `Example Row Detail` even mean and I still have some doubt. Perhaps something like`annotation.html` or `connected-annotation.html` could work at least based on its usage here. Thoughts?

I agree. I think `example-row-detail` is one of those placeholder names that just accidentally became semi-permanent. I think `annotation.html` or `example-annotation.html` makes the sense to me (I'd argue against `connected-annotation.html` to keep the name tied to purpose as opposed to presentation).

I'll make a follow up PR for that to keep this PR focused. Thanks for the feedback.

**Update**: Here's [the PR](https://github.com/StackExchange/Stacks/pull/1863)

giamir: Thanks for the PR, I have merged it now. 
You might want to rebase this branch then. Hopefully there are not nasty conflicts to fix. 🤞

giamir (approved):

PR #1862, opened 2024-11-26
build(netlify): change screenshots extension to .ico to avoid pulling them unnecessarily

giamir:
[STACKS-691](https://stackoverflow.atlassian.net/browse/STACKS-691)

This branch is changing the extension of our screenshots to something different than `.png` (specifically we went with `.ico`) because netlify is not honoring  the `GIT_LFS_FETCH_INCLUDE` env variable: this was causing the netlify agent git checkouts to timeout while trying to pull all our screenshots unnecessarily.


dancormier (approved):
🙏

PR #1870, opened 2024-12-11
fix(pagination): support accessible markup

dancormier:
[STACKS-687](https://stackoverflow.atlassian.net/browse/STACKS-687), [A11Y-122](https://stackoverflow.atlassian.net/browse/A11Y-122)

---

This PR introduces support for more accessible markup of the pagination component. Existing implementations using the previous markup should render as they did before.

## Notable changes

- In order for pagination item heights to render as desired with the new markup, this PR adds `display: inline-block` to these items.
- I've added a deprecation notice that recommends consumers migrate to the new markup.
- In addition to modifying the markup used in visual tests, I've added `legacy` markup for child elements using the deprecated structure. I figure this will help us identify any issues that could arise when rendering the deprecated markup.

## To test

- Visit the [pagination component documentation page](https://deploy-preview-1870--stacks.netlify.app/product/components/pagination/)
- Inspect the example pagination element to verify the new structure

### DIFF
--- a/docs/product/components/pagination.html
+++ b/docs/product/components/pagination.html
@@ -29,72 +29,113 @@ description: Pagination splits content into pages, as seen on questions, tags, u
 </section>
 <section class="stacks-section">
     {% header "h2", "Example" %}
+
+    {% tip, "warning", "mb24" %}
+        <strong>Note:</strong> As of Stacks 2.6.1, the recommended pagination component markup has been updated for better accessibility. Please update any existing implementations to the new markup.
+    {% endtip %}
+
     <div class="stacks-preview">
 {% highlight html %}
-<div class="s-pagination">
-    <a class="s-pagination--item is-selected" href="…" aria-current="page">
-        <span class="v-visible-sr">page</span>
-        1
-    </a>
-    <a class="s-pagination--item" href="…">
-        <span class="v-visible-sr">page</span>
-        2
-    </a>
-    <a class="s-pagination--item" href="…">
-        <span class="v-visible-sr">page</span>
-        3
-    </a>
-    <a class="s-pagination--item" href="…">
-        <span class="v-visible-sr">page</span>
-        4
-    </a>
-    <a class="s-pagination--item" href="…">
-        <span class="v-visible-sr">page</span>
-        5
-    </a>
-    <span class="s-pagination--item s-pagination--item__clear">…</span>
-    <a class="s-pagination--item" href="…">
-        <span class="v-visible-sr">page</span>
-        122386
-    </a>
-    <a class="s-pagination--item" href="…">
-        Next
-        <span class="v-visible-sr">page</span>
-    </a>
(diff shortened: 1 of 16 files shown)


docs/product/components/pagination.html
```diff
@@ -29,72 +29,113 @@
...
-    <a class="s-pagination--item" href="…">
-        <span class="v-visible-sr">page</span>
-        122386
-    </a>
-    <a class="s-pagination--item" href="…">
-        Next
-        <span class="v-visible-sr">page</span>
-    </a>
-</div>
+<nav class="s-pagination">
```
giamir: This is unlikely to be the only navigation landmark in the page so we might want to add an `aria-label="pagination"` in this example that people will copy/paste to prevent them from introducing another accessibility issue.

screenshots/Chromium/baseline/s-pagination-dark-legacy.ico
giamir, line 1: Oh no, github preview doesn't pick up images with the `.ico` extension. This is a bummer. Netlify really need to sort their issue with that `GIT_LFS_FETCH_INCLUDE` env variable not being honored. 😔

docs/product/components/pagination.html
```diff
@@ -29,72 +29,113 @@
 </section>
 <section class="stacks-section">
     {% header "h2", "Example" %}
+
+    {% tip, "warning", "mb24" %}
```
giamir: I don't think this note makes a lot of sense here in isolation since it is unlikely people will remember the old markup. 
Maybe we could have a section called legacy example where we add the deprecation warning and that it will be removed in future releases.

This is what I am suggesting from a content hierarchy perspective:

- Example 
  - New markup example
 
- Legacy Example
  - Deprecation warning message
  - Old markup example 
  
  What do you think?

giamir (approved):
Great work @dancormier. I have only left a couple of suggestions but I would be ok to merge the PR even as it is. I will leave it up to you if you think it makes sense to follow up on them.

Thanks for taking care of this a11y bug. 🎉

docs/product/components/pagination.html
dancormier: Good call. I've added `aria-label="pagination"` to example code and `role="presentation"` to rendered examples.

screenshots/Chromium/baseline/s-pagination-dark-legacy.ico
dancormier: Maybe there's an image format that will render. I might spend a moment messing around with the extension to find out.

docs/product/components/pagination.html
dancormier: Sounds good to me. I've made the change

dancormier:
@giamir thank you for your review. I've made the changes you've suggested and I plan to merge this today. If you see anything you think could use revision, let me know!

PR #1883, opened 2025-02-21
fix(button): fix anchor hover + visited font color

dancormier:
The button component does not receive the correct styling when it meets the following conditions:

- It is in a `:hover` state
- It's rendered as an `<a />`
- It has the `.s-btn__filled` class applied
- It's `:visited`

This is due to the styles from `:not(.s-btn__outlined)` overriding the styles applied to `.s-btn__filled`.
### Relates to

- https://github.com/StackExchange/Stacks/pull/1879
- 🔒 [Slack discussion](https://stackexchange.slack.com/archives/C27RWNQN9/p1737562461312659)


giamir (approved):
I can confirm this fix the hover+visited issue for links that looks like filled buttons.

Maybe the only thing we could in this PR or another one is to have a place in the doc where we document and showcase links that have the same styles as a button. Maybe in the links page.

Thanks @dancormier

PR #1886, opened 2025-02-24
chore(tests): upgrade web test runner and playwright

giamir:
Our stacks pipeline has not run in a while because we have not done active development in this repo for some time. 

We were failing behind on the playwright versions and the new ubuntu runners are not supporting those anymore. This PR is upgrading dependencies related to that.

On top of that it adds a scheduled run of our testing pipeline every working day so that we can address issue early on by receiving email notifications of pipeline failures. 

Apparently the new version of the browsers installed by the new playwright is slightly adjusting the rendering of many of our baseline image and this is the reason there 3k+ file changes.


.github/workflows/workflow.yml
```diff
@@ -16,23 +20,30 @@ jobs:
...
+          needs_lfs: false
           needs_playwright: true
         - command_description: A11y Tests
           command: npm run test:a11y -- --config web-test-runner.config.ci.mjs
+          needs_lfs: false
           needs_playwright: true
         - command_description: Visual Regression Tests
           command: npm run test:visual -- --config web-test-runner.config.ci.mjs
-          needs_playwright: true
+          needs_lfs: true
```
giamir, line 43: Save time on all the other jobs by not pulling down all the images for the visual regressions tests

lib/components/modal/modal.test.ts
```diff
@@ -80,7 +80,11 @@ describe("modal", () => {
 
         await user.click(closeBtn);
 
-        await waitFor(() => expect(modal).not.to.be.visible);
+        // TODO: undestand why only in CI, for webkit the following assertion fails
```
giamir, line 83: This is a odd one. For webkit and in in CI only the visible assertion does not seem to work correctly. Locally it works fine and it's only happening in this test. For now I decided to assert on the attribute aria-hidden that it used by the styles to determine the visibility.

package.json
```diff
@@ -110,6 +110,7 @@
         "not Android > 0"
     ],
     "overrides": {
-        "aria-query": "~5.0.0"
+        "aria-query": "~5.0.0",
+        "@web/dev-server-core": "0.7.1"
```
giamir, line 114: https://github.com/modernweb-dev/web/issues/2772#issuecomment-2301837407

Newer versions of this transitive dependencies make the output of failing tests very hard to debug.

PR #1891, opened 2025-03-10
fix(anchors): ensure anchor styles take priority over `<a>` styles

dancormier:
This PR resolves an issue where styles that should be applied to `.s-anchors a:visited` were being superseded by styles on `a` (see [link.less](https://github.com/StackExchange/Stacks/blob/develop/lib/components/link/link.less#L3-L19)). The issue reports issues specifically with `.s-anchors__inherit` but some investigation revealed that styles were not being applied as expected with _all_ `.s-anchor` variants.

## Screenshots

> [!NOTE]
> All screenshots are within containers with the classes `.s-notice.s-anchors.s-anchors__underline`.

<details>
<summary>Before</summary>

### .s-anchors__inherit
image

### .s-anchors__danger
image

</details>

<details>
<summary>After</summary>

### .s-anchors__inherit
image

### .s-anchors__danger
image

</details>

## Note on tests

This PR does not include any modification of tests despite fixing a styling bug. We currently do not have a way to test links with pseudo classes (in this case, `:visited`) applied.

## How to test

- Ensure the [Stacks docs](https://stackoverflow.design/product/develop/using-stacks/) have been visited in your current browser
- Spin up the docs locally (`npm run start`)
- Open any Stacks docs page (such as [links.html](https://github.com/StackExchange/Stacks/blob/develop/docs/product/components/links.html))
- Insert the below HTML and save the change

```html
    <div class="s-notice s-notice__important s-anchors s-anchors__inherit s-anchors__underlined">
        <a href="https://stackoverflow.design/product/develop/using-stacks/">Link within `.s-notice.s-anchors.s-anchors__inherit`</a>
        <br/>
        <button class="s-btn s-notice--btn js-notice-seen">.s-btn.s-notice--btn</button>
        <br/>
        <a href="https://stackoverflow.design/product/develop/using-stacks/">.s-btn.s-notice--btn visited</a>
        <br/>
        <a href="#unvisited">.s-btn.s-notice--btn unvisited</button>
    </div>
```
- Visit the page you edited
- Observe the links within the added s-notice element inherit the link styles
- (Bonus): Change `.s-anchors__inherit` on the s-notice element to any other anchors variant and observe that links render as expected

---

[STACKS-708](https://stackoverflow.atlassian.net/browse/STACKS-708)

### DIFF
--- a/lib/components/link/link.less
+++ b/lib/components/link/link.less
@@ -12,7 +12,7 @@ a {
         // Otherwise, use the default link colors.
         &:active,
         &:hover {
-            color: var(--_an-a-fc-hover, --_li-fc-hover);
+            color: var(--_an-a-fc-hover, var(--_li-fc-hover));
         }    
 
         &:visited {
@@ -20,7 +20,7 @@ a {
         }
 
         &:hover:visited {
-            color: var(--_an-a-fc-hover-visited, --_li-fc-hover-visited);
+            color: var(--_an-a-fc-hover-visited, var(--_li-fc-hover-visited));
         }
     }
 }
--- a/lib/components/link/link.less
+++ b/lib/components/link/link.less
@@ -1,19 +1,26 @@
 // TODO we *really* shouldn't be apply styles directly onto `<a>` like this, but
 // it's tech debt that'll take some doing in consumer's code to pay down.
 a {
-    &:visited {
-        // We're target these specific selectors to avoid affecting the visited state of stacks components
-        // not specified here. See for https://github.com/StackExchange/Stacks/pull/1740#discussion_r1698389312
-        // TODO remove .post-tag, .badge, .badge-tag reference once core no longer requires them
-        &:not([class*="s-"]):not(.post-tag):not(.badge):not(.badge-tag),
-        &.s-link,
-        &.s-sidebarwidget--action,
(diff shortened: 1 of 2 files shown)


lib/components/link/link.less
```diff
@@ -133,4 +134,10 @@ a,
     p & {
         text-decoration: underline;
     }
+
```
giamir, line 137: @dancormier This is so these interaction styles apply only for elements with the `s-link` class applied and not all `a` elements correct?

giamir (approved):
Work as expected. Thanks @dancormier for taking care of this fix. ❤️

lib/components/link/link.less
dancormier: Yep, the changes in this PR include a separate rule for `a` elements (lines 13-16 in this file) with the reference to the `.s-anchor` PPCP (`--_an-a-fc-hover`):

```less
        &:active,
        &:hover {
            color: var(--_an-a-fc-hover, var(--_li-fc-hover));
        } 
```

In contrast to `a`, we want styles applied to `.s-link` to override `.s-anchor a` styles, we'd want it to override any anchor styles, so I omitted the `.s-anchor` PPCP here.

PR #1893, opened 2025-04-01
fix(a11y): Increase contrast for the navigation component selected muted style

qheaden-stack:
The muted variant of the selected navigation component style does not meet the  WCAG Non-Text Contrast success criterion. This PR updates the colors for that variant to fix this issue. It also re-enables the test for it.

Here are comparisons between the old and new styling:

### Non-Hover
Before:
selected-muted-light-old
After:
selected-muted-light-new

### Hover

Before: 
selected-muted-light-hover-old

After:
selected-muted-light-hover-new

### Non-Hover
Before:
selected-muted-dark-old
After:
selected-muted-dark-new

### Hover
Before:
selected-muted-dark-hover-old

After:
selected-muted-dark-hover-new


CGuindon:
@qheaden-stack Instead of changing the color of the text, let's match what we do on the left navigation items — make them bold when selected/active. The "boldness" is a change that isn't color alone and covers us for the requirement (3:1 ratio no longer needed). It avoids introducing an additional color (different shade of grey) to mean something and will be consistent with some of the existing navigation components (left nav and button group).

Screenshot 2025-04-01 at 10 08 58 AM

CGuindon (changes requested):
(Oops, commented in the wrong spot, see above)

CGuindon:
@qheaden-stack In fact.... (if it's easy) — would it be possible to make all variants of the navigation component (not just muted) have **bold** text for the active/selected state? So the default nav (orange pill on selected/active) would be bold too?

qheaden-stack:
Closing this in favor of https://github.com/StackExchange/Stacks/pull/1894 where we make the text bold instead.

PR #1726, opened 2024-05-01
docs(button-group): Add documentation for `data-text`

dancormier:
Addresses [https://stackoverflow.atlassian.net/browse/STACKS-588](https://stackoverflow.atlassian.net/browse/STACKS-588)

This PR adds a description of the `data-text` attribute to the button group documentation.

### To test

1. Hit up the [button group docs page](https://deploy-preview-1726--stacks.netlify.app/product/components/button-groups/#button-attributes) in the deploy preview.
2. Find and review the "Preventing layout shift" section.

PR #1894, opened 2025-04-04
fix(a11y): Add bold font to selected nav items

qheaden-stack:
[STACKS-755](https://stackoverflow.atlassian.net/browse/STACKS-755), Partially resolves [A11Y-15](https://stackoverflow.atlassian.net/browse/A11Y-15)


dancormier (commented):
@qheaden-stack thank you for taking this on! I neglected to mention that the `data-text` approach requires the addition of a child element to contain the text and the `:before` pseudo-element (I actually totally forgot this) like in the [button group](https://stackoverflow.design/product/components/button-groups/) buttons. 

In any case, I've added a [commit](https://github.com/StackExchange/Stacks/pull/1894/commits/1c6cd65dda8ba15af8c780bb8d562b26d9992cf8)* that adds this element (`.s-navigation--item-text`) and shuffles around the Less as needed to support it. I'm happy to go over these changes with you if you'd like more detail. Feel free to change anything you see fit. Once this is ready for review, I'm happy to do so but I'd also include Giamir since I made a bunch of changes.

### What's needed before merging

- [x] Updated Navigation component documentation describing `data-text` (see [button group](https://github.com/StackExchange/Stacks/pull/1726/files#diff-0e5954c010bf5cdbbcc61d3569dc15e60db51191b51aaa2ac4397d9e998ffe08))
- [x] Updated visual regression images (I haven't updated the visual regression images since I'm not 100% that the design won't get modified)
- [x] Guidance from design about width increase due to this change (see https://github.com/StackExchange/Stacks/pull/1894#issuecomment-2779804322)

**Sidenote**: I just want to say thanks for picking up these a11y issues. You've been one of the few I've seen picking up issues (I know not everyone has enthusiasm and/or bandwidth for improving accessibility) and I'm glad you make the time and effort to improve our accessibility ❤️

---

*you can safe ignore my first two commits in this PR. I missed that all my changes included a change to line endings that created changes on every line. The last commit I made includes my changes _without_ this line changes

docs/_includes/header.html
```diff
@@ -9,12 +9,16 @@
 
         {% if site-navigation %}
             <nav aria-label="Global" class="sm:d-none">
-                <ul class="s-navigation ml8 fw-nowrap sm:d-none">
```
dancormier, line 12: The removal of `sm:d-none` here has nothing to do with the navigation component changes but the class is redundant here since the parent includes it too so I figured I might as well chuck it in the bin now.

lib/components/button-group/button-group.less
```diff
@@ -58,7 +58,6 @@
         & .s-btn--text {
             &:before {
                 content: attr(data-text);
-                content: attr(data-text) / "";
```
dancormier, line 61: I realized that we have a duplicate property here, so I got rid of it in https://github.com/StackExchange/Stacks/pull/1894/commits/56a606c9251371cc4fc1a9bbecd308d79a712442

dancormier:
@CGuindon I wanted to raise a design concern with you and see how/if we want to address it.

With the addition of the hidden bold text to maintain a consistent size for navigation items between selected states, the items are now all a bit wider:

### Before (361px wide)

image

image

### After (378px wide)

image

image

This has two potential issues:

1. This is not the intended design
2. This could possibly cause layout shifts from what we currently have (this could make the update in Core kinda a pain if we need to mitigate any new undesirable overflow/wrapping)

This issue could potentially be partially mitigated by reducing the gap between navigation items:

### With gap reduced from 4px to 1px (366px wide)

image

image

The compromise is that hover state reveals that the navigation items are closer together. I'm happy to brainstorm other ideas or just leave it as-is if that seems right.

CGuindon:
@dancormier I don't think we can reduce the spacing to 1px — I _believe_ this might cause issues with the focus state (the blue ring adds 1-2 px on the outside sometimes?). I'm honestly _not too worried_ about the navigation get a tiny bit wider. Navigations are already wrapping for responsive screen sizes — I'd be curious to first see if there are any pages where the nav perfectly fits right now and it becomes a big issue for a regular desktop size.

giamir (commented):
The changes made sense to me. Thanks for picking this up.
Before merging we need to add a [changeset](https://github.com/StackExchange/Stacks/pull/1894#issuecomment-2830355100).
Also it would be nice to know ahead of time if there is something special we need to watch out for when performing the upgrade in Core. If no side effects are expected we can also cut a release straight away.

dancormier:
**Note**: This change is ready but we're delaying merging until prep work has been completed in Core.

PR #1950, opened 2025-07-11
docs(notice): add suggestion to apply s-anchors

dancormier:
This PR updates the notice docs to suggest using `.s-anchor` classes to style child links.

See also: [🔒 Slack discussion](https://stackexchange.slack.com/archives/C27RWNQN9/p1752247460586519)

### How to test
- View the [notice docs](https://deploy-preview-1950--stacks.netlify.app/product/components/notices/#classes)
- Verify that the tip suggesting `.s-anchors` on notices is appropriate
- Verify that example code snippets are appropriate

### DIFF
--- a/docs/product/components/notices.html
+++ b/docs/product/components/notices.html
@@ -239,13 +239,19 @@ description: Notices deliver <strong>System</strong> and <strong>Engagement</str
         </table>
     </div>
 
+    {% tip, "", "mb24" %}
+        <p><strong>Note:</strong> We recommend including <code class="stacks-code">.s-anchors.s-anchors__inherit.s-anchors__underlined</code> classes for appropriate child link styling for notices containing links.</p>
+    {% endtip %}
+
     {% header "h3", "Filled" %}
     <p class="stacks-copy">Default style used to separate notices from the main content</p>
     <div class="stacks-preview">
 {% highlight html %}
 <div class="s-notice" role="status">…</div>
 <div class="s-notice s-notice__info" role="status">…</div>
-<div class="s-notice s-notice__success" role="status">…</div>
+<div class="s-notice s-notice__success s-anchors s-anchors__inherit s-anchors__underlined" role="status">
+    … <a href="…">…</a>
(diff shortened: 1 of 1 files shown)


giamir (approved):
Happy to see this merged. Thanks @dancormier. 

### Sidenote (out of scope for this PR)
I wonder if the default behavior of `s-anchors` shoud be to make the link underlined to help developers "fall in the pit of success" (Making the link underlined is necessary when it is surrounded by other text for a11y - https://stackoverflow.design/product/components/links/#accessibility)

KyleMit (commented):
Left a comment.  I think this is a reasonable quick fix.  Don't want to draw out scope too much, but I think we could probably do even more to highlight this pattern with the docs, but also agree with @giamir that Pit of Success would be a real win here.

docs/product/components/notices.html
```diff
@@ -239,13 +239,19 @@
         </table>
     </div>
 
+    {% tip, "", "mb24" %}
+        <p><strong>Note:</strong> We recommend including <code class="stacks-code">.s-anchors.s-anchors__inherit.s-anchors__underlined</code> classes for appropriate child link styling for notices containing links.</p>
+    {% endtip %}
+
```
KyleMit: Here's my silly end user feedback on UI / copy.  I probably would have missed this even if it was included.  For some reason, my brain skips right past an informational note (I know bad, but maybe I'm not alone).  I'm usually looking for headings when skimming docs.

## Before

Image

## After

I know this is just a quick fix in the first place, but I think giving it a little more space and examples would help a lot like this

Image

KyleMit: If those changes look helpful and consistent with the docs philosophy, you can pull them into this PR from #1955

dancormier: If it helps our devs find what they need and build UIs in a more robust manner, I'm all for it. I'm merging in your PR then merging this one. Thanks for the extra consideration @KyleMit!

PR #1958, opened 2025-07-21
chore(modal): remove unneeded will-change CSS property

dancormier:
We've had the `will-change` CSS property on our modals for some ~7 years. Seems it's not necessary and possibly detrimental to performance (see [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change)).

This PR moves that CSS property. See also [this Meta post](https://meta.stackexchange.com/questions/411541/stop-using-will-change-in-s-modal-or-give-me-back-my-ram).

### DIFF
--- a/.changeset/petite-beans-start.md
+++ b/.changeset/petite-beans-start.md
@@ -1,0 +1,5 @@
+---
+"@stackoverflow/stacks": patch
+---
+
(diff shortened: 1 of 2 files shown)


giamir (approved):
Tested, I did not find any regression. Thanks @dancormier

PR #1963, opened 2025-08-05
SPARK-8: Fix Modal Focus Issue when using Keyboard

mukunku:
# Summary

This is for [SPARK-8](https://stackoverflow.atlassian.net/browse/SPARK-8).

This is an issue with Firefox. If the modal doesn't have `tabindex` set, when anything not tabbable within the modal is selected/focused, like raw text, and a key is pressed the event target will always be the `body` of the page. This causes a bug with the `keepFocusWithinModal` event handler where no matter what key you press the first tabbable element gets focused within the modal if you're not already focused on a tabbable element, deselecting any text you may have highlighted.

# Solution

The easiest solution is to only handle `Tab` key presses in this handler; not sure if there's a reason to handle any other kind of key presses. This addresses the original reported issue.

Additionally, while unlikely, I also added code to handle the case where the modal has no tabbable items; including the close button in the top-right. Without this additional code, if you have a modal with no `button`s and press `Tab` focus will go through elements in the main page, behind the modal. With the code, the `Tab` keypress becomes a no-op if there's nothing to switch focus to.

# How to Test

To re-create the issue:
* Launch the following codepen in firefox: https://codepen.io/KyleMit/pen/myemPmE?editors=1010
* Open the modal and highlight some text
* Hit any key on your keyboard and notice the text selection get cleared as focus is moved to the link in the modal. 
https://deploy-preview-1963--stacks.netlify.app/product/components/modals/#examples

To test the fix:
* Launch the PR env for this branch in Firefox: https://deploy-preview-1963--stacks.netlify.app/product/components/modals/#examples
* Open a modal and select some text
* Hit any key on your keyboard besides `Tab` and notice nothing changes
* Only `Tab` key presses will mess with focus changes now.


webpack.config.js
```diff
@@ -37,7 +37,7 @@ const baseConfig = (isProd, minify) => ({
             {
                 test: /\.less$/,
                 use: [
-                    MiniCssExtractPlugin.loader,
+                    isProd ? MiniCssExtractPlugin.loader : "",
```
dancormier, line 40: I noticed the commit message for this was to "fix webpack.config.js for windows". I'm curious what was happening initially (for the record, this looks good. I'm just personally curious).

dancormier (approved):
The fix is straightforward and refined, the tests are thorough, and after testing it in Firefox I can report that it works as expected. 

Great work here @mukunku! Happy to see this merged in when you're feeling ready.

webpack.config.js
mukunku: When loading the design site locally none of the css would load with the following error:
```
Refused to apply style from 'http://localhost:8080/xxxxxx' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled.
```
Giamir gave me the solution to disable the css plugin locally as he encountered the same issue when assisting Tavian.

I believe it may be related to this as the root cause: https://github.com/webpack/webpack-dev-server/issues/3168#issuecomment-816580916

PR #1965, opened 2025-08-07
[SPARK-6] - update the notice border color for dark mode and high contrast dark mode

ttaylor-stack:
Updated the notice border color for dark mode and high contrast dark mode as specified in the the story https://stackoverflow.atlassian.net/browse/SPARK-6

Also updated README to correct the port to use with localhost and add `npm i` to build instructions


lib/components/notice/notice.less
```diff
@@ -17,15 +17,9 @@
...
         --_no-code-bg: ~"var(--@{colorName}-200)";
 
-        .dark-mode({
-            --_no-bc: ~"@{darkMode}";
-        });
-
-        .highcontrast-dark-mode({
-                --_no-bc: ~"@{highContrast}"
+        .highcontrast-mode({
+            --_no-bc: ~"var(--@{colorName}-400)";
```
dancormier: @ttaylor-stack since all high contrast mode borders should be the `400` color stop for their given color, we can remove the need for the extra `@highContrast` param and just concat the color name to `-400`.

lib/components/notice/notice.less
```diff
@@ -64,13 +58,6 @@
     --_no-code-fc: var(--_no-fc);
 
     // CONTEXTUAL STYLES
-    .dark-mode({
-        --_no-bc: var(--black-225);
-    });
```
dancormier: We previously had a dark mode override here (https://github.com/StackExchange/Stacks/blob/develop/lib/components/notice/notice.less#L57-L59):

```less
    .dark-mode({
        --_no-bc: var(--_no-bg);
    });
```

Both light and dark modes use `--black-225` for border color (base style when not high contrast). By getting rid of the dark mode override altogether, `var(--black-225)` becomes the default value and is simpler to override elsewhere because it's targeting a less specific selector (`.s-notice` instead of `body.dark-mode .s-notice`).

lib/components/notice/notice.less
```diff
@@ -87,6 +74,8 @@
                 --_no-code-fc: var(--white);
             }
         }
+
+        --_no-bc: var(--black-400);
```
dancormier, line 78: This applies `--_no-bc: var(--black-400);` to the base high contrast style which can then be overridden by the `.highcontrast-mode` style rules defined in `.generate-variant-variables` (line 20).

dancormier (commented):
Really nice work @ttaylor-stack. Working on styles for this specific component is like playing on hard mode.

I think we can achieve the desired result without needing to add params to `.generate-variant-variables`. I was going to make suggestions inline but GitHub didn't make that easy in this case, so I just made a commit (https://github.com/StackExchange/Stacks/pull/1965/commits/678326d06a3bc35e33133fb66cba733900ed37f0). I added a few notes in comments on that commit.

Check out the commit and let me know if anything seems off or could use clarification. If you're feeling good about the changes, we can go ahead and merge this branch.

### Visual regression tests

I've also updated the visual regression tests images to represent the new style (https://github.com/StackExchange/Stacks/pull/1965/commits/0afed48cca6143c384c62ce9beec0044d95ed8b9). There's some instruction in the README on [how to setup and run visual regression tests](https://github.com/StackExchange/Stacks?tab=readme-ov-file#visual-regression-tests). Let me know if you have issues running them (or if you have any other questions) and I'd be happy to help.

> [!NOTE]
> You'll notice lots of new images for both notice _and_ banner components. That's because the banner styles are lifted from notice, so changes applied to notice will also apply to banner. Just a heads up.

README.md
```diff
@@ -42,9 +42,10 @@ To migrate from Stacks v1 to v2, see our [migration guide](/MIGRATION_GUIDE.md).
 
 To contribute to Stacks documentation or its CSS library, you’ll need to build Stacks locally:
 ```sh
+npm i
 npm start
 ```
-This command will pull up the local dev server at http://localhost:8000. You can also view our [building guidelines](https://stackoverflow.design/product/develop/building).
+This command will pull up the local dev server at http://localhost:8080. You can also view our [building guidelines](https://stackoverflow.design/product/develop/building).
```
dancormier, line 48: Thank you @ttaylor-stack! Sorry if this typo caused you any trouble.

PR #1969, opened 2025-08-26
SPARK-12: Move to monorepo setup by splitting Stacks Docs and Classic into NPM workspaces

mukunku:
# Summary

This is for [SPARK-12](https://stackoverflow.atlassian.net/browse/SPARK-12) and is being done in preparation for project SHINE.

This PR transitions the stacks classic and stacks docs projects into the [npm workspaces](https://docs.npmjs.com/cli/v9/using-npm/workspaces) structure. This allows several different packages to be maintained in a single monorepo, which we will need once we start bringing stacks svelte, etc into this repo.

# How to Test

Testing is straightforward: Everything should work the same as it did before.
* Click around the PR environment to confirm it's good
* Run some npm scripts locally like tests to confirm they work
* Run a local build and make some changes to confirm your local build environment behaves the same as it did before

# How to Review (To be updated)

Most of the file changes in this PR are from cheese being moved around. Mainly moving the `docs` and `lib` folders into new npm workspaces. Here are a list of file changes that are not part of this simple move; I left comments on the files themselves to share my thoughts and rationale for the changes.
<pre>
.github/
├─ workflows/
│  ├─ <a href="https://github.com/StackExchange/Stacks/pull/1969/files#diff-126bf89616b7daa3d14ebc882ad18666aaf1c3dae888c4ba306a66ec80758bc1">workflow.yml</a>
<a href="https://github.com/StackExchange/Stacks/pull/1969/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947">.gitignore</a>
adrs/
├─ <a href="https://github.com/StackExchange/Stacks/pull/1969/files#diff-d3e16d03a810103cce46acaa31e0496b380112826466211de4b5e5d4881b636f">0001-pseudo-private-custom-properties.md</a>
├─ <a href="https://github.com/StackExchange/Stacks/pull/1969/files#diff-b5024267ef99faaf1fb02c68719a480b5901d3d786924289820df92618473772">0003-monorepo.md</a>
<a href="https://github.com/StackExchange/Stacks/pull/1969/files#diff-ab8f79b68b7adff7a07db953bf453f3c5aa6ade98d2b1b67d8432b36392489ed">netlify.toml</a>
<a href="https://github.com/StackExchange/Stacks/pull/1969/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519">package-lock.json</a>
<a href="https://github.com/StackExchange/Stacks/pull/1969/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519">package.json</a>
packages/
├─ stacks-classic/
│  ├─ <a href="https://github.com/StackExchange/Stacks/pull/1969/files#diff-b81eb7715492373027a755ed64786bd15421bf009ac014e10af5521aca6fe194">package.json</a>
├─ stacks-docs/
│  ├─ <a href="https://github.com/StackExchange/Stacks/pull/1969/files#diff-f7340a9a5e0c24bb5da1c96c3fd7212bf8ffce26eccbc2824ac2f65f26ca2cad">package.json</a>
<a href="https://github.com/StackExchange/Stacks/pull/1969/files#diff-f4a850e349ee5e31f155cc47316b4db400a848649a7391915d8f446f5d3e4be1">run-test-visual.ps1</a>
<a href="https://github.com/StackExchange/Stacks/pull/1969/files#diff-7bfdd444e5650854e7536869ef3973956e89fbfc8562197c7c13f99cfb3047bf">web-test-runner.config.mjs</a>
</pre>


adrs/0003-monorepo.md
giamir, line 1: > In regards to changelog / release notes generation, we'd continue using Conventional Commits, but moving to use [release-please](https://github.com/googleapis/release-please) which [does support monorepos](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md).

I think this sentence in the RFC should be updated since we have already introduced changesets and we won't move away from it


> For backwards compatibility, we can trivially create a @stackoverflow/stacks "meta-package" that simply imports, then re-exports the dependency packages.

This part I struggle to understand / remember what it means. I would recommend to remove it unless somebody in the team recall what we meant with it.

package.json
```diff
@@ -1,48 +1,39 @@
 {
+    "private": true,
```
giamir, line 2: This repo is not private, I don't think this is necessary here.

package.json
```diff
@@ -1,48 +1,39 @@
...
+    "workspaces": ["packages/*"],
     "name": "@stackoverflow/stacks",
```
giamir: I would actually set the name for the root package.json just `stacks` and reserve the name `@stackoverflow/stacks` to the package.json you have in the `packages/stacks-classic` folder. This is because if we don't do that we will create a breaking change where consumers will need to start consuming `@stackoverflow/stacks-classic` instead.

packages/stacks-classic/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "main": "./dist/js/stacks.js",
+    "types": "./dist/index.d.ts",
+    "style": "./dist/css/stacks.css",
+    "less": "./lib/stacks.less",
+    "unpkg": "dist/css/stacks.min.css",
+    "dependencies": {
+        "@popperjs/core": "^2.11.8"
+    },
+    "peerDependencies": {
+        "@hotwired/stimulus": "*"
```
giamir: I believe `stimulus` is an actual dependency of this package. We don't expect consumers to install stimulus themselves.

packages/stacks-classic/package.json
```diff
@@ -0,0 +1,21 @@
+{
+    "name": "@stackoverflow/stacks-classic",
```
giamir: keep here `@stackoverflow/stacks` to avoid breaking changes for now.

packages/stacks-docs/package.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "@stackoverflow/stacks-docs",
+    "description": "Stacks Design website. Contains all documentation around using Stacks and its components.",
+    "version": "2.8.4",
+    "dependencies": {
+        "@stackoverflow/stacks-classic": "*"
+    },
+    "peerDependencies": {
```
giamir: don't think this is necessary here since stimulus should be a transitive dependencies of stacks-classic (`@stackoverflow/stacks`)

packages/stacks-docs/package.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "@stackoverflow/stacks-docs",
+    "description": "Stacks Design website. Contains all documentation around using Stacks and its components.",
+    "version": "2.8.4",
```
giamir: This package should not need a version since we are not going to release it

package.json
```diff
@@ -116,6 +107,7 @@
     ],
     "overrides": {
         "aria-query": "~5.0.0",
-        "@web/dev-server-core": "0.7.1"
+        "@web/dev-server-core": "0.7.1",
+        "playwright": "1.50.1"
```
giamir, line 89: Any reason why we are adding also playwright to the overrides?

package.json
giamir, line 1: The way that generally monorepo are structured is that workspaces (in our case stacks classic and stacks docs) contains the respective `scripts` and `dependencies` they need to operate. The package.json at the root level is usually pretty shallow and contains only devDependencies that are shared across all workspaces and scripts that are triggering related scripts from all the workspace packages (e.g. `"build": "npm run build --workspaces --if-present"`). 

You can take a look at what we are doing in the accessibility repo as an example to understand what I mean: https://github.com/StackEng/Accessibility/blob/main/package.json
You can also see that sometimes scripts int the package root are "shortcuts" to a command in a specific workspace package.
```
"start": "npm run start -w @a11y/app",
```

package.json
```diff
@@ -1,48 +1,39 @@
...
-        "start:eleventy": "cd ./docs && eleventy --serve",
+        "build:bundle": "webpack --mode=production --config ./packages/stacks-classic/webpack.config.js",
+        "build:docs": "webpack --mode=production --config ./packages/stacks-docs/webpack.config.js && cd ./packages/stacks-docs && eleventy",
+        "start:webpack": "webpack --watch --config ./packages/stacks-docs/webpack.config.js",
+        "start:eleventy": "cd ./packages/stacks-docs && eleventy --serve",
         "test": "npm run test:less && npm run test:unit && npm run test:a11y && npm run test:visual",
-        "test:a11y": "web-test-runner --group=a11y",
-        "test:unit": "web-test-runner --group=unit",
-        "test:unit:watch": "web-test-runner --group=unit --watch",
+        "test:a11y:ci": "cd ./packages/stacks-classic && web-test-runner --group=a11y",
```
giamir: We should avoid `cd`ing into folder and instead use the npm workspace functionalities. I have added a comment to this file that should help simplify all of this. 
I can imagine something like this here
```
"test:a11y": "npm run test:a11y --workspaces --if-present"
```
and keep the original script as it was in the `packages/stacks-classic/package.json`

The same goes with pretty much for all the other scripts: formatting, linting, etc...

giamir (changes requested):
Thanks @mukunku for making great progress on this. ❤️

I have left a few comments to address. The major one is about embracing the [npm workspaces cli](https://docs.npmjs.com/cli/v11/using-npm/workspaces) more to avoid having to cd into packages and changing our workflows more than necessary.

The rest of the stuff is pretty minor. I have not tested yet that the publishing workflow still work as expected. I will do that in a second review after the package.json files get adjusted.

giamir (commented):
Thanks @mukunku 

Notes from Giamir's second review pass (we can discuss and address those sync later):
- Update .gitattributes to point to new location of the screenshots `**/screenshots/**` to make sure the files are still treated using git LFS and don’t blow up the repo
- path in the `test.yml` workflow for uploading the regression test results should be adjusted to `packages/stacks-classic/screenshots`
- Maintain `npm start` on the root package.json as a shortcut to start up the docs site quickly for development
- Need to specify prettier config files also for the format script in stacks classic as well ` "format": "prettier --write ./lib --config ../../.prettierrc --ignore-path ../../.prettierignore"`
- cannot run the visual test via `npm run test:visual -w packages/stacks-classic` - need to specify the config file like we do in the GH workflow `--config ./packages/stacks-classic/web-test-runner.config.mjs` - A better solution would be actually to modify line 32 and 33 in the run-test-visual.ps1  to  -v "${current_dir}/packages/stacks-classic/web-test-runner.config.mjs:/app/web-test-runner.config.mjs" `
  -v "${current_dir}/packages/stacks-classic/web-test-runner.config.ci.mjs:/app/web-test-runner.config.ci.mjs" `
- `jquery`, `list.js` and `docsearch.js`  should move as dependencies of stacks-docs (be removed from root)
- We need to add a changeset with a patch release for stacks-classic and ensure that the release process still run smoothly https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md#i-am-in-a-multi-package-repository-a-mono-repo
- NIT:  for build, format and lint we could use `npm run <command> —workspaces —if-present` syntax
- NIT: we could add a `test` script in the root package.json in a similar fashion to what we do for lint, format and build already

mukunku:
Hey @giamir , just to follow up on our discussion today: I believe I addressed all the items you mentioned in your second review. Kindly take a look and let me know how the PR looks now 🙌🏼 
<hr>
I additionally created the changeset file now. I didn't see any workflow that runs until the PR is merged so I guess we'll know if it works after we merge?
<hr>

Finally, just wanted to share a recap of the `test:visual` script issue we discussed, here, for posterity:


It seems I broke the npm script locally while trying to get it to work in CI. As you mentioned when running visual tests locally the script wasn't picking up the web test runner configuration file. 

The main issue as discussed is caused by the relative node_module paths in [web-test-runner.config.ci.mjs](https://github.com/StackExchange/Stacks/blob/699adb6891723f660099d772713e53c85b564d9a/packages/stacks-classic/web-test-runner.config.ci.mjs#L1-L4)

When we run a different test like a11y tests via: `npm run test:a11y -w packages/stacks-classic -- --config web-test-runner.config.ci.mjs` it works because the `../../node_modules` path is correct in that case.

But when we run `npm run test:visual -w packages/stacks-classic` it fails because `node_modules` is in the same directory as the config files in the docker image: https://github.com/StackExchange/Stacks/blob/2942938a53af896d76929ed63633384c7a455f2f/run-test-visual.ps1#L15-L20

So to accommodate I had to mount the web-test-runner config files in sub-directories within the image so the paths align:
https://github.com/StackExchange/Stacks/blob/699adb6891723f660099d772713e53c85b564d9a/packages/stacks-classic/run-test-visual.ps1#L32-L33
I gave the path a unique name so it's not confused with existing file paths.

Consequently, I adjusted the path we pass in the test workflow:
https://github.com/StackExchange/Stacks/blob/699adb6891723f660099d772713e53c85b564d9a/.github/workflows/workflow.yml#L42
And locally:
https://github.com/StackExchange/Stacks/blob/699adb6891723f660099d772713e53c85b564d9a/packages/stacks-classic/package.json#L22

giamir (approved):
Thanks @mukunku. Let's merge this. 🎉

> I additionally created the changeset file now. I didn't see any workflow that runs until the PR is merged so I guess we'll know if it works after we merge?

Yes, that's fair.

PR #1970, opened 2025-09-03
A11Y-128: Fix aria-current attribute placement in sidebar widget navigation examples

qheaden-stack:
This PR updates the sidebar widget navigation example to place the `aria-current="page"` attribute on the `<a>` tag itself instead of its surrounding `<li>` tag. The [WCAG guidelines](https://w3c.github.io/aria/#aria-current) mention this attribute applies to the relevant item, which in this case would be a link to the current page, and not its container.


dancormier (approved):
Verified the approach by referencing [MDN's `aria-current` example](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-current#example). These changes look good and are in line with my expectations. Thanks @qheaden-stack!

PR #1982, opened 2025-09-12
chore(docs): add readme for each workspace

mukunku:
# Summary

We split the repo into workspaces with #1969 but forgot that each workspace/package now needs its own readme file.

This PR copies the main readme's relevant sections and splits it into separate readme's for the packages.


giamir (approved):
Looks good

PR #1991, opened 2025-10-01
SPARK-48: Publish Stacks-Svelte

mukunku:
# Summary

This PR exposes our internal Svelte implementation of the Stacks design library publicly by adding it to the monorepo.

# How to Test

* Clone the repo and test all the npm scripts in the `stacks-svelte` workspace to confirm the PR is good.
* I already pointed https://svelte.stackoverflow.design to this branch. I will point it to `production` once this PR is merged and releases are performed.
* Click around the deploy previews from Netlify below. They should work as they did before.

# How to Review

Below are the most important files to checkout as part of reviewing this PR:
```
/
├── .changeset
│   └── cyan-teachers-hear.md
├── .github
│   └── workflows
│       └── workflow.yml
├── .gitignore
├── README.md
└── packages
    ├── stacks-docs
    │   └── assets
    │       └── js
    │           └── index.ts
    └── stacks-svelte
        ├── .storybook
        │   └── getting-started
        │       └── Overview.mdx
        ├── README.md
        ├── netlify.toml
        ├── package.json
        ├── vite.config.ts
        └── .storybook
            └── preview.ts
```


packages/stacks-docs/assets/js/index.ts
```diff
@@ -1,7 +1,7 @@
-import "../../../stacks-classic/lib/index";
+import "@stackoverflow/stacks/lib/index";
 import "../less/stacks-documentation.less";
 import "./controllers/docs-resizer";
-import * as Stacks from "../../../stacks-classic/lib/index";
+import * as Stacks from "@stackoverflow/stacks/lib/index";
```
mukunku, line 4: npm creates symbolic links of local peer dependencies inside `node_modules` so it's possible to reference the files as an npm package. Thought that was cool so changed these imports to this. Let me know if we think relative paths were better here.

packages/stacks-svelte/.storybook/preview.ts
```diff
@@ -0,0 +1,26 @@
+import type { Preview } from "@storybook/svelte";
+import "@stackoverflow/stacks/lib/stacks.less";
```
mukunku, line 2: Same as above:

> npm creates symbolic links of local peer dependencies inside node_modules so it's possible to reference the files as an npm package. Thought that was cool so changed these imports to this. Let me know if we think relative paths were better here.

packages/stacks-svelte/package.json
```diff
@@ -0,0 +1,43 @@
+{
+    "name": "@stackoverflow/stacks-svelte",
```
mukunku, line 2: I renamed the package from `@stackeng/stacks-svelte` to `@stackoverflow/stacks-svelte` so it matches the org name for stacks classic. We can update internal consumers to the new package in separate PR's.

packages/stacks-svelte/package.json
```diff
@@ -0,0 +1,43 @@
...
+        "lint": "svelte-check && eslint .",
+        "lint:fix": "eslint . --fix",
+        "test": "npm run build:dependencies && web-test-runner",
+        "test:skipbuild": "web-test-runner",
+        "test:watch": "npm run build:dependencies && web-test-runner --watch",
+        "storybook": "storybook dev -p 6006",
+        "storybook:build": "storybook build -o netlify/dist"
+    },
+    "peerDependencies": {
+        "@stackoverflow/stacks": "^2.8.4",
```
dancormier, line 29: Should we tie the version to whatever is present in the monorepo?

```suggestion
        "@stackoverflow/stacks": "*",
```

dancormier (approved):
Great work with this @mukunku! I have one minor suggestion but nothing so significant that see it unchanged would be a real issue.

packages/stacks-svelte/package.json
mukunku: That's what I thought too but Giamir pointed out stacks-svelte is meant to be a standalone package. So we need to tie it to the version of stacks-classic it actually supports. Otherwise if we were to release stacks-classic v3 at some point and someone downloaded stacks-svelte which targets v2 for stacks-classic, it would break since they would end up pulling v3.

dancormier:
### Context for @giamir / @ttaylor-stack

I've reviewed this PR and found it to be pretty solid. The only necessary further context I think is necessary is the follow up tasks to happen post-merge:

- Archive Stacks-Svelte (✅)
- Update archived Stacks-Svelte repo Readme to point to the Stacks > Stacks-Svelte package (✅)
- Update codebases consuming Stacks-Svelte to get it from this new source (🟨 https://docs.google.com/spreadsheets/d/11vExYW7hWfEkJWepTvbtaRRNe_GFQzV-QxWuR3L8MqA/edit?usp=sharing)
- Build https://svelte.stackoverflow.design/ based on `production`  (✅)
  - (Sal) Need to switch https://stackoverflow.design/ back to `production` as well. Needed to temporarily switch that to my branch as well for builds to be green. (✅)
- (Sal) Add build badges to main readme for stacks-svelte, similar to stacks-classic (✅)
- (Sal) I created a new ticket in our backlog to address the two failing svelte tests that we're currently skipping as a temp workaround: https://stackoverflow.atlassian.net/browse/SPARK-88 (✅)

PR #1993, opened 2025-10-01
SPARK-52: Update Playwright to possibly address Firefox test flakiness

mukunku:
# Summary

This PR bumps playwright to the latest version for Stacks Classic Visual Regression Testing. Found this which seemed relevant to the issues we've been seeing: https://github.com/modernweb-dev/web/issues/2907

~20 tests started failing with <3% visual difference so I updated the baseline images for those.


dancormier (approved):
Seems smart to upgrade playwright. It's odd that some images required updating but it passes, so it should be fine. Thanks @mukunku :shipit:

PR #2008, opened 2025-10-17
feat(button) - update button base styles for SHINE

ttaylor-stack:
PR to update button component to match the new SHINE base styles

Ticket: https://stackoverflow.atlassian.net/browse/SPARK-55
Figma: https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=4272-20325&m=dev

**Questions/Concerns:**

- [x] I've used decimal pixel values to get the button height exactly as specified however in the refinement meeting yesterday it was mentioned that this might not be necessary. Should I round these values up/down to make use of the static pixel values? cc: @dancormier
- [x] I have updated the badge background color to be `red-500` within the danger variant in high contrast mode in order to keep the 7:1 ratio. Let me know if that is okay or needs updating cc: @CGuindon @dancormier


packages/stacks-classic/lib/components/button/button.less
```diff
@@ -214,29 +209,46 @@
             margin-top: -0.3em; // [4]
             margin-bottom: -0.3em; // [4]
             transition: opacity 200ms var(--te-smooth); // Animate the transition to .is-loading
+            
+            &:not(:last-child) {
+                margin-right: var(--_bu-icon-gap, 8px);
```
dancormier: No need for the fallback `8px` value here. Since `--_bu-icon-gap` is defined in a broader scope, it should always have a value.

packages/stacks-classic/lib/components/button/button.less
```diff
@@ -214,29 +209,46 @@
...
+            }
+            
+            &:not(:first-child) {
+                margin-left: var(--_bu-icon-gap, 8px);
```
dancormier: Same here. No need for the fallback `8px` value.

packages/stacks-classic/lib/components/button/button.less
```diff
@@ -3,10 +3,14 @@
     --_bu-baw: var(--su-static1);
     --_bu-bc: transparent;
     // --_bu-bg: inherit; // [1]
-    --_bu-br: var(--br-md);
+    --_bu-br: 20px;
```
dancormier: Instead of using a hardcoded value here, I suggest using a border radius variable (I think `--br-pill` would be best).

Extra context: Design plans on simplifying our [border radii](http://localhost:8080/product/base/border-radius/) to something like _none_ (0), _round_ (10px I think), _pill_ (1000px), and potentially _circle_ (100%).

I suggest adding a todo here to convert this value to an atomic value once border radius is updated in our libraries (see [SPARK-91](https://stackoverflow.atlassian.net/browse/SPARK-91)).

packages/stacks-classic/lib/components/button/button.less
```diff
@@ -3,10 +3,14 @@
...
     --_bu-fc: var(--theme-button-color, var(--theme-secondary-400));
     --_bu-fs: var(--fs-body1);
-    --_bu-p: 0.8em;
+    --_bu-p: 10px;
+    --_bu-px: 16px;
+    --_bu-py: 10px;
```
dancormier: When reasonably possible, I'd like to stick to sizing variables instead of hardcoded values. For example, `10px` would be represented using a `calc(var(--su8) + var(--su2))`.

Extra context: We have sizing units that are [variable](https://github.com/StackExchange/Stacks/blob/develop/packages/stacks-classic/lib/atomic/spacing.less#L6-L36) and [static](https://github.com/StackExchange/Stacks/blob/develop/packages/stacks-classic/lib/atomic/spacing.less#L6-L20). Check out this [ADR](https://github.com/StackExchange/Stacks/blob/develop/adrs/0005-spacing-units.md) for more info about this approach.

packages/stacks-classic/lib/components/button/button.less
```diff
@@ -214,29 +209,46 @@
...
 
     // Size
     &&__xs {
         .size-styles(xs; bu; @styles: fs);
         --_bu-dropdown-bw: calc(var(--su-static4) - var(--su-static1));
-        --_bu-p: 0.6em;
+        --_bu-fs: 12px; // Override size-styles for custom font size
+        --_bu-px: 12px; // 12px padding as per Figma specs
+        --_bu-py: 10px;
+        --_bu-icon-gap: 4px; // 4px spacing as per Figma specs
```
dancormier: This sort of stuff will be tricky until we have base styles updates like typography figured out. You'll see a [`.size-styles` mixin](https://github.com/StackExchange/Stacks/blob/1add86d0987e24bb1cc43022620747b717b8059d/packages/stacks-classic/lib/exports/mixins.less) a couple of lines up. This mixin allows us apply the same styles for a given size across multiple components. We could modify that mixin to account for that but I'm leaning towards just adding a few `todo`s to port things over to that mixin.

packages/stacks-classic/lib/components/button/button.less
```diff
@@ -214,29 +209,46 @@
...
     }
 
     &&__sm {
         .size-styles(sm; bu; @styles: fs);
+        --_bu-fs: 13px; // Override size-styles for custom font size
+        --_bu-px: 12px; // 12px padding as per Figma specs
+        --_bu-py: 10px; // Calculated to achieve 32px height: (32 - 12) / 2 = 10px
+        --_bu-icon-gap: 6px; // 6px spacing as per Figma specs
```
dancormier: In order to achieve the desired button height with the suggested padding, we'll probably need to adjust `line-height` pretty significantly. For instance, the `xs` button can hit ~28px height with `line-height: .5`, but this is a _very_ tight line height that would result in some rough display if it ever needs to wrap.

<details>
<summary>Example images</summary>

### line-height: .5 single line

image

### line-height: .5 multi-line

image



</details>


For the button sizes in general, I suggest asking via a comment on the [Figma](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=753-10567&t=707K7NHN7RETDvwR-0) asking if the overall height is the goal over specific padding values. I suspect hitting ~28px height is more important. If that's true, I'd ignore the recommended padding values in favor of hitting the height while ensuring the text is vertically centered.

packages/stacks-classic/lib/components/button/button.less
```diff
@@ -482,7 +444,7 @@
     cursor: pointer;
     display: inline-block;
     font-family: inherit;
-    font-weight: normal;
+    font-weight: var(--_bu-fw, 600);
```
dancormier: No need for the fallback value here

dancormier (changes requested):
Hey @ttaylor-stack I looked it over and made a few broad suggestions that might help out. I'm happy to discuss more while we pair on Monday. Thanks for taking on this beastly redesign 🙂

packages/stacks-classic/lib/components/button/button.less
ttaylor-stack: Ignore padding specified in Figma and target height OR do recommendation above

CGuindon:
Discussion in this [slack thread](https://stackexchange.slack.com/archives/C096063B1PH/p1761333062219589)

- It's missing the Tonal button (replacing Muted I think?)
- I updated the Facbook icon a month ago and I probably broke something in this Facebook button (extra color class in there maybe?). Can we fix this as part of the new buttons? (nobody has noticed yet — I don't think we use this button in a lot of places)

Screenshot 2025-10-24 at 12 14 04 PM

packages/stacks-classic/lib/components/button/button.less
```diff
@@ -214,29 +198,48 @@
...
     }
 
     // Size
+    // TODO: port the below added styles to the size-styles mixin SPARK-91 and SPARK-102
     &&__xs {
         .size-styles(xs; bu; @styles: fs);
         --_bu-dropdown-bw: calc(var(--su-static4) - var(--su-static1));
-        --_bu-p: 0.6em;
+        --_bu-p: var(--su12);
+        --_bu-py: 6.285px; // Adjust to reach 28px height
```
ttaylor-stack: round 6.285 down or up

CGuindon (changes requested):
1. We're going to keep the focus ring Black on all components so it won't match the featured and danger state button colors. Sorry I thought this was handled by the focus ring component we have (currently blue) and not the buttons. Just updated the figma file to reflect this — buttons in the Figma were out of date from this decision. We're keeping existing logic that the focus is always the same color — black-600.
Screenshot 2025-11-12 at 12 49 10 PM
Screenshot 2025-11-12 at 12 48 32 PM

---

2. `s-btn` and `s-btn__tonal` shouldn't get a background fill for the **focus**  state (one is getting black fill so the text disappears and the other is getting a light fill that makes it look tonal). We should just have the outer ring appear (making it look like an outlined button) — same as `.s-btn .s-btn__danger` but black.
Screenshot 2025-11-12 at 12 51 42 PM 
Screenshot 2025-11-12 at 12 53 21 PM

---

3. The selected state on the clear buttons seem to be missing the gradient/two-toned styling? And the `tonal` selected button state has a gradient that looks much less noticeable than Figma (I'm not sure how to grab the values from the browser). Bottom grey in Figma is HLS 220, 8, 93 and top darker grey is black-200 (HLS 216, 8, 88).
Screenshot 2025-11-12 at 1 01 20 PM
Screenshot 2025-11-12 at 12 57 43 PM

---

4. Can we try to get the default button size to match closer to 40px in height? By reducing the top/bottom padding a bit. I know the other ones aren't exact but it's over 41px right now. This will be a component that gets aligned with other components often and I've made other elements match the 40px height as well so they all line up nice.
Screenshot 2025-11-12 at 1 08 35 PM

---

5. `.s-btn .s-btn__filled` in dark mode doesn't have the two toned/gradient selected state. Is there a way to use color stops to make this gradient or is it using hard coded values? Looping in @dancormier for thoughts on how to approach this. Ideally we use the color stop values so we don't need to update this in the future if we change any color values.
Screenshot 2025-11-12 at 1 13 59 PM
Screenshot 2025-11-12 at 1 12 12 PM

---

6. `.s-btn .s-btn__featured .s-btn__filled` also seems very low contrast between the two tones. That might be dark mode colors. Wondering what approach you took to making all these gradients now. 
Screenshot 2025-11-12 at 1 14 20 PM

CGuindon:
I think we might need a call to talk through the selected state options — I'll set that up.

packages/stacks-classic/lib/components/button/button.less
```diff
@@ -38,31 +45,28 @@
...
-        --_bu-outlined-fc-selected: var(--white);
+        --_bu-bc: var(--black-300);
+        --_bu-bc-selected: var(--black-300);
         --_bu-badge-o: 0.8;
-        --_bu-number-fc-selected: var(--theme-button-primary-number-color, var(--theme-secondary-600));
+        --_bu-number-fc-selected: var(--white);
+    });
+
+    .dark-mode({
+        --_bu-filled-bg-selected: linear-gradient(to bottom, var(--theme-secondary-600) 0%, var(--theme-secondary-600) 50%, color-mix(in srgb, var(--theme-secondary-600) 80%, black 40%) 50%, color-mix(in srgb, var(--theme-secondary-600) 80%, black 40%) 100%);
```
ttaylor-stack: @CGuindon by reversed I'm assuming you meant use black instead of white for the gradient instead of flipping the colors.

image

CGuindon:
@ttaylor-stack I'll have a separate ticket for button badges — so leaving the change you made for HC mode is fine and great for now in this PR.

CGuindon (changes requested):
- Tonal seems to have lost it's _tone_. The default (non-selected) state should have a background color of `black-150`. We don't have an instance of _tonal__clear_ the _tonal__filled_ is the only version. So filled should appear as the default for `s-btn
.s-btn__tonal`
Screenshot 2025-11-14 at 9 19 04 AM

I think it's because it's pulling from `theme-secondary`. Wondering if for the _tonal_ styles we need to stick to the black color stops — @dancormier thoughts?
Screenshot 2025-11-14 at 9 23 52 AM

CGuindon:
Screenshot 2025-11-14 at 9 32 50 AM

For the `.s-btn .s-btn__danger` **selected** let's tweak one value on the gradient (60% to 50% on the bottom color):
`linear-gradient(to bottom, var(--red-100) 0%, var(--red-100) 50%, color-mix(in srgb, var(--red-100) 50%, white 40%) 50%, color-mix(in srgb, var(--red-100) 50%, white 40%) 100%)`

CGuindon:
For dark mode, I think the gradients are using the wrong white or black value or ordering in some cases. I mocked them up in Figma and added notes for what I changed on the gradient where it differs from the Light mode. Ultimatley, all of the **selected** states should have the darker tone on the top layer and light tone on the bottom layer. I'm not sure how the figma gradient values correlate to the code values but hoping this helps.

[Figma dark mode](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=4628-5887&t=S4ocNMHppC9x1E3o-1)

dancormier (changes requested):
This PR is a beast and in order to keep things moving, I took the liberty to refactor the Less in this PR: https://github.com/StackExchange/Stacks/pull/2052

@ttaylor-stack I'd like to take a little time to discuss the PR in our 1:1 and see if it's appropriate to merge into yours. I think we can save ourselves some future pain by structuring the Less and the custom properties a bit differently.

dancormier (approved):
Let's merge it!

PR #2012, opened 2025-10-20
docs(stacks-svelte): generate llms friendly docs from storybook content

giamir:
[SPARK-93](https://stackoverflow.atlassian.net/browse/SPARK-93)

**Attention: 2 vibe coded files are checked in as part of this PR**

This PR introduces a script that analyze the content of the stacks-svelte source and related storybook docs and generates a `llms.txt` summarizing all that content in markdown easily digestible by LLMs.

The only package I could find out there to take storybook docs and make them LLMs friendly was [this one](https://www.npmjs.com/package/@fluentui/storybook-llms-extractor).  Unfortunately though the package support only storybook v7 and v8 and we are using v9. It also uses playwright under the hood.

Until there’s a more established package that supports converting Storybook docs to Markdown, I’ve opted to vibe code what we need in a very isolated tools folder. In general, I’m not in favor of checking in code generated by LLMs unless we fully understand it. However, for this specific use case, I think it’s acceptable — the script is unlikely to need ongoing maintenance, and even if it breaks, the only loss would be something primarily useful for LLMs.

## How to test

- Check out this branch
- Run `npm run storybook:build -w @stackoverflow/stacks-svelte`
- Open `packages/stacks-svelte/netlify/dist/llms.txt` and check the generated markdown

Deployment Preview: https://deploy-preview-2012--stacks-svelte.netlify.app/llms.txt


dancormier (approved):
> Unfortunately though the package support only storybook v7 and v8 and we are using v9. It also uses playwright under the hood.

We're finally on the cutting edge 🎉 

> In general, I’m not in favor of checking in code generated by LLMs unless we fully understand it. However, for this specific use case, I think it’s acceptable — the script is unlikely to need ongoing maintenance, and even if it breaks, the only loss would be something primarily useful for LLMs.

100% agree. Seems like a wise, low-liability use of vibe coding.

---

I followed the PR instructions and `llms.txt` generated exactly as in the preview (besides `**Generated:** 2025-10-21T20:08:03.165Z` of course 🙂). Solid LGTM material!

PR #2013, opened 2025-10-20
chore(bling): update to SHINE styling

dancormier:
[SPARK-58](https://stackoverflow.atlassian.net/browse/SPARK-58)

---

https://deploy-preview-2013--stacks.netlify.app/product/components/bling/

### TODO
- [x] Component styles are updated in Stacks Classic
     - [x] Add default style
     - [x] Add `filled` modifier
     - [x] Add `size` modifiers (`sm`, `lg`, default)
     - [x] Add `activity` variant
     - [x] Add `rep` variant
- [x] Add breaking changes to the v2 → v3 migration guide (see [SPARK-60](https://stackoverflow.atlassian.net/browse/SPARK-60))
- [x] Rename to `bling`
     - [x] Component uses `s-bling` class
     - [x] Documentation refers to component as `Bling`
- [x] Update documentation with new modifier/variant examples
- [x] Update documentation with copy to describe newly-updated component
- [x] Update Svelte component
     - [x] Update name to Bling
     - [x] Add `filled` bool prop
     - [x] Add `size` string prop
     - [x] Add `activity`, `rep`, null (default) to `type` prop
     - [x] Remove `children` snippet prop
     - [x] Update markup to match updated markup
     - [x] Update examples to reflect above changes
- [x] Add changeset 

### Follow up
We'll need to follow up on this PR once a few other changes have been made to the system:

- [x] The `activity` variant will need to be changed to `pink` colors once added
- [x] Visual regression images will need to be regenerated once color palettes are updated

### The docs

I have not put too much consideration into the docs. @CGuindon if you have an idea of how these should be updated, let me know and I'll add those changes.

**Update Mon Nov 10, 2025**: I've updated the docs per the Figma as [suggested](https://github.com/StackExchange/Stacks/pull/2013#pullrequestreview-3377999470).


giamir (commented):
This component looks great @dancormier. ❤️🎉

I was lurking around and I wonder if we should add in the acceptance criteria of our stories also the **documentation of breaking changes**. I know that there is a ticket in the backlog to create a migration doc (v2 to v3). Maybe it would make sense to fill up that document as we go so that our memory is fresh on the breaking changes we made from the point of view of the consumers. 

Also, a friendly reminder to add the changesets for this PR. I think this is the first time we’ll have two changesets for two different libraries in the same PR. In any case, `npx changeset` should guide us through creating these slightly more complex changesets. 🙂

CGuindon (commented):
Docs: I have text in the figma under each style and on the left side for the "header + intro" text.
Screenshot 2025-10-24 at 10 23 42 AM

CGuindon:
I guess the order of this component changes in the list as well with the new name — "Bling" should go under "Banners"
Screenshot 2025-10-24 at 10 25 05 AM

CGuindon:
All of the badge shapes should _fit_ in a 10x10 "box" but I've tweaked the sizing a bit for each to make them visually more balanced when next to each other. 

- `s-bling__bronze` should be 8x8 square (inside the 10x10 box frame) Screenshot 2025-10-24 at 10 26 37 AM

- `s-bling__silver`  `s-bling__gold` both look like they're the correct 9x9 size inside that box ✅
Screenshot 2025-10-24 at 10 30 30 AM
Screenshot 2025-10-24 at 10 32 37 AM

packages/stacks-docs/_data/site-navigation.json
```diff
@@ -222,6 +218,10 @@
               "title": "Banners",
               "url": "/product/components/banners/"
             },
+            {
+              "title": "Bling",
+              "url": "/product/components/bling/"
```
giamir: Set `new` to true here to show the badge

packages/stacks-docs/product/components/bling.html
```diff
@@ -0,0 +1,144 @@
+---
+layout: page
+title: Bling
+razor: https://razor.stackoverflow.design/components/award-bling
```
giamir: There should be a link to svelte here, not to Razor. Probably you lost some context rebasing. We don't have Razor links in the beta branch anymore.

giamir (commented):
Looks good @dancormier. I have only left a couple of minor NIT comments. 

I noticed that the following components depending on Bling are now somewhat broken:
- Badge
- Topbar
- UserCard
- Table

I guess this is on purpose and we will take care of them on follow up tickets. Correct?

Thanks for taking care of this one. ❤️

dancormier:
> I noticed that the following components depending on Bling are now somewhat broken:
> 
>     * Badge
> 
>     * Topbar
> 
>     * UserCard
> 
>     * Table
> 
> 
> I guess this is on purpose and we will take care of them on follow up tickets. Correct?

@giamir Updates to consuming components' _styles_ should usually be left for when those components are being updated to SHINE styles, _but_ I think we should generally update any markup when there are breaking changes like there are here with bling.

I've updated the bling markup in those other components' docs pages and any relevant tests to match the expected markup. I've also updated the migration guide to mention the markup changes. If nothing else, these doc updates just make things look much nicer while we wait to update the other components 🙂 

<details>

image
image
image
image
image
image

</details>

> Thanks for taking care of this one. ❤️

My pleasure and thank you for the review!

giamir (approved):
Looks good Dan. Thanks for the updates. The only thing missing is the link to the svelte component (svelte badge) in the Bling doc page (in the frontmatter `svelte: ...`). Apart from that looks good to be merged from my end. 🙂

CGuindon (changes requested):
In the Svelte component the rep bling is using `black-400` — both **rep** and **default** bling dots use `black-600`. The difference is only if the fill is showing, the background color of the square box is lighter/darker but the dots are 600 in both cases. (Classic looks fine)

Screenshot 2025-11-12 at 1 36 39 PM

dancormier:
> In the Svelte component the rep bling is using `black-400` — both **rep** and **default** bling dots use `black-600`. The difference is only if the fill is showing, the background color of the square box is lighter/darker but the dots are 600 in both cases. (Classic looks fine)

Whoops! Good catch @CGuindon. I've updated the `rep` bling to use `black-600` and I've also updated the [Svelte docs](https://deploy-preview-2013--stacks-svelte.netlify.app/?path=/docs/components-bling--docs) to omit the examples of the non-filled `rep` and `activity` variants (though a consumer could still _technically_ use them.

<details>
<summary>Screenshot</summary>

image

</details>

mukunku:
Should the filled variant mention the filled class in the table below? 

{904B1646-0917-49C3-87AA-1A92FE7EC8F3}

dancormier:
> Should the filled variant mention the filled class in the table below?
> {904B1646-0917-49C3-87AA-1A92FE7EC8F3}

@mukunku yes, nice catch! I'll add that class into the docs

PR #2015, opened 2025-10-22
chore(colors): update color palettes

dancormier:
[SPARK-49](https://stackoverflow.atlassian.net/browse/SPARK-49)

---

[Deploy preview](https://deploy-preview-2015--stacks.netlify.app/product/foundation/colors/)

This PR updates Stacks to use the new SHINE color palettes. 

### TODO
- [x] All colors updated
    - [x] Light
    - [x] Dark
    - [x] Light HC
    - [x] Dark HC
- [x] Pink color set added
- [x] Metallic colors sets removed
- [ ] Documentation updated as needed
- [x] (verify) Code highlight colors updated
    - We should [leave them as-is](https://github.com/StackExchange/Stacks/pull/2015#discussion_r2457375422)
- [x] ~~(verify) Update or remove translucent color sets (these are mostly used for focus rings)~~
    - [x] For now, we should retain these colors for backwards compatibility. See also [Coexistence of Multiple Design System Versions in SHINE UIs doc](https://docs.google.com/document/d/1kdt-JEWdWhxTdvDv1lmIymo5XQKQrcNcqCnpI-qYgs0/edit?tab=t.0)
- [x] Resolve tests
    - [x] Recreate visual regression test images
    - [x] Fix (or skip) accessibility failures on components yet-to-be updated
    - [x] Recreate less test snapshots
- [x] Add changeset

cc @leileitun3


packages/stacks-classic/lib/exports/color-sets.less
```diff
@@ -11,6 +11,7 @@
 //
 // tl;dr: don't put inline after values in this file
 
+// TODO SPARK update default values for white and black sets
```
dancormier, line 14: @CGuindon currently we have `black` and `white` colors with no stop (used like `bg-black` for example. What should those values be?

Here's what they are now for reference:

- white
  - light: `hsl(0, 0%, 100%)`
  - dark: `hsl(210, 3%, 15%)`
  - light HC: `hsl(0, 0%, 100%)`
  - dark HC: `hsl(0, 0%, 0%)`
- black
  - light: `hsl(0, 0%, 0%)`
  - dark: `hsl(0, 0%, 100%)`
  - light HC: `hsl(0, 0%, 0%)`
  - dark HC: `hsl(0, 0%, 100%)`

packages/stacks-classic/lib/exports/color-sets.less
dancormier, line 1: **Note to reviewers**: This file is where the vast majority of changes have happened.

packages/stacks-classic/lib/exports/color-sets.less
```diff
@@ -484,6 +441,7 @@
 
 // highlight
 // TODO consult with design to potentially map static colors to new colors variables
+// TODO SHINE request colors for use in code blocks (or verify these can be used for now)
```
dancormier, line 522: @CGuindon These `highlight` sets are used for syntax-highlighted code (see [Code blocks](https://deploy-preview-2015--stacks.netlify.app/product/components/code-blocks/#html)). Do we expect to change these colors now or should we keep them as-is for the moment?

dancormier (commented):
@CGuindon, @leileitun3 I've popped in a few questions here

packages/stacks-classic/lib/exports/color-sets.less
```diff
@@ -431,15 +386,16 @@
 }
 
 // translucent utility colors (HC set represents both light and dark HC modes)
+// TODO SHINE remove once they're no longer used (probably once focus ring is updated -- currently used )
```
dancormier: @CGuindon we have these `translucent` colors which are used as focus rings (and around the [stepped variant](https://github.com/StackExchange/Stacks/pull/2015#pullrequestreview-3372257300) of the progress bar). Can you remind me of the plan for focus rings?

If we can make them non-translucent across the board (as they currently are on [buttons](https://deploy-preview-2015--stacks.netlify.app/product/components/buttons/)), I think we can get rid of these sets.

packages/stacks-classic/lib/exports/color-sets.less
CGuindon: I'm assuming we can keep them all the same expect for the White > dark which looks custom.

I think that must be our background color when in dark mode? It's not a pure black background (like our pure white background in black). That value is `hsl(0, 0%, 8%)`

packages/stacks-classic/lib/exports/color-sets.less
CGuindon: @dancormier These colors have their own custom hex/hsl code associations? I would keep the same

packages/stacks-classic/lib/exports/color-sets.less
CGuindon: @dancormier Yes our plan would be to transition everything from translucent to use actual color stops. So for that stepped progress bar, we'd use a solid blue-100 on the outer ring instead of something with opacity (if I understood that correctly). Focus-rings should use a solid color (like buttons). Anything with a translucent style (like that stepped progress bar will get removed once it gets redesigned). We removed it from the activity indicator so I see this no longer being part of our visual language moving forward.

Are you able to get a list of all the translucent places? Rule of thumb re-mapping would be to use the 100 stop variant instead for now until we redesign that component altogether.

CGuindon (changes requested):
Looks like Dark mode is using HC Dark mode colors and the reverse for `orange`
Screenshot 2025-10-29 at 11 54 25 AM
Screenshot 2025-10-29 at 11 56 00 AM

CGuindon:
`Pink` in light mode isn't using the right color codes
Screenshot 2025-10-29 at 12 00 13 PM

Screenshot 2025-10-29 at 12 01 16 PM

CGuindon (changes requested):
Color values should be (I think some of these are the ones Lei Lei tweaked recently):

- Light mode `blue` 400: hsl(226, 67%, **95%**)
- Light mode `green`  200: hsv(82, **30%**, 95%)
- HC Light mode `green` 300: hsv(82, 80%, **63%**)
- HC Light mode `green` 400: hsv(82, **93%**, **38%**)
- Dark mode `red` 500: hsv(0, **20%**, 99%)
- Light mode `yellow` 200: hsv(48, **36%**, 96%);
- Light mode `pink`
    100: hsv(**293**, 6%, 99%);
    200: hsv(**293**, 25%, 95%);
    300: hsv(**293**, 45%, 95%);
    400: hsv(**293**, 68%, **91%**);
    500: hsv(**293**, 80%, 70%);
    600: hsv(**293**, 90%, 45%);
- HC Light mode `pink` 300: hsv(293, **68%, 91%**);
- HC Dark mode `pink`  400: hsv(293, **30%, 80%**);

dancormier:
@CGuindon thank you for the detailed suggestions. You saved me a lot of time (your **bold**ing helped a lot!) 🫶  I've made those changes in https://github.com/StackExchange/Stacks/pull/2015/commits/6a9f965c1510b87bad1a4628e9806e82b51e1a68.

I've also:
- Added gold/silver/bronze color sets back in since we might need to retain the CSS variables for legacy purposes
- Updated `theme-secondary` to use `black` values
    - I expect this will require some tweaking as `theme-400` is now a gray tone (in light mode) whereas I'd expect it to be close to pure black. I'm not entirely sure how we should handle this. Any ideas @CGuindon?

> Looks like Dark mode is using HC Dark mode colors and the reverse for `orange`
> Screenshot 2025-10-29 at 11 54 25 AM
> Screenshot 2025-10-29 at 11 56 00 AM

I might be misunderstanding. You're suggesting that the dark and dark HC orange palettes are swapped? To me it looks correct. Can you clarify @CGuindon?

image

CGuindon:
> I might be misunderstanding. You're suggesting that the dark and dark HC orange palettes are swapped? To me it looks correct. Can you clarify @CGuindon?

@dancormier The oranges looked swapped in the preview but I don't think I saw anything that off in the code file you sent me... Would that preview need to be regenerated?

CGuindon:
@dancormier 
If we keep `theme-secondary` to use `blue` values, would that change the buttons on SO to be blue? Or does it just change the Team's default theming to use blue buttons? SE sites too perhaps?

I don't mind having the buttons remain blue in the Teams template/default but ideally SE sites (without custom) use the new black buttons down the line.

dancormier:
> @dancormier If we keep `theme-secondary` to use `blue` values, would that change the buttons on SO to be blue? Or does it just change the Team's default theming to use blue buttons? SE sites too perhaps?
> 
> I don't mind having the buttons remain blue in the Teams template/default but ideally SE sites (without custom) use the new black buttons down the line.

Since blue is no longer being used as a theme color, I'd suggest replacing it with black. I haven't tested it, but I _think_ we could get away with mapping black color stops 1-to-1 with theme secondary stops and then have `theme-secondary` (with no color stop suffix) as pure black. Like this:

image

In this orientation, we'd default to using `theme-secondary` for themable components like button and links. Does this sound reasonable @CGuindon? I'll update this PR to show this so we can get a better idea of how it would look in practice.

dancormier:
I've updated this PR to have `theme-secondary` set to black (see [Color palette](https://deploy-preview-2015--stacks.netlify.app/product/foundation/colors/)). I also switched references of `--theme-secondary-400` to `--theme-secondary` so themed components will get that black color (with the exception of components currently being worked on like [Button](https://github.com/StackExchange/Stacks/pull/2008)). I also took this opportunity to update the [activity indicator](https://deploy-preview-2015--stacks.netlify.app/product/components/activity-indicator/) colors to match the Figma.

I think it looks alright and allows it to be overridden by a custom theme. @CGuindon please give it a look and let me know what you think. Thanks!

CGuindon:
@dancormier I think that makes sense to me. Looping in @andrewmeacham to see if he has any extra insights we're not considering for theming. 

How do I test the theming in this PR?

CGuindon (approved):

CGuindon:
Approved in case you need colors to merge sooner than later since it's blocking other tickets — I think documentation still needs to be updated (add pink to [usage](https://deploy-preview-2015--stacks.netlify.app/product/foundation/color-fundamentals/#color-roles)). Maybe we can spin up a separate PR/ticket for that if needed.

dancormier:
> How do I test the theming in this PR?

Click the button far right end of the navbar (after the search input). There, we the have a toggle within the popover to apply a custom theme. This is a custom theme we've applied in the docs using the same functions one would as a consumer of Stacks.

image



> Approved in case you need colors to merge sooner than later since it's blocking other tickets — I think documentation still needs to be updated (add pink to [usage](https://deploy-preview-2015--stacks.netlify.app/product/foundation/color-fundamentals/#color-roles)). Maybe we can spin up a separate PR/ticket for that if needed.

Thank you! Is there guidance written down anywhere for pink? I looked in the Figma but didn't find it. Happy to add a row for guidance on that. If there are lots of docs changes missing, I'll create a follow up ticket.

dancormier:
I'm merging this in to unblock other work. Created a follow up ticket for doc updates ([SPARK-108](https://stackoverflow.atlassian.net/browse/SPARK-108)).

PR #2025, opened 2025-10-27
chore(deps): integrate stacks icons v7-beta and rename v6 legacy

giamir:
[SPARK-101](https://stackoverflow.atlassian.net/browse/SPARK-101)

This PR integrates the [new beta stacks icons](https://beta.icons.stackoverflow.design/) (and spots) in the monorepo. 

Since we cannot simply remove the old v6 icons the changes create an alias for that package (`@stackoverflow/stacks-icons-legacy`). The main reason this PR is touching many files is because the docs and the stacks svelte components still utilize the old icons for now. 

We will gradually migrate to the new icons as we work on components and updating the stacks docs. 
Ultimately the goal would be to remove every reference of the legacy icon package and uninstall it. 

The only places where we can see the new icons are the following as of now:
- [Stacks Docs Icons Page](https://deploy-preview-2025--stacks.netlify.app/product/foundation/icons/)
- [Stacks Docs Spot Illustration Page](https://deploy-preview-2025--stacks.netlify.app/product/foundation/spots/) (This is marked explicitly as a WIP since we have only 1 Spot)
- [Storybook Stacks Svelte Icon Documentation](https://deploy-preview-2025--stacks-svelte.netlify.app/?path=/docs/components-icon--docs)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,8 +20,8 @@
                 "@rollup/plugin-commonjs": "^26.0.1",
                 "@rollup/plugin-replace": "^6.0.2",
                 "@stackoverflow/prettier-config": "^1.0.0",
-                "@stackoverflow/stacks": "^2.8.4",
-                "@stackoverflow/stacks-icons": "^6.6.1",
+                "@stackoverflow/stacks-icons": "^7.0.0-beta.1",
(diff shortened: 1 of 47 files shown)


dancormier (approved):
I looked through the changes and they all seem reasonable. I also ran this locally to see the docs changes and I think they are good. I appreciate the compromise going with a `legacy` package with the plan to remove it when the time comes. Thanks @giamir

PR #2019, opened 2025-10-24
feat(avatar): update to new visual language

giamir:
[SPARK-61](https://stackoverflow.atlassian.net/browse/SPARK-61)

[Figma](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=610-18797&p=f&t=c9NSyJWdASb80eVC-0)

Here is what changed:

- Stacks Classic: Made avatar border squared. Introduced `.s-avatar--indicator` class for positioning an activity indicator element in relation to the avatar. ([docs](https://deploy-preview-2019--stacks.netlify.app/product/components/avatars/))
- Stacks Svelte: Introduced a new `status` prop. When set to `online` it will show an activity indicator on the avatar. ([storybook](https://deploy-preview-2019--stacks-svelte.netlify.app/?path=/docs/components-avatar--docs))

Note: This PR should be merged after the [activity indicator](https://github.com/StackExchange/Stacks/pull/2027) one get merged.

PR #2026, opened 2025-10-29
feat(br): Update border radius atomic styles

ttaylor-stack:
Update border radius atomic styles as specified in the SPARK-91.

Ticket: https://stackoverflow.atlassian.net/browse/SPARK-91

Figma: https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=2213-7342&p=f&t=czT0zuZbbghSpKgs-0

Notes:

- Anywhere in the docs that used `bar-sm` or `bar-lg` I replaced with `bar-md`
- I left the base atomic styles that are used by the other components in place e.g. `br-sm` and `br-md`. I thought because it's used in so many components it makes sense to update them within the components story instead of this one.


packages/stacks-classic/lib/atomic/border.less
dancormier, line 1: All `sm` and `lg` border radius classes should be removed from this file, not just the ones that affect all sides.

packages/stacks-classic/lib/exports/constants-helpers.less
```diff
@@ -32,7 +32,7 @@ body {
 
     // Border Radius
     --br-sm:                 calc(var(--su-static4));
-    --br-md:                 calc(var(--su-static6));
+    --br-md:                 calc(var(--su-static4) + var(--su-static6)); // 10px;
     --br-lg:                 calc(var(--su-static8));
```
dancormier: I suggest getting rid of these `sm` and `lg` custom properties

packages/stacks-docs/_data/borders.json
dancormier, line 1: All `sm` and `lg` border radius classes should be removed from this file, not just the ones that affect all sides.

dancormier (changes requested):
Thank you for taking this on @ttaylor-stack ❤️ 

The **tl;dr** for my review is that we should get rid of all references of `sm` and `lg` border radius classes and custom properties. This extends to the custom property references in `.less` files across the code base.

> [!WARNING]
> Removing/replacing styles is likely to introduce merge conflicts for any in-flight component changes. I don't expect them to be tough to handle, but I figured it warrants a heads up.

CGuindon:
I'm still seeing all the `sm` and `lg` classes in the documentation — does the preview need to be re-generated? 
Screenshot 2025-10-29 at 12 57 58 PM

dancormier (changes requested):
For breaking changes, we should mention the breaks in the Migration Guide ([example here](https://github.com/StackExchange/Stacks/pull/2028/files#diff-45651f4c3460b0f20be048556615842a07ec86d262fe5d3d346b5e4a9ce56d6a)). I think something general mentioning the removal of `sm`/`lg` border should be sufficient.

Besides that I think this is all set. We'll probably want to wait until the [Avatar PR](https://github.com/StackExchange/Stacks/pull/2019) is merged before we merge this one, but we're close!

 🙂 Thanks @ttaylor-stack

CGuindon (approved):

PR #2030, opened 2025-10-31
feat(popover) - update popover component for SHINE

ttaylor-stack:
Ticket: https://stackoverflow.atlassian.net/browse/SPARK-74

Update to popover component, removed arrow from component in both svelte and classic

### DIFF
--- a/.changeset/thin-days-leave.md
+++ b/.changeset/thin-days-leave.md
@@ -3,3 +3,6 @@
 ---
 
 Update popover component to match SHINE styles
+
+BREAKING CHANGES:
+- `s-popover--arrow` css class has been removed
--- a/.changeset/thin-days-leave.md
+++ b/.changeset/thin-days-leave.md
@@ -1,0 +1,5 @@
+---
+"@stackoverflow/stacks": minor
+---
+
+Update popover component to match SHINE styles
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -16,6 +16,14 @@
 - `.blr-sm` removed
 - `.blr-lg` removed
 
+### Components
+
+#### Popover
+- `s-popover--arrow` removed
+
+##### Svelte component
+-  `ArrowEl` removed from Popover svelte component
+
 # Migrating from Stacks Svelte v0 to v1
 
 Coming soon…
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -19,10 +19,7 @@
 ### Components
 
 #### Popover
-- `s-popover--arrow` removed
-
-##### Svelte component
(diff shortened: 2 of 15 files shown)


CGuindon (approved):

MIGRATION_GUIDE.md
```diff
@@ -16,6 +16,14 @@
 - `.blr-sm` removed
 - `.blr-lg` removed
 
+### Components
+
+#### Popover
+- `s-popover--arrow` removed
+
+##### Svelte component
```
giamir: I don't think this is relevant for consumers, since the Svelte Popover public API did not change therefore I would recommend to remove it.

MIGRATION_GUIDE.md
```diff
@@ -16,6 +16,14 @@
 - `.blr-sm` removed
 - `.blr-lg` removed
 
+### Components
+
+#### Popover
+- `s-popover--arrow` removed
```
giamir: Rather than limit ourselves to write what changed I would recommend to add here what the consumers are expected to do. After all this is not a changelog but a migration guide doc. 🙂

So in this case it could be something like. 
```
The new popovers no longer include an arrow element. The `s-popover--arrow` css class has been removed, and any markup using it (e.g. `<div class="s-popover--arrow"></div>`) should be deleted from the codebases as part of the migration.
```

packages/stacks-docs/product/components/popovers.html
giamir, line 1: The popover example that has "Follow question" as a reference looks quite odd now.
Screenshot 2025-11-03 at 11 04 48
Perhaps we could consider removing it or adding it in a separate stacks preview.

giamir, line 1: NIT: We could add the [new figma file](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=1617-27089&t=jI5jOezjW4KxM6a0-1) in the front matter of this page so that the figma badge redirect to the correct node in the file.

.changeset/thin-days-leave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stackoverflow/stacks": minor
+---
+
+Update popover component to match SHINE styles
```
giamir, line 5: I think we could be slightly more precise and call out the breaking change as well here:
```
BREAKING CHANGES:
- `s-popover--arrow` css class has been removed
```

giamir (approved):
Fantastic work @ttaylor-stack. Love to see a lot of code being removed as part of this PR. I have only left minor comments that you should be able to address swiftly. Feel free to reach out if something I wrote does not make sense. Thank you! ❤️

PR #2035, opened 2025-11-04
feat(pagination) - update pagination component to use new SHINE design system

ttaylor-stack:
Story: https://stackoverflow.atlassian.net/browse/SPARK-73

Updated the pagination component in classic and svelte.

NOTE: I added a new assertion function in assertions.ts. The logic looks sound to me and the test is passing but could it be double checked to ensure its correct. cc @giamir 

NOTE: Will need to update color of nav buttons to tonal once button changes are in


packages/stacks-classic/lib/components/pagination/pagination.less
```diff
@@ -9,28 +9,96 @@
 
     & &--item {
         --_pa-item-bg: transparent;
-        --_pa-item-bc: var(--bc-darker);
-        --_pa-item-fc: var(--fc-medium);
-        --_pa-item-bg-focus: var(--black-400);
-        --_pa-item-fc-focus: var(--white);
-        --_pa-item-bg-hover: var(--black-225);
+        --_pa-item-fc: #6B6D73;
+        --_pa-item-bg-hover: #F0EFED;
```
dancormier: The color PR is merged so you should be able to use the custom properties now

packages/stacks-classic/lib/components/pagination/pagination.less
```diff
@@ -9,28 +9,96 @@
...
         --_pa-item-bc-hover: var(--bc-darker);
         --_pa-item-fc-hover: var(--fc-dark);
+        --_pa-item-bb: var(--su-static0);
```
dancormier: `--su-static0` doesn't exist so `0` should be used instead. That said, I noticed a layout shift related to this border width that can be addressed with a few tweaks.




I suggest always having a bottom border on the pagination item where the color is set with one of these sort of custom properties. I'd suggest having the default color be `transparent` and and set to black\* on the selected pagination item. with a consistent width around the pagination item to prevent layout shift.

---

\* I'm not sure if this updated component is intended to get theme colors. I've asked a question on the figma [here](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ?node-id=1189-1966#1503202016)).

packages/stacks-classic/lib/components/pagination/pagination.less
```diff
@@ -9,28 +9,96 @@
...
+                --_pa-item-fc: var(--black);
+                --_pa-item-bc: var(--black);
+                --_pa-item-bg: var(--black-400);
+            });
+
+            .highcontrast-dark-mode({
+                --_pa-item-fc: var(--black-500);
+                --_pa-item-bc: var(--black-500);
+                --_pa-item-bg: var(--black-300);
+            });
```
dancormier: > Updated dark mode and highcontrast dark mode to keep accessibility score high

I think the `--_pa-item-bg` values should stay `transparent` in all cases except when focused.

### Dark
image

### Light HC
image

### Dark HC
image

packages/stacks-classic/lib/components/pagination/pagination.less
```diff
@@ -9,28 +9,96 @@
...
+                --_pa-item-fc: var(--black-600);
+            });
+
+            .highcontrast-dark-mode({
+                --_pa-item-fc: var(--black-400);
+            });
+        }
+        &.is-next {
+            background-color: #F0EFED;
+            border-radius: var(--br-circle);
```
dancormier: I'd set this as a psuedo private custom property with a value of `unset` at the top of the file and reset the value here to prevent cascade issues.

packages/stacks-classic/lib/components/pagination/pagination.less
CGuindon: Layout shift related to the focus state as well (shifts the number horizontally a bit). I think the focused state width is larger than non-focused width. I'd suggest adding some margin to non-focused state so it matches the same width — like an invisible container/box for each numbered item that's always ~20px and we remove the gutter spacing in between the items (spacing attaching to individual item itself.

CGuindon:
+1 to Dan's comments, will review again after those get resolved

packages/stacks-classic/lib/test/assertions.ts
```diff
@@ -32,5 +32,45 @@ const WCAGNonTextContrast: AdditionalAssertion = {
     },
 };
 
+const WCAGBoxShadowContrast: AdditionalAssertion = {
```
giamir: I am working on the Navigation component which have a similar way to highlight selected state but in there we won't be able to use the box shadow directly to the element. Instead we will likely need a pseudo element. All of this to say that it might be sensible to remove the check all together from this PR and have this code you created in a branch and linked to a new ticket with context. The assertion we have at the moment is too specific and should probably be generalized. If we cannot do it in a sensible way maybe even accept that it is not worth testing. 

In a nutshell my recommendation would be to extract this in a new branch and create a related ticket for the backlog. Thanks @ttaylor-stack for experimenting with this. ❤️

ttaylor-stack: Thank you, I'll do that now

ttaylor-stack: This ticket has been created in the backlog https://stackoverflow.atlassian.net/browse/SPARK-110

CGuindon (changes requested):
1. The "Prev" should be replaced by an arrow button (like we do on the right side) but pointed left.
2. The arrows for Prev/Next should use a button component and be styled with `.s-btn .s-btn__filled .s-btn__muted`. It will have the grey circle on default (not just hover like the numbers).  (only an issue in Svelte not Classic)
Screenshot 2025-11-12 at 1 28 06 PM

packages/stacks-classic/lib/components/pagination/pagination.less
```diff
@@ -9,69 +9,62 @@
...
         }
 
-        background-color: var(--_pa-item-bg);
-        border: 1px solid var(--_pa-item-bc);
+        // Only apply background-color when not a button (button classes handle their own background)
+        &:not(.s-btn) {
+            background-color: var(--_pa-item-bg);
+        }
+        border-radius: var(--_pa-item-br);
+        box-shadow: var(--_pa-item-bs);
```
dancormier: @giamir implemented a [similar selected state for the navigation component](https://github.com/StackExchange/Stacks/pull/2043/files#diff-845a558e834104583bf3479fcdc5df7efdf518047bc7e75275acb2f99692311cR56-R64) that I think could be a more reliable approach.

Instead of using box-shadow, he's using a pseudo-element (`:before`) to achieve the bottom border. The implementation here would be slightly different. We'd want it to:

- Be slightly outside the element (so `top: 100%` instead of `bottom: 0`)
- Only apply when `:not(:hover)` and `:not(focus-visible)`
- We shouldn't need any pseudo-private custom properties for this since the style shouldn't change with different variants/modifiers

Let me know if this needs clarification or if you want any feedback 🙂

packages/stacks-docs/product/components/pagination.html
```diff
@@ -74,8 +75,8 @@
             </a>
         </li>
         <li>
-            <a class="s-pagination--item" href="…">
-                Next
+            <a class="s-pagination--item s-pagination--item__nav" href="…">
+                {% icon "ArrowRight" %}
```
dancormier, line 79: The descriptive text in `span.v-visible-sr` should be something like `next page`

packages/stacks-docs/product/components/pagination.html
dancormier, line 1: Since this new design includes custom styles for next/previous, I'd suggest including a "←" button in the example with a page between the first and last selected. Something like this:

image

dancormier, line 1: @CGuindon should we include an example with ["15 / 30 / 50 items per page"](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=612-18815&p=f&m=dev) like this?

image

MIGRATION_GUIDE.md
```diff
@@ -24,6 +24,9 @@
 
 ### Components
 
+#### Pagination
+- The next button now uses an ArrowRight and ArrowLeft icon instead of text. To apply the new styling, add the `item-nav` class to pagination items that represent the "Next" button.
```
dancormier:
```suggestion
- The next and previous button now uses an `ArrowRight` and `ArrowLeft` icon instead of text. To apply the new styling, use the class `.s-pagination--item__nav`. Since these buttons use icons to represent their behavior, make sure to include descriptive text for screen readers.
```

packages/stacks-classic/lib/components/pagination/pagination.a11y.test.ts
```diff
@@ -33,6 +33,6 @@ describe("pagination", () => {
             `,
         },
         tag: "nav",
-        additionalAssertions: [WCAGNonTextContrast],
+        // additionalAssertions: [WCAGNonTextContrast],
```
dancormier: Since we're not relying on background color to denote the selected state, we can remove this line and the import altogether.

packages/stacks-classic/lib/components/pagination/pagination.less
```diff
@@ -9,69 +9,62 @@
...
-            color: var(--_pa-item-fc-focus);
+            --_pa-item-br: var(--br-md);
         }
 
-        background-color: var(--_pa-item-bg);
-        border: 1px solid var(--_pa-item-bc);
+        // Only apply background-color when not a button (button classes handle their own background)
+        &:not(.s-btn) {
+            background-color: var(--_pa-item-bg);
+        }
```
dancormier: > The arrows for Prev/Next should use a button component and be styled with .s-btn .s-btn__filled .s-btn__muted. It will have the grey circle on default (not just hover like the numbers). (only an issue in Svelte not Classic)

From a technical standpoint, these pagination buttons would just be made more complicated by relying on `.s-btn` styles. We can just apply styles via `.s-pagination` classes.

```suggestion
```

packages/stacks-docs/product/components/pagination.html
```diff
@@ -125,90 +126,14 @@
...
                     <li>
-                        <a class="s-pagination--item" href="#">
-                            Next
-                            <span class="v-visible-sr">page</span>
-                        </a>
+                        <!-- TODO: Change muted to tonal when button PR goes in-->
+                            <a class="s-btn s-btn__muted s-btn__filled s-pagination--item s-pagination--item__nav" href="#">
+                                {% icon "ArrowRight" %}
+                                <span class="v-visible-sr">page</span>
+                            </a>
```
dancormier: Related to my previous comment about not relying on `.s-btn` styles here, the styling of this button should be handled entirely with pagination styles and without any `.s-btn` styles.

packages/stacks-svelte/src/components/Pagination/PaginationController.svelte
```diff
@@ -73,8 +78,10 @@
         <PaginationItem
             url={urlGenerator(page - 1)}
             onclick={onPaginationItemClick(page - 1)}
+            itemNav={true}
         >
-            {i18nPrevText} <span class="v-visible-sr">{i18nPageText}</span>
+            <Icon src={IconArrowLeft} title={i18nPrevText} />
+            <span class="v-visible-sr">{i18nPageText}</span>
```
dancormier: Including both `title` on the icon and the screen reader span will result in some assistive tech reading out the contents of `i18nPrevText` twice. I'd recommend removing the span.

packages/stacks-svelte/src/components/Pagination/PaginationController.svelte
```diff
@@ -95,8 +102,10 @@
         <PaginationItem
             url={urlGenerator(page + 1)}
             onclick={onPaginationItemClick(page + 1)}
+            itemNav={true}
         >
-            {i18nNextText} <span class="v-visible-sr">{i18nPageText}</span>
+            <Icon src={IconArrowRight} title={i18nNextText} />
+            <span class="v-visible-sr">{i18nPageText}</span>
```
dancormier: Same as the previous comment

> Including both `title` on the icon and the screen reader span will result in some assistive tech reading out the contents of `i18nPrevText` twice. I'd recommend removing the span.

packages/stacks-svelte/src/components/Pagination/PaginationItem.svelte
```diff
@@ -20,17 +25,52 @@
...
+        if (selected) {
+            classes += ` is-selected`;
+        }
+        if (itemNav) {
+            classes += ` ${base}__nav`;
+        }
+        return classes;
+    };
+
+    const classes = $derived(getClasses(selected, itemNav));
```
dancormier: If we remove the `Button` below, I think we could remove this function altogether and just rely on conditional classes on the anchor.

packages/stacks-svelte/src/components/Pagination/PaginationItem.svelte
```diff
@@ -20,17 +25,52 @@
...
+        <a
+            class="s-pagination--item"
+            class:is-selected={selected}
+            aria-current={selected ? "page" : undefined}
+            href={url}
+            {onclick}
+        >
+            {@render children?.()}
+        </a>
+    {/if}
```
dancormier: I don't think there's a need for the `Button` element here. We should be able to just use an anchor. We'd probably want to add `class:s-pagination--item__nav={itemNav}` to apply the nav class.

packages/stacks-classic/lib/components/pagination/pagination.less
```diff
@@ -9,69 +9,62 @@
...
+            --_pa-item-fc: var(--black-600);
         }
-        // override hover styles to match base styles
-        &.is-selected,
-        &.s-pagination--item__clear {
-            --_pa-item-bc-hover: var(--_pa-item-bc);
-            --_pa-item-bg-hover: var(--_pa-item-bg);
-            --_pa-item-fc-hover: var(--_pa-item-fc);
+        &.s-pagination--item__nav {
+            --_pa-item-bg: unset;
```
dancormier: I think this is because of https://github.com/StackExchange/Stacks/pull/2035#discussion_r2501028120

> I'd set this as a psuedo private custom property with a value of unset at the top of the file and reset the value here to prevent cascade issues.


I was intending the base background color (line 11) to be set to `unset` and I now realize saying "reset the value here" is unclear. Sorry about that!

For the nav items, `--_pa-item-bg` should be set to `--black-150`.

```suggestion
            --_pa-item-bg: var(--black-150);
```

dancormier (changes requested):
Thanks for your work here @ttaylor-stack. I have a few suggestions mostly around documentation and the "next"/"previous" buttons. Let me know if anything needs clarification or if you want to pair up on my suggestions.

dancormier (approved):
Great work @ttaylor-stack! I'm happy with where this landed 🙂

PR #2037, opened 2025-11-04
fix(br): reinstate br-sm, br-lg CSS variables for fallback

dancormier:
This PR adds the `--br-sm` and `--br-lg` CSS custom properties so they can be inherited in a shadow DOM using Stacks v2. I'm not sure if we need to go with this approach of including all fallback custom properties in Stacks v3, but I figured I'd setup this PR in case we decided to.


giamir (commented):
I am unsure if we should do this within our v3 library bundle or if we should instead create a separate file either here or directly in Core. Maybe we could add all those variables in a separate file in this repo for now. Something like `v2-ccp-compat.less` (we can find a better name). Thoughts?

dancormier:
> I am unsure if we should do this within our v3 library bundle or if we should instead create a separate file either here or directly in Core. Maybe we could add all those variables in a separate file in this repo for now. Something like `v2-ccp-compat.less` (we can find a better name). Thoughts?

For some styles like border radius, having a separate file would be pretty trivial. For some others like metallic colors, those are generated using a series of mixins. We for sure can move them but it will require a little more caution on how we generate these. It could be trivial but I haven't investigated yet. I'll create a follow up ticket for this.

giamir:
> For some styles like border radius, having a separate file would be pretty trivial. For some others like metallic colors, those are generated using a series of mixins. We for sure can move them but it will require a little more caution on how we generate these. It could be trivial but I haven't investigated yet. I'll create a follow up ticket for this.

I would say let's timebox it. If we see it is too complicated I guess we could leave with the leftover v2 variables polluting v3 a little. Thanks Dan

dancormier:
I've created a ticket ([SPARK-107](https://stackoverflow.atlassian.net/browse/SPARK-107)) to tackle separating our legacy custom properties. For now, I'll merge this PR just so we can retain these custom properties and handle these custom properties another way later.

PR #2043, opened 2025-11-12
feat(navigation): update to new SHINE visuals

giamir:
[SPARK-105](https://stackoverflow.atlassian.net/browse/SPARK-105)

[Figma](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=610-18812&t=5ggsDIb9OENEwDl0-1)

Consult the new stacks classic component in the [preview docs](https://deploy-preview-2043--stacks.netlify.app/product/components/navigation/).
And the new stacks svelte component in [storybook](https://deploy-preview-2043--stacks-svelte.netlify.app/?path=/docs/components-navigation--docs)

Unresolved minor points (for another ticket potentially):
- Svelte component does not support yet the role `tablist` and `tab`, we could iterate on it once there is a concrete use case for. It is also a purely accessibility matter.
- We might benefit from a decorative-only version of the avatar component to prevent screen readers from reading text twice (in the context of navigation the avatar could be considered decorative I believe similarly to what we do for icons)

[Complex example using Svelte in Storybook (inspired by the new SHINE sidebar)](https://deploy-preview-2043--stacks-svelte.netlify.app/?path=/docs/components-navigation--docs#trailing)
I spent some time putting this example together because it saved us few ping pongs with the engineers trying to implement it and realize things were missing. @CGuindon could you confirm that this is the desired animation behavior we want in the new SHINE sidebar?

https://github.com/user-attachments/assets/dc0c88db-b231-429f-9b2e-2e1f593e5c9b


CGuindon:
Looking really good! A couple of things that jump out at me seeing this nav in the Stacks docs (small nitpick stuff)

1.**Documentation improvements** optional
These aren't necessary but would be nice to have I think — these could also live in a different type of page I know @phoebe-leung was exploring around showing examples of "larger components" using all the pieces together. We could skip these for now and I'll document below or my own purposes later:
- Including a popover with the [Dropdown](https://deploy-preview-2043--stacks.netlify.app/product/components/navigation/#dropdown) example (feels weird to click the dropdown menu button and nothing happens.
- Including a button on the right of the dropdown example (like in the Figma with the "Today" button right aligned) — common top bar pattern

---

2. **New badge** We would still want to support the option to have the "New" badge that we do sometimes now. I haven't done badges yet but we should still keep an example (maybe with this one?) that has a "new" badge like the activity indicator and it would update once that component gets completed. I didn't have an example of that in Figma — my fault.
Screenshot 2025-11-17 at 4 28 49 PM

---

giamir:
> Including a button on the right of the dropdown example (like in the Figma with the "Today" button right aligned) — common top bar pattern

I added a tonal button in one of the horizontal examples but it won't look great until the button PR is merged in beta.
Screenshot 2025-11-18 at 10 21 22

> New badge We would still want to support the option to have the "New" badge that we do sometimes now. I haven't done badges yet but we should still keep an example (maybe with this one?) that has a "new" badge like the activity indicator and it would update once that component gets completed. I didn't have an example of that in Figma — my fault.

I have added the badge in one of the examples but similarly it won't look good until the badge PR is merged. At the moment the height of the small badge is enough to mess up the 42px overall height of the navigation item containing it. This should not be a problem though with the new badge design that has considerably less height. 

Screenshot 2025-11-18 at 10 23 53 

> Including a popover with the [Dropdown](https://deploy-preview-2043--stacks.netlify.app/product/components/navigation/#dropdown) example (feels weird to click the dropdown menu button and nothing happens.

I left this out for now since it was working this way also in the previous version of the docs. We can always add it at a later stage

CGuindon:
3. **PADDING CHANGE on the selected/hover states**
Can we reduce the padding on the no icons variant (default) to be 6px top/bottom (instead of 8px) but keep the 8px for right/left — that makes a full list of no icon items fit closer together but still gives the selected state some breathing room.

Screenshot 2025-11-18 at 7 04 02 AM

CGuindon (changes requested):
Forgot to add my comment through the request change — see above

giamir:
> Forgot to add my comment through the request change — see above

@CGuindon Addressed in https://github.com/StackExchange/Stacks/pull/2043/commits/3aaf68952f2f47a79f8795441d5b7b2aeac159b8

packages/stacks-svelte/src/components/Navigation/Navigation.stories.svelte
```diff
@@ -0,0 +1,79 @@
...
+
+<Story name="Base">
+    {#snippet template()}
+        <Navigation label="Global">
+            <NavigationItem text="Label 1" selected />
+            <NavigationItem text="Label 2" />
+            <NavigationItem text="Label 3" />
+        </Navigation>
+    {/snippet}
+</Story>
```
dancormier: For the base story to be interactive, I believe we'll need to supply the args via the template.

```suggestion
<Story name="Base" args={{ label: "Global" }}>
    {#snippet template(args)}
        <!-- eslint-disable-next-line @typescript-eslint/no-unused-vars -->
        <Navigation {...(({ children, ...rest }) => rest)(args)}>
            <NavigationItem text="Label 1" selected />
            <NavigationItem text="Label 2" />
            <NavigationItem text="Label 3" />
        </Navigation>
    {/snippet}
</Story>

```

The weird spread approach was the cleanest way I could find to prevent an error caused by including the `children` prop.

dancormier (approved):
The style changes look great overall @giamir. I have one small suggestion for the Svelte docs that I wrote _before_ scrolling to all the `TODO`s. I should've assumed it was a WIP 🤦

Let me know when this is ready for a full review and I'd be happy to do so.

---

FYI, I pushed a [small commit](https://github.com/StackExchange/Stacks/pull/2043/commits/741c49ebe5b45766e2f2e498523eadc241e8d573) to remove some cruft that created duplicate text displayed for some items in our docs.

dancormier:
> The only thing missing should be updating the dropdown caret ([@dc](https://stackexchange.slack.com/team/U0230RJJ136) I would appreciate a suggestion from you on how to proceed since the caret changed significantly).

For the caret, I'd suggest using a pseudo-element much like I did on Menu (see https://github.com/StackExchange/Stacks/pull/2046/commits/8c4c1133e06b79706437e9647d71b6013a4f27e0).

The gist is that the SVG(s) will need to be hard coded into the Less with a Less variable to define the fill on a per-mode basis. I'd expect something like this:

```less
.s-navigation {
…
    // Less variables for check svg fill color
    @me-bg-image-fill: .set-black()[default];
    @me-bg-image-fill-dark: .set-white()[default];
    @me-bg-image-fill-esc: escape("@{me-bg-image-fill}"); // color escaped for URL usage
    @me-bg-image-fill-dark-esc: escape("@{me-bg-image-fill-dark}"); // color escaped for URL usage
    --_na-after-bg-image: url("data:image/svg+xml… fill='@{me-bg-image-fill-esc}'…");
    
        // CONTEXTUAL STYLES
    .dark-mode({
        --_na-after-bg-image: url("data:image/svg+xml… fill='@{me-bg-image-fill-dark-esc}'…");
    });
…
    & &--item {
        &__dropdown {
            &:after {
                background-image: var(--_na-after-bg-image);
                content: "";
                display: inline-block;
                height: var(--su16);
                margin-left: auto;
                width: var(--su16);
            }
…
        }
    }
}
```

I know it seems a little weird, but it's the best approach I've found for applying our colors to background SVGs while still supporting our color modes. It's much more reasonable when the color is monochromatic, like it is here. Let me know if this makes sense and if you'd like any more help. Happy to do so 🙂

packages/stacks-svelte/src/components/Navigation/Navigation.stories.svelte
giamir: Thanks for the suggestion. Will look into it more today. ❤️

giamir:
@dancormier for the caret I think I found a solution that could be slightly more elegant (unless I am missing something). 
See this commit and let me know your thoughts: https://github.com/StackExchange/Stacks/pull/2043/commits/05685a22de1e95e5fefdec2d375ecdfe0e3c93a1 🙂

dancormier (approved):
Great work here @giamir!

CGuindon:
The only thing I noticed was that the background on this bottom part of the documentation example should be white (not grey?) but I'm going to approve anyway since the actual component looks great!

Screenshot 2025-11-21 at 10 02 38 AM

CGuindon (approved):

PR #2045, opened 2025-11-12
fix(tags): add screen reader improvements

mukunku:
# Summary

This PR addresses some accessibility gaps with regards to our Tags component and screen readers: https://stackoverflow.atlassian.net/browse/A11Y-89

It does the following:
* Makes sure tags are tabbable by appending `tabindex` when using `<span>`s instead of `<a>`s
* Make sure we have screen reader text for the dismiss button in our documentation
* Add screen reader text for the four different tag states: ignored, watched, moderator, required

# How to Test

Confirm screen reader improvements in:
* Stacks Docs: https://deploy-preview-2045--stacks.netlify.app/product/components/tags/
* Stacks Svelte: https://deploy-preview-2045--stacks-svelte.netlify.app/?path=/docs/components-tag--docs


packages/stacks-docs/product/components/tags.html
```diff
@@ -37,15 +43,15 @@
 {% highlight html %}
 <div class="d-flex g4">
     <a class="s-tag" href="#">jquery</a>
-    <span class="s-tag">javascript <button class="s-tag--dismiss">@Svg.ClearSm</button></span>
-    <a class="s-tag" href="#"><img class="s-tag--sponsor" src="https://i.stack.imgur.com/tKsDb.png" width="16" height="18" alt="Google Android"> android</a>
+    <span class="s-tag" tabindex="0">javascript <button class="s-tag--dismiss"><span class="v-visible-sr">Dismiss</span>@Svg.ClearSm</button></span>
```
dancormier: NIT: The screen reader text should give an idea of what the focused element does. `Dismiss` is good, but I think something like `Dismiss tag` or maybe even `Dismiss javascript tag` would be provide a good example.

```suggestion
    <span class="s-tag" tabindex="0">javascript <button class="s-tag--dismiss"><span class="v-visible-sr">Dismiss javascript tag</span>@Svg.ClearSm</button></span>
```

packages/stacks-docs/product/components/tags.html
```diff
@@ -37,15 +43,15 @@
...
+    <a class="s-tag" href="#"><img class="s-tag--sponsor" src="https://i.stack.imgur.com/tKsDb.png" width="16" height="18" alt="Google Android"> android <div class="v-visible-sr">Sponsored tag</div></a>
 </div>
 {% endhighlight %}
         <div class="stacks-preview--example">
             <div class="d-flex g4 fw-wrap">
                 <a class="s-tag" href="#">jquery</a>
-                <span class="s-tag" href="#">javascript <button class="s-tag--dismiss">{% icon "ClearSm" %}</button></span>
-                <a class="s-tag" href="#"><img class="s-tag--sponsor" src="https://i.stack.imgur.com/tKsDb.png" width="16" height="18" alt="Google Android"> android</a>
+                <span class="s-tag" tabindex="0">javascript <button class="s-tag--dismiss"><span class="v-visible-sr">Dismiss</span>{% icon "ClearSm" %}</button></span>
+                <a class="s-tag" href="#"><img class="s-tag--sponsor" src="https://i.stack.imgur.com/tKsDb.png" width="16" height="18" alt="Google Android"> android <div class="v-visible-sr">Sponsored tag</div></a>
```
dancormier, line 52:
```
<span class="s-tag" href="#">javascript <button class="s-tag--dismiss">{% icon "ClearSm" %}</button></span>
```
Whoops! Where'd that `href="#"` on that `span` come from? I must've been tired 😆

packages/stacks-docs/product/components/tags.html
dancormier: I'm not sure adding `tabindex="0"` is the best move here. I recommend reserving tabbing for interactive elements. Generally, these dismissable tags will all be grouped together so I think context mixed with my suggestion for slightly more descriptive screen reader text on the dismiss button is preferable. Let me know if I'm overlooking some details.

```suggestion
    <span class="s-tag">javascript <button class="s-tag--dismiss"><span class="v-visible-sr">Dismiss</span>@Svg.ClearSm</button></span>
```

dancormier (changes requested):
Thanks for tackling these accessibility improvements @mukunku ❤️ I'm 100% on board for the watched/ignored/etc updates and the addition of descriptive text for the dismiss button.

The only real reservation I have is the addition of `tabindex="0"` to tags that aren't interactive. Screen readers should be able to generally assist with informing the user of static text and should prioritize tabbing for interactive elements. Generally, I reserve it for either elements that are scrollable (like the code examples in the docs) or elements that are not interactive by default but made interactive via JS (this should almost always be avoided but exceptions occasionally arise).

Let me know if this makes sense or not, or if I'm misinterpreting the purpose

packages/stacks-docs/product/components/tags.html
mukunku: I went with your suggestion as it also resolved the issue I was trying to fix with `tabindex="0"`!

mukunku: Thanks for the feedback Dan. I forgot that these tags aren't interactable, only the dismiss button is. My intention behind adding the tabindex was so I could get my screen reader to read the tag itself. Without the tabindex, when using keyboard navigation, I can only focus the dismiss button and the screen reader just reads "dismiss button"; no way to get which tab I'm dismissing.

But your idea to include the tag's name in the dismiss button's screen reader content solves that concern for me so I removed the tabindex's now 👍🏼 https://github.com/StackExchange/Stacks/pull/2045/commits/89cdf247a210fd870ae7b03424635e91a90f53c3

It does cause a little inconsistency in our Svelte component as [we are not automatically injecting the tag name](https://github.com/StackExchange/Stacks/blob/8e628a58bd6b9d96a0c79637d4540af21114632e/packages/stacks-svelte/src/components/Tag/Tag.svelte#L196) into the screen reader content there 🤷🏼

dancormier: > It does cause a little inconsistency in our Svelte component as [we are not automatically injecting the tag name](https://github.com/StackExchange/Stacks/blob/8e628a58bd6b9d96a0c79637d4540af21114632e/packages/stacks-svelte/src/components/Tag/Tag.svelte#L196) into the screen reader content there 🤷🏼

Yeah, if we were only allowing text to be passed in place of the `children` snippet, we could construct default text for the dismiss button that includes the tag name (like `Dismiss ${tagName} tag`) but I think we'll need to rely on consumers setting the appropriate value for now. I'm ok with that but I agree it's not ideal.

dancormier (approved):
Thanks @mukunku! :shipit:

PR #2046, opened 2025-11-12
chore(menu): update to SHINE styles

dancormier:
[SPARK-71](https://stackoverflow.atlassian.net/browse/SPARK-71)

---

This PR updates the Menu component to match updated SHINE styles ([Figma](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=2021-32195&t=c9NSyJWdASb80eVC-0)) and adds a Svelte version of the component.

For the Svelte component, I made the choice to go with child components much like we do with `PostSummary`. I'm not 100% on this approach and would appreciate any thoughts/opinions here.

### How to test
- View the [component](https://deploy-preview-2046--stacks.netlify.app/product/components/menus/) and verify it renders as expected
  - Please read the docs and provide feedback
- Go to the [Svelte component](https://deploy-preview-2046--stacks-svelte.netlify.app/?path=/docs/components-menu--docs) and verify it renders as expected
- Run tests and verify they pass (excluding visual tests which will be updated before merge)

### TODO before merge
- [ ] Verify docs are reasonably updated
- [ ] Regenerate visual regression test images


packages/stacks-docs/product/components/menus.html
giamir, line 1: Add link to Svelte and Figma in frontmatter

giamir: Missing svelte and figma frontmatter links

packages/stacks-classic/lib/components/menu/menu.less
```diff
@@ -1,41 +1,100 @@
...
+
+            &:focus-visible,
+            &:hover {
+                --_me-link-fc: var(--red-500);
+            }
+        }
+
+        &.is-selected {
+            &:after {
+                content: '✓'; // TODO SHINE replace with proper svg icon
```
giamir: I think we already have this icon available in https://beta.icons.stackoverflow.design/
See `Check16`

packages/stacks-svelte/src/components/Menu/Menu.stories.svelte
```diff
@@ -0,0 +1,189 @@
...
+                options: MenuTags,
+            },
+        },
+    });
+</script>
+
+{#snippet basicChildren()}
+    <MenuItemLink href="#">Share</MenuItemLink>
+    <MenuItemLink href="#">Edit</MenuItemLink>
+    <MenuItemLink>Follow</MenuItemLink>
```
giamir: Is a link with no href a valid link? Should `href` be required?

packages/stacks-svelte/src/components/Menu/Menu.stories.svelte
```diff
@@ -0,0 +1,189 @@
...
+{/snippet}
+
+<Story name="Base">
+    {#snippet template(args)}
+        <div class="s-popover is-visible ps-relative ws2 p8">
```
giamir: Dogfood, use Svelte Popover, not stacks classic

packages/stacks-svelte/src/components/Menu/MenuItem.svelte
```diff
@@ -0,0 +1,104 @@
...
+        return classes;
+    };
+
+    const itemClasses = $derived(getItemClasses(className));
+    const linkClasses = $derived(getLinkClasses("", danger, selected));
+</script>
+
+<li class={itemClasses} role="menuitem">
+    <svelte:element
+        this={href ? "a" : "button"}
```
giamir, line 94: I think the name of this component should just be `MenuItem` since it is not always a link (can also be a button)

packages/stacks-docs/_data/menu.json
```diff
@@ -3,22 +3,37 @@
...
     {
-      "class": ".s-menu--title",
+      "class": ".s-menu--divider",
       "applies": ".s-menu",
-      "description": "Adds appropriate styling for a title within a menu"
+      "description": "Adds a divider line between menu sections."
     },
     {
-      "class": ".s-menu--divider",
+      "class": ".s-menu--item",
```
giamir, line 14: I am unsure about the naming here. Since this can be also a button should we call it something more generic (`s-menu--action`)

packages/stacks-svelte/src/components/Menu/Menu.svelte
giamir, line 1: I don't think there is a need to unit test every single subcomponents like that. The value of those tests is questionable and they add a lot of friction to evolve the overall API. I think it is a better idea to just have a single test file testing the different combination. Check what we do in Popover or Pagination. You will end up with fewer tests to maintain which are much more valuable in my opinion.

packages/stacks-svelte/src/components/Menu/MenuItemControl.svelte
```diff
@@ -0,0 +1,96 @@
...
+    const getRole = (type: ControlType) => {
+        return type === "radio" ? "menuitemradio" : "menuitemcheckbox";
+    };
+
+    const itemClasses = $derived(getItemClasses(className));
+    const inputClass = $derived(getInputClass(type));
+    const role = $derived(getRole(type));
+</script>
+
+<label class={itemClasses} for={id} {...restProps}>
```
giamir: Dogfood Label
https://github.com/StackExchange/Stacks/blob/beta/packages/stacks-svelte/src/components/Label/Label.svelte

packages/stacks-svelte/src/components/Menu/MenuItemControl.svelte
```diff
@@ -0,0 +1,96 @@
...
+    let {
+        type,
+        id,
+        name,
+        checked = false,
+        disabled = false,
+        label,
+        description,
+        class: className = "",
+        ...restProps
```
giamir: Why do we spread these props? It is better to be conservative at this stage. Would recommend to remove this and add only what consumers need.

giamir (changes requested):
Great work @dancormier. I have left few comments and suggestions. 

> The examples for this component show radio group versions as well as a selected state. Both of these are subject to change after this PR is merged and once work is done on the radio component ([SPARK-77](https://stackoverflow.atlassian.net/browse/SPARK-77)).

Do we have a follow up ticket for updating the Menu component once Radio and Checkbox have been implemented?
Crazy thought, should we leave the `MenuItemControl` Svelte subcomponent out of scope for this PR since it will reuse Radio and Checkbox (do the work as part of the follow up ticket)?

packages/stacks-docs/_data/menu.json
giamir: I am talking about `s-menu--link` of course, gh higlighted something different.

packages/stacks-svelte/src/components/Menu/Menu.svelte
dancormier: Addressed in https://github.com/StackExchange/Stacks/pull/2046/commits/32e3619acc687de62ccf2a8339268fd52da4259f

packages/stacks-docs/product/components/menus.html
dancormier: Addressed in https://github.com/StackExchange/Stacks/pull/2046/commits/359b7390916402e171d4cb91d87c26ba1efd4ced

packages/stacks-classic/lib/components/menu/menu.less
dancormier: Addressed in https://github.com/StackExchange/Stacks/pull/2046/commits/f4aa2d8958972e3c4c55609ad44f648d0c888718

packages/stacks-svelte/src/components/Menu/Menu.stories.svelte
dancormier: Renamed to just `MenuItem` to resolve confusion

packages/stacks-svelte/src/components/Menu/Menu.stories.svelte
dancormier: Addressed in https://github.com/StackExchange/Stacks/pull/2046/commits/b364354541faf5304132e46455e096362c434cf4

packages/stacks-svelte/src/components/Menu/MenuItem.svelte
dancormier: Addressed in https://github.com/StackExchange/Stacks/pull/2046/commits/787d6f853215912d31b9ae5b277ab97e20ed02ce

packages/stacks-docs/_data/menu.json
dancormier: Addressed in https://github.com/StackExchange/Stacks/pull/2046/commits/51cdac208b4a74e4bc18a2820b8bd117d6903f67

packages/stacks-svelte/src/components/Menu/MenuItemControl.svelte
dancormier: Removed `MenuItemControl` in https://github.com/StackExchange/Stacks/pull/2046/commits/356ebccd0f3242824dad61b00916d5d58238e497 so this is no longer relevant

packages/stacks-svelte/src/components/Menu/MenuItemControl.svelte
dancormier: Removed MenuItemControl in https://github.com/StackExchange/Stacks/commit/356ebccd0f3242824dad61b00916d5d58238e497 so this is no longer relevant

CGuindon (approved):

packages/stacks-classic/lib/components/menu/menu.less
```diff
@@ -1,39 +1,126 @@
 .s-menu {
-    --_me-divider-bg: var(--bc-light);
-    --_me-label-btc: var(--bc-light);
-    --_me-label-cursor: pointer;
+    --_me-action-bg: unset;
+    --_me-action-fc: var(--black-500);
+    --_me-item-p: var(--su8);
+    // Less variables for check svg fill color
```
giamir: Have you tried the mask approach I used in the navigation PR?
https://github.com/StackExchange/Stacks/pull/2043/files#diff-845a558e834104583bf3479fcdc5df7efdf518047bc7e75275acb2f99692311c

giamir (approved):
Great work @dancormier. ❤️ 
I only had a small clarifying comment. Apart from that I think we should be good to merge. Thank you.

PR #2061, opened 2025-11-21
feat(popover): add the ability to add extra css classes to s-popover--content element

chris-doucette-stack:
## Summary 
This PR adds a new prop `contentClass` to the `PopoverContent` component. This prop add additional classes on the inner content div (s-popover--content) that is otherwise inaccessible. This change is important for popovers with custom padding values to prevent the inner div from overflowing from its parent container.


giamir (approved):
@chris-doucette-stack Thanks for the PR. I took the liberty of pushing a commit with my suggestions directly in this PR branch. I hope that's ok. We can always revert it. 

In a nutshell I think we should treat this escape hatch as we treat any other class addition props. It should not override but it should give you the option to append atomic classes which will ultimately have the override effects you need. I have also added an extra story to storybook to show how custom classes could be applied. 
https://deploy-preview-2061--stacks-svelte.netlify.app/?path=/story/components-popover--custom-classes

Let's wait for a thumbs up from @dancormier and then merge.

dancormier (approved):
LGTM! Thanks @chris-doucette-stack and @giamir!

PR #2062, opened 2025-11-21
feat(vote): add vote component

dancormier:
[SPARK-103](https://stackoverflow.atlassian.net/browse/SPARK-103)

---

[Figma](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=617-18830&p=f&m=dev)
[Docs (Stacks Classic)](https://deploy-preview-2062--stacks.netlify.app/product/components/vote/)
[Storybook (Stacks Svelte)](https://deploy-preview-2062--stacks-svelte.netlify.app/?path=/docs/components-vote--docs)

---

This PR adds a _vote_ component to Stacks Classic and Stacks Svelte.

> [!NOTE]
> At the time of writing, downvote icons are not yet design so are instead shown using the placeholder icons. Once the downvote icons are updated in the stacks-icons packages, the correct icons should show.

<details>

#### Light
image

#### Dark
image

#### Light HC
image

#### Dark HC
image

### Expanded
image

### Horizontal
image

### Voted
image

</details>


mukunku:
Not sure if this is ready for review but noticed the downvote button looks weird: https://deploy-preview-2062--stacks-svelte.netlify.app/?path=/docs/components-vote--docs
{D90C11DB-CAD4-4428-865E-9ACC1050C816}

packages/stacks-docs/product/components/vote.html
```diff
@@ -0,0 +1,242 @@
...
+                        </button>
+                    </div>
+                </div>
+            {% endfor %}
+        </div>
+    </div>
+
+    {% header "h3", "Expanded" %}
+    <p class="stacks-copy">
+        Include the <code class="stacks-code">.s-vote__expanded</code> modifier to show upvote and downvote counts in addition to the total vote count. This modifier hides <code class="stacks-code">.s-vote--total</code> and shows <code class="stacks-code">.s-vote--upvotes</code> and <code class="stacks-code">.s-vote--downvotes</code> instead.
```
mukunku: Should this be "instead of" instead of "in addition to"? Seems like the very next sentence contradicts the first one.

packages/stacks-docs/product/components/vote.html
```diff
@@ -0,0 +1,242 @@
...
+                        </button>
+                    </div>
+                </div>
+            {% endfor %}
+        </div>
+    </div>
+
+    {% header "h3", "Horizontal" %}
+    <p class="stacks-copy">
+        Apply the <code class="stacks-code">.s-vote__horizontal</code> modifier to arrange the vote buttons and counts in a horizontal layout. This layout does not officially support expanded vote count. This configuration is best suited scenarios such as comment voting, where a more compact design is preferred.
```
mukunku: small nit: missing "for" 
```suggestion
        Apply the <code class="stacks-code">.s-vote__horizontal</code> modifier to arrange the vote buttons and counts in a horizontal layout. This layout does not officially support expanded vote count. This configuration is best suited for scenarios such as comment voting, where a more compact design is preferred.
```

mukunku (commented):
Changes are looking great. Noticed one thing: the selected state for horizontal layout doesn't match the one for comments in designs. Similarly, you can't click on the whole thing to upvote which I'm guessing isn't the intention 🤔 

{279EFF0C-0FD5-4DC7-8971-7C1DA5E9CDFF}
{26D5D4B6-493D-4237-9F94-0AD33ACBB63A}

Do we need an `upvote-only` mode or something maybe?

CGuindon:
Downvote icons are available now (the vote icon but direction set to down)

CGuindon:
[Confirming](https://github.com/StackExchange/Stacks/pull/2062#pullrequestreview-3572999327) we don't have a use case currently where we have a horizontal version with both up and downvotes. We only have the horizontal single vote so wondering if we should just **not** show the version with 2 votes on horizontal (people shouldn't use this). 

I'm ok keeping the version you have though where both up/down are possible and the focus/hover stays on one side. I think it still works fine even with just the upvote and it would save us time later if product decides to add in downvotes to comments/replies.

CGuindon (changes requested):
New downvote icon needed (overall looks good though!)

dancormier:
> the selected state for horizontal layout doesn't match the one for comments in designs

Nice catch @mukunku! I built it with this in mind originally but completely spaced when making the examples and Svelte component. Fixed!

---

> I'm ok keeping the version you have though where both up/down are possible and the focus/hover stays on one side. I think it still works fine even with just the upvote and it would save us time later if product decides to add in downvotes to comments/replies.

@CGuindon Sal pointed out that I missed the configuration in the Figma. I've updated the horizontal orientation to match what's in Figma where the element is just one upvote button that includes the vote count. If/when we want to allow for downvotes on the horizontal orientation, we'll be able to do that pretty trivially.

> Downvote icons are available now (the vote icon but direction set to down)

The standard (20x20) downvote icon is available but we need the 16px ones for this component. We can still merge without them and once they're published and the icons library dependency is updated, this component will display them with no further changes besides regenerating the test images.

mukunku (approved):
LGTM!

CGuindon (approved):

giamir (approved):
Great work @dancormier. I really like this component. 

The only thing I noticed is that we automatically update the total number but not the upvotes and downvotes numbers. 

Screenshot 2025-12-17 at 10 46 47

I am actually unsure if updating the different counts is something we should do in the design system component directly or rather have the consumers take care of it.  We don't have control over what the server returns and how the API looks like and we risk to show an incorrect value to the user (imagine a call is done to the server to register an upvote and the server come back with a different score because in the meantime somebody else has also casted a vote, etc...).

That said I am happy to merge it as is and maybe have those conversations with the first team implementing the component. 

A small nit. the prop we call `total` in the component is called `score` in our APIs (check for [example](https://sem1.dev.stackenterprise.co/api/v3/swagger/index.html) the schema of the response for `questions/id`). It would be nice probably to stick to score also here but total is ok too. 🙂

CGuindon:
The Svelte component was built to update the count and wait for a promise to resolve to confirm the vote count was updated on the server-side. It makes sense to do the same thing for the separate counts. 

[Slack thread](https://stackexchange.slack.com/archives/C097YESGSMB/p1765987695613159)

PR #2064, opened 2025-11-22
Update TextInput.svelte to not emit a self-closing tag on the `<input>` element

kylejrp:
Was running some output from a Svelte component that used `<TextInput/>` through W3C's HTML validator and got this informational message:

image

It links to this guidance:
https://github.com/validator/validator/wiki/Markup-%C2%BB-Void-elements#trailing-slashes-in-void-element-start-tags-do-not-mark-the-start-tags-as-self-closing

TL;DR: it does nothing so we shouldn't emit it

It doesn't bother me too much personally, but it does check off a validation box 🤷

### DIFF
--- a/packages/stacks-svelte/src/components/TextInput/TextInput.svelte
+++ b/packages/stacks-svelte/src/components/TextInput/TextInput.svelte
@@ -222,7 +222,7 @@
                 on:focus
                 on:blur
                 on:paste
-            />
+            >
 
             {#if state}
                 <div class="s-input-icon">


mukunku (approved):

mukunku (changes requested):
After deeper investigation, unfortunately [this breaks linting](https://github.com/StackExchange/Stacks/actions/runs/19588661033/job/56102619855?pr=2064) because prettier **requires** self terminating elements and this can't be changed: https://github.com/prettier/prettier/issues/5641

PR #2066, opened 2025-11-24
chore(activity-indicator): rm HC overrides

dancormier:
This PR removes all the HC overrides for the Activity Indicator component. A11Y tests pass, so I think we're no longer need the overrides.

### Code changes

All code changes are localized to activity-indicator.less so that should be all the code you need to look at. All other changes are visual regression images being regenerated.

### Visual changes

1. Go to the [activity indicator component docs](https://deploy-preview-2066--stacks.netlify.app/product/components/activity-indicator/)
2. Turn on High Contrast mode using the mode switcher in the top right of the topbar
3. Notice that it displays the HC mode versions of the base component/variant colors (i.e. no overrides)

### Before

image
image

### After

image
image


mukunku (approved):
LGTM 👍🏾

CGuindon:
Not _totally_ related to activity indicator and probably only an issue in the docs but the G in the avatar example goes dark when in Dark HC mode (it's fine in Light HC mode)
Screenshot 2025-11-24 at 9 00 45 AM

CGuindon (approved):
Activity indicator LGTM

giamir (approved):

PR #2068, opened 2025-11-25
feat(modal): update to the new SHINE visuals

giamir:
[SPARK-89](https://stackoverflow.atlassian.net/browse/SPARK-89)

[Figma](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=5405-11871&m=dev)

This PR upgrade the modal to new SHINE visual styles.

Most of the change cascade down from the border radius, typography and color changes.
The only noticeable change is the swap of the old `ClearSm` legacy icon with the new `Cross` icon. padding and positioning of the close button has been adjusted as well as per figma specs. 

Docs Preview:
- [Stacks Classic](https://deploy-preview-2068--stacks.netlify.app/product/components/modals/)
- [Stacks Svelte](https://deploy-preview-2068--stacks-svelte.netlify.app/?path=/docs/components-modal--docs)


ttaylor-stack (approved):

CGuindon (changes requested):
1. The danger modal should have black/white text as the title like the default modal (not red). Only the button uses the danger red color.
Screenshot 2025-11-25 at 9 26 51 AM

---

2. Customizations per mode: the only tricky thing about modal is with the background color and dark/dark HC mode. We need to tweak/customize in dark modes in order to make them visible enough.

**A. For dark mode and Dark HC**
Background of the modal should be `black-100` (instead of 225)
Screenshot 2025-11-25 at 9 22 16 AM

**B. For Dark/Light HC mode**
We have an old ticket from Stacks to fix HC modals for a11y. They need to have a 2px outline of `black-050` added around the modal itself. [Details in this old ticket here.](https://stackoverflow.atlassian.net/browse/STACKS-476)

giamir:
> 1. The danger modal should have black/white text as the title like the default modal (not red). Only the button uses the danger red color.
> 
> Screenshot 2025-11-25 at 9 26 51 AM
> 2. Customizations per mode: the only tricky thing about modal is with the background color and dark/dark HC mode. We need to tweak/customize in dark modes in order to make them visible enough.
> 
> **A. For dark mode and Dark HC** Background of the modal should be `black-100` (instead of 225) Screenshot 2025-11-25 at 9 22 16 AM
> 
> **B. For Dark/Light HC mode** We have an old ticket from Stacks to fix HC modals for a11y. They need to have a 2px outline of `black-050` added around the modal itself. [Details in this old ticket here.](https://stackoverflow.atlassian.net/browse/STACKS-476)

@CGuindon I have applied all the changes you requested. For the outline I used `black-500` because `black-050` was not visible at all. I think you might have inverted the scale. Have a look and let me know if I need to perform further changes. Thank you.

CGuindon (approved):

PR #2069, opened 2025-11-25
feat(select): update Select component to SHINE styles

ttaylor-stack:
Updated border radius and removed md and lg sizes. Updated the Chevron and Check icons. 

ticket: https://stackoverflow.atlassian.net/browse/SPARK-78
Figma: https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=720-3056&p=f&m=dev

NOTE: I have created this ticket to update the Alert icons as they are not ready -> https://stackoverflow.atlassian.net/browse/SPARK-125


packages/stacks-classic/lib/components/input-icon/input-icon.less
```diff
@@ -38,7 +38,7 @@
     color: var(--_ii-fc);
     right: var(--_ii-r);
 
-    margin-top: calc((var(--su-static8) + var(--su-static1)) * -1); // -9px - Half the icon's height at 18px for centering;
+    margin-top: calc((var(--su-static8) + var(--su-static2)) * -1); // -10px - Half the icon's height at 20px for centering
```
ttaylor-stack, line 41: @CGuindon I have assumed we are using the 20px version of the Check and Alert icons (I've kept Chevron at 16px). The old icons are 18px so the formatting of the alert icons currently looks slightly off, once we have the updated icons and update the icons the formatting should return to normal

Example: https://deploy-preview-2069--stacks.netlify.app/product/components/select/#error

giamir (approved):
Overall looks good to me. The only thing I noticed is that the check icon in the Svelte component has not been updated to the new one: https://github.com/StackExchange/Stacks/blob/06ee846046e2f5e6706dc73ffbe3588ce66acdce/packages/stacks-svelte/src/components/Select/Select.svelte#L189

I approved but let's also wait for design feedback from Caro before merging.

Great work @ttaylor-stack 🎉

CGuindon (changes requested):
I'm going to suggest a few tweaks to the spacing for each size. Padding should be the same for the input as well (not picked up yet but ideally they match heights when next to each other). I can see that the large size has additional overriding classes, I think you'll have to do that for the small size as well. Tagging in @mukunku since he's working on that ticket now.

**Default size**
Screenshot 2025-11-25 at 8 55 55 AM
Screenshot 2025-11-25 at 8 58 54 AM

---

**Small size**
Screenshot 2025-11-25 at 9 08 32 AM
Screenshot 2025-11-25 at 9 08 50 AM

---

**Large size**
Screenshot 2025-11-25 at 9 06 33 AM
Screenshot 2025-11-25 at 9 07 03 AM

packages/stacks-classic/lib/components/select/select.less
```diff
@@ -23,30 +23,15 @@
...
     &&__sm {
         .size-styles(sm; se-select; @styles: fs);
-        // --_se-select-fs: var(--fs-caption);
-    }
-
-    &&__md {
-        .size-styles(md; se-select; @styles: br, fs);
-        --_se-select-py: 0.5em;
+        --_se-select-lh: var(--lh-sm);
+        --_se-select-py: calc(var(--su6) - var(--su-static1) / 2);
```
ttaylor-stack, line 29: @dancormier , This is the best I could do with the small variant to get within .5 pixels of 28 pixels. Every other line height + padding combo gave me a height that too big or too small

CGuindon (approved):

PR #2072, opened 2025-11-25
docs(storybook): support switching between light, dark and high constrast modes

giamir:
This is a storybook setup improvements PR to easily test our svelte components in dark and high constrast modes. It is not perfect but I think it is better to what we have now. 

Things to look at:
- ~iframes of story preview in Docs pages is fixed to 400px height (how easily we can make it dynamic based on the content of the iframe?)~
   - I managed to get the iframes height to get updated to fit the content after the story load - it is not perfect but I think it is acceptable
  
- ~Somehow in regular story pages (e.g. Base story) the mode does not change as soon as we select it from the drop down, instead also a manual page refresh is necessary.~
   - Also this one now seems to work. 

Test it out here:
https://deploy-preview-2072--stacks-svelte.netlify.app/


mukunku:
This is really cool and useful! 

I don't quite get why we need to specify a height now (400px), when we weren't before. It doesn't look too great with the static height:
{752318C4-B0E4-47D2-8E9A-D0F2A8562B7E}

Regarding the component not updating to user selection changes, any ideas why that might be happening?

packages/stacks-svelte/src/components/Menu/Menu.stories.svelte
giamir, line 1: The menu story had to be adjusted because the way the max height of the popover is calculated is based off the viewport height which in the iframes is initially low. `--_po-content-mxh: calc(100vh - var(--_po-topbar-height) - var(--su48))`

packages/stacks-svelte/.storybook/manager-head.html
giamir, line 1: Unfortunately we need to handle the resizing of the iframe height based on content ourselves. There is an open issue for the storybook addon docs: https://github.com/storybookjs/storybook/issues/8112

ttaylor-stack (approved):
LGTM!

dancormier:
@giamir since this change breaks the controls for the primary story, I feel like we should only include the mode switcher once the controls work. I've reverted this change in https://github.com/StackExchange/Stacks/commit/d2e8a000555afb9c34283aee90070bff3444c394

PR #2073, opened 2025-11-25
Input tweaks

dancormier:
This PR contains minor tweaks to the input component that I think gets reasonably close to our desired input sizes/layouts while keeping them pretty maintainable and flexible.

### Sizing

This PR modifies the line-height and padding properties to achieve our desired input heights for each size within .5px of the mocks. Here's a table showing the difference in heights between the Figma mocks and the input sizes in this PR:

| Size | Figma | This PR | Diff | 
|-|-|-|-|
| `sm` | 28px | 27.7px | -0.3px |
| default | 36px | 35.9px | -0.1px |
| `lg` | 46px | 46.2px | +0.2px |

We should be able to apply the same sizing to other similar element to achieve the same heights

<details>

image
image
image

</details>

### Nested input

I tried to keep the extra CSS to support this as minimal as possible while providing sensible defaults that account for the different likely states of this configuration. Below, you can see a tag input with in various states (one tag, no tags, many tags):

image

This would still require some atomic classes on the element containing the tags, which I feel is a reasonable compromise.

### Icons within/over inputs

Since we're switching to 20px² icons, I've updated some sizing/positioning to support this change and I've added TODOs to the docs to update examples to the appropriate icons once they're available.

### Alias colors

We have various aliased colors for different properties in the form of custom properties. They look like `--bc-darker`, `fc-light`, etc. I'd like the library to move away from using these aliased properties as I think they are a bit intuitive and used inconsistently. For these reasons, I've replaced them with the actual color variable they reference (e.g. `--bc-darker` becomes `--black-300`).

### DIFF
--- a/packages/stacks-classic/lib/components/input_textarea/input_textarea.less
+++ b/packages/stacks-classic/lib/components/input_textarea/input_textarea.less
@@ -47,12 +47,14 @@
     &&__sm {
         --_in-fs: var(--fs-caption);
         --_in-lh: var(--lh-sm);
+        --_in-px: calc(var(--su12) - var(--su2)); // 10px
         --_in-py: calc(var(--su4) + var(--su1)); // 5px
     }
 
     &&__lg {
         --_in-fs: var(--fs-body3);
         --_in-lh: var(--lh-lg);
+        --_in-px: calc(var(--su12) + var(--su1)); // 13px
     }
     
     // CHILD ELEMENTS
--- a/packages/stacks-classic/lib/components/input_textarea/input_textarea.less
+++ b/packages/stacks-classic/lib/components/input_textarea/input_textarea.less
@@ -1,18 +1,16 @@
 .s-input,
 .s-textarea {
-    --_in-bc: var(--bc-darker);
+    --_in-bc: var(--black-300);
     --_in-bg: var(--white);
     --_in-br: var(--br-md);
     --_in-bw: var(--su-static1);
     --_in-c: unset;
-    --_in-fc: var(--fc-dark);
+    --_in-fc: var(--black-600);
     --_in-fs: var(--fs-body1);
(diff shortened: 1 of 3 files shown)


ttaylor-stack:
Not entirely relevant to this PR but I've created this ticket to update the alert icons when we get them https://stackoverflow.atlassian.net/browse/SPARK-125

mukunku (approved):

packages/stacks-docs/product/components/inputs.html
```diff
@@ -262,7 +264,7 @@ <h1 class="fs-headline1 fw-normal mb16">
                     <label class="flex--item s-label" for="example-success">Username</label>
                     <div class="d-flex ps-relative">
                         <input class="s-input" id="example-success" type="text" value="aaronshekey" aria-describedby="example-success-desc" />
-                        {% icon "Checkmark", "s-input-icon" %}
+                        {% icon "Check", "s-input-icon" %}
```
mukunku, line 267: I could've sworn I couldn't find this on https://beta.stackoverflow.design/product/foundation/icons/ this week. Thanks for fixing

packages/stacks-classic/lib/components/input_textarea/input_textarea.less
```diff
@@ -48,15 +45,14 @@
 
     // Sizes
     &&__sm {
-        .size-styles(sm; in; @styles: fs);
-        --_in-min-h: calc(var(--su32) - var(--su4)); // 28px
-        --_in-px: calc(var(--su8) + var(--su2)); // 10px
```
mukunku, line 53: Looks like you removed the `px` variable here.  In the designs the x paddings are supposed to be 10, 11, and 13 respectively.

mukunku (changes requested):

PR #2074, opened 2025-11-25
chore(navigation): move transition to NavigationItem

dancormier:
This PR adds `transition:slide` to the `NavigationItem` component. This will allow for simpler markup for toggling Navigation sections and increases the likelihood that markup will be valid.

See this [Slack thread](https://stackexchange.slack.com/archives/C096063B1PH/p1764109633371739) for more details. cc @kylejrp

> [!NOTE]
> @giamir Should we consider adding a prop to add/remove the transition? I'm not sure what the potential unintended effects of including the transition on every `NavigationItem` might be.


giamir (approved):
I took the liberty to make the transition an opt in feature. That helps also with odd behavior I noticed in the trailing example where animations where triggered also when elements were selected.

Good to merge and release for me so that @kylejrp can use it. 

Thank you both.

PR #2075, opened 2025-11-26
chore(storybook) - Add labs label to skeleton and expandinginput

ttaylor-stack:
Adds labs label to the title of the Skeleton and ExpandingInput page, unfortunately it is not possible to add it to the sidebar too without creating a new Storybook addon, which seems like a lot for small documentation changes

story: https://stackoverflow.atlassian.net/browse/SPARK-119
Skeleton page: https://deploy-preview-2075--stacks-svelte.netlify.app/?path=/docs/components-skeleton--docs
ExpandingInput page: https://deploy-preview-2075--stacks-svelte.netlify.app/?path=/docs/components-expandinginput--docs


mukunku (approved):
LGTM!

packages/stacks-svelte/.storybook/manager-head.html
```diff
@@ -0,0 +1,2 @@
+<link rel="stylesheet" href="https://unpkg.com/@stackoverflow/stacks@latest/dist/css/stacks.min.css" />
```
giamir: We should not rely on unpkg if possible. 

The simplest approach would be to build stacks classic like we do for the test script:
https://github.com/StackExchange/Stacks/blob/beta/packages/stacks-svelte/package.json#L22

The alternative (probably slightly more complex) approach would be to understand if we can import the `stacks.less` directly and see if the less can be parsed on the fly to css.

giamir (approved):
Works well. Let's merge it. Thanks @ttaylor-stack 🎉

PR #2070, opened 2025-11-25
feat(input): update input styles for SHINE

mukunku:
# Summary

This PR makes some small tweaks to update input components to match the [new designs](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=713-24991&p=f&m=dev).

I also took a look at the [expanding input svelte component](https://deploy-preview-2070--stacks-svelte.netlify.app/?path=/docs/components-expandinginput--docs) and it didn't look like it needed any changes to me. I tried adding more size variations but ended up deciding it wasn't worth it.

## Webkit Overrides

I removed all the webkit specific css overrides as in my testing they don't seem to be needed anymore on modern IOS Safari. I tested it on my Iphone 14 using the PR environment; screenshots are below.

<details>
<summary>Iphone 14 Test Screenshots📱</summary>

<h2>Before Screenshots</h2>

<h2>After Screenshots</h2>

</details>

# Breaking changes
Following sizes have been removed:
* `s-input__md`
* `s-input__xl`
* [Nested inputs](https://deploy-preview-2070--stacks.netlify.app/product/components/inputs/#nested-inputs) html will require slight tweaking on consumers' side

## Out of Scope

[Text Area](https://deploy-preview-2070--stacks.netlify.app/product/components/textarea/) is out of scope for this PR. I have a separate PR for it here: https://github.com/StackExchange/Stacks/pull/2077 

# How to Test

Take a look the Input styles in Stacks Docs and Stacks Svelte:
* Stacks Docs: https://deploy-preview-2070--stacks.netlify.app/product/components/inputs/
* Stacks Svelte
  * TextInput: https://deploy-preview-2070--stacks-svelte.netlify.app/?path=/docs/components-textinput--docs
  * ExpandingInput: https://deploy-preview-2070--stacks-svelte.netlify.app/?path=/docs/components-expandinginput--docs

PR #2077, opened 2025-11-26
feat(textarea): new SHINE styles and Svelte component

mukunku:
# Summary

This PR makes some small tweaks to update the textarea component to match the [new designs](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=4537-49735&t=EjYIAcArFjb08HwE-1). It builds on the new input designs in #2070 so this should be merged after that one.

Additionally, this PR exposes a new `TextArea` Svelte component. This is a clone of the TextInput component minus the Input component specific bits.

# Breaking changes
Following sizes have been removed:
* `s-textarea__md`
* `s-textarea__xl`

# How to Test

* Take a look at the TextArea styles in Stacks Docs: https://deploy-preview-2077--stacks.netlify.app/product/components/textarea/
* Try out the new TextArea Svelte component in Stacks Svelte: https://deploy-preview-2077--stacks-svelte.netlify.app/?path=/docs/components-textarea--docs


packages/stacks-classic/lib/components/input_textarea/input_textarea.less
```diff
@@ -1,5 +1,7 @@
 .s-input,
-.s-textarea {
+.s-textarea,
+//The :has selector here allows us to use these vars for sibling icons in textarea components
+:has(> .s-textarea) {
```
dancormier, line 4: Really clever @mukunku!

packages/stacks-classic/lib/components/input_textarea/input_textarea.less
```diff
@@ -122,9 +128,19 @@
     });
 
     --_in-py: var(--su8);
+    --_in-min-h: calc(var(--su32) + var(--su4)); // 36px
```
dancormier: I don't believe the `--_in-min-h` custom property is being applied. You'll need to have `min-height: var(--_in-min-h)`.

### Nit notes

Since it's only applying to `.s-textarea`, my initial thought is to name it specific to this component (e.g. `--_te-hmn`\*) to telegraph that it's not being used on `.s-input`. This way, we remove the need for the unset value on both `.s-input` and `.s-textarea`.

\* I've tended to name the properties to mirror their atomic class counterparts for consistency, hence [`hmn` instead of `min-h`](https://beta.stackoverflow.design/product/base/width-height/#min-height-classes). That said, `min-h` might be more intuitive, so I'll leave this up to you.

packages/stacks-classic/lib/components/input_textarea/input_textarea.less
```diff
@@ -122,9 +128,19 @@
...
+
+    &&__sm {
+        --_in-min-h: calc(var(--su32) - var(--su4)); // 28px
+    }
+
+    &&__lg {
+        --_in-min-h: calc(var(--su48) - var(--su4)); // 44px
```
dancormier: ### Follow up from previous comment

These values will need to be updated to match the [Figma values](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=4537-49735&m=dev).

- `sm`: 55px
- default: 69px
- `lg`: 96px

<details>

#### `sm`

Image

#### Default

Image

#### `lg`

Image

</details>

```suggestion
    --_in-min-h: calc(var(--su64) - var(--su4) + var(--su1)); // 69px

    &&__sm {
        --_in-min-h: calc(var(--su48) + var(--su6) + var(--su1)); // 55px
    }

    &&__lg {
        --_in-min-h: var(--su96);
    }
```

---

@CGuindon @mukunku I think we should consider some different values here. It seems like the target is moreso getting three lines of text within the textarea without introducing a scrollbar. If that's correct, here are values that might achieve what we're looking for:

- `sm`: `58px`
- default: `72px`
- `lg`: `104px`

Image

> [!NOTE]
> I have not tested this outside of my Mac, so these values might need to be tweaked for the same scroll-less three-liner textareas on other OSes.

packages/stacks-docs/product/components/textarea.html
```diff
@@ -11,13 +12,17 @@
 {% highlight html %}
 <div class="d-flex fd-column gy4">
     <label class="flex--item s-label" for="example-item">Question body</label>
-    <textarea class="flex--item s-textarea" id="example-item" placeholder="…"></textarea>
+    <div class="d-flex ps-relative">
+        <textarea class="flex--item s-textarea" id="example-item" placeholder="…"></textarea>
+    </div>
```
dancormier, line 17: We should simplify this example by removing the wrapping `div` here.

```suggestion
    <textarea class="s-textarea" id="example-item" placeholder="…"></textarea>
```

---

**Sidenote**: `.flex--item` is largely vestigial at this stage. At some point, we can deprecate the class and removing all references of it from examples but for now I just suggest removing it when updating a given component's docs.

packages/stacks-docs/product/components/textarea.html
```diff
@@ -11,13 +12,17 @@
...
 </div>
 {% endhighlight %}
         <div class="stacks-preview--example">
             <div class="d-flex gy4 fd-column">
                 <label class="flex--item s-label" for="example-item7">Question body</label>
-                <textarea class="flex--item s-textarea" id="example-item7" placeholder="e.g. The Stack Overflow glyph used to have 6 bars, but now…"></textarea>
+                <div class="d-flex ps-relative">
+                    <textarea class="flex--item s-textarea ps-relative" id="example-item7" placeholder="e.g. The Stack Overflow glyph used to have 6 bars, but now…"></textarea>
+                </div>
```
dancormier, line 25: We should simplify this example by removing the wrapping `div` here.

```suggestion
                <textarea class="s-textarea" id="example-item7" placeholder="e.g. The Stack Overflow glyph used to have 6 bars, but now…"></textarea>
```

packages/stacks-svelte/src/components/TextArea/TextArea.svelte
```diff
@@ -0,0 +1,187 @@
...
+    </Label>
+
+    {#if $$slots.description}
+        <!-- Renders a description between the label and input. -->
+        <p class="s-description mb0 mtn2" id={`${id}-description`}>
+            <slot name="description" />
+        </p>
+    {/if}
+
+    <div class="d-flex">
```
dancormier: No need for this wrapper `div`, right?

packages/stacks-docs/product/components/textarea.html
dancormier, line 1: @mukunku Can you add an `Accessibility` section with a sentence or two on marking required textareas with the `.s-required-symbol`? I think we can get away with just linking to the ["Required input fields" section on the `Inputs` docs](https://beta.stackoverflow.design/product/components/inputs/#required-input-fields).

packages/stacks-svelte/src/components/TextArea/TextArea.svelte
dancormier, line 1: I think we can safely remove `optional`, `i18nOptionalText` props from Svelte component.

@giamir confirming: we don't need these props anymore since we have similar `required` props, right?

dancormier (changes requested):
I made a handful of suggestions but nothing too significant. Overall, we're looking good. Once my comments are addressed\* and `"new": true` is added for textarea in [navigation.json](https://github.com/StackExchange/Stacks/blob/f5007e50f1a37359c4b5d7de7e610483f085494f/packages/stacks-docs/_data/site-navigation.json#L195-L198), this should be good to go.

@mukunku Thanks for the nice work and for that `:has(> .s-textarea)` trick I look forward to stealing at a later date 🙂

---

\* one comment could use some design input. I'm fine with it being handled in a follow up if design needs a moment to address.

packages/stacks-docs/product/components/textarea.html
mukunku: Hey Dan, it doesn't work without some kind of wrapper element. Can you take a look maybe? All the other examples have a wrapper div so I thought making it the "default" might be okay.
{B9A41ADF-D580-4F95-8159-9F9EA2FC493A}

Seems the `d-flex` and `ps-relative` aren't needed, though.

packages/stacks-classic/lib/components/input_textarea/input_textarea.less
mukunku: My intention was to make the min-height as tall as a single line just so the user can't make the input very small and mess up the icon (since textarea is user resizable):
{0760059E-DB63-4CF3-9F37-BDD8D87067C0}

Do we really want to restrict developers/users from setting a smaller height than minimum 3 lines? Maybe they'd want the input to start small and expand as needed?

packages/stacks-docs/product/components/textarea.html
mukunku: Without the wrapper div the textarea doesn't respect the `width: 100%` css style. Which allows the user to resize the horizontal size of the input:
resize

I thought it would be okay to add this wrapper here since all the other examples use it anyway. Can you take a look and let me know what you think @dancormier ?

Additionally, I removed all usages of `flex--item` from the Input and Textarea docs: https://github.com/StackExchange/Stacks/pull/2077/commits/f08863c1e21894bdd814880b01bcfb9bbb0b1ba4

packages/stacks-classic/lib/components/input_textarea/input_textarea.less
mukunku: I think the min-height was removed during a merge by mistake. Added it back now 👍🏾 

I also renamed the var and made it textarea specific as you suggested: https://github.com/StackExchange/Stacks/pull/2077/commits/fd3dda7b636d6911dbf7555e456396e5a00f8d5e + https://github.com/StackExchange/Stacks/pull/2077/commits/f697e471e684f0d2568186004b35782e3ff2fe86

packages/stacks-svelte/src/components/TextArea/TextArea.svelte
mukunku: Seems not! Removed it now: https://github.com/StackExchange/Stacks/pull/2077/commits/27683cfddfaa6d8d169d6a19e31e16b351aaa5a7

packages/stacks-docs/product/components/textarea.html
mukunku: Added an accessibility section now. Let me know what you think: https://github.com/StackExchange/Stacks/pull/2077/commits/c00d05601390a30c55654c1b584bdf92f7ecb930

packages/stacks-svelte/src/components/TextArea/TextArea.svelte
mukunku: Removed now: https://github.com/StackExchange/Stacks/pull/2077/commits/80b5944a41ae33fed93d6a23d43e436cda065907

packages/stacks-classic/lib/components/input_textarea/input_textarea.less
dancormier: > Do we really want to restrict developers/users from setting a smaller height than minimum 3 lines? Maybe they'd want the input to start small and expand as needed?

I think we want to make the right/consistent thing the easy thing while providing escape hatches that create a little friction but still allow for flexibility. Since the design shows these sizes, I'd expect to make these sizes the easy thing. A developer still could set whatever height/max-height they'd need by overriding the max-height by adding [a `.hmx-*` class](https://beta.stackoverflow.design/product/base/width-height/#max-height-classes). They'd have to hop into the inspector to understand that there's a max-height set, but it's doable. Most of our atomic classes include `!important`, so they'll always take precedence over the component styles.

packages/stacks-docs/product/components/textarea.html
dancormier: Ah, I hadn't considered that! That makes sense to wrap it.

packages/stacks-classic/lib/components/input_textarea/input_textarea.less
dancormier: Considering this a bit more, I think it could make sense to set the min-height the way you did @mukunku and then setting a `height` for each size of textarea.

This way, a user could only resize it down to a single line and not beyond while having the initial size reflect the designs in the Figma. Everything could still be overridden by a dev but the defaults would be sensible and consistent. Does this seem reasonable @mukunku?

packages/stacks-docs/product/components/textarea.html
dancormier: 💯

dancormier (approved):
This PR is in a good place IMO. I've responded to your comments @mukunku and a couple things might need very minor tweaking but nothing that needs to hold up this PR. I'm happy to leave any further changes up to your discretion. 

Nice work here ❤️

packages/stacks-classic/lib/components/input_textarea/input_textarea.less
mukunku: That's a great compromise and I think would work great! Made the change now: https://github.com/StackExchange/Stacks/pull/2077/commits/f49e103800c4f0843be7a0e2535039d10e4136df

I went with the following sizes which are almost the same as the Figma's:
* sm: 54px
* default: 70px
* lg: 96px

Here's how it looks on my PC:
{E8DC742E-B64C-42FA-AB39-46FD2AB308E9}

I can adjust to your recommendation as well if we think that's better. Let me know what you think @dancormier 🙏🏾

mukunku: Going to merge as-is but happy to change as a follow-up 👍🏾

PR #2082, opened 2025-11-28
feat(badges): update to new SHINE visuals - part 2

giamir:
[SPARK-113](https://stackoverflow.atlassian.net/browse/SPARK-113)

This PR updates the remainder of the badges to the new SHINE visuals. It also clean up the related svelte component api.

- [Figma Specs](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=610-18799&p=f&m=dev)
- [Docs (Stacks Classic)](https://deploy-preview-2082--stacks.netlify.app/product/components/badges/)
- [Storybook (Stacks Svelte)](https://deploy-preview-2082--stacks-svelte.netlify.app/?path=/docs/components-badge--docs)

Tip use the filter feature of GH to get a better diff experience
Screenshot 2025-12-04 at 17 03 19


CGuindon:
Structure looks great — we would be deprecating Icon and Number count sections so you can get rid of those. I'll add notes to the Mapping section in Figma for that. @giamir

mukunku:
Code changes overall look good to me. Noticed a couple things though, let me know what you think:

* The Sizes example and code sample don't match (It's not just the colors, the sizes are all the same as well)
   
* Should we apply some negative margin (-1px, -2px) on icons to better align them with the text inside the badges?
   image
* Noticed a `state` option in the `type` dropdown in Svelte which I think shouldn't be there?

giamir:
> The Sizes example and code sample don't match (It's not just the colors, the sizes are all the same as well)

Good catch I have updated that now. 

> Should we apply some negative margin (-1px, -2px) on icons to better align them with the text inside the badges?

This is an OS specific issue similar to what we had for the activity indicator component. I don't have that in macos so adding a negative margins will make it look odd in mac. I am honestly unsure how to mitigate that. Maybe for now we can merge it as is and we can have a follow up ticket to investigate further. We could probably play around with the line-height. I don't think using margins will cut it. 

> Noticed a state option in the type dropdown in Svelte which I think shouldn't be there?

That's how the API currently works. 
You need to specify that badge is of `type="state"` otherwise the `state` prop is ignored.

mukunku (approved):
LGTM. 

Regarding the negative margin for icons: since we have an upcoming change to adjust line heights probably best to leave it for later as you said 👍🏾

dancormier (approved):

CGuindon (changes requested):
1. I think there are sizing tweaks missing for the `s-badge__squared` variants for `lg` and `sm`. Looks like the square and icon don't change so the square is either too small or too big when the badge changes size.
Screenshot 2025-12-05 at 4 29 30 PM

---

2. For the icon badges, we should have 4px padding between the icon and the text label and 6px on the right of the text (most icons don't go to the edge of the container so I reduced the space there to make it more visually balanced in most icon cases). Looks like it's 6px on both sides right now. (purple gutter and far right green padding side)
Screenshot 2025-12-05 at 4 32 06 PM

---

3. For the Squared variant (not for `important` modifyer, just regular), we should add an outline in HC mode around the square (the background colors because the same so we loose the box effect and they just look like non-squared badges).

Screenshot 2025-12-05 at 4 38 24 PM
Screenshot 2025-12-05 at 4 38 20 PM
Screenshot 2025-12-05 at 4 40 24 PM

---

4. Feels weird that there's an empty option in the sizes (but maybe that's something we need to live with with storybook for the "default" size?
Screenshot 2025-12-05 at 4 41 22 PM

CGuindon:
tagging @dancormier since Giarmir is out — found a few things above

mukunku:
This is ready for re-review now 🙏🏾

CGuindon (changes requested):
Changes look good, the only thing I'm finding struggling with is the order of the size examples for each type. I know our _default_ size makes sense to put first but it looks really weird to have default, small then large in order and all the heights looks different. Can we order them to have small, default then large? Makes it easier to see the difference. (in Stacks Classic docs)

CGuindon (approved):

CGuindon:
moot point about the sizes if they'll be removed from docs — but any areas where you keep all three, I would order them from smallest to largest size (not mixed)

PR #2087, opened 2025-12-02
Default width/height in Avatar.svelte

kylejrp:
## Summary

This PR adds a default height/width to the <Avatar> Stacks-Svelte component based on the `size` prop to prevent layout shifts as the image loads.

(I haven't done any testing, I just did this edit from the GitHub UI)


giamir:
Closing for now. Cannot reproduce the issue.

PR #2088, opened 2025-12-03
chore(input): fix non text input styles

mukunku:
# Summary

This PR fixes a bug I introduced in #2070 which breaks the non-text type inputs styles by pushing the date picker etc to the left: https://beta.svelte.stackoverflow.design/?path=/docs/components-textinput--docs

# How to Test

* Confirm all the inputs looks okay for the TextInput Svelte Component. Most importantly the various **Types** we have: https://deploy-preview-2088--stacks-svelte.netlify.app/?path=/story/components-textinput--types
* The icons should be properly right aligned now
* Confirm Nested inputs still look okay and aren't broken visually after this change: https://deploy-preview-2088--stacks.netlify.app/product/components/inputs/#nested-inputs

### DIFF
--- a/.changeset/sour-jobs-collect.md
+++ b/.changeset/sour-jobs-collect.md
@@ -1,0 +1,5 @@
+---
+"@stackoverflow/stacks": patch
+---
+
+chore(input): fix non text input styles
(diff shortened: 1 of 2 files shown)


giamir (approved):
This fixes the issue. I am not familiar enough with input stylesheet to comment on the exact rule. I find the nested input a bit odd honestly. It feels like it should be a separate component but it was already there so I guess it is fair to leave it for now. 

Other design system don't call that an input and we should probably follow their lead
https://open-ui.org/components/combobox.research/

Anyway that's a topic for a separate ticket potentially. :) 

Thanks @mukunku

PR #2092, opened 2025-12-04
feat(user card) - Update User card to new SHINE design (part 1)

ttaylor-stack:
This ticket cover part 1 of the user card updates.

The following things will be updated in Part 2:
- States
- Additional Bling

[Ticket](https://stackoverflow.atlassian.net/browse/SPARK-85)
[Figma](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=5235-41606&m=dev)

notes:

- I decided to keep the Usercard as one Svelte component instead of dividing it up even though the large and sm/default sizes have different markup. I thought it would be more user friendly to have 1 components users interact with.

- I decided split the different parts of the Usercard into different Svelte subcomponents


mukunku:
Hey @ttaylor-stack ,

I did a quick glance through the code and it overall looked good to me. Looking at the UI elements I noticed an issue with gaps, though:
image

Here's what I see in the designs:
image

CGuindon (changes requested):
Base should be using the 24x24 avatar size.

Screenshot 2025-12-09 at 8 16 22 AM

---

The padding in between the bling and the badge/rep number is created with a space character before the number. Should this be done with padding/gutter spacing instead? Would be 2px.

Screenshot 2025-12-09 at 8 19 22 AM

---

Is there any way to have 4px spacing in between badges when the count is showing vs 2px spacing when the count number isn't?

Screenshot 2025-12-09 at 8 21 43 AM

---

For responsive, when we need to wrap, I think we should keep certain blocks of info together (so the whole block wraps when the screen size is too short)


Screenshot 2025-12-09 at 8 24 15 AM

---

Got confused by my own section title — can we change this to "with user badges"?

OR — I'm wondering if we should also have a "with Bling" section OR should both of those just be in the Base section?

Screenshot 2025-12-09 at 11 05 54 AM

---

Should / could the Sizes be in the same section? Maybe in a table to get both descriptions or just the descriptions above separate by a paragraph and both sizes in the same code box?

CGuindon (changes requested):
- The gutter space between the avatar image and the username should be 6px — looks like it's using the same value as the gutter spacing in between the bling/rep — that one can remain 4px (currently 4px everywhere).
Screenshot 2026-01-07 at 1 34 52 PM


- Similarly, the gutter spacing before the _asked {timestamp}_ should be 8px but the spacing in between username and rep/bling is 6px (currently 6px everywhere).
Screenshot 2026-01-07 at 1 40 09 PM

- rep # should be `black-600` text color (using 500)
Screenshot 2026-01-07 at 1 47 40 PM

- In the example, we show that the username is a link — the mouse changes the cursor  to the pointer when I hover and it's a dead link when I click. The timestamp "asked 2h ago" would also be a link in production. Could we/should we show this in the code example as well? Have it be a similar dead link but the cursor changes on hover.


- Now that I see it, I think we can remove the example with the mod badge from the Base section. We have the "With Badge" right below where we show those variants. Similarly, we'll have "With bling" in user card part two that will show the trophy variation too down below.
Screenshot 2026-01-07 at 2 00 01 PM


- Nice work on the responsive breakdown of all the pieces. There is the odd user that likes to have a sentence in their username. This is more rare but wondering if you have any suggestions on if/how we could wrap the username if we need to (instead of it cutting off). This is lower priority
Screenshot 2026-01-07 at 2 02 50 PM

- For the large size, there's different sizes for all the padding in there (currently all 6px). Some should be 12px or 4px.
Screenshot 2026-01-07 at 2 08 33 PM

dancormier:
@CGuindon The spacing is on me! I tweaked it in [3b915e5](https://github.com/StackExchange/Stacks/pull/2092/commits/3b915e5596b012cbab6323a43bf6f1116a1e8b9c). Can you check it out again and let me know if it seems correct?

CGuindon (changes requested):
Missed one small one, these gaps are both 6px. The left gap should be 6 but the right gap (before "asked 2h ago" should be 8px.
Screenshot 2026-01-08 at 7 33 13 AM

I had this note from before the holidays — it's lower priority since we won't have just the badges showing alone without the numbers for Beta but curious if this is an easy fix. If it takes too long, I can descope for now.
Screenshot 2026-01-08 at 7 38 18 AM

packages/stacks-classic/lib/components/user-card/user-card.less
```diff
@@ -1,129 +1,118 @@
...
+        font-weight: 600;
+        color: var(--black-600);
     }
 
-    & &--link {
-        font-size: var(--_uc-link-fs);
-        white-space: var(--_uc-link-ws);
+    & &--time {
+        color: var(--black-400);
+        margin-left: var(--su2);
```
dancormier: @ttaylor-stack we can't apply a left margin here because when it wraps, we'll wind up with an extra 2px gap to the left of the timestamp

dancormier:
> I had this note from before the holidays — it's lower priority since we won't have just the badges showing alone without the numbers for Beta but curious if this is an easy fix. If it takes too long, I can descope for now.

@CGuindon I vote for descope since it would require some tricky CSS

dancormier:
> Missed one small one, these gaps are both 6px. The left gap should be 6 but the right gap (before "asked 2h ago" should be 8px.
Screenshot 2026-01-08 at 7 33 13 AM

@CGuindon I think the gaps should be good now. See the screenshots in the toggle section below (my tool shows an extra 1px for each value. No clue why and I'm choosing to ignore that and just assume that the fact that the timestamp is +2px from the previous item is good enough 😂)

<details>

image
image
image
image

</details>

CGuindon:
[Happy to descope the tricky CSS](https://github.com/StackExchange/Stacks/pull/2092#issuecomment-3726011317)

CGuindon (approved):

giamir (changes requested):
@ttaylor-stack and myself will go through few things on Monday regarding the Svelte API. Let's hold off merging for now. 

Here are my review notes for our call on Monday

- It looks like the markup example between `default` and `with badges` don't differ in the docs. I think it is important that we try to map what we show in the example to the markup so that engineer can easily copy/paste what they need.
- I am a bit confused about the `Note on timestamps` guidance we provide. At the moment all the examples show `asked ...`. I think it would be appropriate to show concrete example of what we meant with that guidance (e.g. `modified ...` or even the actual precise date in the tooltip). If we don't do this we risk to confuse engineers and then it would be better to remove that guidance all together.
- Updating the migration guide
- Updating markup structure of visual and accessibility tests to reflect the new classes, etc...  
- It looks like all the subcomponents created UserCardAwards, UserCardBadges, etc… are used only internally and we don't expose them, therefore their API should not be published in the story for consumers to see

Then for the `UserCard.svelte` api here are some preliminary thoughts that we will discuss together:
```
<UserCard>

Props

- name (string - required)
- avatar? (string - src to the user avatar)
- profileUrl? (string - url to the user profile page - previously called href)
- size? (sm/default/lg - default: default)
- reputation? (string | number - default: undefined)
- role? string // can be confusing | this is not an html attribute | find a different name
- location? string

// Badges (these are all badges, can we just create a badges snippet and give up control to the consumer? - create more flexibility)
admin?
moderator?
staff?

// Awards (Those are all blings, should we just have an awards snippet instead and let the consumer slot them instead of passing single props)
- goldAwards?
- silverAwards?
- bronzeAwards?

// Timestamp (extra child component??? - bad naming - maybe we should just call it UserCardTime like we do in the CSS class —time - would be placed in a time snippet)
- timestamp?
- i18nTimestampTooltip?
- timestampHref?

// Recognition (can it be a snippet? can we give up control here and and let the consumer take care of it)
text
href
linkText
Icon

// Can this just be a bio snippet
bio

### Coming up in the follow up tickets

original poster (probably a modifier class in CSS and could be a boolean prop for the svelte component - op and opTooltipText)

new (just another badge in the badges snippet - consumers take care of it)

deleted? (boolean - do we leave to the consumer to make sure the avatar prop is changed to deleted avatar or do we hardcode that specific avatar in the component?) 

additional blings - could they just be a snippet for the consumer to fill? what level of control do we need? 
```

dancormier:
Thank you very much for this review @giamir!

> * It looks like the markup example between `default` and `with badges` don't differ in the docs. I think it is important that we try to map what we show in the example to the markup so that engineer can easily copy/paste what they need.

I think this is my mistake. I updated the example but not the markup example. I've removed the extra markup in the example in [36a9d83](https://github.com/StackExchange/Stacks/pull/2092/commits/36a9d8332a50a172da1e1eaf721f6dba96e168ff)

The examples are still very similar, but I included both `Default` and `With badges` as distinct usage guidelines as opposed to examples of significantly different markup following the distinction in the [Figma](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=5235-45601&m=dev)

> * I am a bit confused about the `Note on timestamps` guidance we provide. At the moment all the examples show `asked ...`. I think it would be appropriate to show concrete example of what we meant with that guidance (e.g. `modified ...` or even the actual precise date in the tooltip). If we don't do this we risk to confuse engineers and then it would be better to remove that guidance all together.

This is also on me. I added that tip based on the callout in the [Figma](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=5375-13938&m=dev) without scrutinizing it. Now that I _really_ read it, I'm a bit confused by the intention, both in the docs and with the UI. I've asked a few questions directly on the Figma.

> 
> * Updating the migration guide

I will take care of adding the markup changes to the migration guide but I'll hold on anything beyond that.


> * Updating markup structure of visual and accessibility tests to reflect the new classes, etc...
> * It looks like all the subcomponents created UserCardAwards, UserCardBadges, etc… are used only internally and we don't expose them, therefore their API should not be published in the story for consumers to see

➕ 

> Then for the `UserCard.svelte` api here are some preliminary thoughts that we will discuss together:
> 
> ```
> <UserCard>
> 
> Props
> 
> - name (string - required)
> - avatar? (string - src to the user avatar)
> - profileUrl? (string - url to the user profile page - previously called href)
> - size? (sm/default/lg - default: default)
> - reputation? (string | number - default: undefined)
> - role? string // can be confusing | this is not an html attribute | find a different name
> - location? string
> ```

Seems reasonable. For `role`, maybe `designation` or `jobTitle`?

> ```
> // Badges (these are all badges, can we just create a badges snippet and give up control to the consumer? - create more flexibility)
> admin?
> moderator?
> staff?
> 
> // Awards (Those are all blings, should we just have an awards snippet instead and let the consumer slot them instead of passing single props)
> - goldAwards?
> - silverAwards?
> - bronzeAwards?
> 
> // Timestamp (extra child component??? - bad naming - maybe we should just call it UserCardTime like we do in the CSS class —time - would be placed in a time snippet)
> - timestamp?
> - i18nTimestampTooltip?
> - timestampHref?
> 
> // Recognition (can it be a snippet? can we give up control here and and let the consumer take care of it)
> text
> href
> linkText
> Icon
> 
> // Can this just be a bio snippet
> bio
> ```

➕ to all of this 🙂

> ```
> ### Coming up in the follow up tickets
> 
> original poster (probably a modifier class in CSS and could be a boolean prop for the svelte component - op and opTooltipText)
> 
> new (just another badge in the badges snippet - consumers take care of it)
> 
> deleted? (boolean - do we leave to the consumer to make sure the avatar prop is changed to deleted avatar or do we hardcode that specific avatar in the component?) 
> 
> additional blings - could they just be a snippet for the consumer to fill? what level of control do we need? 
> ```

Agreed. For deleted, I vote we hardcode the deleted avatar and render it without the need for consumers to pass it.

---

I broadly agree with these notes on the Svelte component. I think we should approach architecting these components with a heuristic that's something like:
- Removing the simpler subcomponents (like `bio`) in favor of including their markup directly in the main `UserCard` component
- Including props with a type of `Snippet` for components that only contain other components
- Providing subcomponents for elements that introduce more than a couple of props


I'd expect something like:
```js
// UserCard props
    interface Props {
        name: string;
        avatar?: string;
        bio?: Snippet;
        designation?: string;
        location?: string;
        profileUrl?: string;
        reputation?: string | number;
        size?: Size;

        // Expose subcomponents for these snippets
        awards?: Snippet;
        badges?: Snippet;
        recognition?: Snippet;
        time?: Snippet;

        i18nReputationLabel?: string;
        class?: string;
    }
```

I'm happy to help brainstorm on this more. Let me know what you think and feel free to schedule a call (or @ttaylor-stack would you like to add @giamir to our 1:1 on Monday?)

CGuindon (changes requested):
NEVER MIND THIS (wrong PR)

packages/stacks-svelte/src/components/UserCard/UserCardTime.svelte
```diff
@@ -0,0 +1,39 @@
+<script lang="ts">
+    import Popover from "../Popover/Popover.svelte";
+    import PopoverContent from "../Popover/PopoverContent.svelte";
+    import PopoverReference from "../Popover/PopoverReference.svelte";
+
+    interface Props {
+        /**
+         * Timestamp displayed in the user card generally used to indicate when a comment was posted UPDATE THIS
```
giamir: this description needs to be updated. 
Maybe we could even provide an e.g. in the description (e.g. asked 2h ago) to make it clearer.

packages/stacks-svelte/src/components/UserCard/UserCardTime.svelte
```diff
@@ -0,0 +1,39 @@
...
+         */
+        text?: string;
+
+        /**
+         * Tooltip text for the timestamp
+         */
+        preciseTimestamp?: string;
```
giamir: maybe we could just call the prop `timestamp` here and update the description to something like `Precise date and time of the post in ISO format`

packages/stacks-svelte/src/components/UserCard/UserCard.test.ts
```diff
@@ -27,38 +19,33 @@ describe("UserCard", () => {
...
-
-    it("should render the user location", () => {
-        render(UserCard, {
-            name: "John Doe",
-            avatar: "https://picsum.photos/128",
-            location: "New York",
-        });
-        expect(screen.getByText("New York")).to.exist;
-    });
+    it("should render the user reputation in blings", () => {
```
giamir: Are these tests updated?
I think you could take inspiration from what we do in Popover to test its subcomponets
https://github.com/StackExchange/Stacks/blob/beta/packages/stacks-svelte/src/components/Popover/Popover.test.ts#L8

packages/stacks-svelte/src/components/UserCard/UserCardBadge.svelte
```diff
@@ -0,0 +1,14 @@
...
+        /**
+         * The type of user badge
+         */
+        type: UserType;
+    }
+    const { type }: Props = $props();
+</script>
+
+<li>
+    <Badge text={type} type="user" userType={type} size="sm" />
```
giamir, line 13: Maybe I am getting ahead of myself but if we are hardcode the type to "user" would not we have potentially issue with the `new` badge coming up in the next story? Happy to postpone this conversation to a follow up PR though.

packages/stacks-svelte/src/components/UserCard/UserCardTime.svelte
```diff
@@ -0,0 +1,39 @@
...
+        /**
+         * Link to post activity
+         */
+        href?: string;
+    }
+
+    const { text, preciseTimestamp, href }: Props = $props();
+</script>
+
+{#if text}
```
giamir: I think text should be a required prop and we should drop the if here

packages/stacks-svelte/src/components/UserCard/UserCard.svelte
```diff
@@ -162,107 +102,113 @@
...
+                    <ul class="s-user-card--group">
+                        {@render blings()}
+                    </ul>
                 {/if}
-                {#if gold}
-                    <Bling type="gold" name={`${gold} gold awards`} />
-                    <span class="v-hidden">{gold}</span>
+            </div>
+        </div>
+        {#if recognition || designation || location || bio}
```
giamir, line 190: all these nested conditional statements in the component are something we could test for in unit tests. Maybe you can have the AI help you out to create a set of cases and review them. They are a lot and I think they could help to make changes with the confidence of not breaking things in the future

giamir (commented):
Thanks @ttaylor-stack for making a lot of progress on the review comments. Personally I think we are almost ready to merge. I left few more comments but I think the largest point is about getting the unit tests in a good enough state. I left a suggestion to look at how we do things in the Popover that has similar subcomponents. 

Thanks and let me know if you would like to sync. 🙂

packages/stacks-svelte/src/components/UserCard/UserCardBadge.svelte
ttaylor-stack: In uc part 2 I have made the `new` badge a user badge, can have a more thorough conversation during part2

giamir (approved):
I think we are good to go. Thanks @ttaylor-stack for adding few more unit tests. 
We can merge I believe. 🚀

PR #2093, opened 2025-12-05
chore(colors): add initial brand color set

dancormier:
[SPARK-134](https://stackoverflow.atlassian.net/browse/SPARK-134)

---

This PR adds a set of brand colors purely as CSS custom properties. See https://deploy-preview-2093--stacks.netlify.app/product/foundation/colors/#brand

image

### TODO

- [ ] Cleanup docs (consider splitting brand colors into color groups like `blue` `green`, etc)

#### Verify

- [x] Do we just need custom properties or do we also need to expose them as utility classes (.bg-brand-green, .fc-brand-off-black, etc)?
    -  @abovedave yes, added
- [x] Do we have values for the dark mode versions (and do we need HC versions as well)
    -  @abovedave no, not needed yet.
- [x] Will we need to consider color contrast for these colors?
  - "These colors would be used by Brand only — ideally they are checking color contrast for their site but I think in most cases it's for illustrations and logos so that's fine." - @CGuindon ([slack](https://stackexchange.slack.com/archives/C096063B1PH/p1764958543798649?thread_ts=1764947598.360819&cid=C096063B1PH))
- [x] How solidified is this list of colors? Should we anticipate more colors added?
   -  @abovedave this is it for the time being
- [x] Should we reconsider the naming structure?
    - @abovedave this follows the same conventions as our [brand guidelines ](https://alpha.stackoverflow.design/brand/color)
- [x] YAGNI?
  - [x] Do we need `--brand-black`? (It's almost exactly `--black-600`)
      - @abovedave I think for completeness yes
  - [x] Do we need `--brand-white`? (It's exactly `--white`)
      - @abovedave removed
- [x] Reconsider naming accuracy
  - [x] Can we name `--stack-orange` as just `--brand` (it would be the default brand color)?
    - @abovedave yes I like this idea
  - [x] Should `orange-medium` be `red` instead?
  - [x] Should `orange-dark` be `red-dark` instead?
  - [x] Should `pink` instead be `light-pink` instead?
  - [x] Should `purple` be `purple-light` instead?
  - [x] Should `purple-dark` be `purple` instead?


abovedave:
@dancormier modified this pr as discussed

abovedave (approved):

PR #2107, opened 2025-12-18
Bindable TextInput value

chris-doucette-stack:
## Summary
This PR explicitly declares a value prop on TextInput and makes it bindable, I also added a Storybook story, mainly for testing purposes

## How to test
1. Open TextInput's story book
2. Go down to the "Bindable Value" story and update the inputs value

### DIFF
--- a/.changeset/heavy-years-attack.md
+++ b/.changeset/heavy-years-attack.md
@@ -1,0 +1,5 @@
+---
+"@stackoverflow/stacks-svelte": minor
+---
+
+Made TextInputs value prop bindable
--- a/packages/stacks-svelte/src/components/TextInput/TextInput.stories.svelte
+++ b/packages/stacks-svelte/src/components/TextInput/TextInput.stories.svelte
@@ -53,6 +53,8 @@
             },
         },
     });
+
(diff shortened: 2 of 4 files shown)


giamir (approved):
Thanks @chris-doucette-stack. I am ok to see this merged. The only thing that would be nice to have is to add a unit test case if it is not too complicated.

@dancormier and I recently discussed supporting both two-way and one-way binding for the RadioGroup and Checkbox components, so this PR comes at the right time. It brings TextInput in line with what we’re aiming to do across the rest of the form components.

I will let Dan chime in as well but apart from that small test case I don't see why we could not merge this in an express way. 

Don't forget to add a changeset (run `npx changeset` in your terminal and follow the instructions) for the PR @chris-doucette-stack so that our pipelines will automatically create a new release for you once merged. Thank you

dancormier (approved):
@giamir's comment pretty much matches my thoughts. Seems like a nice enhancement (and one that we're working on getting in for other inputs). Besides the changeset, I think this is good to get merged.

Thanks for the contribution @chris-doucette-stack ❤️

PR #2108, opened 2025-12-19
chore(navigation): make avatar background transparent by default

giamir:
For navigation items there is no need to enforce a white static background for avatars. In fact we have a need for the transparency of the images to be respected. This PR simply specify that for avatar included in navigation items background should remain transparent.

See [this slack thread](https://stackexchange.slack.com/archives/C096063B1PH/p1766136670129519?thread_ts=1766101431.015539&cid=C096063B1PH) for full context.


dancormier (approved):
👍

PR #2109, opened 2025-12-19
feat(notice/toast): update styles for SHINE

mukunku:
# Summary

This PR updates the Notice and Toast components according to the latest [SHINE designs](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=612-18813&p=f&m=dev).

# How to Test

To test the Notice component:
* Stacks Docs: https://deploy-preview-2109--stacks.netlify.app/product/components/notices/
* Stacks Svelte: https://deploy-preview-2109--stacks-svelte.netlify.app/?path=/docs/components-notice--docs

To test the Toast component:
* Stacks Docs: https://deploy-preview-2109--stacks.netlify.app/product/components/notices/#toast
* Stacks Svelte: https://deploy-preview-2109--stacks-svelte.netlify.app/?path=/docs/components-toast--docs


packages/stacks-classic/lib/components/notice/notice.less
```diff
@@ -157,41 +147,54 @@
...
         outline: var(--su-static1) solid var(--_no-code-bc);
-
-        border-radius: var(--br-md);
         padding-left: var(--su2);
         padding-right: var(--su2);
     }
 
-    & &--btn {
-        // TODO: decouple .s-notice--btn from .s-btn
+    button&--dismiss {
```
mukunku, line 155: Introduced `s-notice--dismiss` class here to make the dismiss button more explicit and style it accordingly.

packages/stacks-svelte/src/components/Notice/Notice.svelte
```diff
@@ -44,6 +53,21 @@
...
+        /**
+         * Optional dismiss event handler
+         */
+        onDismiss?: MouseEventHandler<HTMLButtonElement>;
+
+        /**
+         * Dismiss button label override
+         */
+        i18nDismissButtonLabel?: string;
+
```
mukunku, line 70: I made the dismiss button a boolean flag here with additional props instead of a special NoticeAction case like it was before. I think this make more sense and improves usability. Hopefully other folks agree 🙏🏾

packages/stacks-classic/lib/components/notice/notice.less
```diff
@@ -157,41 +147,54 @@
...
     .construct-notice-component(s-notice);
 
-    border-radius: var(--br-md);
-    border-width: var(--su-static1);
-    padding: var(--su16);
+    display: flex;
+    align-items: center;
+    gap: var(--su4);
+
+    &--icon {
```
mukunku, line 191: Introduced the `s-notice--icon` class for the icons

packages/stacks-docs/product/components/notices.html
```diff
@@ -21,19 +19,12 @@
...
-                aria-label="Dismiss"
-                data-action="s-toast#hide">
-                {% icon "ClearSm", "m0" %}
-            </button>
+    <div class="s-notice s-notice__info d-flex ai-center wmn4">
+        <span class="s-notice--icon">{% icon "Info" %}</span>
+        Toast notice message with an undo button
+        <div class="d-flex ml-auto">
+            <button type="button" class="s-link s-link__underlined">Undo</button>
+            <button type="button" class="s-link s-notice--dismiss js-toast-close" aria-label="Dismiss" data-action="s-toast#hide">{% icon "ClearSm" %}</button>
```
mukunku: I opted to style the buttons with `s-link` but not sure if I should've used `s-btn s-btn__link` instead like we do in the Svelte component for Notices.

CGuindon (changes requested):
Looking great, just noticing some weirdness when I resize the screen smaller:
- the whole browser starts to shrink instead of resize when I go below 600px width — feels like a docs issue but not sure if the toast is affecting this somehow
- Notices can wrap to multiple lines when needed:
   - Padding right should maintain a minimum of 12px (even when there's no X or under button, the description text shouldn't touch the right edge of the notice)
   - Links on the end shouldn't wrap within itself (keep as single line unless the whole description is wrapping)
   - Code snipets shouldn't wrap within itself either

Screenshot 2025-12-19 at 9 59 44 AM


_Added an extra example in the variations section to make that a bit clearer_

Screenshot 2025-12-19 at 10 02 32 AM


---

For the toast not banner (possibly for the other ticket):
- I noticed when I add the `s-notice__important` to the toast example, the undo link and close icon stay black instead of turning white as well.
Screenshot 2025-12-19 at 9 52 18 AM

CGuindon:
I also had a note to myself to write a warning _note_ to add to the docs right below the Toast description text. Can you use a warning in-page notice there that says:

> **Usage Warning**: We are phasing out Toasts due to significant accessibility barriers. Avoid this component for new features. Instead, prioritize integrated alternatives—such as component state changes or inline messages—to provide accessible feedback directly where the user is focused.

Screenshot 2025-12-19 at 10 13 57 AM

mukunku:
Thanks for the review @CGuindon ! I believe I addressed all your findings around responsiveness and multi-line content. Please take a look and let me know how it looks now.

I also decided to do Toasts as part of this PR because the changes left were so small. I additionally added the warning you mentioned:
{EF8D2E39-B3F5-4EA2-BD4B-A62E0492A649}

CGuindon (changes requested):
I think you're using the old X icon, you can use the new `cross` for the dismiss icon button. Otherwise LGTM

Screenshot 2026-01-07 at 2 19 12 PM

CGuindon (approved):
OOOooooooooh even some fancy stacking and animations (not that we see this often) — looking good!

Screenshot 2026-01-09 at 10 43 43 AM

mukunku:
> OOOooooooooh even some fancy stacking and animations (not that we see this often) — looking good!
> 
> Screenshot 2026-01-09 at 10 43 43 AM

FWIW That was already there; I didn't add that feature 😅

dancormier (approved):
Looking sharp 😎 I'd be happy to see this merged. 

The only issue worth mentioning that I can see is with the margin-left on the button. In the [Svelte Toast example for "With actions"](https://deploy-preview-2109--stacks-svelte.netlify.app/?path=/story/components-toast--with-actions), I noticed the actions butting up against one another.
Image

This is the sort of thing I'd expect to be resolved with a `gap` applied. The issue with this is it will add _another_ `x`px to the `22px` margin-left on the notice button. I'd make a suggestion but I didn't have time to dig into it and I don't think this needs to hold up merging this PR.

Thanks for the nice work here @mukunku

mukunku:
> Looking sharp 😎 I'd be happy to see this merged.
> 
> The only issue worth mentioning that I can see is with the margin-left on the button. In the [Svelte Toast example for "With actions"](https://deploy-preview-2109--stacks-svelte.netlify.app/?path=/story/components-toast--with-actions), I noticed the actions butting up against one another. Image
> 
> This is the sort of thing I'd expect to be resolved with a `gap` applied. The issue with this is it will add _another_ `x`px to the `22px` margin-left on the notice button. I'd make a suggestion but I didn't have time to dig into it and I don't think this needs to hold up merging this PR.
> 
> Thanks for the nice work here @mukunku

@dancormier I actually thought that was a single button with the text "YesNo" 🤦🏾 Let me know what you think of this solution I came up with: https://github.com/StackExchange/Stacks/pull/2109/commits/9e855e3db8a2fe4e5375284623e9d6aaa26aae22 . As mentioned in today's standup, this is probably another example of where in V3 we're potentially being more disruptive with our selectors. 

Additionally, I noticed NoticeActions didn't look right in Svelte due to a css issue. Which is why I also added a patch for that in the same commit:
image
Adding the `ow-initial` class directly on the button didn't work 🙃

CGuindon (changes requested):
Looks like there's some extra padding on the Toast only that snuck in (I didn't notice this before if it was there). 
Screenshot 2026-01-12 at 4 00 42 PM


Should be 24px padding on the right side (between text and button) and 12px padding on the left side (between icon box and text).
Screenshot 2026-01-12 at 4 03 27 PM
Screenshot 2026-01-12 at 4 03 23 PM

mukunku:
Thanks for the catch @CGuindon , It seems I broke it with my latest changes. I fixed the gaps again now, please let me know how it looks 🙏🏾

CGuindon (approved):

PR #2118, opened 2026-01-09
feat(banner): update styles for SHINE

mukunku:
# Summary

This PR Updates the Banner component styles to match the latest SHINE designs: https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=5982-7648&m=dev

This PR should be merged after #2109

# How to Test

Check out the Banner component: https://deploy-preview-2118--stacks.netlify.app/product/components/banners/

There is no Svelte component for banners.


mukunku:
This should be ready for review @CGuindon ! Let me know what you think 🙏🏾 

I also saw your comment about maybe removing the pinned option. Didn't know if we should do that so I kept it for now.

CGuindon (changes requested):
Screenshot 2026-01-09 at 10 48 22 AM

When I fixed the ellipsis icon to be the Cross, I noticed it had also reverted to it's default #000 everywhere. I fixed this morning as well in the Figma, for the **Important** variants, the Cross/X icon should be white like the text (to pass a11y) — except for yellow/warning.

mukunku:
> When I fixed the ellipsis icon to be the Cross, I noticed it had also reverted to it's default #000 everywhere. I fixed this morning as well in the Figma, for the **Important** variants, the Cross/X icon should be white like the text (to pass a11y) — except for yellow/warning.

I think that was an oversight by me. It should be fixed now @CGuindon  👍🏾

CGuindon (approved):

packages/stacks-classic/lib/components/notice/notice.less
```diff
@@ -164,16 +164,26 @@
             background-color: var(--_no-fc);
             color: var(--_no-bg);
         }
+        color: var(--_no-fc);
     }
 
-    :has(>button&--dismiss) {
+    &--actions {
```
mukunku, line 170: I created `s-notice--actions` and `s-banner--actions` to define this container. It was getting a bit unruly with the selector we had before.

CGuindon (approved):

packages/stacks-docs/product/components/banners.html
```diff
@@ -265,6 +259,30 @@
                     {% bannerExample "danger", "true", "mt8" %}
                 </div>
             </div>
+        
+        {% header "h3", "Featured" %}
+        <div class="stacks-preview">
+    {% highlight html %}
+    <div class="s-banner s-banner__featured" role="alert" aria-hidden="false">…</div>
+    <div class="s-banner s-banner__featured s-banner__important" role="alert" aria-hidden="false">…</div>
+    {% endhighlight %}
```
dancormier: **Very minor issue**: There's a few of these highlight sections that are indented but shouldn't which results in weirdly offset code.

<details>
<summary>Screenshot</summary>

Image

</details>

dancormier (changes requested):
Thanks for these changes @mukunku. The styling seems good and I only found one very minor issue with the docs. I have a couple of other notes.

## A few suggested changes

- I'd recommend adding the new variants to the visual/a11y test files
  - I think it mainly be updating the array of `variants` in those test files
  - We should probably add them for Notice/Toast as well
- I think this would be a good time to add a `Banner` Svelte component
  - I would expect this to be pretty trivial since it's really similar to `Notice`
  - I wonder if it would be easiest to make a private component that accepts a class name that we could just import into `Banner.svelte`, `Notice.svelte`, and maybe `Toast.svelte`

Let me know if you'd like any help or clarification. Thanks again 🫶

mukunku:
Hey @dancormier , thanks for the review! I fixed the banners page spacing and updated the visual/a11y tests for banners/notices/toasts. I had forgotten about those 🤦🏾 

It was tricky updating the icons dynamically based on the test variant so it took me a bit longer than anticipated.

Additionally, as discussed during Monday's standup, I created a follow-up ticket to create the banner svelte component: https://stackoverflow.atlassian.net/browse/SPARK-168

giamir (approved):

PR #2119, opened 2026-01-09
fix(font-size): properly set rem values

dancormier:
This PR fixes a bug with our font sizing that had downstream effects on styles that leveraged our `--s-full` custom property. Previously, we were setting our base font size to a static `14px` value at the `html` level. This resulted in values uses in classes such as our atomic static width/height classes to be inflated.

For example, check out the width of the `.s-topbar--container` in our [docs](https://beta.stackoverflow.design/product/develop/using-stacks/):

image

Notice that it is `1361.23px` wide. It has `width: var(--s-full)`, which should result in a width of `1264px`. With the changes in this PR, we get the intended width:

image

As a bonus, this simplifies our usage of rem values dramatically since sizing more often tends to be divisible by `16px` (the default base font size set on the `html` tag - keep reading for more details on this). This should also make the upcoming atomic size class updates simpler 🤞 

## How this works

Standard browser default font size is `16px`. By setting `font-size: 100%` on `html`, you're saying _use 100% of the size of the browser's set font size_. If the user overrides that font size, the user's preference will instead be applied.

On the `body` tag, a `rem`-based font size can be set. This will be relative to the font size set on `html`. Since the default is `16px`, we can achieve our desired base font-size by setting the `body` font size to `0.825rem` (which computes to `14px`).

From [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-size#pixels):
> If a font-size has not been set on any of the `<p>`'s ancestors, then `1em` will equal the default browser `font-size`, which is usually `16px`. So, by default `1em` is equivalent to `16px`, and `2em` is equivalent to `32px`.[…]

> [!NOTE]
> The above quote mentions `em` but the same logic applies to `rem` when operating at the `body` tag level.

## Notable changes

- As mentioned, `html` now has a `font-size` or `100%`
- I've replace the single usage of `--fs-base` with `--fs-body1` and removed `--fs-base` altogether
- I've fixed a bug where `.s-prose` of different sizes should have had font sizes that scaled relative to the parent's font size. We were using `rem` values when we should use `em`.
  - This also allowed me to remove all `--fs-*-relative` custom properties. They we're working as intended and were all either referenced in `.s-prose` or not at all
- I've updated all `rem` values through our codebase to their base-16 equivelent values (this [px to rem converter](https://nekocalc.com/px-to-rem-converter) is handy if you'd like to check my math)

## Other details

This should result nearly zero changes to the look of components (I think the `.s-prose` sizes are the most notable change). I'm relieved to see that the visual regression tests have passed 😅 

## How to test

I suggest hitting the local instance (or deploy preview) and just poking around a bunch. Notice that the docs site is no longer _too wide_. Verify that the font sizes compute to what is expected. Verify that any modified component still renders as expected.

For bonus points, try scaling your browser font size and notice the text scale with it (more work will need to be done in [SPARK-154](https://stackoverflow.atlassian.net/browse/SPARK-154) to have all scaling live in harmony).

### DIFF
--- a/.changeset/four-planets-own.md
+++ b/.changeset/four-planets-own.md
@@ -1,0 +1,5 @@
+---
+"@stackoverflow/stacks": patch
+---
+
+Fix rem font size values
--- a/packages/stacks-classic/lib/base/body.less
+++ b/packages/stacks-classic/lib/base/body.less
@@ -20,7 +20,7 @@ html,
 body {
     color: var(--theme-body-font-color, var(--black-600));
     font-family: var(--theme-body-font-family);
-    font-size: var(--fs-base);
+    font-size: var(--fs-body1);
     line-height: var(--lh-base);
 }
 
--- a/packages/stacks-classic/lib/components/popover/popover.less
+++ b/packages/stacks-classic/lib/components/popover/popover.less
@@ -3,7 +3,7 @@
     --_po-bc: var(--bc-medium);
     --_po-bs: var(--bs-md);
     --_po-d: none;
-    --_po-wmn: 12rem;
+    --_po-wmn: 10.5rem;
     --_po-w: 100%;
     // content
     // --_po-topbar-height assumes the topbar height based on topbar styles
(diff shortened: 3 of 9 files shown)


giamir (approved):
Thanks Dan for taking the time to research and fix this issue. ❤️
It looks good to be merged for me.

mukunku (approved):

PR #2123, opened 2026-01-14
feat(user card) - Update User Card to SHINE styles (part 2)

ttaylor-stack:
[SPARK-127](https://stackoverflow.atlassian.net/browse/SPARK-127)

[Figma](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=7688-60373&m=dev)

This PR covers part 2 of the User Card component which includes:

- States:
    - New contributor
    - Original poster
    - Deleted
- Additional bling
    - Recognized member (collectives)
    - top 3 leaderboard (collectives)


packages/stacks-classic/lib/components/badge/badge.less
```diff
@@ -126,6 +127,32 @@
...
+        --_ba-bc: var(--purple-400);
+        --_ba-gap: var(--su2); // 2px gap between icon and text
+        --_ba-before-h: calc(var(--su12) + var(--su2)); // 14px
+        --_ba-before-w: calc(var(--su12) + var(--su2)); // 14px
+        --_ba-before-icon: url("data:image/svg+xml;,%3Csvg width='14' height='14' viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M8 1a7 7 0 1 1 0 14A7 7 0 0 1 8 1m0 1.25a5.75 5.75 0 1 0 0 11.5 5.75 5.75 0 0 0 0-11.5m1 10.06H7V10.6h2zM8.14 4c1.78 0 2.77.96 2.77 2.5 0 1.79-2.1 2.2-2.1 3.25h-1.6c0-2 2-2.08 2-3.26 0-.54-.27-.96-1.02-.96-.7 0-1.03.46-1.1 1.23H5.4A2.75 2.75 0 0 1 8.13 4' fill='%23fff'/%3E%3C/svg%3E");
+
+        &:before {
+            height: var(--_ba-before-h);
+            margin-top: 0;
+            margin-left: calc(var(--su2) * -1); // -2px margin
```
giamir: We should have `--sun2` now after Dan's PR got merged

packages/stacks-classic/lib/components/user-card/user-card.less
```diff
@@ -50,14 +58,36 @@
         color: var(--black-400);
     }
 
+    & &--original-poster {
```
giamir: I wonder if this should be named `s-user-card--username__op`? cc @dancormier

packages/stacks-classic/lib/components/user-card/user-card.less
```diff
@@ -50,14 +58,36 @@
...
+        color: var(--_uc-username-fc);
         font-size: var(--_uc-username-fs);
         font-weight: 500;
         min-width: 0;
         overflow-wrap: break-word;
+        padding: var(--_uc-username-p);
         word-break: break-all;
     }
 
+    & &--awarded-third {
```
giamir: How are those used? I cannot find them in the docs snippets?

packages/stacks-classic/lib/components/user-card/user-card.less
giamir: The design differ compared to figma

PR
Screenshot 2026-01-16 at 15 47 38


Figma
Screenshot 2026-01-16 at 15 49 11

packages/stacks-svelte/src/components/UserCard/UserCard.svelte
```diff
@@ -1,12 +1,21 @@
...
-    import { IconStarVerifiedSm } from "@stackoverflow/stacks-icons-legacy/icons";
+    import {
+        IconStarVerifiedSm,
+        IconAchievementsSm,
+    } from "@stackoverflow/stacks-icons-legacy/icons";
     import type { Snippet } from "svelte";
+    import Popover from "../Popover/Popover.svelte";
+    import PopoverReference from "../Popover/PopoverReference.svelte";
+    import PopoverContent from "../Popover/PopoverContent.svelte";
+    import avatarDeleted16 from "../../assets/img/avatar-deleted-16.svg?url";
```
giamir: Does this inline the asset otherwise it won't work after things get bundled?

giamir: I think we could probably create an AvatarDeleted16.svelte and an AvatarDeleted24.svelte components in the `UserCard` folder for now with the svg in it.

packages/stacks-svelte/src/components/UserCard/UserCard.svelte
```diff
@@ -131,28 +170,101 @@
...
+                                class="s-user-card--username s-user-card--original-poster"
+                                >{name}</span
+                            >
+                        </PopoverReference>
+                        <PopoverContent>
+                            <span>
+                                <a
+                                    href={profileUrl}
+                                    class="s-link s-link__underlined">{name}</a
+                                > is the original poster.
```
giamir: should this part be paramaterized i18nOpTooltipText? Default `is the orginal poster.`

packages/stacks-svelte/src/components/UserCard/UserCard.svelte
```diff
@@ -131,28 +170,101 @@
...
+                {@render recognition()}
+            </PopoverContent>
+        </Popover>
+    {/if}
+    {#if badges}
+        <ul class="s-user-card--group">
+            {@render badges()}
+        </ul>
+    {/if}
+    {#if award}
```
giamir: I think we would be better off to create `UserCardAdditionalBling` subcomponent here for consumer to use. We are inferring too much from the value of award (tooltip text, class, etc...). And then just have a `additionalBling` snippet in the root `UserCard` component.

packages/stacks-svelte/src/components/UserCard/UserCard.svelte
```diff
@@ -131,28 +170,101 @@
...
+                    <span class="s-user-card--username">{name}</span>
+                {/if}
             {/if}
         </svelte:element>
     {/if}
 {/snippet}
 
+{#snippet userCardMainContent()}
+    {@render avatarAndName()}
+    {#if recognition && size === "sm"}
```
giamir: Why do we need `recognitionHref`? 
Can we actually make the `UserCardAdditionalBling` subcomponent generic so that the consumer can decide if they want to show the "little cup icon" bling or the "recognition icon" bling instead? It would be nice to keep the additions on the UserCard.svelte component to a minimum. As I said in a previous comment maybe just a `additionalBling` (or `additionalBlings`) snippet depending if more than one bling can be displayed together or not. Keep things generic and then have any additional logic in the `UserCardAdditionalBling` subcomponent.

packages/stacks-docs/_data/user-card.json
```diff
@@ -218,5 +218,210 @@
...
+  ],
+  "states": [
+    {
+      "name": "Original Poster",
+      "class": "s-user-card--original-poster",
+      "description": "This label identifies the author of the primary post (such as the Question asker) when they appear in secondary contexts, like comment threads."
+    },
+    {
+      "name": "New Contributor",
+      "class": "s-user-card--new-contributor",
```
giamir: This class does not exist in less, we should probably remove it from the docs and the examples

packages/stacks-svelte/src/components/UserCard/UserCard.svelte
```diff
@@ -131,28 +170,101 @@
...
+        <Popover id="user-card-recognition-popover" tooltip>
+            <PopoverReference>
+                <svelte:element
+                    this={recognitionHref ? "a" : "div"}
+                    href={recognitionHref}
+                    class="s-user-card--group s-user-card--recognition"
```
giamir: The same `s-user-card--recognition` is used for 2 purposes:
- for the container element "Recognized by..."
- for the recognition icon in the additional blings section

I think this is confusing, we are probably better off create another class for the additional blings part

packages/stacks-svelte/src/components/UserCard/UserCard.svelte
```diff
@@ -131,28 +170,101 @@
...
+            </PopoverContent>
+        </Popover>
+    {/if}
+    {#if badges}
+        <ul class="s-user-card--group">
+            {@render badges()}
+        </ul>
+    {/if}
+    {#if award}
+        <Popover id="user-card-award-popover" tooltip>
```
giamir: popover ids need to be unique for the whole page so if we generate them inside svelte we need to either create a random text we can append to them or ask the consumer to do that

giamir (changes requested):
Thanks Tavian for the PR. As you can see I am asking for few changes. Hopefully they are all understandable if not let me know and we can pair.

For the svelte component I suggest to create a `UserCardAdditionalBling` subcomponent so that we don't have to add a lot of logic and props in the existing root `UserCard` component but just a snippet. 

**Sidenote**
_I think `AdditionalBling` is a pretty poor name to describe the little extra icons that can be added to between the badges and the blings but it is what we have in Figma so I guess we can roll with it. I don't have a better name for now.  cc @CGuindon_

packages/stacks-classic/lib/components/user-card/user-card.less
ttaylor-stack: It's used to set the colors for the additional bling trophies. It's used in the awarded snippet like so

` <div class="s-user-card--group s-user-card--awarded-third" title="…" data-controller="s-tooltip">
            @Svg.Icon.AchievementsSm.With("native")
        </div>`

giamir: Thanks for clarifying. I would like to get @dancormier opinions on the API naming here. There is a part of me that still wonder if this "additional blings" are worthy to be included already in the central library or we should just have the consumer define them for now and only centralize when we see the pattern is used across several scenarios (not just collectives). What I mean is that instead of creating a `s-user-card--awarded-third` class to add just the color consumers could use atomic color classes and take care of it themselves (we could still keep an example in the docs but we won't have any dedicated less). The equivalent in the svelte component world would be just to have a snippet (e.g. `additionalBlings` - it would be nice to find a better name but it is what it is)  to plug things in it but also there we will allow consumers to pass in it whatever they want. 

Also cc @CGuindon for visibility.

giamir: By the way what I am suggesting is some version of the [last responsible moment](https://blog.codinghorror.com/the-last-responsible-moment/). Do we have enough confidence to nail the API at this stage, with the information and the use case we have or is it better to keep things flexible for the time being and defer centralization of some styling to a later moment.

packages/stacks-svelte/src/components/UserCard/UserCard.svelte
ttaylor-stack: In the Figma it states that the additional bling icons need to provide links to either the collective leaderboard or the collective landing page, this is a way for the user to provide this. I've removed it from UserCard.Svelte and placed it into the additional bling subcomponent

image

packages/stacks-classic/lib/components/user-card/user-card.less
```diff
@@ -76,6 +81,17 @@
...
     }
 
+    &--username__op {
+        --_uc-username-p: var(--su2) var(--su4) var(--su2) var(--su4);
+        --_uc-username-bg: var(--blue-100);
+        --_uc-username-bl: var(--su4) solid var(--blue-300);
+        --_uc-username-d: flex;
+        --_uc-username-ai: center;
+        .s-user-card__sm & {
+        --_uc-username-p: 0px var(--su4);
```
giamir: NIT: incorrect indentation and `px` (unit of measure) not necessary

packages/stacks-docs/_data/user-card.json
giamir: @ttaylor-stack it looks like there are still references of the `s-user-card--new-contributor` class across the codebase, Maybe you want to do an old fashion CTRL+F and make sure everything is cleaned up. Thank you. 🙂

packages/stacks-svelte/package.json
```diff
@@ -24,7 +24,7 @@
         "test": "npm run build:dependencies && web-test-runner && npm run test:tools",
         "test:watch": "npm run build:dependencies && web-test-runner --watch",
         "test:tools": "npm run storybook:build && node --test tools/*.test.js",
-        "storybook": "storybook dev -p 6006",
+        "storybook": "npm run build:dependencies && storybook dev -p 6006",
```
giamir: Why this bit @ttaylor-stack? We shouldn't need it since the `less` files of stacks classic are compiled on the fly.

packages/stacks-svelte/src/components/UserCard/AvatarDeleted16.svelte
giamir, line 1: I was actually suggesting something dumber than this. HTML (like SVGs) are valid svelte component. So `AvatarDeleted16.svelte` would basically be as simple as:

```svelte

<svg ...>
  ...source code now under `assets/img` folder
</svg>
```
If you prefer we could name the component `AvatarDeleted16Svg.svelte` or `AvatarDeleted16Raw.svelte` but it does not really matter since they are internal only to the `UserCard`. Nothing else can or should use it for now. 
That way we don't need to create an assets folder at all for now.

giamir: Maybe we should sync about this. Perhaps I don't fully see all limitations/issues we would have with the approach I am suggesting.

packages/stacks-svelte/src/components/UserCard/UserCardAdditionalBling.svelte
```diff
@@ -0,0 +1,80 @@
...
+
+        /**
+         * The text for the tooltip
+         */
+        tooltipText: string;
+
+        /**
+         * Unique id for the popover
+         */
+        popoverId: string;
```
giamir: no need to call this popoverId, we just need an `id` for the component and should only knows that. That we use that for a popover internally is an implementation detail. 🙂

packages/stacks-svelte/src/components/UserCard/UserCardAdditionalBling.svelte
```diff
@@ -0,0 +1,80 @@
...
+
+        /**
+         * The additional bling icon
+         */
+        icon: string;
+
+        /**
+         * href for the link to leaderboard or collective landing page
```
giamir: I think the description should be more generic
`link for the additional bling element (e.g. leaderboard or collective landing page`)

packages/stacks-svelte/src/components/UserCard/UserCard.svelte
```diff
@@ -69,25 +69,30 @@
         recognition?: Snippet;
 
         /**
-         * Snippet used to display user bio
+         * Snippet used to display user recognized member
```
giamir: Why do we need a dedicated snippet for the recognized member bling? Can everything fall under the `additionalBlings` snippet umbrella?

It looks like all additionalBlings are always rendered after the badges according to figma
Screenshot 2026-01-19 at 15 26 25

At the moment it seems we are placing the recognized member one before the badges snippet which should not be the case. I think one more reason to have a single snippet called `additionalBlings`.

packages/stacks-svelte/package.json
ttaylor-stack: I was getting this error `[ERROR] Failed to resolve entry for package "@stackoverflow/stacks-utils". The package may have incorrect main/module/exports specified in its package.json. [plugin vite:dep-scan]`

I've removed it and it seems to be working atm

packages/stacks-svelte/src/components/UserCard/AvatarDeleted.svelte
```diff
@@ -0,0 +1,40 @@
+<script lang="ts">
+    type Size = 16 | 24;
+
+    interface Props {
+        /**
+         * Account name for screen readers
```
giamir: NIT: wrong description. It's an internal only component, we could also skip jsdoc all together.

packages/stacks-svelte/src/components/UserCard/AvatarDeleted.svelte
```diff
@@ -0,0 +1,40 @@
...
+            width="72"
+            height="72"
+            preserveAspectRatio="none"
+            xlink:href="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAEgAAABICAYAAABV7bNHAAAACXBIWXMAACE4AAAhOAFFljFgAAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8YQUAAAKZSURBVHgB7dxBbtpAFAbg92wv6I4jtDfIEWAT0UVbo4rQrBpOEOUEVU8Q9QRxV22gEo66oav6CO0NOALLLIJfPJEtHOTxMyZYIfyflMVocGx+yR4znmf+OQmFVqJPA79LBX5Mph0m/ksWyXZs60v2YbbrFPWJI6PTj/2gcJ/j6RkzX1Ex67Gm+xRbn5B0Twf9iCocq0NQCgEpvOQvyhrJKfTP9kHXo0V8t/rsRlj+k1jOwJjmts0chxYixfssO9ZUZOtwhRa2vtJjBQAA2Cts7pCzhhn+Tk762vD5oo3H06MlUztre/mfDzE/3Dt06YDFzJeMnxrVISAFAlIgIAUCUiAggJ2qPTN0/Nb3HVm2aUuz2e+grL/Xe3dGW3Pns1kYUQ0e1cSyPI8tE/EbCso6k31c0ZaYlhFRvdlQXKQVCEiBgBRe+qDsgXlSMBx8uKiy4auW27+9pa0v0hqH3DfUoOvJzWXyYPEoa5uLdCdrJB1UVRiG5tHJgnYsGX3m1KA0nE7WximmQEAKBKRAQAoEpPBEZJQ1HNn9qPTcyZ18JZe+EwAAwN5js9R21aK5bXnsobj+NfXjOL944fE65IhqTk2+FCJ8zoxf85UhIAUCUiAgBQJSVC5F2ITv++0mJvTzWi1apPPk21krRdhJUcJx732QDJifqUFMTvBndjOiJ4ZTTIGAFAhIUXt1RxmH4iC5KkTUKJ4TwLPDVaueDwWqnjeEgBQISIGAFAhIgYAUnnnRR9ZwsXiBHJGLfEkmAADA/uKqpQjmjQSm6N72j8pmAdaX9z86gFi+DYf9sKgvfW/al8LtlLKJ/PdaZ4Zy2xsmapcimHsDrlkftr68Py/2ShZMCr0mtrwcTi+b6Ng6yu5zUIqwIQSkuAe53NsA2WVtFwAAAABJRU5ErkJggg=="
```
giamir: @CGuindon this svg is actually embedding a png, is that on purpose? Didn't we have a regular svg?

packages/stacks-svelte/src/vite-env.d.ts
giamir, line 1: I don't think we need this file anymore @ttaylor-stack

packages/stacks-svelte/src/components/UserCard/AvatarDeleted.svelte
CGuindon: @ttaylor-stack Yes! Sorry that's a figma thing, I can't add an SVG to the figma component so I added the PNG for the design mockups. The SVG for the deleted user [is here](https://www.figma.com/design/Z5yoO4WH58QDHvmxwMWhr0/Stacks-Icons--beta-?node-id=896-208&t=tUzDErhCH5kHpqJ6-1) (also exported and attached)
Deleted

packages/stacks-classic/lib/components/user-card/user-card.less
CGuindon: I'm good with "just have the consumer define them for now" but I think we wanted @dancormier to weigh in as well

CGuindon (changes requested):
Is there an easy way to have the height of the container that the username is in be the same as the avatar height for both the Original poster and New contributor on the larger size. It's missing 1px of padding above and below I think to match.
Screenshot 2026-01-21 at 16 56 53

ttaylor-stack:
> Is there an easy way to have the height of the container that the username is in be the same as the avatar height for both the Original poster and New contributor on the larger size. It's missing 1px of padding above and below I think to match. Screenshot 2026-01-21 at 16 56 53

@CGuindon updating original poster is fine I can do that now however with the new contributor in the figma it matches the height of the other user badges not the avatar. Do we still want to increase the height to match the avatars? Also if that's the case do we want to update all user badges to match the height of the avatar?

packages/stacks-svelte/src/components/UserCard/AvatarDeleted.svelte
```diff
@@ -0,0 +1,30 @@
...
+    <rect
+        x="1.44543"
+        y="1.44543"
+        width="45.1091"
+        height="45.1091"
+        stroke="#A8ABB0"
+        stroke-width="2.89086"
+        stroke-dasharray="5.78 5.78"
+    />
+    <rect x="10" y="14" width="12" height="3.12" fill="#46484D" />
```
giamir: @ttaylor-stack the color for the `fill` and `stroke` attributes need to use our color custom properties (e.g. `var(--black-400)`) otherwise it only looks good in light mode. @CGuindon could you help Tavian map the correct colors to the different part of the svg?

ttaylor-stack: In the [Figma](https://www.figma.com/design/Z5yoO4WH58QDHvmxwMWhr0/Stacks-Icons--beta-?node-id=896-163&m=dev) it looks like black-300 for the border and black-500 for the face. I'll update that now. @CGuindon can your confirm that's right?

packages/stacks-classic/lib/components/user-card/user-card.less
dancormier: >instead of creating a s-user-card--awarded-third class to add just the color consumers could use atomic color classes and take care of it themselves

💯 I would much prefer to push this onto the consumer as to not over complicate our API if we're not certain of our approach

CGuindon:
> @CGuindon updating original poster is fine I can do that now however with the new contributor in the figma it matches the height of the other user badges not the avatar. Do we still want to increase the height to match the avatars? Also if that's the case do we want to update all user badges to match the height of the avatar?

No my mistake, just original poster is fine.

packages/stacks-svelte/src/components/UserCard/AvatarDeleted.svelte
CGuindon: Yes correct! @ttaylor-stack

packages/stacks-svelte/src/components/UserCard/UserCardBadge.svelte
```diff
@@ -1,14 +1,34 @@
...
+         */
+        tooltip?: Snippet;
     }
-    const { type }: Props = $props();
+    const { type, tooltip }: Props = $props();
 </script>
 
 <li>
-    <Badge text={type} type="user" userType={type} size="sm" />
+    {#if tooltip && type === "new"}
```
giamir: @ttaylor-stack NIT, no need here to check also for `type ==="new"`. This can be left generic. If tooltip is defined we show the tooltip otherwise we don't.

packages/stacks-svelte/src/components/UserCard/UserCard.svelte
```diff
@@ -1,12 +1,17 @@
 <script lang="ts" module>
     export type Size = "sm" | "lg" | undefined;
+    export type AwardBlings = "first" | "second" | "third" | undefined;
```
giamir: @ttaylor-stack Leftover?

giamir (approved):
I have left a couple of minor comments. Once those are addressed I think this is good enough to be merged. Thanks @ttaylor-stack

PR #2124, opened 2026-01-14
feat(stacks-utils): new @stackoverflow/stacks-utils package

mukunku:
# Summary

This PR creates a new workspace and npm package `stacks-utils`. This name is available to us on https://npmjs.org and should get published once we merge this PR.

The new package exposes 1:1 clones of the following functions: 
* [formatTime(utcTime: string)](https://github.com/StackEng/StackOverflow/blob/main/StackOverflow/_Scripts/islands/%40stackeng/stacks-svelte-labs/src/utils/time-formatter.ts)
* [formatReputation(rep: number)](https://github.com/StackEng/StackOverflow/blob/main/StackOverflow/_Scripts/islands/%40stackeng/stacks-svelte-labs/src/utils/reputation-formatter.ts)
  * Renamed to `formatCount(count: number)` because we intend to use this for reputation _and_ votes.

# How to Test

Not much to test for the new package itself, the github workflow [already ran the new tests](https://github.com/StackExchange/Stacks/actions/runs/21010991724/job/60405543636?pr=2124).

For the vote component the before and after can be compared:
* https://beta.svelte.stackoverflow.design/?path=/docs/components-vote--docs#counts
* https://deploy-preview-2124--stacks-svelte.netlify.app/?path=/docs/components-vote--docs#counts
image


packages/stacks-utils/package.json
```diff
@@ -0,0 +1,29 @@
+{
+    "name": "@stackoverflow/stacks-utils",
+    "description": "Stack Overflow utility functions for formatting and data manipulation",
+    "version": "0.0.1",
```
mukunku, line 4: I opted to go with `0.0.1` as the starting version in case we need to push hotfix versions since this is a new package and everything might not work smoothly. This way we don't end up with a version like 1.0.4 right off the bat. But open to suggestions; I can put 1.0.0 and hope for the best.

packages/stacks-utils/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "scripts": {
+        "build": "tsc",
+        "test": "vitest run",
+        "test:watch": "vitest watch",
+        "format": "prettier --write .",
+        "lint": "eslint . && prettier --check .",
+        "lint:fix": "eslint . --fix"
+    },
+    "dependencies": {
+        "dayjs": "^1.11.19"
```
mukunku, line 27: I checked and `dayjs` seems like it's kept up-to-date and used actively.

packages/stacks-utils/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+    "compilerOptions": {
+        "strict": true,
+        "target": "ES2020",
```
mukunku: CommonJS for max compatibility.

giamir: No need to support this old version and get our code transpiled with polyfill or old syntax. Let's use `esnext` for both target and module like we do for the new svelte package. 🙂

packages/stacks-utils/vitest.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+    test: {
+        globals: true,
```
giamir: I don't think you need to set globals to true. That will expose the testing api (describe, it, etc...) to the global namespace of the test but we don't need that since we `import` them from the vitest package. Let's remove it.

packages/stacks-utils/vitest.config.ts
```diff
@@ -0,0 +1,8 @@
...
+        environment: "node",
```
giamir: Also I think environment node is not necessary. The source code will need to run both in the browser and node (server side rendering). We probably can simply delete the configuration file given how simple the setup is and just keep the vitest default for the time being.

packages/stacks-utils/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+        "module": "ES2020",
+        "moduleResolution": "bundler",
+        "declaration": true,
+        "declarationMap": true,
+        "outDir": "./dist",
+        "rootDir": "./src",
+        "skipLibCheck": true,
+        "esModuleInterop": true,
```
giamir: Do we need to set `esModuleInterop` because of `dayjs`? If not, it is probably ok to leave it out.

packages/stacks-utils/package.json
giamir: I think we can leave 0.0.1 but given we are in a prerelease mode branch let's mark the changeset as major instead of patch so that the first release will be `1.0.0-beta.0` and then `1.0.0-beta.1` and so on. The version `1.0.0` will only be released once we merge this `beta` branch back in `develop`.

.changeset/three-humans-search.md
```diff
@@ -0,0 +1,5 @@
+---
+"@stackoverflow/stacks-utils": patch
```
giamir: > I think we can leave 0.0.1 but given we are in a prerelease mode branch let's mark the changeset as major instead of patch so that the first release will be 1.0.0-beta.0 and then 1.0.0-beta.1 and so on. The version 1.0.0 will only be released once we merge this beta branch back in develop.

Swap it to major will give us a first release that looks like `1.0.0-beta.0`

giamir (approved):
Great work @mukunku. 🚀
I have left only minor comments to address/clarify. 
Apart from that I think we are good to merge.

mukunku:
Thanks for the review @giamir I believe I addressed all your comments, can you kindly take a look again?

giamir (approved):
Thanks @mukunku. Let's ship it 🚀

PR #2137, opened 2026-01-20
feat(post-summary): update to SHINE styles

dancormier:
[SPARK-75](https://stackoverflow.atlassian.net/browse/SPARK-75), [Figma](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=612-18817&p=f&m=dev)

---

This PR updates the Post Summary to the latest SHINE styles. cc @ycantu 

image

> [!NOTE]
> See also [this PR](https://github.com/StackExchange/Stacks/pull/2162) which includes updated visual regression test images.

## Deploy previews
- [Stacks Classic docs](https://deploy-preview-2137--stacks.netlify.app/product/components/post-summary/)
- [Stacks Svelte docs](https://deploy-preview-2137--stacks-svelte.netlify.app/?path=/docs/components-postsummary--docs)

## TODO
- [x] Revisit title line height (should be ~20px)
- [x] Revisit excerpt line height (should be ~20px)
- [x] Update gutters for content rows
  - [x] `8px`: Usercard/stats to title
  - [x] `4px`: Title to excerpt
  - [x] `10px`: Excerpt to tags
- [x] Change tag gap to `8px`
- [x] Change score to votes gutter to `8px`
- [x] Change unaccepted answers color to `black-400`
- [x] Docs: Add `v-truncate` class table
- [x] Change answer left border color to `black-200`

## Open questions/requests
- [x] How should we handle the `:visited` title given that [we cannot manipulate the `font-weight` property](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:visited?utm_source=devtools&utm_medium=inspector-inactive-css#privacy_restrictions) in this context?
  - Punting on this
- [ ] We need the following icons
  - [ ] `Answer16`
  - [ ] `Answer16Fill`
  - [x] `Shield` (using legacy icon for now)


MIGRATION_GUIDE.md
```diff
@@ -137,6 +137,9 @@ The menu component has been updated to use new class names and structure. The fo
 #### Pagination
 - The next and previous button now uses an `ArrowRight` and `ArrowLeft` icon instead of text. To apply the new styling, use the class `.s-pagination--item__nav`. Since these buttons use icons to represent their behavior, make sure to include descriptive text for screen readers.
 
+#### Post Summary
+The Post Summary component has changed dramatically. Please see refer to the docs for complete guidance.
```
giamir:
```suggestion
The Post Summary component has changed dramatically. Please refer to the docs for complete guidance.
```

packages/stacks-classic/lib/components/post-summary/post-summary.less
```diff
@@ -1,458 +1,288 @@
...
-    & &--answer {
-        --_ps-stats-ai: center;
-        --_ps-stats-fd: row;
-        --_ps-stats-w: auto;
-    }
-
-    &&__minimal {
-        .s-post-summary--content {
-            width: 100%;
+    @psx-container-sm: 28rem;
```
giamir, line 2: Curiosity: How did we settle on 448px as a breakpoint?

giamir (approved):
I skimmed through it and everything seems in order from a code perspective @dancormier. Thanks for tackling this huge component. 

The only thing I observed is that for the Svelte component the `readableTimestamp` tooltip is somehow not tabbable via keyboard.

CGuindon (changes requested):
YAY!!! Looking good! I only have small nit picky stuff:

1. I noticed the line-height on the description and title are 18px — which feels a bit tight. We increased the base line-height to 140% which _should put the line-height of the description to around 19.6px (we probably rounded up to 20px as that's what I'm seeing on the typography [page](https://deploy-preview-2137--stacks.netlify.app/product/foundation/typography/#size-examples) for Body1)_. I think we had the line-height not centralized before so wondering if this is a case if it inheriting the custom old vs new here.

Image

---

2. I'm seeing a 6px gutter used as spacing in between each line. _Unfortunately_ (sorry), the spacing here isn't all the same. We have 8px, 4px and 10px amounts between each line. (Note the bottom tags container does have a sneaky 2px margin top if that helps or not with the math).

3. There should be an 8px gap between each tag (instead of 4).

Image
Image

---

4. The gutter between the score box and the container with the answer data should be 8px (instead of 6).

5. The font and icon color for the answer data (when not accepted) should be `black-400`

Image

---

6. Similar to how you have the classes listed out at the top, I think it might be worth listing out all the class options above each example (within each section). For example, showing the `truncate` classes and the state classes (danger, warning, etc) all listed out in a table between description and code example. I found it a bit hard sometimes to see find the "different" classes within the code examples that show everything.

---

7. Border left on answers should be `black-200` (instead of 150)

Image

packages/stacks-classic/lib/components/post-summary/post-summary.less
dancormier: I chose it after some trial and error. I figured we needed a breakpoint that was larger than the sidebar width (21.5rem, 344px). I tried that but felt like it looked pretty cluttered at just above that breakpoint. This image is at with the container right at 345px with the breakpoint at 344px:

image


For that reason, I chose the next defined size up in our library (the 28rem). That said, I could be convinced that this breakpoint should match up with the sidebar size. Any strong opinions on that?

dancormier:
Thanks for the review @CGuindon! I've address everything you mentioned. Let me know if I missed anything!

CGuindon:
@dancormier Sorry I probably miscommunicate the line-height. It looks good on the description now but still needs that 140% on the title of the post as well. Should be 25px based on [our new sizing](https://deploy-preview-2137--stacks.netlify.app/product/foundation/typography/#size-examples). Looks like you have 20px for both

CGuindon:
****UPDATE**: Ellen preferred above and it won't be as common so leaving as is for now.**

~~For the Responsive (small screen) version of the States (with badges like Review, Draft): I'm tempted to keep the badge in line with the metadata and not make the height of the whole post bigger. (Just want to confirm with Ellen she feels the same). My thinking would be to do something more like the Draft example than what we have now — I also realize we didn't have this in the Figma mocked up. (sorry)~~
Screenshot 2026-02-02 at 1 42 32 PM

Screenshot 2026-02-02 at 1 42 47 PM

packages/stacks-classic/lib/components/post-summary/post-summary.less
giamir: Sounds good to me, it is just not one of our "official" breakpoints meaning that it will change the layout at a custom viewport size compared to other things. I guess it is fine. I wonder though if the `sm` breakpoint (780px one) could have suffice. In the sense that for viewports below that size we show the compact version even if in theory there is the estate to show more. 
Anyway I am good with what we have. This is really a NIT. Thanks Dan.

dancormier:
@CGuindon sorry for the oversight! I've updated the title to have a 25px (aka 1.563rem) line height.

CGuindon (approved):
Just the Answer icon missing (should be released now in the icons package) but otherwise LGTM 🙌

CGuindon (approved):
🙌

packages/stacks-docs/_includes/post-summary-item.html
```diff
@@ -0,0 +1,199 @@
...
+                    {{ data.votes | default: 24 }}
+                </div>
+                <!-- Answers -->
+                <div class="s-post-summary--stats-answers">
+                    {% if data.answered %}
+                        {% icon "Answer16Fill", "s-post-summary--stats-answers--icon" %}
+                    {% else %}
+                        {% icon "Answer16" %}
+                    {% endif %}
+                    {{ data.answers | default: 1 }}
```
mukunku: Should this be `answerCount`? On the smallest page size the answer count gets rendered as `[object Object]`
```suggestion
                    {{ data.answerCount | default: 1 }}
```

Image

mukunku (commented):
Wow this is amazing stuff. Great work, Dan! 

Left one small comment below but otherwise flawless 🤌

packages/stacks-docs/_includes/post-summary-item.html
dancormier: Ah, thanks for the nice catch here! Committed ✅

PR #2138, opened 2026-01-21
fix(user-card): export child components

dancormier:
We need to export the user card child components that are used as the content for snippets h/t @chris-doucette-stack ([slack message](https://stackexchange.slack.com/archives/C096063B1PH/p1769025081177999))


chris-doucette-stack (approved):
Thank you!

PR #2142, opened 2026-01-23
feat(loader) - Update Spinner to SHINE styles; rename to Loader

ttaylor-stack:
[SPARK-117](https://stackoverflow.atlassian.net/browse/SPARK-117)

[Figma](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=617-18823&p=f&t=41n7ExYMHp4Qbm6a-0)

As mentioned in the ticket we will not be implementing the Logo Loader in this PR.

NOTES:
- The animation matches what is in the Figma however I think it slightly differs from the icon in the [icons package](https://beta.icons.stackoverflow.design/). @CGuindon can you confirm if that's alright?
- I wasn't sure on exact timings of the animation so I tried to match what was in the beta package. cc: @CGuindon 
- I've added in a `type` property in anticipation for the next part of the component. Is this the right choice or should we remove it until the the next part of the component is implemented? @giamir @dancormier
- For some reason when the loader is a child of a button element the animation stopped working correctly so I created a second animation for that. There is probably a better way to do this. So I'm happy to hear any suggestions @dancormier @giamir


CGuindon (changes requested):
I would try to match what's in in the package that @abovedave did. A couple differences I'm noticing for the animation:
- Only 1 block is moving at a time, is it possible to have the second block start to go up before the first one comes down? To get more of a wave effect.
- In the package, each block goes up and down without stopping. In your version they seem to have a pause before going up again, it should be a more fluid up/down non-stop.

(This might be what you asked Dan's help for)
- The docs are overlapping for the button loader instead of scrolling.
Screenshot 2026-01-26 at 7 28 49 AM

abovedave:
When I was implementing the icon version, I was wondering if could just be used like a typical icon with the modifier `s-btn__icon` which could also be resizable. Might be a chance to remove some code from the css bundle.

dancormier (changes requested):
@ttaylor-stack This PR seems somewhat removed from what's specified in the ticket. I've created a [PR](https://github.com/StackExchange/Stacks/pull/2150) that points to this one to make these changes line up with the requested changes in SPARK-117.

Please merge it into your PR if you feel comfortable with that

packages/stacks-classic/lib/components/loader/loader.a11y.test.ts
giamir, line 1: Aren't we missing the `s-loader--block` element in both a11y and visual tests?

packages/stacks-classic/lib/components/loader/loader.less
```diff
@@ -0,0 +1,78 @@
...
+        margin-left: var(--su1);
+        margin-right: var(--su1);
+    }
+
+    &__lg {
+        --_ld-size: var(--su8);
+    }
+
+    // CHILD ELEMENTS
+    & &--block,
```
giamir: Adding an extra `s-loader--block` element feels a bit odd. Given we can assume that there will always be an element with text for the screen readers ` <div class="v-visible-sr">Loading…</div>` how crazy would it be to use the :before or :after pseudo element of that to create the middle block? We could even call it `s-loader--sr-text` or something
```
<div class="s-loader">
    <div class="s-loader--sr-text">Loading…</div>
</div>
```
Thoughts?

giamir (commented):
Overall looks good. I left an observation about the API. Maybe we can sync as a team about it

packages/stacks-classic/lib/components/loader/loader.less
dancormier: I considered that but scrapped it because I assumed the visibility of that element would make it tricky to pull off. That said, I'm good to test that assumption and, if it's not too much of a pain, implement it that way. Seems cleaner

ttaylor-stack: @giamir with that approach would it the final markup (minus the icon) look like this:

`<div class="s-loader">`
`    <div class="s-loader--sr-text">Loading…</div>`
`</div>`

or like this

`<div class="s-loader">`
`    <div class="v-visible-sr s-loader--sr-text">Loading…</div>`
`</div>`

giamir: I think like this
```
<div class="s-loader">
    <div class="s-loader--sr-text">Loading…</div>
</div>
```
No need to have the user add the visibility class since we are already calling that `s-loader--sr-text` and we could add the necessary extra properties directly in it.

packages/stacks-docs/_data/site-navigation.json
```diff
@@ -287,6 +287,10 @@
                             "title": "Link previews",
                             "url": "/product/components/link-previews/"
                         },
+                        {
+                            "title": "Loader",
+                            "url": "/product/components/loader/"
```
giamir: add` new: true`

giamir (approved):
Thanks @ttaylor-stack. I think we are good to merge now.

PR #2144, opened 2026-01-23
feat(tag): allow for dismiss btn and linked tag

dancormier:
This PR introduces support for a tag having both a dismiss button and linked tag text simultaneously. cc @ycantu 

See also: [Slack](https://stackexchange.slack.com/archives/C097YESGSMB/p1769162634880809?thread_ts=1768928502.861409&cid=C097YESGSMB)


giamir (approved):
Apart from adding a patch changeset for the 2 libraries I think this is good to be merged. I noticed that the focus outline of the link and the dismissable button looks a tad misaligned in the lg and sm variant but I personally think what we have is good enough. Thanks @dancormier

PR #2145, opened 2026-01-26
fix(button): adjust badge variant

giamir:
[SPARK-115](https://stackoverflow.atlassian.net/browse/SPARK-115)

This PR updates the badge variant of the buttons to reflect the colors and style in [this figma file](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=7491-52932&m=dev).

You can check how all the different combinations looks like in storybook:
https://deploy-preview-2145--stacks-svelte.netlify.app/?path=/story/components-button--badges

~I used my best judgment for the disabled state since there was no guidance in the figma.~

I moved the visual test fix in this separate PR to keep the diff smaller for this one:
https://github.com/StackExchange/Stacks/pull/2147


CGuindon:
"I used my best judgment for the disabled state since there was no guidance in the figma."

[Figma](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=7491-52932&t=vc1BewSXOJTvCKMS-1) had the disabled states... did I miss something?

Screenshot 2026-01-26 at 10 26 48 AM

CGuindon (commented):
LGTM other than the disabled states. I have them in the figma (scroll to the right to see all modes). 

I'm not sure if it's worth doing [this ticket](https://stackoverflow.atlassian.net/browse/SPARK-143) at the same time or only doing the disabled badges when we tackle that ticket.

CGuindon (approved):

dancormier (changes requested):
Thanks for these badge tweaks @giamir. Looks pretty good overall, but I noticed a few minor issues:

### Badge font size and padding adjustments
I noticed that the font size for the badge stays the same (`fs-caption`) for all button sizes but it should be set to `fs-fine` for `xs` and `sm` variants. Making this change will probably also require minor adjustments to the badge padding within those variants.

I also noticed a slight difference in the height of the badge in the `lg` button (~21px in Stacks vs 24px in Figma).

I think something like this gets us pretty close to the mocks:

```diff
.s-btn {
…
+  --_bu-badge-fs: var(--fs-caption);
…
  &&__xs,
  &&__sm {
…
+        --_bu-badge-fs: var(--fs-fine);
+        --_bu-badge-py: var(--su1);
+        --_bu-badge-px: calc(var(--su4) - var(--su1)); // 3px
}
…
  &&__lg {
…
+        --_bu-badge-py: calc(var(--su2) + var(--su1)); // 3px

  }
…
    & &--badge {
…
+        font-size: var(--_bu-badge-fs);
…
}
```


### TODO can be removed
On line 235 of `button.less`, the comment `// TODO SHINE fine-tune badge styles` is still there but can be removed.

---

**Note**: I noticed that the selected state background colors have been swapped vertically. I created a [ticket](https://stackoverflow.atlassian.net/browse/SPARK-169) to capture correcting this issue.

giamir:
@dancormier Thanks for the review. I addressed your comments. 
https://github.com/StackExchange/Stacks/pull/2145/commits/2df8a17667071c766fb84a3f46942dd839420271

It should be now ready for a quick re-review. Thanks

dancormier (approved):
Thanks for the changes @giamir. Looks great

PR #2146, opened 2026-01-26
fix(usercard) - remove nested <a> tags in user card

ttaylor-stack:
[SPARK-163](https://stackoverflow.atlassian.net/browse/SPARK-163)

PR to remove the nested a tag in user card to fix the warning `node_invalid_placement_ssr`

Updated tests too.

This fix cannot be tested within Stacks. I will reach out to a core team to get them to test this when a new version of Stacks has been shipped

### DIFF
--- a/packages/stacks-svelte/src/components/UserCard/UserCard.test.ts
+++ b/packages/stacks-svelte/src/components/UserCard/UserCard.test.ts
@@ -54,18 +54,6 @@ describe("UserCard", () => {
         ).to.have.class("s-user-card__sm");
     });
 
-    it("should render the avatar and name as links", () => {
-        render(UserCard, {
-            name: "John Doe",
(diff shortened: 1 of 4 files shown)


packages/stacks-svelte/src/components/UserCard/UserCard.test.ts
```diff
@@ -54,18 +54,6 @@ describe("UserCard", () => {
         ).to.have.class("s-user-card__sm");
     });
 
-    it("should render the avatar and name as links", () => {
```
giamir, line 57: I think instead of removing the test, it would make sense to change it to just find one individual link instead of 2 separates one.

giamir (approved):
Looks good, only one minor observation about the test. Feel free to address that and merge. Thanks @ttaylor-stack

PR #2148, opened 2026-01-27
feat(vote): allow downvoting for horizontal layout

giamir:
[SPARK-155](https://stackoverflow.atlassian.net/browse/SPARK-155)

This PR:
- Documents and showcase downvoting as an option for horizontal layouts in stacks docs 
- Adds an `upvoteOnly` boolean in the stacks svelte vote component that allow consumers to render vote components with only the upvote button. By default now for both vertical and horizontal layout, downvoting and upvoting buttons are shown
- Update visual and a11y tests for vote component

# How to test
- Go to vote page in the stacks docs and check the markup and example to upvote/downvote in horizontal layout
- Go to the vote component in storybook and check out the new upvote only story


mukunku (approved):
LGTM. The only thing that caught my eye was that there seems to be a large gap between the arrow and the number for the vertical upvote only component, but maybe that's intentional:

Image

giamir:
> The only thing that caught my eye was that there seems to be a large gap between the arrow and the number for the vertical upvote only component, but maybe that's intentional

Good point, I am actually unsure if we should officially support that variant. 
The API allows for that combination but maybe it is not something we will see in the design.

CGuindon:
@mukunku @giamir No, we shouldn't have a use case where we use vertical upvote only button. The vertical variant is for up and down only.

PR #2150, opened 2026-01-27
Refactor loader to loading, match styles to spec

dancormier:
@ttaylor-stack this PR refactors your [PR](https://github.com/StackExchange/Stacks/pull/2142) to match the specifications in [SPARK-117](https://stackoverflow.atlassian.net/browse/SPARK-117). Most notably:

- The component is named `Loading` instead of `Loader`
- The animation is lifted from the icon (if you inspect it in [Stacks-Icons](https://beta.icons.stackoverflow.design/), you can see inline styles for the icon)
- The base component name was `.s-loader--block`. The `--*` suffixes are reserved for component child elements.
- Sizing is fixed
- `CurrentColor` is used for background color so it matches context

### What's still needed to complete ticket
- Button docs need to be accurately updated
- Visual regression images need to be updated
- Follow up ticket is created for the logo-based loading animation
  - **Note:** I anticipate this will be a different component in Stacks Classic if included there at all.


packages/stacks-classic/lib/components/loader/loader.less
```diff
@@ -1,61 +1,78 @@
-.s-loader--block {
-    // BASE COMPONENT-SPECIFIC CUSTOM PROPERTIES
-    --_ld-size: var(--su-static24);
-    --_ld-block-size: calc(var(--su-static8) - var(--su-static1));
-    --_ld-gap: calc(var(--su-static2) + var(--su-static1));
-    --_ld-start-offset: var(--su-static2);
+.s-loader {
```
ttaylor-stack, line 1: I spoke to Caro because the Figma called it Loader but the story said Loading, she confirmed it is Loader. I forgot to update the ticket sorry! I've changed it back to loader

packages/stacks-svelte/src/components/Button/Button.svelte
```diff
@@ -183,14 +184,14 @@
     disabled={(!href && disabled) || null}
     aria-disabled={href && disabled ? "true" : null}
     {...restProps}
-    >{#if loader}{@render loader()}{/if}{#if !badge}
-        {@render children()}
-    {:else}
+    >{#if loading}<Loader
+            size="sm"
```
ttaylor-stack, line 188: how would the user update the label for internationalization? should we add another prop if we're not gonna use a Snippet?

ttaylor-stack: Add another prop for label

PR #2152, opened 2026-01-28
feat(vote): extract the vote state logic in an opt in svelte module

giamir:
[SPARK-155](https://stackoverflow.atlassian.net/browse/SPARK-155)

This PR is making our Vote.svelte component much dumber than it was and leave to consumers the job of managing the state (e.g. react to downvote and upvote clicks and update the other props accordingly).

There are also part of the vote state logic though (such as optimistic updates) that will end up being repeated over and over by product teams with potentially slightly different behavior which would only end up confusing the user. 

For this reason we also distribute a svelte module (`vote-state.svelte.ts`) which contains a `createVoteState` factory which takes initial state values and return reactive props that can be spread in the Vote component.

Example

```svelte
<script lang="ts">
  import Vote from "./Vote.svelte";
  import { createVoteState } from "./vote-state.svelte";

  const voteState = createVoteState({
    total: 12,
    onUpvote: async () => {
      const res = await fetch("/api/vote", {
        method: "POST",
        body: JSON.stringify({ vote: "up" })
      });
      const data = await res.json();
      // Return server values to override optimistic update
      return { total: data.total };
    },
    onDownvote: async () => {
      const res = await fetch("/api/vote", {
        method: "POST",
        body: JSON.stringify({ vote: "down" })
      });
      const data = await res.json();
      return { total: data.total };
    },
  });
</script>

<Vote {...voteState} />
```

[Storybook preview](https://deploy-preview-2152--stacks-svelte.netlify.app/?path=/docs/components-vote--docs)


mukunku (approved):
Lgtm

dancormier (approved):
⬆️ 🎉

PR #2154, opened 2026-01-28
fix(user-card): additional bling tooltip

ttaylor-stack:
Bug to fix tooltip in the user card docs. This also fixes a bug where the UserCardTime component had a static ID

Steps to test
- Go to usercard page
- Go to awards section
- Hover over icon
- Tooltip should have the correct text

### DIFF
--- a/.changeset/empty-suits-serve.md
+++ b/.changeset/empty-suits-serve.md
@@ -1,0 +1,5 @@
+---
+"@stackoverflow/stacks-svelte": patch
+---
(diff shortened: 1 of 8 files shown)


mukunku (approved):
Tooltips look good to me.

Noticed a weird issue with the Svelte docs, though. I can't type into any of the inputs without losing input focus 🤔 

Image

I don't mind fixing it in a follow-up PR.

PR #2155, opened 2026-01-28
feat(sidebar-widget): initial SHINE design changes

mukunku:
# Summary

This PR removes a few of the existing sidebar widget functionality while salvaging what I could and updating it according to the new [SHINE designs](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=617-18821&p=f&t=c9NSyJWdASb80eVC-0).

A lot of the old variations might not be relevant anymore so the idea is to add them back on a case-by-case basis as needed. The reason they were removed was because it didn't make sense to spend effort on them without proper designs.

# How to Test

Check out the Sidebar Widgets page: https://deploy-preview-2155--stacks.netlify.app/product/components/sidebar-widgets/


packages/stacks-docs/product/components/sidebar-widgets.html
```diff
@@ -9,667 +9,66 @@
...
-    <div class="s-sidebarwidget--content">
-        <img src="…" class="bar-md mr8">
-        <div class="lh4">
-            <strong>balpha</strong><br>
-            contributed 6 edits<br>
-            joined 8 months ago
-        </div>
+    <div class="s-sidebarwidget--header">
+        <h2>…</h2>
+        <a class="s-sidebarwidget--action">…</a>
```
giamir: We should add an `href` to this anchor otherwise we cannot test properly the focus when tabbing with keyboard

giamir (approved):
Thanks @mukunku

CGuindon (changes requested):
1. The sidebar is 344px without any padding so the sidebar widgets shouldn't have left/right padding. That spacing will be handled on the outside of the whole right sidebar.

Image

---

2. Add `s-btn s-btn__xs s-btn__clear` to the `s-sidebarwidget--action` (the + track button) <a> element so that it gets the button stylings for height and hover states. 

3. Then you're padding between the header and middle container can be a normal 12px class (instead of 16 + 2, because the button will make the header container taller).

_(screenshot shows hover state)_

Image

---

4. Should the text inside `s-sidebarwidget--content` be in a <p> element? (not 100% but wondering about a11y or html standards with text if it's needed here or not)

---

5. Title: There's a space before "Community Achievements" that shouldn't be there (made me think there was extra padding) and the title should be `fw-medium` (instead of bold).

6. (more rare use case but thinking of future proofing) Can we have the icon be it's own "column" on the left in the header? So that if the title _needs_ to wrap to a second line, the text wouldn't wrap under the icon.

7. AND make sure there's at least 6px padding in between the title and the button on the right (so if the title does get long, it won't touch the button's edge).

Screenshot 2026-01-30 at 8 56 13 AM
Screenshot 2026-01-30 at 8 56 24 AM

_Popped that second example in the Figma_

---

8. I _feel like_ the warning note should be above the example not after (easier to miss) — that means 2 notices before, I think that's fine.
Screenshot 2026-01-30 at 8 51 46 AM

mukunku:
Thanks for the review @CGuindon ! I believe I addressed all your comments:
1. I removed the x padding so we get the full 344px width now
2. I updated the header action style 👍🏾 
3. I played around with the paddings to match what you have in the figma. Please let me know what you think.
4. It doesn't have to be; it's valid html and accessible to screen readers. It's just not semantically correct, and is more proper for it to be wrapped with something. I wrapped it in `<span>` now just because...
5. Removed the space and set the font-weight to 500
6. I made the icon its own column but there's no way to top align the icon and header cta if the title ends up being multiple lines. Meaning the icon and header cta will be centered always. Still looks okay with 2 line headers I think.
7. Added 6px padding 👍🏾 
8. Moved the warning above now.

CGuindon (approved):
There's no Svelte component right? Should there be? or is this a different ticket? Approving the Stacks CLassic version so you can merge

PR #2157, opened 2026-01-30
Add s-loader--sr-text

ttaylor-stack:
Added s-loader--sr-text and removed --block


giamir (approved):
It looks good to me. I feel the API is much simpler this way for consumers.
Remember to update also the tests with the correct new markup.
I also noticed that the docs snippet is not formatted well
 Screenshot 2026-01-30 at 11 37 17

Thanks @ttaylor-stack

PR #2160, opened 2026-01-30
fix(button): adjust disabled state style in HC modes

giamir:
[SPARK-143](https://stackoverflow.atlassian.net/browse/SPARK-143)

This PR updates the styles of button disabled states in HC modes according to the [Figma spec](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=7732-16891&m=dev)

It also make sure that for the `base` and `base clear` button variant we exclusively use `theme-secondary` color step which don't have `x50`s steps. Therefore for those 2 variants you could observe some slight difference compared to what is in Figma.

[Buttons Docs Page (PR Env)](https://deploy-preview-2160--stacks.netlify.app/product/components/buttons/)


mukunku (commented):
I noticed the border color is black-600 in the designs for HC light mode for tonal and base but you have black-300. Other than that everything seems to match the designs.

Image

Image

mukunku (approved):
Opened a new PR to fix the border color: https://github.com/StackExchange/Stacks/pull/2161

CGuindon (changes requested):
I like most of the changes you made with the non-50 stops. 

1. For HC Dark and HC Light mode on `s-btn s-btn--badge` — the badge color should be black-300 (I think you went one stop the other direction). That will match what we have now on Dark/Light mode as well which is nice.

Image
Image

CGuindon (approved):

PR #2161, opened 2026-01-30
patch(button): fix hc border color for clear and tonal

mukunku:
This PR fixes button border color for `clear` and `tonal` variations in HC:
image
 
Figma: https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=7732-16891&m=dev


packages/stacks-classic/lib/components/button/button.less
```diff
@@ -45,7 +45,8 @@
             --_bu-badge-fc-disabled: var(--theme-secondary-300);
 
             .highcontrast-mode({
-                --_bu-bc: var(--theme-secondary-300);
+                --_bu-bc: var(--theme-secondary-600);
+                --_bu-bc-disabled: var(--theme-secondary-300);
```
CGuindon, line 49: I actually kind of liked the more subtle outline but black-300 doesn't pass 3:1 ratio for HC so we still need to bump up the contrast. Can we use black-400 here (instead of 600) and I'll update the designs. @mukunku

mukunku: Made the switch to 400 now

PR #2163, opened 2026-02-02
feat(label): update styles to match the new SHINE specs

giamir:
[SPARK-69](https://stackoverflow.atlassian.net/browse/SPARK-69)

[Figma](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=7925-51791&m=dev)

This PR:
- Remove `s-label__md`, `s-label__xl`
- Remove `s-label--status`, `s-label--status__required`, `s-label--status__new`, `s-label--status__beta` in favor of using the new `s-badge` states instead.
- Adjust the private `Label.svelte` component accordingly and the components using it internally (TextInput, TextArea and Select)
- Reactivate the label a11y and visual tests

I took the liberty of reducing the padding bottom between label and input for `sm` variant from `--su8` to `--su4` even if in Figma there was no explicit guidance regarding sizes and related paddings.

[Label Docs](https://deploy-preview-2163--stacks.netlify.app/product/components/labels/)

Storybook ([TextInput](https://deploy-preview-2163--stacks-svelte.netlify.app/?path=/story/components-textinput--label-status), [TextArea](https://deploy-preview-2163--stacks-svelte.netlify.app/?path=/story/components-textarea--label-status), [Select](https://deploy-preview-2163--stacks-svelte.netlify.app/?path=/story/components-select--label-status))


packages/stacks-classic/lib/components/label/label.less
```diff
@@ -110,7 +81,17 @@
     text-decoration: none !important;
 }
 
-// [1] In Core, we have *many* instances of `.s-label--status` used without the `.s-label` parent.
-//     While I'd prefer to enforce the requirement of the parent class, it's too much of a lift at this moment.
-//     We'll come back to it, hopefully when we have a pill component to replace the current usage of `.s-label--status`
-//     without the base label class.
+// Helper mixin to apply styles to elements
```
giamir: @dancormier I found it pretty difficult to add the new `--su8` gap between label and inputs. This is because the label component is an entirely separate element. I ended up creating this mixin to avoid some duplication. It goes off and apply the padding if it finds a `s-input` `s-textarea` or `s-select` as a sibling (or nested sibling). Any thoughts on how we could make this better? 

There is always the option of leaving this up to the consumers (they add themselves the appropriate atomic classes) but I feel it defeats the purpose of centralizing this component for consistency across pages.

CGuindon (changes requested):
1. I noticed you have 8px padding below the `label` but there's also a 4px gutter spacing. If you need to keep the gutter you could do 4+4 so that the total spacing is only 8.
Image
- On the small size, reduce the gutter between the label and input to 2 (instead of 4) and it will make 6px total with the 4px padding that's already there.
- Large has a total of 10px right now (8px gutter + 2px padding) – that's good, leave as is.

---

2. Similarly with the description line, there's both a 4px padding and 4px gutter space doubling the amount (should be 4px total).
Image
Image

---

3. For the Status variants, the badge component isn't perfectly aligned vertically in the middle. It looks like the it's aligned with the baseline of the text but because the badge font size is smaller, it aligns everything else lower. 
Image

---

4. There's a "space" added after each label title before the badge in the docs. Makes it seem like there's too much padding, we can delete this character.

packages/stacks-classic/lib/components/label/label.less
dancormier: Ah, yeah this seems tricky.

My expectations would be that components like this don't handle the spacing themselves if possible.

I think the _optimal_ (but possibly more effortful) approach would be to leverage [`.s-form-group`](https://github.com/StackExchange/Stacks/blob/SPARK-69/labels/packages/stacks-classic/lib/components/form-group/form-group.less) in as a wrapper for label + any other input/textarea/description/etc. This would allow for tighter control and would lift any contextual styles up out of the component where they might result in confusing spacing for the consumer.

I whipped up [this PR](https://github.com/StackExchange/Stacks/pull/2165) to demonstrate what I think might work for us in this regard. Let me know what you think about that approach.

dancormier (changes requested):
@giamir thanks for tackling the label. Everything besides spacing looks good to me. I left a comment on your question with a link to a PR that I think might be helpful. Feel free to take what is helpful and leave the rest. And let me know if you want to discuss this or pair up.

CGuindon:
New spacing looks good except on the description example. I can see there's an 8px gutter that's getting ignored in between the title and the description (but applied below the description). Which means it looks like there is 0 spacing in between title and description — that should be 4px.

Label title
[4px - missing]
Label description
[8px - good]
Input

Screenshot 2026-02-04 at 6 59 34 AM

dancormier:
> New spacing looks good except on the description example. I can see there's an 8px gutter that's getting ignored in between the title and the description (but applied below the description). Which means it looks like there is 0 spacing in between title and description — that should be 4px.
> 
> Label title [4px - missing] Label description [8px - good] Input
> Screenshot 2026-02-04 at 6 59 34 AM

@CGuindon The gutter is 8px but the description has a top margin of -4px to achieve a 4px gap between the label and description

CGuindon (approved):

PR #2164, opened 2026-02-03
feat(link): Update link component to SHINE designs

ttaylor-stack:
[SPARK-70](https://stackoverflow.atlassian.net/browse/SPARK-70)

[Figma](https://www.figma.com/design/do4Ug0Yws8xCfRjHe9cJfZ/Project-SHINE---Product-UI?node-id=610-18808&p=f&m=dev)

Along with updating the link component I also:

- Removed a tag styling from s-link
- Updated s-anchors to fix a11y failing tests
- Updated Link Svelte component to Svelte 5
- Updated Link stories to match the formatting/styling we use throughout the rest of the Storybook

Questions/Concerns:

- @CGuindon do we need to force visited styling any more (using __visited)? If not can we remove it from the from Stacks?


CGuindon (changes requested):
+ 1 to Giarmir's comments about the underlines:

> Looks like all the navigation elements now are underlined. Probably something we don’t want I imagine. Regular `a`  tag also now get a strange blue color which is probably something we don’t want.

**New request** (I've updated the Figma)
I have two small changes I'd like to make (sorry) to the original spec since we can't change the bold to unbold for visited (security reasons from the browser).
- Default visited link should be `black-400` now (used to be 500)
- Muted visited link should be `black-400` now as well (used to be 500 too, muted visited won't have a visual difference anymore. This isn't used in as many places so I'd like to test it and see how important the visited state really is for these types of links).

dancormier:
@ttaylor-stack I've removed the `__visited` modifier ([16ba78f](https://github.com/StackExchange/Stacks/pull/2164/commits/16ba78f07251f0d3496f9221ff11bd18cbf5dda7)) and have added styles to handle anchors on a per-component basis ([7e7bb5d](https://github.com/StackExchange/Stacks/pull/2164/commits/7e7bb5d7ce798c31606f98709f83990ffb85e0e0)).

dancormier:
@CGuindon I updated the colors and have this now:

### Default
image

### Hover
image

### Visited
image

Does this look correct?

CGuindon:
@dancormier Colors look correct — ~~is it possible to get the underline on hover (for all of them) as well? If that over complicates I can live without out.~~ Seeing the underline in the docs

CGuindon:
Links page looks good, I'm seeing extra underlines in a few places where they shouldn't be like:
- Menus
- Breadcrumbs (although we haven't changed that component yet)

Screenshot 2026-02-03 at 4 18 04 PM
Screenshot 2026-02-03 at 4 17 24 PM

CGuindon (approved):

dancormier:
> Links page looks good, I'm seeing extra underlines in a few places where they shouldn't be like:
> 
>     * Menus
> 
>     * Breadcrumbs (although we haven't changed that component yet)
> 
> 
> Screenshot 2026-02-03 at 4 18 04 PM Screenshot 2026-02-03 at 4 17 24 PM

@CGuindon was that in high contrast mode or standard light mode? If it's HC, all anchors should include underlines (which seems to be the case in the menu since the non-underlined elements are buttons)

dancormier (approved):
🔗👍

CGuindon:
It was in regular mode originally but looks fixed now @dancormier so all good.

PR #2178, opened 2026-02-06
fix(avatars): Use black background in dark mode for avatars

mukunku:
# Summary

Our avatars are hardcoded to use a white background which doesn't look great in dark mode when used with transparent images. This PR makes the background color dynamic (white/black) based on the user's selected theme.

## Considerations

* I believe the main reason we have a background color is to frame the avatar and make its boundaries distinct in case of a different background color. Preventing any transparent avatars from blending with the background. Examples:
   * image
   * image

* Another thing to consider is the common practice of images that aren't transparent originally and were aliased against a white background being made transparent after the fact. With the background color we are applying there isn't much we can do about this: users will need to use proper transparent PNG images if they want it to look good.
   * Example: image
   * If we were to ditch the background color and let the image use its parent's background color we **can** utilize `drop-shadow` to try soften the rough edges a bit but I feel it's a hack at best as the color needs to be set for the shadow and it might not look right in all cases.
* I also briefly looked into `mix-blend-mode` but doesn't seem like that's what we are looking for.

# How to Test

* Quickly refresh the page in dark mode and confirm there is no "flash of white" anymore before images are loaded: https://deploy-preview-2178--stacks.netlify.app/product/components/avatars/#users


giamir (approved):
Thanks @mukunku for the research. I can get behind this choice. I cannot think honestly of better ways to go about it. All have compromises.

CGuindon (approved):
Ok let's do it!

dancormier (approved):

PR #2181, opened 2026-02-12
fix(navigation): make vertical navigation more accessible

mukunku:
# Summary

This PR suggests some small but breaking changes to the markup so we can have an accessible vertical navigation component.

**Before:**
```html
<nav>
	<ul class="s-navigation s-navigation__vertical">
		<li class="s-navigation--title">Resources</li>
		<li><a href="#" class="s-navigation--item is-selected">Icons</a></li>
		<li><a href="#" class="s-navigation--item">Spot illustrations</a></li>
	</ul>
</nav>
```
**After:**
```html
<nav>
	<ul class="s-navigation s-navigation__vertical">
		<li>
			<h4 class="s-navigation--title" id="nav-section1">Resources</h4>
			<ul aria-labelledby="nav-section1">
				<li><a href="#" class="s-navigation--item is-selected">Icons</a></li>
				<li><a href="#" class="s-navigation--item">Spot illustrations</a></li>
			</ul>
		</li>
	</ul>
</nav>
```
## Motivation
This PR was created because the current beta Navigation Svelte component, when in vertical mode, has some accessibility issues around its usage of `role="separator"`:
* The different navigation group titles are currently `<li>` items themselves which causes screen readers to incorrectly count how many navigation items are present.
* There is no semantic markup to represent the implied parent-child relationship between different groups
* When this role is assigned to an element all its children [implicitly receive the `presentation` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/separator_role#all_descendants_are_presentational) which isn't something we necessarily want in our navigation component. 

## Rationale

[This question](https://stackoverflow.com/questions/69021634/the-correct-way-to-add-separators-in-lists) does a good breakdown of why I went with breaking the different navigation groups into actual separate `<ul>` elements. This does mean I had to make some assumptions around nested `<ul>` elements in the `.less` code but I'm hoping this is an acceptable trade-off to have an accessible component. 

Additionally, I used an `<h4>` (configurable in svelte) element as the group title as that is the [recommended approach for screen readers](https://ux.stackexchange.com/questions/59947/what-is-the-most-usable-navigation-mark-up-for-screen-readers).

# How to Test

## Stacks Docs

The `Titles` variant is what has changed.

* Confirm there are no visual differences between the previous version and this PR:
   * Current: https://beta.stackoverflow.design/product/components/navigation/#titles
   * This PR: https://deploy-preview-2181--stacks.netlify.app/product/components/navigation/#titles
* Notice the new markup contains multiple `<ul>` elements for the different navigation groups and each list contains the various navigation items now.

## Stacks Svelte

* Confirm there are no visual differences between the previous version and this PR:
   * Current: https://beta.svelte.stackoverflow.design/?path=/story/components-navigation--titles
   * This PR: https://deploy-preview-2181--stacks-svelte.netlify.app/?path=/story/components-navigation--titles
* Notice that `NavigationTitle` has been renamed to `NavigationGroup` and now encapsulates its child `NavigationItem`s.


packages/stacks-docs/_includes/layouts/home.html
```diff
@@ -202,11 +202,11 @@ <h1 class="fw-bold mln2 mb0 mtn4 fs-display1 lh-xs">Stacks</h1>
     <div class="bg-black-100 ai-center d-flex fd-column jc-center w100">
         <nav class="d-grid grid__2 fs-body2 g16 px64 py32 w100 wmx12 lg:px32 md:py24 sm:grid__1 sm:g8 sm:p24" aria-label="Global">
             {% for card in home.cards %}
-                <a class="s-card ai-center d-flex g32 p16" href="{{ card.url }}">
+                <a class="s-card s-link ai-center d-flex g32 p16" href="{{ card.url }}">
```
dancormier, line 205: Thank you! I've been bothered by the ugly browser-default blue/purple each time I hit the docs.

dancormier (approved):
Really nice work here @mukunku! Thanks for this fix. I like the API you landed on with `NavigationGroup`. Makes a lot of sense to me.

:shipit:

PR #2182, opened 2026-02-13
Remove & redirect outdated Content & Brand elements

abovedave:
As [the new docs site](https://alpha.stackoverflow.design/) is taking shape this PR removes and redirects older content.

Screenshot 2026-02-13 at 15 34 42


giamir (commented):
I think we can merge this, it would probably be nice to actually adjust also the beta branch accordingly after this is merged.

cc @dancormier

PR #2187, opened 2026-02-19
chore(components): remove deprecated components

dancormier:
This PR:
- Removes deprecated components from our CSS bundle
- Adds a deprecation notice to the top of the docs pages for deprecated components with a link to v2 docs
- Hides deprecated component docs page content

### Screenshot
image

> [!NOTE]
> I left the right rail table of contents in tact for these pages since it seemed like a bit of a lift to hide them without causing issues on rendering them on non-deprecated component pages when navigating around.

> [!NOTE]
> I've left the visual regression test images in place but can remove them if you feel it's necessary @giamir. Just let me know


packages/stacks-classic/lib/stacks-static.less
giamir, line 1: This is the part I care the most about.

giamir (approved):
Thanks @dancormier. 

This looks like a solid start! To keep this PR focused and cut a new release for our beta env, let’s handle the documentation updates and the removal of dead code and tests in a dedicated follow-up.

I’m a big advocate for keeping the codebase lean. We should feel confident deleting cruft knowing that git history is our safety net if we ever need to reference or restore that logic later.

NIT: On the empty pages, could we update the warning text from 'deprecated' to 'removed'? Since these components are no longer functional in v3, 'removed' gives our users more clarity on the current state of the API.

NIT: The uploader component has a stacks.js controller that should be removed/deleted from the bundle.

CGuindon (changes requested):
Visually work for me! I'm not sure if this was intentional but the link to the v2 docs in the yellow notice from any of the deprecated components leads to [Breadcrumbs](https://stackoverflow.design/product/components/breadcrumbs/). We could just default to the product landing page instead if we syncing up each component to the right page is too much work. (Card should point to v2 card page, Page title should point to v2 page title...)

dancormier:
Nice catch @CGuindon! I've updated the link and I'm gonna go ahead and merge so we can cut a new release first thing Monday.

@giamir early next week, I'll put together a follow up PR removing all the cruft.

Thanks for the reviews!

PR #2191, opened 2026-02-24
chore: remove cruft left over from removing components

dancormier:
This PR removes cruft left behind when removing components in https://github.com/StackExchange/Stacks/pull/2187.

## Notable changes
- I decided to port styles from the Card and Topbar components in our docs-specific Less as `.docs-card` and `.docs-topbar` so we can fully remove those components
- I gave the topbar a fresh coat of paint to a) more closely match our beta site and b) simplify the styles a little
- The `s-anchor` visual regression tests used `.s-card` previously. I replaced `.s-card` there with atomic classes that result in a slightly different style than previous, requiring those test images to be regenerated


giamir (commented):
Thanks @dancormier. Changes look good to me. 

I noticed 2 minor issues

Icons in the docs topbar when in dark mode are not visible anymore

Image

For the pages using the new "removed" layout I wonder if we can get rid of the irrelevant table of content or at least leave it empty rather than showing wrong info.

Image

dancormier:
@giamir Thanks for the review. I made your suggested changes and _I think_ this should be good to go.

**Sidenote**: I'm very happy this repo is open source so I can still get _some_ work done on my personal machine while I wait for a replacement work laptop 🙂

giamir (approved):
Looks good. Thanks @dancormier 🚀

PR #2193, opened 2026-02-25
fix(expandable): expose expandable controller for legacy UI

dancormier:
This PR includes the expandable controller in the JS bundle for backwards compatibility. See https://github.com/StackEng/StackOverflow.UIComposer/pull/149 for more context.


giamir:
Suggesting to close in favor of https://github.com/StackEng/StackOverflow/pull/24975

PR #2206, opened 2026-03-04
`<Avatar>` dimensions and screen reader text

### LINKED ISSUE #2205: `<Avatar>` Svelte component img should have width and height defined
The `` tag has zero width and height causing a brief layout shift before the image loads.

https://github.com/StackExchange/Stacks/blob/dc19e487ac9f06ee00d751f3c58ec4c4369de4f0/packages/stacks-svelte/src/components/Avatar/Avatar.svelte#L92

abovedave:
* Closes #2205 to avoid the layout shift before an image loads.
* Begins to address A11Y-126 by using the title prop i.e.,`<Icon title="Private">`, which removes the `aria-hidden="true"`.

### DIFF
--- a/.changeset/strong-news-float.md
+++ b/.changeset/strong-news-float.md
@@ -1,0 +1,6 @@
+---
+"@stackoverflow/stacks-svelte": patch
+---
+
+* `<Avatar>` nested `<img>` dimensions.
+* Screen reader text for the private badge variant.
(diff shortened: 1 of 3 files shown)


giamir (commented):
@abovedave We should update the related unit tests to account for the changes you made. I also would not want to hardcode the word "Private" in the component. We use the `i18nxxx` prop pattern to allow consumers to override default values.  Apart from that looks good.  Thanks for the PR. ❤️

dancormier (approved):
@abovedave this PR looks good! I added a few tests for these changes and I think it should be all good to merge

PR #2240, opened 2026-04-21
feat(docs-next): Collapsable table of contents

abovedave:
[STACKS-869](https://stackoverflow.atlassian.net/browse/STACKS-869)

Adds back a more refined table of contents treatment for smaller screens.

Screenshot 2026-04-22 at 17 35 32


packages/stacks-docs-next/src/components/Contents.svelte
```diff
@@ -178,4 +192,13 @@
     backdrop-filter: invert(1);
     -webkit-backdrop-filter: invert(1);
   }
+  @media (max-width: 71.875rem) {
```
abovedave, line 195: There's no CSS var for `md` breakpoint, right?

dancormier: That's right, CSS vars can't be used in media query conditions

packages/stacks-docs-next/src/components/Contents.svelte
```diff
@@ -167,15 +162,27 @@
...
 
 <style>
-  .contents-indicator {
-    background-color: var(--black-100);
-    /* background-color: rgba(255, 255, 255, 0.1); */
-    backdrop-filter: invert(1);
-    -webkit-backdrop-filter: invert(1);
+  @media (max-width: 71.875rem) {
+    .contents-inner-mobile {
+        max-height: calc(100vh - 200px);
```
dancormier: If possible, I'd suggest we used a sizing unit custom property here (`--su*`) since those custom properties use `rem` which is more accessible and will allow the sizing to scale organically based on user preferences
```suggestion
        max-height: calc(100vh - var(--su256));
```

dancormier (dismissed):
Thanks for this @abovedave

I saw some opportunity for cleanup so I made some changes to split out the mobile and standard contents to prevent most of the component-specific styles. I also simplified the scrolling logic and made some accessibility improvements, plus I resolved the merge conflicts from the `stacks-docs-next` to `stacks-docs` [shuffle](https://github.com/StackExchange/Stacks/pull/2250).

Gone mergin'!

dancormier (approved):

PR #2245, opened 2026-04-23
fix(docs-next): External url paths in sidebar are truncated

abovedave:
Currently they are passed to [`resolve()`](https://svelte.dev/docs/kit/$app-paths#resolve) which removes the first char of the string and prefixes with the base url.

Screenshot 2026-04-24 at 12 06 35


dancormier (dismissed):
Looking good @abovedave I pushed a commit with some minor formatting changes and a change to ensure the `rel` value syncs up with the `href` value on these items.

It's good to go! 🚀

dancormier (approved):

PR #2248, opened 2026-04-28
ci: gate visual regression tests by path

dancormier:
## Summary

Gates the visual regression suite so it only runs when visual-test-relevant files change. It also updates our workflows to use node v24 (via [actions/setup-node@v5](https://github.com/actions/setup-node/releases/tag/v5.0.0))

## Related Issue

[STACKS-887](https://stackoverflow.atlassian.net/browse/STACKS-887)

## Changes

- Adds a `dorny/paths-filter` check for visual regression relevant paths.
- Moves visual regression tests into a dedicated gated workflow job.
- Preserves release behavior when the visual regression job is intentionally skipped.

#### Verify visual regression tests complete successfully when changes are present in `packages/stacks-classic`

- Change to `packages/stacks-classic/README.md`: https://github.com/StackExchange/Stacks/pull/2248/commits/bd7620330a05afec11a78d08b41076478ac7d7b5
- Commit with completed workflow that includes visual regression test run: https://github.com/StackExchange/Stacks/pull/2248/commits/b69ece2fbe1eda4adb3e785cc92d2197eec58fab
- [Successful visual regression test run](https://github.com/StackExchange/Stacks/actions/runs/25127941398/job/73646167242)
  - **Note**: it runs "Detect changes", which reports that there are changes to Stacks Classic, with the visual regression tests running after

#### Verify visual regression tests are skipped when changes are not present in `packages/stacks-classic`

- Commit reverting changes to `packages/stacks-classic/README.md`: https://github.com/StackExchange/Stacks/pull/2248/commits/ec365b478a9ac21907dc1f6909a3109663300f42
- [Skipped visual regression test run](https://github.com/StackExchange/Stacks/actions/runs/25128448500/job/73647910209)

### Manually testing

- In this PR, push a small change to a file within `packages/stacks-classic` (I suggest adding a comment to `README.md`
- Verify the visual regression job runs and completes successfully
- Revert your change and verify that the visual regression job is skipped

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,17 +7,6 @@ on:
     branches: [develop, beta, main]
 
 jobs:
-  build:
-    name: Build
-    uses: ./.github/workflows/test.yml
-    with:
-      command: npm run build
-      command_description: Build
-      needs_playwright: false
-      needs_lfs: false
-      detect_visual_regression_changes: true
-    secrets: inherit
-
   build-and-test:
     name: ${{ matrix.command_description }}
     uses: ./.github/workflows/test.yml
@@ -25,6 +14,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
+        - command_description: Build
+          command: npm run build
+          needs_lfs: false
+          needs_playwright: false
         - command_description: Lint
           command: npm run lint
           needs_lfs: false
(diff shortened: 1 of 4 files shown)


giamir (commented):
If we go this path, would it make more sense to group tests based on the packages that have changed? Accessibility tests at the moment are also only relevant for stacks classic, etc...

I wonder if there is a monorepo tool that can help us to do the heavy lifting here instead of going adhoc. `pnpm --filter "[origin/main]" test`  https://pnpm.io/filtering#--filter-since

It feels a lot of extra code that we are adding here to do something rather simple. Maybe it is that time where we finally switch to pnpm or any other serious monorepo tool  :D

PR #2250, opened 2026-05-01
chore(docs): change stacks-docs-next to stacks-docs

ttaylor-stack:
## Summary
- Replaces the legacy docs package with the SvelteKit docs app from `stacks-docs-next`.
- Removes obsolete legacy docs dependencies, generated outputs, ignore rules, and old release/deploy workflow references.
- Updates docs tooling so lint includes Svelte checking and clean CI builds `@stackoverflow/stacks-svelte` before checking docs.
- Keeps private docs out of git and clones them during local/deploy setup when credentials are available.
- Simplifies `/llms.txt` generation and removes the old legacy docs fetch.

## Testing
- `mise x node@lts -- npm run lint -w packages/stacks-docs`


dancormier (approved):
Thanks for spinning this up @ttaylor-stack.

Follow-up changes made during review:
- Removed obsolete legacy docs dependencies, generated outputs, old ignore rules, and stale docs guidance.
- Cleaned up package scripts so docs lint now includes `svelte-check`.
- Fixed clean-CI docs checking by building `@stackoverflow/stacks-svelte` before running docs checks.
- Removed the old private-docs submodule setup and replaced it with an ignored `src/docs/private` clone step for local/deploy builds.
- Removed `.gitmodules` because the repo no longer tracks private docs as a git submodule.
- Simplified `/llms.txt` generation by using `turndown` and removing the legacy docs fetch.
- Updated local docs instructions and release/deploy workflow references from the old `develop`/`production` flow to the current `main`/`beta` flow.
- Updated auth config so Okta redirects use `https://stackoverflow.design` instead of the retired alpha subdomain. (see also https://github.com/StackEng/okta-terraform/pull/1229)

PR #2262, opened 2026-05-01
patch(svelte): update components to accept objects and arrays

dancormier:
## Summary
- Recreates #2213 against `main` now that `beta` has been deleted.
- Updates Stacks Svelte component `class` props to support object and array class syntax.
- Adds Storybook coverage/examples for object and array class values across affected components.

## Original PR
- #2213
- STACKS-799: https://stackoverflow.atlassian.net/browse/STACKS-799


ttaylor-stack (approved):
LGTM, pushed a small fix to the control in the navigation and textarea storybook controls

PR #2272, opened 2026-05-19
fix(docs): Responsive & dark mode adjustments

abovedave:
[STACKS-885](https://stackoverflow.atlassian.net/browse/STACKS-885)

Addresses some minor style issues when using a narrow screen.


dancormier (approved):
Thanks for these fixes @abovedave!

PR #2273, opened 2026-05-20
fix(docs): point v2 site URLs at v2.stackoverflow.design

asblanco:
## Summary
- The v2 docs site is served from `v2.stackoverflow.design`, but the eleventy-plugin-llms-txt `siteUrl` and `_data/site.json` `url` were both set to the v3 host. As a result, `v2.stackoverflow.design/llms.txt` emitted v3-domain links for every page and `og:url` meta tags pointed off-site.
- Update both values to `https://v2.stackoverflow.design` so generated links and OG metadata match the actual host.

## Test plan
- [ ] Build the v2 docs and confirm `/llms.txt` header reads `# Site URL: https://v2.stackoverflow.design` and every `URL:` line uses the v2 host.
- [ ] Inspect any page's `<head>` and confirm `<meta property="og:url" content="https://v2.stackoverflow.design">`.
- [ ] Post-deploy: `curl https://v2.stackoverflow.design/llms.txt | head` shows the v2 host.

JIRA: [STACKS-864](https://stackoverflow.atlassian.net/browse/STACKS-864)

### DIFF
--- a/packages/stacks-docs/.eleventy.js
+++ b/packages/stacks-docs/.eleventy.js
@@ -19,7 +19,7 @@ module.exports = function(eleventyConfig) {
   eleventyConfig.addPlugin(headerPlugin);
   eleventyConfig.addPlugin(tipPlugin);
   eleventyConfig.addPlugin(llmsTxtPlugin, {
-    siteUrl: 'https://stackoverflow.design',
+    siteUrl: 'https://v2.stackoverflow.design',
     collections: ['base', 'components', 'develop', 'foundation'],
     additionalMetadata: ['description'],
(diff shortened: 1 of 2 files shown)


dancormier (approved):
This change looks good! Thank you @asblanco!

dancormier:
Turns out the v2 workflows need to be updated to run successfully, so we're stuck with this for now:

image


I've updated [STACKS-854](https://stackoverflow.atlassian.net/browse/STACKS-854) to ensure all workflows run successfully. I'll go ahead and bypass the checks here since I feel this change is safe.

PR #2275, opened 2026-05-20
fix(docs): resolve accessibility report issues

dancormier:
## Summary

- Fixes a few Stacks docs accessibility report issues found at https://accessibility.stackoverflow.net/products/stacks/automated
- Adds accessible names/roles for affected examples and improves docs-only contrast on some text and links.
- Wraps the toggle button demo in the standard example container (not strictly accessibility but just cleanup)

## Testing

- Load up the deploy preview
- Ensure `N/A` in class tables on our component docs pages (such as [Avatars](https://deploy-preview-2275--stacks.netlify.app/system/components/avatars)) meet our contrast minimums
- Ensure page description (below title) meets our contrast minimums
- Ensure links within docs page text content (such as the Popper.js link in the [Popover docs](https://deploy-preview-2275--stacks.netlify.app/system/components/popovers#interactive-popovers)) meet our contrast minimums
- Ensure each scrollable [overflow](https://deploy-preview-2275--stacks.netlify.app/system/base/overflow) atomic class example can be tabbed to and scrolled with arrow keys as appropriate (not all will be tabbable/scrollable)


packages/stacks-docs/src/app.css
```diff
@@ -478,6 +478,7 @@ h1 {
     list-style: none;
 }
 .docs-link {
+    color: var(--theme-link-color, var(--black-600));
```
asblanco, line 482: Generated markdown prose links still seem to inherit the lower-contrast paragraph color. `addDocClasses` gives paragraphs `docs-copy` and anchors `docs-link`, and the existing `.docs-copy .docs-link { color: inherit; }` rule is more specific than this new `.docs-link` rule. That means links like the Popper.js example stay on the inherited paragraph color instead of using `--theme-link-color` / `--black-600`.

Could we either remove/update the earlier inherited-color rule, or include the prose selector here so it wins?

```suggestion
.docs-copy .docs-link,
.docs-link {
    color: var(--theme-link-color, var(--black-600));
    font-weight: 500;
}
```

asblanco (changes requested):

asblanco:
I’ve updated the branch with `origin/main` and resolved the resulting conflicts. The previous review feedback is now addressed.

The accessibility fixes include:

- Documentation links had insufficient contrast because their CSS rule was incorrectly nested. The selector is now defined at the correct level, restoring the intended contrast.
- Selected anchor examples used button semantics and an invalid `aria-pressed` attribute. These attributes were removed so the examples retain valid link semantics.
- The regression tests inspected CSS selectors and source markup instead of behavior. They were replaced with a rendered-component test verifying that selected anchors remain links without invalid ARIA attributes.

The focused Stacks Svelte and documentation test suites pass.

asblanco (approved):

PR #2292, opened 2026-06-01
Add email docs and `stacks-email` package

abovedave:
[STACKS-905](https://stackoverflow.atlassian.net/browse/STACKS-905)

Introduces `@stackoverflow/stacks-email`, a new (experimental) [MJML](https://mjml.io/) powered email package for building, previewing, documenting, and compiling Stack Overflow email templates.

- A reusable email component library
- A template gallery for complete email examples
- A shared MJML compile pipeline
- Target-specific token replacement for preview, .NET/Razor, and Braze
- A flexible Node/TypeScript package API
- A narrower SvelteKit HTTP compile endpoint for service integration
- Static compiled artifacts for use in `stacks-docs`

## Background

Historically, Stack Overflow's email design system lived across several separate implementations, including the monolith and external marketing tools. Since 2020, the system has had limited modernization or maintenance.

This work is part of the 2026 rebrand rollout. The goal is to create a shared source of truth for email design patterns, implementation, previewing, and downstream consumption.

We selected MJML because it is widely adopted, MIT licensed, and abstracts much of the client-specific complexity of writing production email HTML.

### Components

`~/packages/stacks-email/components`

Adds reusable email building blocks such as `Button`, `Header`, `Footer`, `Card` etc. Each defines their variants, options, tokens, and MJML render output in one place.

### Templates

`~/packages/stacks-email/templates`

Adds example and reusable template definitions including `Transactional`, `Newsletter`, `Promotional`. These compose components into full MJML documents and can expose variants and props.

###  Compile APIs

The package can be consumed in several ways:

- As a local/gallery app for previewing emails (this branch is currently live on [email.stackoverflow.com](https://email.stackoverflow.design/))
- As a Node/TypeScript library via `@stackoverflow/stacks-email`
- As a deployed SvelteKit endpoint via `POST /api/compile`
- As static compiled artifacts for docs previews
- As raw MJML or compiled HTML output

The package API and HTTP API intentionally serve different use cases.

The package API is the full in-process integration surface for trusted TypeScript consumers. It supports catalog discovery, component compilation, template compilation, renderable dispatch, and lower-level MJML compilation primitives.

The HTTP endpoint is a narrower service integration surface. It currently focuses on composing transactional emails from a validated JSON block list, then returning compiled MJML and HTML. This keeps the cross-service contract small while still enabling non-TypeScript consumers to generate branded email output without embedding the package.

###  Documentation

Adds a new Email section to `stacks-docs`, including:

- Email overview
- Template documentation
- Component documentation
- Embedded email previews
- MJML/HTML output tabs
- Usage examples
- Generated options tables

### As a package

```ts
import {
    getEmailCatalog,
    compileEmailTemplate,
    compileEmailComponent,
    compileEmailRenderable,
} from "@stackoverflow/stacks-email";

const email = compileEmailTemplate({
    slug: "transactional",
    target: "braze",
    props: {
        headlineText: "Reset your password",
        ctaText: "Reset password",
    },
});

console.log(email.html);
```

### As an endpoint

```bash
curl -X POST https://email.stackoverflow.design/api/compile \
  -H "Content-Type: application/json" \
  --data '{
    "template": "transactional",
    "target": "preview",
    "previewText": "Reset your password",
    "blocks": [
      {
        "type": "headline",
        "props": {
          "textContent": "Reset your password"
        }
      },
      {
        "type": "text",
        "props": {
          "textContent": "Click below to continue."
        }
      },
      {
        "type": "button",
        "props": {
          "href": "[[BUTTON_URL]]",
          "text": "Reset password"
        }
      }
    ]
  }'
```

### As static docs artifacts

`@stackoverflow/stacks-email/sveltekit` exports route handlers used by `stacks-docs` to serve precompiled email HTML, MJML, and manifest data.

## Compile targets

Every compile supports a target:

- `preview`: replaces tokens with sample values
- `dotnet`: replaces tokens with Razor model bindings
- `braze`: replaces tokens with Braze/Liquid-style values

This allows the same email source to produce output for different downstream systems.

## Notes for reviewers

Suggested review path:

1. Start with `packages/stacks-email/README.md`
2. Review the component structure in `packages/stacks-email/components`
3. Review template composition in `packages/stacks-email/templates`
4. Review the compile pipeline in `packages/stacks-email/src/lib/pipeline`
5. Review public API exports in `packages/stacks-email/src/lib/api`
6. Review the HTTP endpoint in `packages/stacks-email/src/routes/api/compile/+server.ts`
7. Review docs integration in `packages/stacks-docs/src/components/StacksEmailEmbed.svelte`
8. Review static artifact handlers under `packages/stacks-docs/src/routes/email/compiled`

## Current limitations / Follow-ups

- This is mostly experimental but wanted to at least get the design elements and static markup available for consumption.
- Open to ideas for a testing strategy for a follow-up PR.
- The endpoint is intentionally narrower than the package API.
- The package compiles email markup; it does not send emails.
- Real inbox/client testing is still required before using output in production sends (Litmus?).
- Additional templates, variants, and target mappings can be added as rollout needs become clearer.
- Production hosting and ownership model should be finalized separately from this package implementation (hi platform team!)
- Add [Figma Code Connect](https://developers.figma.com/docs/code-connect/)
- Sync with tools e.g., [Braze API](https://www.braze.com/docs/api/endpoints/templates/content_blocks_templates/post_create_email_content_block)


packages/stacks-docs/src/routes/+page.svelte
```diff
@@ -49,29 +65,24 @@
...
-    >
-      <Icon src={IconArrowUpRightBox} />
-      <h2 class="mt-auto fs-headline1 lh-sm mb12">Looking for the previous version of Stacks?</h2>
-      <p class="mb0 fs-body2 fc-black-400">If you are working on a classic experience you can visit the v2 documentation.</p>
-    </a>
   </div>
+
+  <Notice variant="info" class="s-anchors s-anchors__inherit s-anchors__underlined mt32">
+      <p class="mb0">Looking for the previous version of Stacks? <a href="https://v2.stackoverflow.design/product/develop/using-stacks/">Go to the classic v2 documentation</a>.</p>
+  </Notice>
```
dancormier, line 87: Looks much nicer this way

packages/stacks-docs/src/app.css
```diff
@@ -307,7 +307,7 @@ body {
 
 .ff-stack-sans-headline,
 .ff-stack-sans-headline-notch {
-    font-family: "Stack Sans Headline";
+    font-family: "Stack Sans Headline" !important;
```
dancormier: Was the heading font being overridden by something? I'm not necessarily against adding `!important` here but just curious about the issue.

packages/stacks-email/src/lib/tokens.ts
```diff
@@ -0,0 +1,195 @@
...
+            FOOTER_REASON: "you subscribed to Stack Overflow updates.",
+            UNSUBSCRIBE_URL: "https://example.com/unsubscribe",
+            COMPANY_NAME: "Acme Corp",
+        },
+    },
+    dotnet: {
+        tokens: {
+            FIRST_NAME: "@Model.FirstName",
+            BUTTON_LABEL: "@Model.ButtonLabel",
+            BUTTON_URL: "@Model.ButtonText",
```
dancormier: Should `BUTTON_URL` use `@Model.ButtonText`? Seems like `@Model.ButtonText` would be the expected text content of the button.

packages/stacks-docs/src/components/StacksEmailEmbed.svelte
```diff
@@ -0,0 +1,461 @@
...
+                    text={tab.label}
+                    selected={activeTab === tab.id}
+                    onclick={() => (activeTab = tab.id)}
+                />
+            {/each}
+        </Navigation>
+
+        <Select
+            id={`email-target-${kind}-${slug}`}
+            label={null}
```
dancormier: Nit: this select should have an accessible name

packages/stacks-email/components/header.ts
```diff
@@ -0,0 +1,101 @@
...
+        defineOption({
+            name: "sectionClass",
+            type: "string",
+            initialValue: "bg-block",
+            description: "Header section mj-class override.",
+        }),
+        defineOption({
+            name: "logoSrc",
+            type: "string",
+            initialValue: "/email/stack-overflow-logo.png",
```
dancormier, line 41: The README describes email.html as “ready-to-send HTML”, but the default image sources are root-relative `/email/…` URLs. These images would likely break once sent outside the preview host. 

Should we make asset URLs absolute/configurable or clarify that consumers must rewrite/package image assets before sending?

packages/stacks-email/templates/transactional.ts
```diff
@@ -0,0 +1,153 @@
...
+                            textContent: props.headlineText,
+                        }),
+                        ...(graphicPath.length > 0
+                            ? [Graphic("spot", { imageSrc: graphicPath })]
+                            : []),
+                        Text("body", { textContent: bodyContent }),
+                        Section(
+                            [
+                                Button("primary", {
+                                    href: "[[CTA_URL]]",
```
dancormier: `[[CTA_URL]]` doesn’t appear in the compile target token maps, so this `href` will remain literal in preview/.NET/Braze output. Can we either use the existing `[[BUTTON_URL]]` token here or add `CTA_URL` replacements for every target in `src/lib/tokens.ts`?

packages/stacks-docs/src/structure.yaml
```diff
@@ -366,6 +411,10 @@ navigation:
             slug: "icons"
             image: "/images/heros/product.svg"
 
+          - title: "Email gallery"
+            slug: "emails"
+            externalUrl: https://email.stackoverflow.design
+
```
dancormier, line 423: Looks like we could use an image here

Image

dancormier (changes requested):
Really nice work here @abovedave. Adding `stacks-email` seems reasonable and I think this is looking pretty solid. I noticed a few bugs that I think shouldn't be too tough to handle

> [!NOTE]
> I noticed a bunch of whitespace-only changes to a couple of `yaml` files and undid just the whitespace changes to reduce the noise.

packages/stacks-docs/src/routes/+page.svelte
abovedave: Thanks. Note to self I'll need to update this again for #2299 as that adds an extra block here which will need addressing in the layout.

packages/stacks-docs/src/components/StacksEmailEmbed.svelte
abovedave: Updated! I missed the `hideLabel` setting is available

packages/stacks-docs/src/app.css
abovedave: Good catch - reverted.

I think it was the original way I was constructing these cards they were not inheriting the classes but looks fixed with the way I ended up doing them.

Screenshot 2026-07-14 at 12 52 00

packages/stacks-email/components/header.ts
abovedave: Ah yes, **great** point. I've added a concept of a `assetBaseUrl` which defaults to https://email.stackoverflow.design but can be configured at the server and consumption layers.

I might need to amend this once I get into how tools like Braze ingest these external urls (or say, want assets POST'd).

abovedave: https://github.com/StackExchange/Stacks/pull/2292/commits/f19c383647d831b2d4b4f0155c652534474c34d9

dancormier (approved):
I pushed a few follow-up fixes:
- Made `@stackoverflow/stacks-email` publishable outside the repo, including a changeset
- Safely validated custom asset URLs
- Fixed component props being silently ignored
- Made gallery MJML reflect the selected target
- Added regression and catalog-wide tests, now run in CI

The package, docs build, type checks, lint, tests, and external-consumer smoke test all pass.

@abovedave when you have a moment, can you take a moment to verify that these changes match the intention of the PR (especially the publishing of `@stackoverflow/stacks-email` and the gallery MJML change)? If it all looks good, feel free to merge!

PR #2297, opened 2026-06-15
fix(popover): Fix popover focus-leave dismissal

alexwarren:
Stacks fixes to address A11Y-160

## Summary
- Close classic `s-popover` when keyboard focus leaves both the trigger and popover, but only for popover content marked with `role="menu"`.
- Apply the same role-derived focus-leave behavior to Svelte Popover while preserving tooltip, non-menu, and controlled behavior.
- Add regression tests for aria-expanded, menu focus transitions, outside focus, null relatedTarget, controlled close requests, and non-menu popovers staying open.

## Why
A11Y-160 reports that dropdown/popover menus remain visually and programmatically expanded after users tab away. This keeps `aria-expanded` true even though focus has moved outside the trigger/popover pair.

Stacks popovers remain a low-level primitive, so the new dismissal behavior is derived from menu semantics instead of applying to dialogs or generic popover content.

## Validation
- `npm run test:unit` in `packages/stacks-classic`: 30 passing across Chromium, Firefox, and WebKit.
- `npx web-test-runner --config web-test-runner.config.mjs --files src/components/Popover/Popover.test.ts` in `packages/stacks-svelte`: 43 passing.
- `npx web-test-runner --config web-test-runner.config.mjs` in `packages/stacks-svelte`: 572 passing.
- `npx svelte-check --tsconfig ./tsconfig.json`: 0 errors/warnings.
- Package-scoped ESLint and Prettier checks passed.

### DIFF
--- a/.changeset/two-goats-press.md
+++ b/.changeset/two-goats-press.md
@@ -1,0 +1,6 @@
+---
+"@stackoverflow/stacks": minor
+"@stackoverflow/stacks-svelte": minor
+---
+
+Fix popover focus-leave dismissal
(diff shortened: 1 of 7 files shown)


alexwarren:
Replaced by #2298

PR #2298, opened 2026-06-15
fix(popover): Fix popover focus-leave dismissal

alexwarren:
Stacks fixes to address [A11Y-160](https://stackoverflow.atlassian.net/browse/A11Y-160)

## Summary
- Close classic `s-popover` when keyboard focus leaves both the trigger and popover, but only for popover content marked with `role="menu"`.
- Apply the same role-derived focus-leave behavior to Svelte Popover while preserving tooltip, non-menu, and controlled behavior.
- Add regression tests for aria-expanded, menu focus transitions, outside focus, null relatedTarget, controlled close requests, and non-menu popovers staying open.

## Why
A11Y-160 reports that dropdown/popover menus remain visually and programmatically expanded after users tab away. This keeps `aria-expanded` true even though focus has moved outside the trigger/popover pair.

Stacks popovers remain a low-level primitive, so the new dismissal behavior is derived from menu semantics instead of applying to dialogs or generic popover content.

## Validation
- `npm run test:unit` in `packages/stacks-classic`: 30 passing across Chromium, Firefox, and WebKit.
- `npx web-test-runner --config web-test-runner.config.mjs --files src/components/Popover/Popover.test.ts` in `packages/stacks-svelte`: 43 passing.
- `npx web-test-runner --config web-test-runner.config.mjs` in `packages/stacks-svelte`: 572 passing.
- `npx svelte-check --tsconfig ./tsconfig.json`: 0 errors/warnings.
- Package-scoped ESLint and Prettier checks passed.

### DIFF
--- a/packages/stacks-classic/lib/components/popover/popover.ts
+++ b/packages/stacks-classic/lib/components/popover/popover.ts
@@ -89,7 +89,11 @@ export abstract class BasePopoverController extends Stacks.StacksController {
      * Only menu popovers dismiss when focus leaves the reference/popover pair.
      */
     protected get shouldHideOnFocusLeave() {
-        return this.popoverElement?.getAttribute("role") === "menu";
+        return (
+            this.shouldHideOnOutsideClick &&
+            (this.popoverElement?.getAttribute("role") === "menu" ||
+                !!this.popoverElement?.querySelector('[role="menu"]'))
+        );
     }
 
     /**
--- a/packages/stacks-classic/lib/components/popover/popover.ts
+++ b/packages/stacks-classic/lib/components/popover/popover.ts
@@ -316,6 +316,7 @@ export class PopoverController extends BasePopoverController {
 
     private boundHideOnOutsideClick!: (event: MouseEvent) => void;
     private boundHideOnEscapePress!: (event: KeyboardEvent) => void;
+    private boundHideOnFocusOut!: (event: FocusEvent) => void;
 
     /**
      * Toggles optional classes and accessibility attributes in addition to BasePopoverController.shown
@@ -352,9 +353,19 @@ export class PopoverController extends BasePopoverController {
             this.boundHideOnOutsideClick || this.hideOnOutsideClick.bind(this);
         this.boundHideOnEscapePress =
             this.boundHideOnEscapePress || this.hideOnEscapePress.bind(this);
+        this.boundHideOnFocusOut =
+            this.boundHideOnFocusOut || this.hideOnFocusOut.bind(this);
 
         document.addEventListener("mousedown", this.boundHideOnOutsideClick);
         document.addEventListener("keyup", this.boundHideOnEscapePress);
+        this.referenceElement.addEventListener(
+            "focusout",
+            this.boundHideOnFocusOut
+        );
+        this.popoverElement.addEventListener(
+            "focusout",
+            this.boundHideOnFocusOut
+        );
     }
 
     /**
@@ -363,6 +374,14 @@ export class PopoverController extends BasePopoverController {
     protected unbindDocumentEvents() {
         document.removeEventListener("mousedown", this.boundHideOnOutsideClick);
         document.removeEventListener("keyup", this.boundHideOnEscapePress);
+        this.referenceElement.removeEventListener(
+            "focusout",
+            this.boundHideOnFocusOut
+        );
+        this.popoverElement.removeEventListener(
+            "focusout",
+            this.boundHideOnFocusOut
+        );
     }
 
     /**
@@ -397,6 +416,24 @@ export class PopoverController extends BasePopoverController {
         // note: .contains also returns true if the node itself matches the target element
         if (this.popoverElement.contains(<Node>e.target)) {
             this.referenceElement.focus();
+        }
+
+        this.hide(e);
+    }
+
+    /**
+     * Forces the popover to hide if keyboard focus leaves both the reference element and the popover.
+     * @param {FocusEvent} e - The focusout event from the reference or popover element
+     */
+    private hideOnFocusOut(e: FocusEvent) {
+        const relatedTarget = e.relatedTarget;
+
+        if (
+            relatedTarget instanceof Node &&
+            (this.referenceElement.contains(relatedTarget) ||
+                this.popoverElement.contains(relatedTarget))
+        ) {
+            return;
         }
 
         this.hide(e);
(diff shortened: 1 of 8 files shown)


packages/stacks-classic/lib/components/popover/popover.ts
```diff
@@ -85,6 +85,13 @@ export abstract class BasePopoverController extends Stacks.StacksController {
         }
     }
 
+    /**
+     * Only menu popovers dismiss when focus leaves the reference/popover pair.
+     */
+    protected get shouldHideOnFocusLeave() {
+        return this.popoverElement?.getAttribute("role") === "menu";
+    }
```
asblanco, line 97: `shouldHideOnFocusLeave` reads `role` from `popoverElement` (the `.s-popover` div). But the documented menu-in-popover markup ([menus.md](packages/stacks-docs/src/docs/public/system/components/menus.md)) puts `role="menu"` on the inner `<ul class="s-menu">`, with no role on `.s-popover`:

```html
<div class="s-popover"><ul class="s-menu" role="menu">…</ul></div>
```

I loaded the built bundle in a browser to confirm: with `role="menu"` on the inner `<ul>`, this getter returns `false` and the popover stays open on focus-leave (SO_07 not fixed); with `role="menu"` on `.s-popover` it works. The new test passes only because its fixture puts the role on `.s-popover`, which differs from how we document menus.

Can you confirm what the real dropdowns render? If the role lives on the inner list, we should detect it there. Otherwise we at least need to document that focus-leave dismissal requires `role="menu"` on the `.s-popover` root.

If real markup keeps `role="menu"` on the inner list, an option that stays backward-compatible:

```suggestion
    protected get shouldHideOnFocusLeave() {
        return (
            this.popoverElement?.getAttribute("role") === "menu" ||
            !!this.popoverElement?.querySelector('[role="menu"]')
        );
    }
```

packages/stacks-svelte/src/components/Popover/Popover.svelte
```diff
@@ -198,6 +209,28 @@
         close();
     };
 
+    const onFocusOut = (e: FocusEvent) => {
+        if (tooltip || !pstate.visible || !pstate.closeOnFocusLeave) {
```
asblanco: `onFocusOut` doesn't check `dismissible`, but `onOutclick` (`:204`) and `onKeypress` (`:235`) both do. A `role="menu"` popover with `dismissible={false}` will still close on focus-leave. Same asymmetry on the classic side: `shouldHideOnFocusLeave` (`popover.ts:91`) ignores the `hide-on-outside-click="never"` opt-out that `shouldHideOnOutsideClick` (`popover.ts:73`) respects, so a menu that explicitly disabled auto-dismissal has no way to opt out of focus-leave dismissal. Intended? If not, gate both on the existing flags.

```suggestion
        if (tooltip || !pstate.visible || !pstate.closeOnFocusLeave || !dismissible) {
```

packages/stacks-classic/lib/components/popover/popover.ts
```diff
@@ -402,6 +428,28 @@ export class PopoverController extends BasePopoverController {
...
+    /**
+     * Forces the popover to hide if keyboard focus leaves both the reference element and the popover.
+     * @param {FocusEvent} e - The focusout event from the reference or popover element
+     */
+    private hideOnFocusOut(e: FocusEvent) {
+        if (!this.shouldHideOnFocusLeave) {
+            return;
+        }
+
+        const relatedTarget = e.relatedTarget;
```
asblanco, line 444: When `relatedTarget` isn't a contained `Node` the handler dismisses. That's correct for "focus left to nowhere" (and it's the behavior the null-relatedTarget test pins), but it also means: clicking a non-focusable dead-space region *inside* the menu blurs the active item with `relatedTarget=null` and closes the menu, and alt-tabbing/window-blur closes it too. Acceptable to ship as-is, but flagging as a known tradeoff in case QA reports a menu closing on an in-menu click. No change requested. (Same on the Svelte side, `Popover.svelte:217`.)

asblanco (commented):
Reviewed against [A11Y-160](https://stackoverflow.atlassian.net/browse/A11Y-160). This PR cleanly tackles the close-on-blur part of the ticket. Mapping the diff to the ticket's three ACs:

| A11Y-160 AC | This PR | Notes |
|---|---|---|
| Announce expanded/collapsed (`aria-expanded`) | ✅ pre-existing + new regression test | already set in `toggleAccessibilityAttributes` (`popover.ts:477`); the PR adds a guard test, not the implementation |
| Announce item count on open (`aria-setsize`/`aria-posinset` or `aria-label`) | ❌ out of scope | not touched in the diff |
| Menu closes when Tab moves focus outside (SO_07) | ✅ implemented | classic + Svelte, gated to `role="menu"` |

So this is a focused fix for **SO_07** plus a regression test for the `aria-expanded` AC — item-count (SO_08) is still outstanding.

👌 Nice incidental catch: returning the `setup*` teardown from `PopoverReference`'s `onMount` fixes a real listener leak where the old code created the cleanup closures but never registered them.

asblanco:
Every popover demo in the docs uses the default `role="dialog"`, which is excluded from this behavior — so there's nothing on the deploy preview that shows the fix working. Consider adding a "Menu popovers" example to `popovers.md` (a `role="menu"` popover next to a default one, with an outside focus target) so the dismissal is demoable and the `role="menu"` placement requirement is documented.

Heads-up: this only demos the **Svelte** path — the docs site doesn't run classic JS, so classic dismissal still relies on the unit tests.

asblanco (changes requested):

asblanco:
The classic fix now matches `role="menu"` on the `.s-popover` or an inner `[role="menu"]`, but the Svelte side only checks PopoverContent's own role. Since the Menu component renders `role="menu"` on its own `<ul>` (Menu.svelte:40), the idiomatic composition diverges from classic: 

| Markup | Dismisses on focus-leave? |
| :--- | :---: |
| Classic `<div class="s-popover"><ul role="menu">` | ✅ |
| Svelte `<PopoverContent><Menu>…</Menu></PopoverContent>` | ❌ |

Suggest mirroring the classic detection so a contained menu also enables dismissal: 

```js
// bind the root element (bind:this={contentElement}) and:
$effect(() => {
    pstate.closeOnFocusLeave =
        !pstate.tooltip &&
        (computedRole === "menu" ||
            !!contentElement?.querySelector('[role="menu"]'));
});
```

asblanco (approved):
LGTM! Now just pending on STACKS-901 to fix CI

PR #2302, opened 2026-06-22
fix(ci): pin Node to 24.15.0 to stop Playwright install hang

asblanco:
## [STACKS-901 CI: Playwright jobs hang ~6h on `playwright install`](https://stackoverflow.atlassian.net/browse/STACKS-901)

Fix the CI hang where every Playwright-dependent job (`Unit Tests`, `A11y Tests`, `Unit Tests (stacks-svelte)`) ran until GitHub's 6-hour ceiling.

**Started failing** 
between [2026-05-28](https://github.com/StackExchange/Stacks/actions/runs/26552520145) (last green, Node 24.15.0) and [2026-06-01](https://github.com/StackExchange/Stacks/actions/runs/26756110626) (first 6h hang, Node 24.16.0), when the runner's `lts/*` rolled 24.15.0 → 24.16.0.

**Root cause** 
`playwright install` hangs extracting the browser archive on **Node 24.16.0**, which `node-version: 'lts/*'` silently rolled onto (last good runtime: 24.15.0). 

**Solution** 
Pinning `node-version` to **24.15.0** restores extraction without touching Playwright or browser versions. A job-level `timeout-minutes: 30` caps any future stall well short of the 6h ceiling.

**Follow-up:** [STACKS-902](https://stackoverflow.atlassian.net/browse/STACKS-902) — upgrade Playwright so the Node pin can float again.

### Manual testing
Confirm `Unit Tests`, `A11y Tests`, `Unit Tests (stacks-svelte)`, and `Visual Regression Tests` pass (~2–8 min; previously hung ~6h).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,9 +35,10 @@ jobs:
       - name: ⎔ Setup node
         uses: actions/setup-node@v4
         with:
-          # Pin to an explicit major rather than lts/* so the runtime can't
-          # silently drift onto an untested Node version (see STACKS-901).
-          node-version: '24'
+          # Pin to Node 22 (explicit, not lts/*): Node 24 hangs `playwright
+          # install` while extracting the browser archive. Pinning an explicit
+          # major also stops the runtime silently drifting. See STACKS-901.
+          node-version: '22'
 
       - name: 🏗 Install Dependencies
         run: npm ci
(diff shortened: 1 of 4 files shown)


giamir (approved):
Nice stop gap @asblanco. Thanks for this!

PR #2303, opened 2026-06-22
ci: upgrade Playwright to 1.61.0 and re-float Node to lts/*

asblanco:
## [STACKS-902 Upgrade Playwright so CI can return to a floating Node pin](https://stackoverflow.atlassian.net/browse/STACKS-902)

Upgrades Playwright to 1.61.0 and re-floats CI's Node pin to `lts/*`, restoring a current, patched toolchain after the STACKS-901 exact-version (`24.15.0`) stopgap.

Acceptance criteria:

- [x] **AC 1 – Current runtime** — Playwright-dependent CI jobs run on a current Node (no exact `24.15.0` pin) and pass.
- [x] **AC 2 – Aligned visual environment** — Playwright and the visual-test browser environment are on matching, current versions (`playwright@1.61.0` + `mcr.microsoft.com/playwright:v1.61.0-jammy`).
- [x] **AC 3 – Trustworthy baselines** — Visual-regression baselines are regenerated and reviewed so the suite passes without masking real visual regressions.

### Manual testing

1. Confirm CI is green on this PR — in particular `Unit Tests`, `A11y Tests`, and `Unit Tests (stacks-svelte)` (Playwright install completes in minutes, no multi-hour hang) and `Visual Regression Tests`.
2. Review the regenerated baselines. The render shift is **Webkit-only (1235 baselines)** from the newer browser; Chromium and Firefox are byte-identical. To view the image diffs locally:
   ```sh
   git config diff.lfs.textconv cat
   ```
   The baseline `.ico` files then diff as images; spot-check a sample to confirm benign render differences rather than masked regressions.
3. The 120 `s-notice-*-longText` → `longtext` baseline renames align filenames with the snapshot name emitted by `notice.visual.test.ts`; the camelCase files were stale orphans the current suite no longer produces.


.prettierrc
```diff
@@ -5,5 +5,14 @@
...
+    "endOfLine": "auto",
+    "overrides": [
+        {
+            "files": ".github/**/*.{yml,yaml}",
+            "options": {
+                "tabWidth": 2,
+                "singleQuote": true
+            }
+        }
+    ]
```
asblanco, line 17: I had to add this otherwise my editor was reformatting the whole file to 4 tabs ([line 2](https://github.com/StackExchange/Stacks/pull/2303/changes#diff-663ade211b3a1552162de21c4031fcd16be99407aae5ceecbb491a2efc43d5d2L2)). I thought it's ok to add this as part of the PR, but let me know otherwise

giamir (approved):

PR #2304, opened 2026-06-22
fix(popover): backport focus-leave dismissal to v2

alexwarren:
## Summary

Backports the popover focus-leave dismissal behavior from PR #2298 to the v2 branch.

- Adds Classic popover focusout handling so menu popovers close when keyboard focus leaves the reference/popover pair.
- Adds Svelte popover focus-leave handling, including controlled popover close requests and nested role="menu" detection.
- Adds v2 docs guidance for menu popovers and restores the PR #2298 changeset.

## Validation

- npm run lint:ts -w packages/stacks-classic passed.
- npm exec -w packages/stacks-svelte -- eslint src/components/Popover/Popover.svelte src/components/Popover/PopoverContent.svelte src/components/Popover/PopoverReference.svelte src/components/Popover/Popover.test.ts passed.
- npm run build:dependencies -w packages/stacks-svelte passed with existing Less and bundle-size warnings.

## Local validation blockers

- Classic web-test-runner starts but fails before assertions because existing unit test imports hit local dependency module export errors from aria-query/lz-string.
- Svelte web-test-runner fails before assertions because the local icon package does not export existing symbols such as IconShieldXSm.
- npm run lint -w packages/stacks-svelte is blocked by the same existing missing icon exports across unrelated Svelte components/stories.
- Root npx prettier could not parse the existing v2 docs HTML because of a pre-existing unmatched </kbd>; Svelte files were formatted via the workspace Prettier command instead.

### DIFF
--- a/.changeset/two-goats-press.md
+++ b/.changeset/two-goats-press.md
@@ -1,0 +1,6 @@
+---
+"@stackoverflow/stacks": minor
+"@stackoverflow/stacks-svelte": minor
+---
+
+Fix popover focus-leave dismissal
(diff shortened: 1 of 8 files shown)


asblanco (approved):
LGTM 🐐 🐐 📰

PR #2309, opened 2026-06-23
Update styleguide.md

someobscurereference:
Minor edits to conform with the CM department version for writing on meta sites

### DIFF
--- a/packages/stacks-docs/src/docs/public/copy/styleguide.md
+++ b/packages/stacks-docs/src/docs/public/copy/styleguide.md
@@ -1,5 +1,5 @@
 ---
-updated: 2026-04-29
+updated: 2026-06-23
 ---
 
 <script>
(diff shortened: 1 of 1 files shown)


someobscurereference:
A few things that need attention, @abovedave , just to make sure the stuff from the CM department's makes it to somewhere in the doc:

- In the "quotes" section, we had a bit that read: "Use blockquote formatting for larger/longer pieces of quoted text, as appropriate." — is there a place in which we can fit this that makes sense, given that this section is for product copy, and the guidance may only apply to meta writing?
- The CMs' version of the guidance for date and time, for writing on Meta, differs fundamentally from this document's. In particular, for writing on meta, the guidance is:
    > The Stack Exchange network uses UTC time wherever exact dates and times are used. If you specify a time in your text, you should generally list that time in UTC (though it’s okay to list it in another timezone in addition to UTC, or to link to a website that automatically converts that time to the user’s local time).
    > In general, avoid relative temporal references (such as “next month”) in a post’s title; it’s fine to mention those in the body of the post, but it’s even better to specify the exact date where possible (since the user may be reading the post long after it was originally published).
- There's a section on the CM guidance for writing on meta sites that is missing here, but I'm unclear as to where to include it: "**Tenses:** Be consistent in your usage of verb tenses in paragraphs. If you’ve been writing in one tense and are switching to a new tense, consider breaking your text into a new paragraph to make it less confusing to read."
- I also left a comment inline about abbreviations: the guidance in this section for "internal communications" also applies to writing on meta sites, despite that being an external communication channel. How can we somehow highlight that the "internal communications" guidance also applies when writing to meta sites?

PR #2315, opened 2026-07-01
fix(gap): target axis gap properties

dancormier:
### What this PR does

1. **Adds the `!important` declaration to all `gap` property atomic classes**
  We expect all of our atomic classes to include `!important` so they override the given style of a component in all cases.
2. **Updates the `.gx*` and `.gy*` atomic classes so they _only_ impact the the CSS property for a single axis, as intended.**
  Prior to this PR, those properties set a `gap` property which included _some_ value for both axes. This would result in the property potentially overwriting an existing gap value for the axis _not_ intended to be set by the atomic class applied.
3. **Updates the gap docs and Less snapshot to reflect these changes**

## Related Issue

[STACKS-908](https://stackoverflow.atlassian.net/browse/STACKS-908)

## Changes

- Updates `.g*` classes to emit `gap: var(--_gap-y, 0) var(--_gap-x, 0) !important`.
- Updates `.gx*` classes to emit `row-gap: var(--_gap-x) !important`.
- Updates `.gy*` classes to emit `column-gap: var(--_gap-y, 0) !important`.
- Keeps the explicit selector lists in `gap.less`.
- Updates the gap docs tables to reflect the new emitted properties.

## Testing

- `npm run lint:css -w packages/stacks-classic`
- `npx vitest run lib/atomic/gap.less.test.ts -u`
- `npm run test:less -w packages/stacks-classic`


packages/stacks-classic/lib/atomic/gap.less
```diff
@@ -39,8 +39,14 @@
...
 
-.gx0, .gx1, .gx2, .gx4, .gx6, .gx8, .gx12, .gx16, .gx24, .gx32, .gx48, .gx64,
-.gy0, .gy1, .gy2, .gy4, .gy6, .gy8, .gy12, .gy16, .gy24, .gy32, .gy48, .gy64,
 .g0, .g1, .g2, .g4, .g6, .g8, .g12, .g16, .g24, .g32, .g48, .g64 {
-    gap: var(--_gap-y, 0) var(--_gap-x, 0);
+    gap: var(--_gap-y, 0) var(--_gap-x, 0) !important;
+}
+
+.gx0, .gx1, .gx2, .gx4, .gx6, .gx8, .gx12, .gx16, .gx24, .gx32, .gx48, .gx64 {
+    row-gap: var(--_gap-x) !important;
```
asblanco:
```suggestion
    column-gap: var(--_gap-x) !important;
```

packages/stacks-classic/lib/atomic/gap.less
```diff
@@ -39,8 +39,14 @@
...
+}
+
+.gy0, .gy1, .gy2, .gy4, .gy6, .gy8, .gy12, .gy16, .gy24, .gy32, .gy48, .gy64 {
+    column-gap: var(--_gap-y, 0) !important;
```
asblanco:
```suggestion
    row-gap: var(--_gap-y, 0) !important;
```

asblanco (changes requested):
I believe we have inverted the x/y axes. We had `gap: <row-gap | gap-y> <column-gap | gap-x>` and now we're setting row-gap to gx and column-gap to gy 👀

dancormier:
> I believe we have inverted the x/y axes. We had `gap: <row-gap | gap-y> <column-gap | gap-x>` and now we're setting row-gap to gx and column-gap to gy 👀

Oh wow, I can't believe I did that 🤦 Thank you for catching this! I've accepted your changes and have updated the Less snapshots so this should be ready for another review

asblanco (approved):

PR #2323, opened 2026-07-08
chore(docs): deploy docs on release publish

dancormier:
## Summary

Deploy production docs from the release flow instead of ordinary `main` pushes. When Changesets publishes packages from `main`, the release workflow calls the Netlify build hook in `NETLIFY_DOCS_BUILD_HOOK_URL`.

This also adds a separate manual `Deploy docs` workflow for the rare case where we intentionally want to deploy the latest `packages/stacks-docs` content from `main` without publishing a package release.

### Why

Netlify continuous deployment is still active so PR previews and the `v2` branch deploy keep working. Because `main` is also the production branch, normal `main` pushes could otherwise deploy production docs even when no package was released.

This PR keeps Netlify active, but adds a Netlify ignore guard so production builds only proceed for:

- release-triggered build hooks
- explicit manual docs deploys
- deploy previews and branch deploys

## Changes

- Adds a release-only Netlify build hook call in `.github/workflows/main.yml`.
- Adds `.github/workflows/deploy-docs.yml` for manual latest-`main` docs deploys.
- Adds bounded retries to both build-hook calls.
- Adds `packages/stacks-docs/netlify.toml` and `netlify-ignore.sh` for Netlify build gating.
- Adds automated coverage for production, deploy-preview, and branch-deploy contexts.
- Updates `README.md` with the release/manual docs deploy behavior.

## Required Configuration

- [x] Add `NETLIFY_DOCS_BUILD_HOOK_URL` as a GitHub Actions secret containing the Netlify build hook URL for the docs site. See [Netlify](https://app.netlify.com/projects/stacks/configuration/deploys#build-hooks)

## Testing

- `npm run test -w packages/stacks-docs` passes the existing docs tests and the Netlify ignore tests.
- `npm run check -w packages/stacks-docs` completes with existing warnings and no errors.
- `npm run lint -w packages/stacks-docs` completes with existing warnings and no errors.
- `bash -n packages/stacks-docs/netlify-ignore.sh` verifies the shell script syntax.
- The Netlify deploy preview confirms the `deploy-preview` context remains enabled.

Netlify documents that build hooks bypass the ignore result, so release and manual hook builds proceed while automatic production builds are skipped.

## Testing Limits

The full release path cannot be tested before merge without performing a real release from `main`; `steps.changesets.outputs.published == 'true'` only happens when Changesets publishes packages after version bumps land on `main`.

The first complete production validation of the release path will happen on the next actual package release. The manual workflow also cannot be run from `main` until this workflow file exists on `main`.

## Netlify/Terraform Follow-Up

This change uses a Netlify build hook named `GitHub release deploy`. It may need to be moved into Terraform as part of [STACKS-860](https://stackoverflow.atlassian.net/browse/STACKS-860). @ttaylor-stack

## Related Issue

[STACKS-862](https://stackoverflow.atlassian.net/browse/STACKS-862)


dancormier:
> The full release path cannot be tested before merge without performing a real release from main; steps.changesets.outputs.published == 'true' only happens when Changesets publishes packages after version bumps land on main.

I'm merging to test this out 🤞

PR #2332, opened 2026-07-09
fix(ci): dependabot CI

asblanco:
Dependabot-authored workflow runs do not have access to the private docs SSH key. The workflow was still creating an empty key file, which made the docs build try to clone private docs and fail with exit 128.

This skips SSH setup when the secret is unavailable and makes the docs script ignore empty key files.

### DIFF
--- a/packages/stacks-docs/package.json
+++ b/packages/stacks-docs/package.json
@@ -12,7 +12,7 @@
         "check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
         "lint": "npm run check && eslint .",
         "format": "prettier --write .",
-        "init-private-content": "if [ -d src/docs/private/.git ]; then git -C src/docs/private pull --ff-only; elif [ -f ~/.ssh/id_rsa ]; then GIT_SSH_COMMAND=\"ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no\" git clone --depth 1 git@github.com:StackEng/StacksDocsPrivate.git src/docs/private; else echo 'Skipping private docs; no SSH key found'; fi"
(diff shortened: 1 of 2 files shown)


packages/stacks-docs/package.json
```diff
@@ -12,7 +12,7 @@
         "check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
         "lint": "npm run check && eslint .",
         "format": "prettier --write .",
-        "init-private-content": "if [ -d src/docs/private/.git ]; then git -C src/docs/private pull --ff-only; elif [ -f ~/.ssh/id_rsa ]; then GIT_SSH_COMMAND=\"ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no\" git clone --depth 1 git@github.com:StackEng/StacksDocsPrivate.git src/docs/private; else echo 'Skipping private docs; no SSH key found'; fi"
+        "init-private-content": "if [ -d src/docs/private/.git ]; then git -C src/docs/private pull --ff-only; elif [ -s ~/.ssh/id_rsa ]; then GIT_SSH_COMMAND=\"ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no\" git clone --depth 1 git@github.com:StackEng/StacksDocsPrivate.git src/docs/private; else echo 'Skipping private docs; no SSH key found'; fi"
```
asblanco, line 15: `-f` means: "Does this file exist and is it a regular file (not a directory)?"
`-s` means: "Does this file exist and is its size greater than 0 bytes?"

dancormier (approved):
Nice fix @asblanco Thanks for hopping on this! Seems like a reasonable fix to me 🙂

PR #2377, opened 2026-08-24
fix(popover): support touch-triggered tooltips

dancormier:
## Summary

- toggle Svelte tooltips when their references are activated by touch
- limit hover behavior to hover-capable pointers and respond to capability changes
- preserve keyboard focus behavior and existing dismissal behavior
- document touch, mouse, keyboard, cursor, and dismissal expectations
- add regression coverage for touch, hybrid devices, mouse input, and capability changes

## Jira

[STACKS-913](https://stackoverflow.atlassian.net/browse/STACKS-913)

[NXT-303](https://stackoverflow.atlassian.net/browse/NXT-303)

## Testing locally

1. Install dependencies from the repository root with `npm ci`.
2. Start Stacks Svelte Storybook with `npm run storybook -w packages/stacks-svelte`.
3. Open [the Popover Tooltip story](http://localhost:6006/?path=/story/components-popover--tooltip).
4. In Chrome DevTools, enable the device toolbar and select a mobile device so pointer events are simulated as touch input.
5. Tap the **Tooltip** reference. The tooltip should open on the first tap and close on the second tap.
6. Open it again and tap outside the tooltip. The tooltip should close.
7. Disable device emulation and verify that mouse hover and keyboard focus still open the tooltip.

## Accessibility

- keyboard focus continues to open tooltips
- touch users can open and dismiss tooltip content
- Escape, focus leave, outside click, and repeat touch dismissal remain supported
- mouse clicks do not independently trigger tooltips

## Changeset

Patch release for `@stackoverflow/stacks-svelte`.


asblanco:
**1. Medium — Component contract still documents tooltip mode as hover-only**

At [Popover.svelte:85](Stacks/packages/stacks-svelte/src/components/Popover/Popover.svelte):
[STACKS-913](https://stackoverflow.atlassian.net/browse/STACKS-913) requires this interaction contract to be documented, but the public tooltip prop still describes tooltip mode as hover-only. Should we document touch toggling, keyboard focus, mouse behavior, and supported dismissal methods in the prop or component documentation?

packages/stacks-svelte/src/components/Popover/Popover.test.ts
```diff
@@ -1225,6 +1269,107 @@ describe("Popover", () => {
...
+            await clock.runAllAsync();
+            await tick();
+            expect(screen.queryByRole("tooltip")).not.to.exist;
+
+            await userEvent.click(button);
+            await clock.runAllAsync();
+            await tick();
+            expect(screen.queryByRole("tooltip")).not.to.exist;
+
+            tap(button);
```
asblanco, line 1300: **2. Low — Touch-specific outside dismissal is untested**

Could this test also open through tap(button) and dismiss through an outside tap? Touch-triggered outside dismissal is part of the agreed behavior, but the new coverage currently verifies only repeated-tap dismissal.

packages/stacks-svelte/src/components/Popover/PopoverReference.svelte
```diff
@@ -74,15 +74,68 @@
...
-        ref.addEventListener("mouseleave", pstate.closeTooltip);
-        ref.addEventListener("focusin", pstate.openTooltip);
+        const hoverMedia = window.matchMedia("(hover: hover)");
+        let pointerDown = false;
+        let pointerType = "";
+        const onPointerDown = (event: PointerEvent) => {
+            pointerDown = true;
+            pointerType = event.pointerType;
+        };
+        const onPointerUp = () => {
```
asblanco, line 86: **3. Low — An aborted touch can affect a later keyboard click**

`pointerType` survives a `pointerup` when no click reaches the reference—for example, when a touch ends elsewhere. A later keyboard-generated click can therefore be treated as touch. Could we reset this state after the current click-dispatch window or use the click event pointer metadata directly?

asblanco:
I was able to test it on my iPhone 16 Pro on Safari and on Qwant browsers and it worked well on both 🎉 

https://github.com/user-attachments/assets/0a82e0be-ede2-4109-a5da-1a5a9bc0c697

I just posted with Codex 3 findings, have a look at them if they make sense for you to act on

dancormier:
Addressed review finding 1 in `118e54669`. The public tooltip prop now documents hover-capable pointer behavior, keyboard focus, touch toggling, mouse-click behavior, and supported dismissal methods.

packages/stacks-svelte/src/components/Popover/Popover.test.ts
dancormier: Addressed in `118e54669`. The regression now opens through touch, dismisses through an outside tap, then reopens to retain repeat-tap coverage.

packages/stacks-svelte/src/components/Popover/PopoverReference.svelte
dancormier: Addressed in `118e54669`. Touch pointer metadata now expires after the current click-dispatch window and is cleaned up. A regression test covers an aborted touch followed by a keyboard click.

asblanco (approved):
LGTM!

